PR #32, opened 2014-12-31
More hyphenation files

cyocum:
Here are the last of the hyphenation files.  The only ones that I did not include are the Greek monotonic and polytonic as I am unsure how Sile handles that when using language codes.  I would assume it is just a hashtable look up but some guide on this would be helpful.


cyocum:
Hi,

This has sat here for a bit.  I was wondering if it got forgotten?

simoncozens:
It did! Sorry!

PR #48, opened 2015-02-13
If we don't do this gTotal.stretch will always be zero 

Nathan22Miles:
and "\vfill .. blah .. \vfill" will not be centered on the page


simoncozens:
I've merged the relevant commit. (I haven't added the lectionary class to the main SILE repo as yet.)

PR #49, opened 2015-02-14
Put core libraries in /usr/lib/sile instead of /usr/share/sile.

oleastre:
/usr/share should only contain architecture independant data.
This is a packaging requirement for various distributions.


simoncozens:
Thank you!

PR #51, opened 2015-02-15
Forgot to modify all-local dependency.

oleastre:
Sorry for this, but I forgot to change all-local definition in src/Makefile.am
This prevent parallel build to work.


simoncozens:
Thanks very much!

PR #57, opened 2015-03-02
Initialize pdf_files to allow the libtexpdf outputter to use PDF files as images.

oleastre:
This is allows me to include a pdf file using `SILE.outputter.drawImage` .
It's a first step towards pdf image support; currently, pdf files cannot be included a using the image package (imagesize lacks pdf support).

### DIFF
--- a/src/justenoughlibtexpdf.c
+++ b/src/justenoughlibtexpdf.c
@@ -24,6 +24,7 @@ int pdf_init (lua_State *L) {
   mediabox.lly = 0.0;
   mediabox.urx = w;
   mediabox.ury = height;
+  texpdf_files_init();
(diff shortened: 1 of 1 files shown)


simoncozens:
Thank you so much for this! This looks great. Maybe we can add some code to `justenoughlibtexpdf` to return bounding boxes rather than using `imagesize`, as that will give us PDF support. The relevant code to steal is here: https://github.com/simoncozens/dvipdfm-x/blob/libtexpdf/xbb.c

oleastre:
I was just looking in that direction as there is already a lot of code dealing with file types and dimension in libtexpdf. 
I will try to solve this in the coming days.

PR #58, opened 2015-03-02
In case of libtexpdf outputter, use it to compute imagesize.

oleastre:
This enables pdf images support in the libtexpdf case.
Other outputters fallback to the imagesize package.


oleastre:
I think I did it... 
I have tested it on a simple file including a pdf image, as well as the sile documentation.

I hope I did not introduced regressions in the cairo and podofo cases.

simoncozens:
Looks great, thank you! I am not altogether too bothered about Cairo at the moment - since it gives horrible output on Linux - and podofo should be considered dead. (It requires various patches to the podofo library to even work at all.) libtexpdf is what we should be aiming at, and anything else is a bonus to help us to get the correct levels of abstraction in our implementation.

PR #59, opened 2015-03-04
Do not silent lua errors

oleastre:
It's easier to debug the class file you're working on, if errors are shown when you test it.


oleastre:
I see travis tests fails; so I tried to run the tests locally but can not reproduce the problem.

simoncozens:
I want travis to fail if there were Lua errors. Previously we were silencing them so travis was passing, but it should not be! :-)

PR #63, opened 2015-03-21
Tools and chordmode

oleastre:
Here is the chordmode code split in:
- tools for the generic text manipulation
- chordmode that uses the tools package to place the text above the current line.

Simple tools package example that defines an uppercase command that uses the transformContent function:
```
\begin[papersize=a4,class=plain]{document}
\script[src=packages/color]
\begin{script}
local tools = SILE.require("packages/tools").exports
SILE.registerCommand("uppercase", function(options, content)
  SILE.process(tools.transformContent(content, string.upper))
end, "Typeset the enclosed text as uppercase")
\end{script}

First line of text

\begin{uppercase}
This one should be uppercase 
  \color[color=#000090]{blue}, 
  \color[color=#900000]{red}, 
  \color[color=#009000]{green}.
\end{uppercase}
\end{document}
```

This is rendered as: 
tools-example

Note how the transformContent function traverse the whole content tree to transform all contained text.

### DIFF
--- a/packages/chordmode.lua
+++ b/packages/chordmode.lua
@@ -1,0 +1,72 @@
+local tools = SILE.require("packages/tools").exports
+
+local function addChords(text, content)
+  local start, stop, chord, char
+  local index = 1
+  local chordText = nil
+  local result = {}
+  local strlen = string.len(text)
+  
+  repeat
+    start, stop, chord = string.find(text, "<([^>]+)>", index)
+    if stop==nil then
+      table.insert(result, string.sub(text, index))
+      index = strlen+1
+    else
+      -- Parse a new chord
+      if (start-index) > 1 then
+        table.insert(result, string.sub(text, index, start - 1 ))
+      end
+      index = stop + 1
+      start, stop, char = string.find(text, "([<\n])", index)
+      if stop==nil then
+        chordText = string.sub(text, index)
+        index = strlen+1
+      else
+        if(char=='<') then 
+          start = start - 1
+        end
+        chordText = string.sub(text, index, start)
+        index = start
+      end
+      table.insert(result, tools.createCommand(content.pos, content.col, content.line, "ch", {name=chord}, chordText))
+    end
+  until index > strlen
+
+  return result
+end
+
+SILE.registerCommand("ch", function(options, content)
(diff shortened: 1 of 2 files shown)


simoncozens:
Neat. Can you explain what createCommand is for?

simoncozens:
Ignore previous comment, I worked it out. :-)

simoncozens:
OK, a few thoughts; in general I like this a lot. The way you've done transformation by manipulating the parsed AST is excellent, and a lot smarter than I was thinking of. (I was just going to transform the incoming text but that way you can't do stuff like your uppercase example.)

However:

* I'm afraid I don't like the name `tools`. It's the sort of package name that anything and everything will end up getting dumped into. Can we call it `inputmanipulation` or something? (I can't think of a smart name right now but I'm sure there is one.)

* Your chords example works fine (which is the most important thing!) but I think there are simpler ways to implement it. For instance, look at the `SU.gtoke` function in `utilities.lua` and how it is used by `SILE.shapers.harfbuzz.shape`; it provides a generic tokeniser which you can use to pick out the chord tokens.

* We could also simplify the implementation a lot if we had a package which provided commands like TeX's `\smash` `\phantom` `\llap` `\rlap` and so on. These would, e.g. create a hbox which has no width but wraps another hbox, calling the wrapped hbox's output routine from its own output routine. You could then implement `\ch` as something like `\collapse{\raise[height=2.5mm]{\color[color=red]{\process}}}`.

oleastre:
First, thanks for the feed back; I knew there should be smarter ways to implement the chordmode but was in need of some directions. I will have a look a all of this and come back with an improved implementation and probably a separate package for the suggested TeX like commands.

For the "tools" package renaming, what do you think about something short like 'ITM' for 'input tree manipulation' ?

thanks

simoncozens:
ITM only makes sense if you already know what it does. :-) Can we call it "inputfilter"? I'll close this PR and look forward to seeing what you come up with. :-)

oleastre:
Ok will do a pull request just for the inputfilter and will followup with an other one for the chordmode (I still have to figure out how to correctly use gtoke for what I want to do)

PR #64, opened 2015-03-25
Add input AST transformation tools.

oleastre:
Initial version of the inputfilter package with a simple example.


simoncozens:
Lovely, thanks very much! I'll add some documentation later on.

PR #65, opened 2015-03-31
Add a package to typeset chord name above text.

oleastre:
Here is a new version of the chordmode package using SU.gtoke for text parsing and a kind of state machine to process the found tokens. Let me know what you think about that part.

The ``\ch`` command implementation is still buggy (See the line starting with "Complex line" in the example) and as you suggested should be implemented on top of some simple TeX like commands (to be added in another package).

I am working to fix this.

### DIFF
--- a/examples/chordmode.sil
+++ b/examples/chordmode.sil
@@ -1,0 +1,25 @@
+\begin[papersize=a4,class=songbook]{document}
+
+\script[src=packages/chordmode]
+
+\begin{chordmode}
+
+  I've be<G>en a wild rover for many's a <C>year
+
+  And I <G>spent all me <D7>money on whiskey and <G>beer
+
+  But now I'm returning with gold in great <C>store
+
+  And I <G>never will <D7>play The Wild Rover no <G>more.
+
+  Complex line <Am7>b<C5dim>d<Em>
+
+  Line without any chord
+
+  <F> starts with a chord.
+
+  ends with a chord<G>
+
+\end{chordmode}
+
+\end{document}
--- a/packages/chordmode.lua
+++ b/packages/chordmode.lua
@@ -1,0 +1,105 @@
+local inputfilter = SILE.require("packages/inputfilter").exports
+
+local function addChords(text, content)
+  local result = {}
+  local chordName = nil
+  local currentText = ""
+  local process = nil
+  local processText, processChordName, processChordText
+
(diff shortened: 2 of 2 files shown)


simoncozens:
Thanks very much for this! I've included this, and simplified the code somewhat. This will be helpful for something else I am thinking about, which is putting *furigana* (also known as "ruby", phonetic readings) on top of Japanese text.

oleastre:
I've seen your modifications... and I hope it's not over simplified for my goal. If you look at the exemple; in the line starting with "Complex Line", the different chords text overlaps. That's something I would like to solve by either adding spacing between letters or even better, add some dash.

To explain a little bit further what I am trying to achieve, I've already done something similar in LaTeX; but with an external parser that generates the tex source.
Here is my parser input:
```
Will I [Sol]play The Wild [Do]Rover

No, ne[Sol]ver, [Ré7]no [Sol]more
```
This is translated to the following LaTeX code:
```latex
Will I \ch{Sol}{play }The Wild \ch{Do}{Rover}

No, ne\ch{Sol}{ver, }\ch{Ré7}{no }\ch{Sol}{more}
```
Note that, part of the text is added as argument to the `\ch` command(my parser uses up to 10 characters); this is to say "that chord applies to that part of the text".

And here is the LaTeX output (with some annotations).
rect4187

I added a representation of the generated LaTeX boxes around the text and the chord; note how LaTeX adds some space between "no" and "more" to keep the "Sol" chord separated from the "Ré7" one (I add always spaces after the chord in the `\ch` command). It can do so because I've grouped the chord name with the text in the same vbox.

And here is the LaTeX definition of the `\ch` command (really simple):
```latex
\newcommand{\ch}[2]{\strut\vbox{\hbox{\chordfont#1\ \ }\hbox{#2}}}
```

That's what I'm trying to reproduce using SILE, if you want to have a look at the LaTeX version, here is a pdf of my current songbook: http://www.oleastre.be/export/chansonnier.pdf

simoncozens:
Ah, you're right. I have oversimplified it. You do need to apply some of the following text to the command.

simoncozens:
OK, I have reverted the parser change, and instead have fixed it such that the lyric box is at least the width of the chord box.

oleastre:
Thanks a lot. 
I will see if I can improve it by adding some settings and commands to avoid hard coded values.

Issue #44, opened 2015-02-04
Support SVG images


PanderMusubi:
Please support including SVG images for \img.

simoncozens:
This is working now - see `examples/converters.sil`.

PR #67, opened 2015-04-06
Add converters to prepare files before inclusion by SILE.

oleastre:
This is a first idea of a file conversion package for SILE.
It allows you to register a converter based on file extensions, and a conversion command.

This package adds an enhanced \include command that convert a file before inclusion.
When called with a source file, it's extension is checked and the file is converted if a matching converter is found. Then the generated file is included instead of the original one.

To avoid conversion on each execution, I added a dependency on LuaFileSystem to check if the source file is newer than the target one. If needed, I can make it optional.

In fact, I have the same problem as #44 : I'm working with svg file and need to convert them to pdf using inkscape before using them in SILE. So, if I provides SILE with an enhanced \image command that uses the converter, I can avoid using a Makefile to prepare my document.

Let me know what you think about that idea.

### DIFF
--- a/examples/conv-test.lua
+++ b/examples/conv-test.lua
@@ -1,0 +1,13 @@
+#!/usr/bin/env lua
+local output = io.open(arg[2], "w+")
+output:write("\\begin{document}\n")
+for line in io.lines(arg[1]) do
+  if(string.sub(line, 1, 2)=="= ") then
+    output:write("\\center{"..string.sub(line, 3).."}")
+  else
+    output:write(line)
+  end
+  output:write("\n")
+end
+output:write("\\end{document}\n")
+output:close()
--- a/examples/conv-test.txt
+++ b/examples/conv-test.txt
@@ -1,0 +1,3 @@
+= Centered text
+
+Left aligned text 
--- a/examples/converters.sil
+++ b/examples/converters.sil
@@ -1,0 +1,12 @@
+\begin[papersize=a4,class=plain]{document}
+
+\script[src=packages/converters]
+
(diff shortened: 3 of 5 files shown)


oleastre:
I added an "extendCommand" function that allows you to replace an existing command with an enhanced version that can callback the original one. 
This can maybe be moved in baseclass.lua as a generic SILE extension method.

Using this, I added a version of the \img command that will try to convert images before inserting them.
This way, #44 is almost solved by converting svg files automatically to pdf using inkscape(see example).

aguai:
how about the MathTeX, BioTex...and maybe Music Notation???

oleastre:
If you have something that can either convert your MathTex or music notation to sile code or pdf, it's just a matter of adding a converter.

I use this to preprocess lilypond code and include it as pdf in a sile document with something like:
```
\converters:register[from=.ly, to=.pdf, command=lilypond --pdf -o "$TARGET" "$SOURCE"]
```
And then in your sile document:
```
\img[src=mymusic.ly]
```

alerque:
This is some kind of genius. I'm definitely interested in seeing this smoothed out and developed a little further. A lilypond package that works something like `lilypond-book` does for TeX is high on the list of things I want to accomplish, but this approach alleviates the need for some packages altogether and can allow packages like the one I have in mind to focus on more specialized tasks (e.g. in the case of lilypond that would be splitting the included content into systems so it can break across pages rather than just dropping in a whole PDF).

simoncozens:
Thanks very much for this, and sorry for sitting on it so long.

PR #71, opened 2015-04-22
Ignore dirty state of the libtexpdf submodule

khaledhosny:
There mere act of running ./bootstrap.sh make the submodule dirty
(regenerating auto generated files and so on), which is uninteresting so
better just ignore it.


simoncozens:
Thanks very much, merged!

Issue #77, opened 2015-04-27
\thisframeRTL from bidi package not working across page boundaries


deepakjois:
Here is the source for an article in Urdu that spans across multiple pages.

https://gist.github.com/deepakjois/79e89961b60de2e61e44

Note the call to `\eject\par` followed by `\thisframeRTL` in the middle. If I remove that, the next page is typeset LTR.

simoncozens:
That's by design - `\thisframe...` really does mean *this* frame. What we probably need is an option on the document class (perhaps the `plain` class) which declares all its default frames to be RTL. Would you be interested in trying to implement that?

deepakjois:
Yes, sure I could give it a try.

simoncozens:
I've just implemented this in the simplest way possible with 97aedbe

Issue #76, opened 2015-04-27
BIDI algorithm not working properly?


deepakjois:
Here is a sample document. The numbers within the script should be formatted LTR, as per Unicode BIDI rules, but they are not.

I did notice that in the example docs (like showoff.sil) the numbers are explicitly formatted LTR using a `\font` directive to wrap them. But I believe that should not be required. Any idea what is causing this?

```
\begin[papersize=a4]{document}
\script[src=packages/bidi]
\thisframeRTL%
\font[size=18pt]
سعادت حسن منٹو اردو کے ایک جانے مانے افسانہ نگار ہیں۔ وہ پنجاب کے لدھانہ ضلع میں  1912 میں پیدا ہوئے۔ ان کے ماں
 باپ کشمیری مسلمان تھے۔ وہ علی گڑھ مسلم یونیورسٹی سے تعلیم حاصل کر نے کے بعد انجمن ترقی پسند مصنفین ہند کے
 ساتھ جڑے۔ 1936 میں وہ بمبئی شہر جا کر بس گئے جہاں انہوں نے اپنے کئی مشہور افسانے اور فلم کے سکرپٹ لکھے
۔1947 میں بٹوارے کے بعد وہ لاہور چلے گئے۔ وہاں ان کی ملاقات فیض احمد فیض، سید ناصر رضا كاظمی
 اور احمد ندیم قاسمی‎ سے ہوئی اور وہ اخباروں کے لئے مضمون لکھنے لگے۔ 1955 میں ان کا انتقال ہوا۔
\end{document}
```

simoncozens:
Not sure what you mean; here is what I see:
![bug-76](https://cloud.githubusercontent.com/assets/106728/7341717/fa93dfa0-ece1-11e4-9ccc-4a9445b04078.png)

Although I set the font family explicitly. (`\font[size=18pt,family=Scheherazade]`)

I think what might be happening is that SILE has a concept of font direction and frame direction. You've set the frame direction to RTL, which is fine, but if you're still using the default font, (Gentium, English) then the expected output is LTR. So things might be getting reversed because of that.

simoncozens:
More to the point I wonder how, without explicitly selecting an Arabic font, you're getting any arabic output at all. Could you run with the latest repository and the command line option `--debug=fonts`?

deepakjois:
Sorry, I have a typo above. Let me try to explain again with the code.

[...]

PR #78, opened 2015-04-27
 \setLTR and \setRTL support

deepakjois:
This is a first attempt at adding support for bidirectional text across multiple frames (see #77) . It seems to have at least one serious bug. Compare [intended output][1] to [actual output][2]. Notice the header on top of the second page just vanishes unexpectedly. I can’t quite figure out why that is the case.

[1]: https://dl.dropboxusercontent.com/u/953/site_assets/urdu_sile.pdf
[2]: https://dl.dropboxusercontent.com/u/953/urdu_assg.pdf

Some other points for discussion:

* The idea of using `\setLTR` and `\setRTL` for switching the direction of paragraphs is borrowed from TeX’s [bidi][3] package
* Given SILE is intended to be a next-gen typesetting package, bidi support should be built-in to one of the default document classes, probably `plain`. We should not need to import a package to get bidi support.
* There should be support for typesetting short RTL text inside LTR paragraphs as well (and vice versa), something equivalent to the TeX bidi pacakages `\LRE` and `\RLE`

[3]: https://www.ctan.org/pkg/bidi?lang=en

Please let me know what you think of this approach, or if you have any alternative suggestions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ ltmain.sh
 subfiles.mk
 
 sile
+config.log
+core/justenoughharfbuzz.so
+core/justenoughlibtexpdf.so
--- a/classes/plain.lua
+++ b/classes/plain.lua
@@ -12,6 +12,14 @@ plain.endPage = function(self)
   return SILE.baseClass.endPage(self)
 end
 
+plain.newPage = function(self)
+  local f = SILE.baseClass.newPage(self)
+  print("setting direction to "..SILE.scratch.bidi["currentDir"])
+  f.direction = SILE.scratch.bidi["currentDir"]
+  f:newLine()
+  return f
+end
+
 SILE.registerCommand("noindent", function ( options, content )
   SILE.settings.set("current.parindent", SILE.nodefactory.zeroGlue)
   SILE.process(content)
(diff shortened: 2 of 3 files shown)


deepakjois:
Ok. I made another commit, where I try to set the frame direction much earlier than when `newPar` is called. I also moved the methods to the `plain.lua` class.

However, I am getting some weird output again. The x-offset of some nodes seem to be way off. [Here][1] is how the doc looks like now. Compare with expected (linked in previous comment) output and see how the document header disappears, and the header on the second paragraph is not center aligned.

It seems that there might be some problem in the `nodefactory.lua` methods where some `moveX` commands are being issued, based on the frame direction. Could you explain what is going on there. That will help me debug further.

[1]: https://dl.dropboxusercontent.com/u/953/urdu_assg_1.pdf

deepakjois:
I made another commit. This fixes the centering on the first page, by setting `cursorX` properly. However, I am still not able to figure out why the heading on the second page does not center.

deepakjois:
Now that #76 is fixed, I just wanted to bring back focus to this issue. The problem here seems to be that setting the direction in the `newPage` hook is a bit too late for all the X and Y alignments to happen properly. I can’t think of a good way to maintain directionality smoothly when switching from one frame to another.

simoncozens:
OK, things have now changed (see #83) and the document has a base direction set in `<document class>.pageTemplate.direction`. So although this patch will need to be redone it should be very easy now. I'll discuss the other points in #83.

PR #87, opened 2015-06-05
Unicode PDF bookmark

shirat74:
Convert PDF bookmark string from UTF-8 to UTF-16BE as specified in PDF
spec.

### DIFF
--- a/packages/pdf.lua
+++ b/packages/pdf.lua
@@ -3,6 +3,62 @@ if SILE.outputter ~= SILE.outputters.libtexpdf then
 end
 local pdf = require("justenoughlibtexpdf")
 
+-- Added UTF8 to UTF16-BE conversion
+-- Compatibility UTF8 decoding function implemented in Lua
+if (not pcall(function() utf8 = require("utf8") end)) then
+  local function utf8_code(ustr)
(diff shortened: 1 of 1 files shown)


simoncozens:
Thanks very much for this. Just a couple of questions: is `utf8.code` the same as `SU.codepoint` from `utilities.lua`? (Get Unicode codepoint for UTF-8 string) I would also prefer to see UTF8-to-UTF16 conversion be implemented as a utility function in `utilities.lua`. 

By the way, thanks for your work translating the manual. It looks great! I am currently working on JIS X 4051 support (including 縦書き), so I am sorry if everything breaks while I am working on that! Any help or advice would be appreciated!

I also have started a Japanese language users group: https://groups.google.com/forum/#!forum/sile-users-ja

shirat74:
It does the same thing. I haven't noticed that there is a code for that...It's obvious though...

The implementation is quite different as I wanted to make it compatible to Lua 5.3's utf8 functions and not to rely on bitwise operations. 

It's great to hear that you are working on JIS X 4051 support.  I will check it later. I'm not expertise in typography but I think I can help on the part of PDF output.

PR #88, opened 2015-06-08
Remove `mod` and `getn` dependency and add LuaJIT and Lua5.3 support

pkulchenko:


simoncozens:
Thanks, that's great!

siffiejoe:
There still seems to be a hidden dependency on the "bit" library (LuaBitOp probably) in local function `to_bits`.

pkulchenko:
@siffiejoe, you are right; it needs to use `bit_not` instead; I haven't changed/noticed that part, so it was this way originally. SILE doesn't use that method, so should be fine, but technically, the lib needs to be updated.

PR #93, opened 2015-06-18
Cleanup trailing whitespace

alerque:
I know some programmers don't care much about this, but trailing whitespace really bugs me. Even in languages where it doesn't make any semantic difference it just gets in the way when you're copying stuff around and clutters up diffs when you got to review. As I started poking around the code I started to run into quite a few instances.

Rather than clean it up on a case by case basis I find it best to do it all at once. This way little fixes here and there don't end up creeping into future diffs and cluttering up the code review process. Also by batching them you easily verify this whole PR shouldn't break anything by running `git diff --ignore-all-space alerque/feature/normalize-whitespace...master`.

I split this into two commits because one of them touches files that were clearly automatically processed. By having that set in a separate commit they could be reverted in order to get a clean comparison if that set was ever re-imported. Just makes the potential git magic that much easier to do.


simoncozens:
Thank you very much!

PR #94, opened 2015-06-18
fix Turkish hyphenation rules

alerque:
The white paper published in TUG on Turkish hyphenation specifically
mentions the special -ecek case and similar suffixes in the context of
dealing with trailing two letter suffixes. The code, however, only
matches the two letter variants with a special case for -ecek, but
overlooks -acak which is a grammatically identical variant with the same
issues.

http://www.tug.org/TUGboat/Articles/tb09-1/tb20mackay.pdf

I also submitted this tweak to the author of the upstream LaTeX package
on CPAN where these rules originated. I don't know how slow moving that
fix will be or if there will eventually be a re-import or not, but until
then I don't see a reason not to patch it here too.

### DIFF
--- a/languages/tr.lua
+++ b/languages/tr.lua
@@ -34,7 +34,8 @@ SILE.hyphenator.languages["tr"].patterns =
 "1v1",
 "1y1",
 "1z1",
--- prevent e-cek at end of word
+-- prevent a-cak/e-cek at end of word
+"2a2cak.",
 "2e2cek.",
 -- prohibit hyphen before pair of consonants
 -- many pairs generated here are impossible anyway


simoncozens:
Thanks! I'm not planning on doing a reimport from LaTeX - one of the goals for the next release is improving the way that language-specific typesetting information (not just hyphenation, but line breaking and other typographic conventions) gets encoded, which is the reason for the Japanese madness right now. I will probably stick with the hyphenation patterns idea from TeX, but there will be lots of other stuff added as well that means the language support files will diverge from TeX.

alerque:
I follow you. It would be nice to be able to 'track' language related packages out of TeX at least to some extent as lots of little issues like this surface and get fixed across a broad spectrum of languages. In the years I've been using it I've seen even English language edge cases get fixed. As SILE is somewhat behind the adoption curve it would be nice to take advantage of the number of eyeballs on documents that the TeX world offers. But I wouldn't put that very high next to the need to being able to implement the backend in a more modern way—perhaps that means we just need to keep a weather eye out on diffs between TeX language releases and periodically manually port any relevant items.

PR #96, opened 2015-06-18
update ignore list with all relics of current build system

alerque:
This is some minor housekeeping but the git ignore rules aren't catching some of the build relics. This just makes it a touch easier to keep track of what is going on during development.

I noticed somebody else had started to fix this as well so I kept their commit to credit them. Incidentally their [other in-progress feature branch](https://github.com/deepakjois/sile/tree/exp) could then be rebased for an easier comparison of what is being contributed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 tests/*.actual
 tests/*.pdf
+tests/*.toc
+examples/*.pdf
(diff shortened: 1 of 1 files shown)


simoncozens:
Looks good, thank you!

PR #97, opened 2015-06-18
Touchup instalation script and readme

alerque:
These are some relatively minor touchups to the build process mitigating points that confused be and hopefully making it easier for others to follow the process. Please see commit messages for more detailed explanations behind each change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 [![Build Status](https://travis-ci.org/simoncozens/sile.svg?branch=master)](https://travis-ci.org/simoncozens/sile)
+[Program exited, press any key to close]
 
 ## What is SILE?
 
@@ -39,12 +40,14 @@ You also need to install the following Lua libraries using [luarocks][] (downloa
 * `luafilesystem`
 * `lgi` (required for Pango-Cairo only)
 
-Once your dependencies are installed, run
+If you are building from a a git clone, start by running the script to setup your environment (if you are using the source tarball this is unnecessary):
 
-* `./configure ; make install`
+    $ ./bootstrap.sh
 
-(If you are running SILE from git, first run `bootstrap.sh` and then
-configure and make.)
+Once your dependencies are installed, run:
+
+    $ ./configure
+    $ make install
 
 This will place the SILE libraries and executable in a sensible location.
 
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,6 +1,14 @@
 #!/usr/bin/env sh
 
-git submodule update --init --recursive --remote
+# We neet a local copy of the libtexpdf library to compile. If this was
+# downloaded as a src distibution package this will exist already, but if not
+# and we are part of a git repository that the user has not fully initialized,
+# go ahead and do the step of fetching the the submodule so the compile process
+# can run.
+if [[ ! -f libtexpdf/configure.ac && -d .git ]]; then
+    git submodule update --init --recursive --remote
+fi
+
 autoreconf --install
 (cd libtexpdf; autoreconf -I m4)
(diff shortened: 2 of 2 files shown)


simoncozens:
The first patch looks fine. The second one, I think, misunderstands the point of the bootstrap script. End-users shouldn't need to run it, because the distribution will ship working copies `./configure` and other generated automake dependencies. The point is to avoid having autogenerated code in the repository, and also to make things easy for end-users. (so they don't have to have the right - or even any - versions of autotools installed) The source releases won't contain `bootstrap.sh`.

alerque:
@khaledhosny [Debian for one](https://wiki.debian.org/Autoreconf).

> Autoreconfing on build is good practice in Debian. There used to be good reasons to stick with the autotools files shipped in the upstream tarball (mostly incompatibility with newer versions), but these days good forward compatibility and the advantages of getting the latest updates for new architectures or other fixes means that autoreconfing on build should be the default in packages.

I believe Arch Linux recommends it too but I can't find where right now. I've also seen it done in Gentoo packages (but am unaware whether it is recommended or not). It's only pointless if you system is sufficiently similar to the one on which the source package is prepared for it to effectively be a noop, but can actually serve a purpose on across divergent systems.

I would actually lean towards ditching autoreconf (and the bootstrap.sh script) altogether in favor of not having some of this crud in the source package and making the source package just a tarball of the repo (and needing to run `autoconf` or `make configure` or whatever) but that is a subject for another issue.

alerque:
@simoncozens See my other comment a second ago, but I do (sort of) get what the issue is.

I thoroughly agree about not having auto-generated code in the repo. No argument from me there. Where I would differ is that I'm totally fine with requiring people that are going to build the source package to have autotools installed as a build dependency. As an upstream package manager for several disros running (or re-running) those tools is standard practice anyway, and most users will be using their distro's packages eventually anyway. In the case of compiling manually, having your distro's -devel package set installed is pretty much a given anyway in most cases. My preference would be not to include any of the baggage related to that in either the repo _or_ the source package and just expect users to follow a standard compile routine. One major advantage of this is that you can distribute the source packages just by snapshotting the repo (or letting Github do this for you) with no extra work involved.

That all being said—I've now voiced my preference but I'm not going to argue it further. If that's the direction you want it to head I will adapt my second commit and fix the docs.

simoncozens:
Let me think about it for a bit. :-) The real factors for me are (a) getting the subrepository bit working, (b) impact on Windows users - @pkulchenko has noted some annoyances getting the configure script correctly generated on mingw32, (c) my personal traditional bias towards `./configure && make` as what you do when you download software. If we can overcome all of those it might be OK...

alerque:
I hadn't considered the Windows situation. Frankly I work pretty hard to live in a bubble where Windows does not exist. However I do see an advantage to being able to use the same source tarball as the canonical "source" package for both platforms instead of going through the two processes of letting the repo snapshot be the source for one platform but having to monkey with a second one for Windows anyway.

Once you consult with whoever is actually using Windows and decide on a direction you want it to go do ping me here and I will rewrite the relevant commit to adapt to whatever the decision is.

khaledhosny:
I never had to auroreconf a released tarball of an autotools-using package, and I believe if `make distcheck` passes it should just work (unless you are building on a brand new architecture not supported by the version of autotools used to prepare the package, but this rather an edge case that happens infrequently).

alerque:
I've rebased this on master again so it can be merged and tweaked what I did to the README (and code comments) to keep the status-quo for now.

alerque:
Rebased to dodge duplicate work that showed up in eea5d0c ...

Issue #95, opened 2015-06-18
Throw error message on missing fonts


alerque:
At the risk of moving towards the pointless verbosity of LaTeX compilers, I think SILE needs to throw an error to the console when a font specified in a document is not found on the system. Currently the console is silent on the matter and only reviewing the document and knowing what it should look like would tip anybody off to a problem.

Thankfully in some cases this is blanant. For example the `example/test.sil` looks like this when compiled on my system:

![test pdf_014](https://cloud.githubusercontent.com/assets/173595/8225769/30a64096-15a1-11e5-8133-09023bdcf983.png)

I've never heard of it, but something tells me that isn't _Frutiger LT Std_. Just saying.

As a side issue I'm curious how the callback system actually worked. That's the keycaps variant of Linux Biolinum there, but that font isn't in my usual desktop stack for anything. If you ask fontconfig what font to use, it gives a pretty plain fallback:

    # fc-match 'Frutiger LT Std'
    DejaVuSans.ttf: "DejaVu Sans" "Book"

My suggestion would be to throw an error along the lines of 

    # sile examples/test.sil
    This is SILE 0.9.2
    <examples/test.sil><examples/macros.sil>
    Warning: Font 'Frutiger LT Std' not available, falling back to 'Linux Biolinum Keyboard O'
    [1] [2] [3] [4] [5] [6] [7] [8] [9] [10] [11] [12] [13] [14] [15] [16] [17] [18] [19] [20] [21] [22] [23] [24] [25] [26] [27] [28]

simoncozens:
Your wish is my command!

alerque:
Wow, spiffy. I will stop hunting for where that would be done. Eventually my lua will be up to speed to make meaningful contributions.

pkulchenko:
Simon, it turned out that font family is not enough as `Bodoni MT` seems to be substituted for many (absent) fonts on my (Windows) machine and this leads to font overlaps (as the font resolver loads `Bodoni MT Poster Compressed`, but the family is reported as `Bodoni MT`):

```
[fonts] Resolved font family Frutiger FT Std -> C:/Windows/fonts/BOD_PSTC.TTF
[fonts] Resolved font family Inconsolata -> C:/Windows/fonts/BOD_PSTC.TTF
[fonts] Resolved font family Monaco -> C:/Windows/fonts/BOD_PSTC.TTF
[wxlua-output]  Set font        Frutiger FT Std;20;200;normal;normal;    => Bodoni MT
```

[...]

PR #114, opened 2015-06-22
Added font fullname to face fields (to include font style).

pkulchenko:
Simon, per the discussion in #95, this adds `fullname` fields to the `face` result for the font substitution, which allows to properly load the font (to use with alternative viewers) as `family` only returns the name, but doesn't return the style.

From the documentation:

```
 fullname       FC_FULLNAME            String  Font face full name where
                                               different from family and
                                               family + style
```

### DIFF
--- a/src/justenoughharfbuzz.c
+++ b/src/justenoughharfbuzz.c
@@ -61,7 +61,7 @@ void calculate_extents(box* b, hb_glyph_info_t glyph_info, hb_glyph_position_t g
 
 int face_from_options(lua_State* L) {
   FT_Face face;
-  FcChar8 * font_path;
+  FcChar8 * font_path, * fullname, * familyname;
   FcPattern* p;
   FcPattern* matched;
   FcResult result;
@@ -145,15 +145,22 @@ int face_from_options(lua_State* L) {
   FcPatternGetInteger(matched, FC_WEIGHT, 0, &weight);
 
   /* Find out which family we did actually pick up */
-  if (FcPatternGetString (matched, FC_FAMILY, 0, &family) != FcResultMatch)
+  if (FcPatternGetString (matched, FC_FAMILY, 0, &familyname) != FcResultMatch)
     return 0;
+  if (FcPatternGetString (matched, FC_FULLNAME, 0, &fullname) != FcResultMatch)
+    return 0;
+
   lua_newtable(L);
   lua_pushstring(L, "filename");
   lua_pushstring(L, (char*)font_path);
   lua_settable(L, -3);
 
   lua_pushstring(L, "family");
-  lua_pushstring(L, (char*)(family));
+  lua_pushstring(L, (char*)(familyname));
+  lua_settable(L, -3);
+
+  lua_pushstring(L, "fullname");
+  lua_pushstring(L, (char*)(fullname));
   lua_settable(L, -3);
 
   FcPatternDestroy (matched);
--- a/src/justenoughharfbuzz.c
+++ b/src/justenoughharfbuzz.c
@@ -147,8 +147,6 @@ int face_from_options(lua_State* L) {
   /* Find out which family we did actually pick up */
   if (FcPatternGetString (matched, FC_FAMILY, 0, &familyname) != FcResultMatch)
     return 0;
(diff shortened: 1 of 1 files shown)


alerque:
@pkulchenko Out of curiosity were you able to replicate the error Travis CI ran into locally?

	### Compiling examples/showoff.sil
	This is SILE 0.9.3-unreleased
	<examples/showoff.sil>
	Error detected:
	core/harfbuzz-shaper.lua:28: attempt to index local 'face' (a nil value)
	The command "perl tests/attack.pl" exited with 1.

Just wondering if this is a problem with the tests or with something showing up in edge cases that your changes broke.

pkulchenko:
@alerque, I've seen this error before, but I don't think it's related to the changes I made. I have two different configurations of fontconfig and one of them doesn't have any config; this is what I get when I run the example with that configuration:

```
<examples\showoff.sil>Fontconfig error: Cannot load default config file

Error detected:
core/harfbuzz-shaper.lua:28: attempt to index local 'face' (a nil value)
```

However, when I run with the complete fontconfig setup, I don't have any issues and everything works as expected. I also tried to complete roll back the changes and I get the same error, so it doesn't seem to be related to what I'm doing.

It's strange that Travis reports this error, so there is still a chance that I'm doing something wrong. Have you tried to compile/run this branch locally?

pkulchenko:
@simoncozens, can you review the changes when you get a chance? They are quite simple and I'm not sure why the build would be failing as it compiles and runs for me locally without any issues (and I can get the information I need based on the changes).

simoncozens:
I'm getting the error and I think I see why. If the match totally fails then fullname is empty and so `return 0` happens without returning a table to the Lua stack. Here's more debugging output:

```
[fonts]	Resolved font family Gill Sans -> /Library/Fonts/GillSans.ttc
Gill Sans;10;200;normal;normal;;LTR
[fonts]	Resolved font family Gill Sans -> nil

! Could not find requested font {direction=LTR,features=,font=Gill Sans,language=en,script=,size=10,style=normal,variant=normal,weight=200} or any suitable substitutes at examples/showoff.sil
```

simoncozens:
OK, the match *is* working and finding Gill Sans Light, but the call to get `FC_FULLNAME` is failing. I still don't understand why that call is failing but rewriting it as

```C
  if (FcPatternGetString (matched, FC_FULLNAME, 0, &fullname) != FcResultMatch) {
    lua_pushstring(L, "fullname");
    lua_pushstring(L, (char*)(fullname));
    lua_settable(L, -3);
  }
```

does the right thing.

pkulchenko:
Ok; that explains the error. Strange though as it seems like it may match `familyname`, but not `fullname`? I was about to suggest the change, but you seem already implemented it ;). Do you want me to update the pull request?

simoncozens:
Yes please!

simoncozens:
Also: I strongly suspect the fact that Gill Sans is a TTC file is the problem here. But you have `index` so you can choose the right font within the collection.

pkulchenko:
`index` won't help in my case as wxwidgets doesn't give me access to this information, but I can always do `face.fullname or face.family`.

Pushed the changes...

pkulchenko:
It passes both Travis and my own tests, so we should be good with the change.

pkulchenko:
Simon, I noticed one interesting issue. `fullname` reports one font that is not available on my system:

```
! Font 'Frutiger LT Std' not available, falling back to 'Agency FB'
[wxlua-output]	Ignored substituted font 'Agency FB Negreta' as it's missing
```

Note that the family name is `Agency FB`, but the full name is reported as `Agency FB Negreta`. I don't have such a font installed (and it doesn't load when I try loading it using the full name).

I fixed this in my output module by checking against the list of known fonts (and added a warning message as shown above), but I thought I should mention this. I checked the C code and it looks correct, but maybe `Negreta` is some leftover from a previous call in the buffer? I don't know where else it would come from...

simoncozens:
No, you're fine. It's this: https://bugs.freedesktop.org/show_bug.cgi?id=27765

pkulchenko:
ok; so it should go away with the next fontconfig update (or I just need to compile it from sources).

Issue #105, opened 2015-06-20
Paragraph indentation is lost after commands


pkulchenko:
This can be seen on this document:

```
\begin[class=book,papersize=129mm x 198mm]{document}
\script[src=packages/rules]
I have seldom heard him mention her under any other name.

\hrule[width=100mm,height=0.25mm]

More text here
\end{document}
```

`More text here` continues on the same line as the horizontal rule, but should start on a new line.

I've bisected it to this commit 269963ad, more specifically to `lineEndLineStartSpace` in `command =((P("\\")-P("\\begin")) * Cg(myID, "tag") * Cg(parameters,"attr") * V"bracketed_stuff"^0 * lineEndLineStartSpace)-P("\\end{")`.

simoncozens:
Urgh. Working out which spaces should and shouldn't be ignored after commands seems like it is beyond me. :-(

> `A\relax B`

should render as `A B`. But

> `A\relax`
> `B`

should render as `AB`, and (presumably) so should

> `A\relax `
> `B`

and possibly even

> `A\relax `
> ` B`

But 

> `A\relax`
> ` `
> `B`

should render as 

> `A`
> ` `
> `B`

(I think.) If this is correct, and someone could put the rule into words, we can write tests for it. At the moment I'm just too confused. :-(

pkulchenko:
I think the logic should be absolutely simple: `\command\s*EOL` should be removed. This means that for the purpose of formatting, if the line ends with command and some optional spaces, the command is removed (along with the spaces and the end-of-line) as if it never existed in the text.

I think it should handle nicely all the test cases you have in your example, but I tried to change `lineEndLineStartSpace` to `(lpeg.S(" ")^0 * lpeg.S("\r\n"))^-1`, which expresses what I'm looking for, but it still doesn't work. Any idea why?

pkulchenko:
I think I got it; this is the diff against the current master:

```diff
diff --git a/core/inputs-texlike.lua b/core/inputs-texlike.lua
index 4805208..fe7eb11 100644
--- a/core/inputs-texlike.lua
+++ b/core/inputs-texlike.lua
@@ -14,7 +14,7 @@ SILE.inputs.TeXlike.parser = function (_ENV)
   local list = lpeg.Cf(lpeg.Ct("") * pair^0, rawset)
   local parameters = (P("[") * list * P("]")) ^-1 / function (a) return type(a)=="table" and a or {} end
[...]

PR #115, opened 2015-06-27
Update documentation for various package managers and generally fleshing things out.

alerque:
I've been poking at the documentation along the way but was holding off on making a PR for all this stuff (largely related to installation procedures) until SILE actually hit Homebrew. That has now happened so I think it's time to try these on for size.

Note this includes updates to both the README and the Documentation source. It does not yet include the actual PDF yet, largely because it doesn't compile right using the HEAD version. See whitespace debacle in #105 and elsewhere.


simoncozens:
Beautiful, thank you!

PR #116, opened 2015-06-29
Clean up Travis recipie

alerque:
This is just some odds and eds to clean up the shell scripting involved in the Travis test runs. This should make it easier to maintain and easier to read the debug log and make sense out of what step fails when they do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,49 +4,37 @@ compiler:
 env:
   global:
     - LUAROCKS_BASE=luarocks-2.2.0
+    - HARFBUZZ_BASE=harfbuzz-0.9.26
   matrix:
-    - LUA_PKG=lua5.1 LUA_DEV=liblua5.1-dev LUA_VER=5.1 LUA_SFX=5.1 LUA_INCDIR=/usr/include/lua5.1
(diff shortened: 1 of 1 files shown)


simoncozens:
While you're messing, it might be good - perhaps as another option in the build matrix - to build lua 5.3 with compatibility-grandfathering turned off, (remove `-DLUA_COMPAT_5_2` from the Makefile) so we are prepared for whatever surprises the Lua community decide to spring on us in the next release.

alerque:
@simoncozens Good idea. I've added a test to the matrix for that.

I think I pissed off Travis running so many jobs. I'm stuck behind some queue now but when it does get around to testing this one last time I think I'm about done with this bit of hacking.

alerque:
I lied. I have a couple more tweaks to test when Travis starts minding me again.

alerque:
Alright Travis says green, I say green.

Things I learned today:

1. Debuging shell scripts inside Travis build scripts is way more of a pain than it should be. They are mucking around with every line and it screws up normal shell quoting, variable scoping and basic syntax.
2. Lua has changed the way they do a lot of things and the different versions are pretty different to test. Case in point: the output of `lua -v` is on STDERR for 5.1 and STDOUT for 5.2 and 5.3. At some point it might be necessary to write write different build tests for each version. Meh.

simoncozens:
Lovely, thank you!

PR #122, opened 2015-06-30
Update regression tests

alerque:
So this includes a whole bunch of changes to get the test suite gearing running a bit smoother. Here's the basic idea.

* Enable `--regression` testing for the build tests in Travis. If changes to the core change output of existing documents, there should be a reason for it. This will mean updating the expected outputs a lot more as things change, but that extra review work _should_ be done. I have some tooling I've been using locally for updating tests that I'll try to adapt for broader use and submit later.
* Update tests that have fallen behind. This includes a number of them that just didn't have the current `debug-output` format or that the default margins for the class or the way the folio was arranged changed for one reason or another.
* Add tests that were just plain missing.
* Reformat tests to A5 where possible. The smaller (and consistent) size is much easier to test. Besides the debug output I've been using visual PDF comparison tools (including one I wrote a while back for another project) that make it easier to scan for problems. When doing this on a whole bunch of tests (and across a whole bunch of commits with git-bisect!) its much easier if the size is small and consistent. Note I didn't change them all yet because it wasn't apparent to me how some of them might be affected by a layout change.
* Strip non-essential parts of the tests out. With the idea that each test should be checking for exactly one problem, I tried to eliminate as many extra factors and code as possible.
* Use only freely available fonts. So far these changes are only in `tests/*`, but I would like to be able to compile anything coming out of the core repository without needing proprietary fonts on my system. None of the items in the test suite seem to depend on the licensed fonts so I've replaced them with reasonable free ones that can be installed in the Travis environment as well as by anyone wanting them. The only exception to this is the OSX specific font issue. As this font format is only used on OSX and it's a default system font I left that one alone and marked the test to only run on Darwin (which means it won't be part of Travis's tests for now).

Most of these are fully tested in that I actually reproduced the errors running old versions of SILE and made sure the expected output was actually the desired thing. There are a couple exceptions (marked "fail state untested") when I couldn't figure out the fail state (or it looked like I didn't have a _working_ state).

Most of the commits have been isolated to individual tests. I squashed a few of them that were closely related, but if problems ever come up with the tests I wanted `git blame` and `git revert` to be useful so they are mostly test-specific.


alerque:
Oley oley! Happy dance—I see green! This should be ready for merge. Finally.

But @simoncozens, you might say, “It looks red to me?!” And you'd be right. But if you look closer you'll find it's actually _supposed_ to be red because there _actually is_ a regression issue between Lua versions. Lua 5.3 handles float math differently than 5.3 and `SILE.outputters.debug` needs to be fixed somehow so that the output of the debugger is deterministic. It doesn't really matter which way it is adjusted (although I'd be inclined to polyfill for 5.1 and 5.2 and keep the default behavior of 5.3) as long as it is fixed. The test suite itself is, for the first time, actually testing for regressions and expected to succeed. As soon as the math bug is fixed it should show green across the whole matrix.

As fixing the actual existing regression is something for the master branch rather than than the test code I'll leave that to be fixed in master or a different PR. This set of changes is, to the best of my knowledge, doing what it should be doing and merging it to master makes sense even with the red x because the issue it found actually exists in master.

simoncozens:
This is incredibly cool (and pretty scary). Thank you very much for all this work. I came up with the idea of regression tests when I was mucking about with the linebreaking algorithm - that is some very scary code and I didn't want to disturb it unintentionally - but I think this will also help us catch plenty of other problems along the way. (Or we could be drowning in a sea of red crosses for the near future. I don't know. Let's find out. :-) )

alerque:
The sea of red crosses is a definite possibility. I think it can be manged however. I have some shell scripts I wrote while setting this up that might help. One of them for example runs through regression tests, catches ones that fail, generates overlay images for visual review, pops up the diff and the overlay image and asks if it's okay, then regenerates the expected output if it gets an okay.

The debug output is pretty cool actually and I think as long as we pay attention to keeping it deterministic and have the right tooling on hand it shouldn't be too bad to manage. In any event it will be easier if we start now rather than trying to add this on farther down the road.

PR #126, opened 2015-07-02
update regression test to match a4ae37f

### LINKED ISSUE #125: Float math in debug output different between Lua versions
The output of `SILE.outputters.debug` is only useful for regression testing in so far as it is deterministic. Unfortunately a different in the way Lua versions handle floating point match is causing it to produce different output on Lua 5.3 than in 5.2 or 5.1.

As mentioned in #122 (and can be seen in the output of [this test run][test]), somehow in the case of numbers that work out to be whole integers, the print statement in Lua 5.3 is giving `n.0` decimal format number while the older Lua versions round it to `n`.

	### Regression testing tests/image.sil
	--- tests/image.expected	2015-07-02 05:47:32.076248986 +0000
	+++ tests/image.actual	2015-07-02 05:52:12.051692919 +0000
	@@ -11 +11 @@
	-Draw image	documentation/fig1.png	77.985571410771	29.76377985	50.0	69.242902208202
	+Draw image	documentation/fig1.png	77.985571410771	29.76377985	50	69.242902208202
	@@ -19 +19 @@
	-Draw image	documentation/fig1.png	77.988630333249	117.1531664332	36.104783599089	50.0
	+Draw image	documentation/fig1.png	77.988630333249	117.1531664332	36.104783599089	50

It doesn't really matter which way it is adjusted (although I'd be inclined to polyfill for 5.1 and 5.2 and keep the default behavior of 5.3) as long as it is fixed so that the output is deterministic. If the older Lua format is preferred, the `.expected` output of two tests will need to be updated after fixing the main issue `tests/image.sil` and `tests/balanced.sil` both exhibit this bug.

  [test]: https://travis-ci.org/simoncozens/sile/builds/69240823

alerque:
Fixes #125

This particular test went the opposite way as the other ones you fixed. In d0dd50f you forced five digit accuracy, in a4ae37f you rounded. Both make sense for the circumstances, but I had saved all the tests based on the default Lua 5.3 behavior. This updates the test to be the way you normalized the output.

### DIFF
--- a/tests/balanced.expected
+++ b/tests/balanced.expected
@@ -797,7 +797,7 @@ Mx 	213.88398
 T	88 74 79 70 84 76 74 79 84 15 1761	(wineskins.”)
 Mx 	170.62378
 My 	365.70583
-Set font 	Linux Biolinum O;18;800.0;normal;normal;;LTR
+Set font 	Linux Biolinum O;18;800;normal;normal;;LTR
 T	36 73 66 81 85 70 83	(Chapter)
 Mx 	239.65178
 T	20	(3)


alerque:
Green times 4, mmm yummy.

Issue #85, opened 2015-05-28
It's better to not include *.pdf in the repo


Passw:
@simoncozens There are some *.pdf files, which are the output of  ```sile *.sil```.
I think it's better to remove this *.pdf files, to take down the size of repo :)

simoncozens:
To be honest the PDFs are not all that big - the biggest is 1.6M. There was a King James' Bible in there which wasn't a good example of anything so I've removed it now, but I think it's important to have the PDF output in there so that (a) people can quickly explore SILE's capabilities before they try playing with it, and (b) they know what they should be getting when they do play with it. So I shall be keeping them!

PR #129, opened 2015-07-04
Reword documentantion for easy reading

alerque:
As I get time I've been reviewing the documentation. I keep coming across bits that sounded stilted of left me hanging so decided to just start fixing them.

1. These can wait for a merge if you are in no hurry, but I wanted to throw them out there in case anybody else starts working on docs as well.

1. Github's diff viewer isn't very good for this purpose. Anybody reviewing documentation changes might want to checkout the branch on their own and run:

        git diff --word-diff master..feature/documentation

   …for a much more useful view into what I changed. Alternatively if you are on this branch with the last commit (see next point) you can use:

        diffpdf =(git show v0.9.2:documentation/sile.pdf) documentation/sile.pdf

   …and set the compare mode to "Words" to highlight differences since the last released version of the manual. You can of course substitute a branch or commit name for that tag.

1. I don't know what you want to do about committing PDF's in-between releases. My inclination would be to not have it in the repo at all and only include a rendered copy as part of the release process, but that brings complications for people that do keep track of master but might not necessarily be able to compile the manual themselves to actually read the current status.

1. These are in chunks to make it easier to comment on commits or pick and choose the bits you like vs. anything you think I messed up.

   Feel free to merge this this branch without that last commit (or let me know and I'll remove it) and let me know how you want to handle PDF generation.


simoncozens:
Thanks very much for these!

alerque:
@simoncozens Any preference about how/when to include updated copies of the compiled manual in the repo?

simoncozens:
I think it is helpful to have at least some kind of PDF in the repository, but with the understanding that there may be layout "surprises" in it between releases. Normally the last thing I do when preparing for a release is to update the documentation with the changes, and so most of the time there won't be any new information in the docs, but I think it's important that if you download the software (and people *do* download from the repository, not just releases) that you should be able to read the documentation without any fuss. So I think there should be a PDF in there, updated as the content is improved.

alerque:
Sounds alright to me. Normally I would be on the side of not having the PDF's in the repo at all (ala #85) but this being a typesetting engine I think makes it a little different. The fact that the manual is the best place to read the inline doc snippets being put out by the lua packages that you may be working on or with during development is another argument for having an up to date-ish PDF in the repo.

PR #134, opened 2015-07-06
Package/pullquote

alerque:
I'm attempting to create a package that creates pull quotes formatted much like this:

pullquote

Most of the process was pretty easy, but I'm kind of baffled about how to position the quote marks.

I messed around with using frame tricks to break the frame up various ways, but never got it to quite work. At best the left one works right but I couldn't get the next paragraph to go back to normal flow correctly and the right mark completely stumped me.

It seems like I could also do a bunch of math and shove a collection of glues around to move in and out of the margin, but that seems like a hack.

What would be the best way to proceed conceptually from here? Note that each commit in this sit has a matching PDF so you can easily review the progress (even directly on Github by viewing each commit: ([block formatting](https://github.com/alerque/sile/blob/0bd13d9f987645cb62d35878783b5d410ce91732/examples/pullquote.pdf), [special quote marks](https://github.com/alerque/sile/blob/303e4684edb1d153a46ee0d81fff84cf35e59afd/examples/pullquote.pdf)).


simoncozens:
I'm getting more and more convinced that in almost all cases, frametricks is more trouble than it's worth. When creating new frames on the fly you have to be very careful not to mess up footnotes and insertions, which are making calculations based on existing frames; you have to be very careful to avoid problems with going onto the next page; it's very much a "now you have two problems" scenario. I think the way to do it would be either to work with a collection of glues (you shouldn't have to do much math) or to use zero-width hboxes.

simoncozens:
Looks nice. Are you happy with it? Good enough to merge?

alerque:
@simoncozens The documentation is broken, I'm fixing that right now so hold off a second. I'm happy with the output if you think the technique is reasonable.

alerque:
@simoncozens The example I have here looks great but something about how I did positioning of the quotes breaks if there is no author line. Before this positioning business the `\par` business worked and the typesetter went on with a new paragraph including an appropriate parskip distance if any. Now it doesn't do that. Any idea what might be up with that?

simoncozens:
I'll take a look.

alerque:
@simoncozens I found the problem.

alerque:
I think we're good now.

alerque:
Scratch that, that "fix" broke the alignment of the top side.

alerque:
For real this time. It works both ways now, but I don't really understand _why_.

simoncozens:
OK, I think this needs work. To me `typesetter:chuck()` is a warning sign that something has gone wrong, and I don't think it's wise to rely on cursor values while a command is executing. For example: if you're using the best-fit pagebuilder (which collects up *all* the lines in the document and then works out the best way to arrange the page breaks) then you really don't want commands to be flushing the typesetter's state, because then you won't have the whole picture to work with. And cursor values really "belong" to the output phase. (Yes, I'm aware there are some bad examples in the SILE Book, which I wrote before I realised their full significance.)

Here is how I would personally implement it:

```lua
SILE.require("packages/color")
SILE.require("packages/raiselower")
SILE.require("packages/rebox")
SILE.registerCommand("pullquote:font", function(options, content)
end, "The font chosen for the pullquote environment")
SILE.registerCommand("pullquote:author-font", function(options, content)
  SILE.settings.set("font.style", "italic")
end, "The font style with which to typeset the author attribution.")
SILE.registerCommand("pullquote:mark-font", function(options, content)
  SILE.settings.set("font.family", "Linux Libertine O")
end, "The font from which to pull the quotation marks.")

local typesetMark = function (open, setback, scale, color, mark)
  SILE.settings.temporarily(function()
    SILE.call("pullquote:mark-font")
    local setwidth = SILE.length.new({ length = SILE.toPoints(setback) })
    SILE.typesetter:pushGlue({width = open and 0-setwidth or setwidth })
    SILE.call("raise", {height = "-"..(scale -1) .. "ex"}, function()
      SILE.settings.set("font.size", SILE.settings.get("font.size") * scale)
      SILE.call("color", {color = color}, function ()
        SILE.call("rebox", {width = 0, height = 0}, {mark})
      end)
    end)
    SILE.typesetter:pushGlue({width = open and setwidth or 0-setwidth })
  end)
end

SILE.registerCommand("pullquote", function(options, content)
  local author = options.author or nil
  local setback = options.setback or "2em"
  local scale = options.scale or 3
  local color = options.color or "#999999"
  SILE.settings.temporarily(function()
    SILE.settings.set("document.rskip", SILE.nodefactory.newGlue(setback))
    SILE.settings.set("document.lskip", SILE.nodefactory.newGlue(setback))
    SILE.settings.set("typesetter.parfillskip", SILE.nodefactory.zeroGlue)

    SILE.settings.set("current.parindent", SILE.nodefactory.zeroGlue)
    SILE.call("pullquote:font")
    typesetMark(true, setback, scale, color, "“")
    SILE.process(content)
    SILE.typesetter:pushGlue(SILE.nodefactory.hfillGlue)
    typesetMark(false, setback, scale, color, "”")
    if author then
      SILE.settings.temporarily(function()
        SILE.typesetter:leaveHmode()
        SILE.call("pullquote:author-font")
        SILE.call("raggedleft", {}, function ()
          SILE.typesetter:typeset("— " .. author)
        end)
      end)
    else
      SILE.call("par")
      SILE.typesetter:chuck()
    end
  end)
end, "Typesets its contents in a formatted blockquote with decorative quotation\
      marks in the margins.")

return { documentation = [[\begin{document}

The \code{pullquote} command formats longer quotations in an indented
blockquote block with decorative quotation marks in the margins.

Here is some text set in a pullquote environment:

\begin[author=Anatole France]{pullquote}
An education is not how much you have committed to memory, or even how much you
know. It is being able to differentiate between what you do know and what you
do not know.
\end{pullquote}

Optional values are available for:

\listitem \code{author} to add an attribution line
\listitem \code{setback} to set the bilateral margins around the block
\listitem \code{color} to change the color of the quote marks
\listitem \code{scale} to change the relative size of the quote marks

If you want to specify what font the pullquote environment should use, you
can redefine the \code{pullquote:font} command. By default it will be the same
as the surrounding document. The font style used for the attribution line
can likewise be set using \code{pullquote:author-font} and the font used for
the quote marks can be set using \code{pullquote:mark-font}.

\end{document}]] }
```

alerque:
Thanks! I'll review when I get a chance. The logic for not using cursor position makes sense now that you explain it, I just couldn't make anything else work and that example in the manual was just so handy ;-)

alerque:
It looks good to be @simoncozens. I tweaked the math a little bit to keep the opening mark down farther and scale differently, but the layout is the same as your version. With a little help from the fix you have for `rebox.lua` in 40612d9 this should work alright now. I tested across page breaks too and it behaved as expected, which is more than I can say for most ways I've done this in LaTeX.

simoncozens:
Lovely. Let's do it!

PR #136, opened 2015-07-10
Use relative paths for includes

### LINKED ISSUE #91: The way relative paths are treated should be sanitized
The way relative paths inside .sil documents are treated leaves something to be desired. On the one hand I don't want to see SILE getting into the mess that LaTeX has by doing searches through a huge path stack, but on the other it does need a touch more intelligence than it has now.

Currently by default SILE uses the CWD for searches. This, for example, means that in order to compile the documentation you must be in the repo root folder and run `sile documentation/sile.sil` but in order to compile most of the things in the tests directory you must be _in_ the tests directory, or to compile the sbgnt example you must be in examples/sbgnt. That the CWD is going to matter is neither apparent nor consistent with the way most non-shell applications handle paths. In the case of shell scripts that have dependencies on other scripts is common practice to retrieve the script's location using variations of `basename` and reference all relative paths on the computed value.

**I propose that relative paths in .sil documents be treated as relative to FILE rather than to CWD by default.** This way I could be in the root directory and run `sile examples/*.sil` or be in the examples directory and run `sile *.sil` and achieve exactly the same thing even in the case of documents that import other documents.

I further suggest that the sile executable accept an include path search argument (in fact accepting it multiple times) that adds paths to a stack to look in if files are not found relative to the document being compiled. This would enable additionally flexibility at document compile time.

alerque:
This should fix #91 by first searching for includes and requires from the directory of the input file before moving on down the path chain. Additionally a few other odds and ends I ran into are touched up as well. See commit messages for details.

### DIFF
--- a/core/inputs-common.lua
+++ b/core/inputs-common.lua
@@ -9,6 +9,10 @@ SILE.inputs.common = {
       end
     end
 
+    -- Prepend the dirname of the input file to the Lua search path
+    local dirname = SILE.masterFilename:match("(.-)[^%/]+$")
+    package.path = dirname.."?;"..dirname.."?.lua;"..package.path
+
     if not SILE.outputFilename and SILE.masterFilename then
       SILE.outputFilename = string.gsub(SILE.masterFilename,"%..-$", "").. ".pdf"
     end
--- a/core/inputs-common.lua
+++ b/core/inputs-common.lua
@@ -9,8 +9,8 @@ SILE.inputs.common = {
       end
     end
 
-    if not SILE.outputFilename and SILE.masterFileName then
-      SILE.outputFilename = string.gsub(SILE.masterFileName,"%..-$", "").. ".pdf"
+    if not SILE.outputFilename and SILE.masterFilename then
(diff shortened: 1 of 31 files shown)


alerque:
@pkulchenko and @shirat74 it looks like both of you have [at least tried] running SILE under Windows. Any input on how to properly paths input from the command line per the comments above?

pkulchenko:
I think the mechanism you have should work just fine. As far as I understand, this is to add `foo/bar` to the search path when the command is for the file `foo\bar\file.sil`, right?

alerque:
@pkulchenko Yes, that was exactly my intention. An edge case I did think of is when including a drive letter. I seem to recall this working in Cygwin but I don't know what Lua's behavior is going to be. Are you able to confirm that Lua knows what to do with a path like `d:/path/with/forward/slashes`?

pkulchenko:
> Are you able to confirm that Lua knows what to do with a path like d:/path/with/forward/slashes?

Yes; it opens the file just fine. This is Lua/LuaJIT compiled with mingw and running outside of cygwin.

I use a similar method in my Lua applications (switching to '/' separator internally for all the paths the app generates) and haven't seen any problems with 5.1-5.3 running on Windows.

PR #140, opened 2015-07-24
fix bootstrap.sh script failing on sed

anthrotype:
the bootstrap.sh script return non-zero exit code, making it impossible to install sile from `--HEAD` using brew.

```
sed: -i: No such file or directory
```

Unlike GNU sed, on OSX the sed command requires that the `-i` (inplace) option be followed by an argument for the extension for the backup file.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -11,4 +11,5 @@ fi
 
 autoreconf --install
 (cd libtexpdf; autoreconf -I m4)
-sed 's/rm -f core/rm -f/' -i configure
+sed -e 's/rm -f core/rm -f/' -i.bak configure
+rm -f configure.bak


alerque:
This isn't quite right. BSD sed can in fact work without creating a temp file. But it is right in so far as I didn't get the argument order right for it to do the sensible thing.

The correct fix would be to make that line read:

    sed -i -e 's/rm -f core/rm -f/' configure

And of course removing the .bak thing will be unnecessary. 

From the BSD sed man page:

     -i extension
             Edit files in-place, saving backups with the specified
             extension.  If a zero-length extension is given, no backup
             will be saved.  It is not recommended to give a zero-length
             extension when in-place editing files, as you risk corrup-
             tion or partial content in situations where disk space is
             exhausted, etc.

The trick is to use the `-i` flag before another explicit flag so that it knows it doesn't have an argument. In that way it behaves differently that GNU sed which assumes the last arg is a file name ever if a flag was looking for an optional parameter.

alerque:
This isn't quite right. BSD sed can in fact work without creating a temp file. But it is right in so far as I didn't get the argument order right for it to do the sensible thing.

The correct fix would be to make that line read:

    sed -i -e 's/rm -f core/rm -f/' configure

And of course removing the .bak thing will be unnecessary. 

From the BSD sed man page:

     -i extension
             Edit files in-place, saving backups with the specified
             extension.  If a zero-length extension is given, no backup
             will be saved.  It is not recommended to give a zero-length
             extension when in-place editing files, as you risk corrup-
             tion or partial content in situations where disk space is
             exhausted, etc.

The trick is to use the `-i` flag before another explicit flag so that it knows it doesn't have an argument. In that way it behaves differently that GNU sed which assumes the last arg is a file name ever if a flag was looking for an optional parameter.

alerque:
Alternatively if you'd rather not go to the trouble of amending the commit and reposting it; you can [just open this PR instead](https://github.com/simoncozens/sile/compare/master...alerque:anthrotype/fix_bootstrap?expand=1) which has your commit but edited ready to go...

anthrotype:
ops.. you're right! :)
I shall close this PR and open the one you suggested.

PR #141, opened 2015-07-24
fix bootstrap.sh script failing on sed

anthrotype:
the bootstrap.sh script return non-zero exit code, making it impossible
to install sile from --HEAD using brew.

```
sed: -i: No such file or directory
```

Thanks @alerque for suggesting a better fix!

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -11,4 +11,4 @@ fi
 
 autoreconf --install
 (cd libtexpdf; autoreconf -I m4)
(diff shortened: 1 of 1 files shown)


alerque:
See discussion in #140...

simoncozens:
Thanks both!

alerque:
And with that, `brew install sile --HEAD` should be working again for peaple with the default toolchain.

Issue #144, opened 2015-08-03
Make papersize optional


alerque:
I think the sile file format should accept a `\begin{document}` declaration without requiring a papersize.

So far I've run into two issues with it being required:

1. Many of my documents are designed for multiple output formats. Sure I can hook with Lua and change it after the fact via an included script, but it doesn't make sense to be required to set something you know is going to get overwritten.
2. Other programs that output to the sile format are not always in the best position to choose a paper size. I'm specifically thinking of `pandoc` at the moment. I've written a basic write module for sile and while pandoc has an optional variable it _can_ pass for paper size, there is no default value being used in its internal format. I can of course code one in, but I think the typesetter is in a better position to choose what the default should be than the document conversion tool. Not setting a value at all unless the source document passes one seems like the best possible behavior.

To this end I think that document-classes would be the best place to put sensible defaults for the intended output and the typesetter itself should have an even more primitive default papersize (probably A4) if none is set in the document class. This way classes that are tailored to the needs of a region can also specify the most likely paper to be used in that region and documents with no settings and using agnostic classes fallback to something that still puts words on a page.

alerque:
This should be complete as of the merger of #145.

PR #145, opened 2015-08-03
Set default papersize to A4

alerque:
See issue #144 for _why_ I started on this.

As for implementation, so far this only does the final fallback step.

I tried setting something like `plain:declareOption("papersize", "a4")` in the plain class file but got unexpected results. I'm not sure yet how to code that part.

 - [x] ~~Set sensible defaults in document classes~~
 - [x] Set sensible default fallback for any clases that don't specify
 - [x] Update documentation


alerque:
The remaining todo items on this have been finished and I rebased it on master. Commit e97b2f8 actually removes a default rather than adds one as in my first todo item, but I couldn't convince myself that any of the default classes needs an override for this. The nearest miss is the "book" class which could conceivably default to quarto or some other common book size, but I rather think that keeping the A4 default across the classes will make it easier for people to understand what _is_ changing (e.g. the page layout changes) without also changing other parameters to arbitrary values that are likely not the final values anyway.

alerque:
Any feedback on this one?

alerque:
Rebased to apply cleanly on master.

simoncozens:
Thanks very much for this - lots of hard work!

PR #147, opened 2015-08-07
Correct path to packages in test

alerque:


simoncozens:
Urgh, there is a bigger issue here. It looked like the regression test passed even though the document didn't successfully compile:

```
$ perl tests/attack.pl --regression tests/space-after-command.sil
### Regression testing tests/space-after-command.sil
This is SILE 0.9.3-unreleased
$
```

alerque:
Say what‽

I'll have a look. My question was how you even managed to play with this and generate the regression test in the first place with this in there.

simoncozens:
I initially generated the expected output from a very simple version with lots of newline-eating comments and so on, and then I made the test more complex by removing the comments and adding the `\script` call to make sure it still produced the right output. I didn't realise that the `\script` call was causing the compilation to die, because our test framework didn't report that. Oops.

alerque:
The regression test fails for me...but that's because I'm actually testing the exit code. Your sample running it from the shell is not.

    ./tests/attack.pl --regression tests/space-after-command.sil || echo FAIL

That will show that it failed. The actual vs. expected test is never getting run and you aren't seeing the error message from SILE because perl is eating it, but it is returning the correct exit code.

alerque:
It looks like the issue here is that SILE is dumping the error to STDOUT rather than STDERR.

PR #148, opened 2015-08-07
Write error messages to STDERR channel

alerque:
See discussion in #147

This should pass through errors to the terminal rather than being caught
by IO redirects. Notably running the regression tests will show errors
on the console rather than them ending up in the test output file.


simoncozens:
Sorry, I beat you to this...

PR #149, opened 2015-08-07
Fix tests

alerque:
Some of this is generic, some of it is specific to the whitespace handling discussed in #105. All but the last commit should match the current state of the code in the master branch.

The last commit has a test for the intended behavior that doesn't pass yet and I'm including it here in the spirit of test driven development ;-)


alerque:
One green one red. Gotta love it when tests run as expected.

simoncozens:
I've now moved to a more Harfbuzz-y way to measure and scale glyphs, which should also increase the accuracy. The downside of this is now the tests are now all failing; I think the only one which shouldn't is your last one, `process-commands`. Could I have another update to the expectations please?

alerque:
Sure. I'm on mobile only today but I'll be at a computer tomorrow and review the whole test suite and update the expectations. Its probably about time I add some of my other PDF review scripts to the test suite too. I guess we'll see how ambitious I get.

PR #150, opened 2015-08-08
Expand test case for whitespace-around-commands

alerque:
See issue #105

Add two new test cases:

 1. The case of \em vs. \font is currently handled differently and one
    eats indentation at the start of a paragraph and the other doesn't.

 2. A command trailing a line and another leading a new line does not
    insert a space between sequential words as expected.

Additionally I documented what is expected of the test a little better
for ease of manual review.


alerque:
These new test cases have been rebased on master and regenerated per issue #152. The test is still failing due to unresolved bugs is the features being tested, but the test itself should be valid again and could be merged any time.

PR #152, opened 2015-08-14
Update tests for new glyph measurement system

alerque:
As [mentioned](https://github.com/simoncozens/sile/pull/149#issuecomment-130439367) by @simoncozens in post-merger of #149, the tests need updating:

>  I've now moved to a more Harfbuzz-y way to measure and scale glyphs, which should also increase the accuracy. The downside of this is now the tests are now all failing; I think the only one which shouldn't is your last one, `process-commands`. Could I have another update to the expectations please?

This PR is a placeholder for those updated tests and putting Travis through the paces.

Todo:

* [x] Fix paths
* [x] Document which tests are not passing and why (see #155, #156)
* [x] Raise other issues for outstanding test failures (see #153, #154, #157, #158)
* [x] Squash commits for approved tests
* [x] <strike>Add scripts used to automate this to makefile?</strike>

Tests passing on my machine but not on Travis for unexplained reasons:

* [x] tests/arabic-scripts.sil (see #159)
* [x] tests/bug-132.sil (see #159)
* [x] tests/bug-76.sil (see #158)
* [x] tests/bug-79.sil (see #159)
* [x] tests/kannada.sil (see #159)
* [x] tests/vertical.sil (added CJK JP font to Travis)


alerque:
@simoncozens Do you have any particular attachment to the test runner being written in Perl?

simoncozens:
Not really, but what are you thinking?

alerque:
@simoncozens I was thinking that automating a bunch of shell commands is a really poor use case for perl (and I _like_ perl as a language so don't think I'm just ragging on it) and it would be a lot simpler and to do the same thing in bash. The bonus would be it would also be easier to integrate into the make environment so that we could get standard practice things like `make tests`. Not that it can't be done with perl too, but it will be an eyesore.

simoncozens:
I'm not sure about moving the test harness to shell. It seems like a move in the wrong direction to me - the code is starting to pick up more and more logic for running different kinds of tests, running named tests from the command line, reporting results, etc. I don't think it's fair to call it a wrapper around a bunch of commands any more, and I think keeping it in a higher level language will help us to maintain those features better than shell.

alerque:
I think you are wrong about appropriate us cases for shell scripting, but at the very least I won't mess with it as part of this PR. If/when I do fix that up I'll make it a independent PR and you can judge for yourself whether the change is for the better or not.

alerque:
Tests passing on my machine but not on Travis:

* [ ] tests/arabic-scripts.sil
* [ ] tests/bug-132.sil
* [ ] tests/bug-76.sil
* [ ] tests/bug-79.sil
* [ ] tests/kannada.sil
* [ ] tests/vertical.sil

alerque:
@simoncozens I think this PR (as well as #150) are ready to be merged. The remaining test failures seem to be unrelated to the main issue being addressed here and have all been reported separately.

Issue #159, opened 2015-08-14
Tokenizing differs between systems


alerque:
So my Linux and Mac test rigs are producing different outputs for the same (current master branch) version of SILE. The notable difference seems to be that tokening is happening differently at transitions between scripts.

For example when testing `tests/arabic-scripts.sil` on Linux I get:

    T       640 778 905     (ههه)

…while the same test on Mac produces:

    T  640 778 905 36  (ههه;)

Both systems claim to have packages installed for icu-55.1.

See also #152.

simoncozens:
Obviously one is wrong, so we will fix this, but could you check the output of configure on the Mac - is it picking up the icu library? It's keg-only on Homebrew so you have to do environment gymnastics to make it available to pkg-config.

alerque:
Looks like that may be the culprit:

> checking for ICU... no

As if our situation with Howebrew wasn't complicated enough.

Incidentally Travis is giving results similar to what I'm seeing on the Mac rather than my local Linux rig.

simoncozens:
I think I basically don't care about this. ICU gives complete support for folks that want that and SILE has built-in support for most stuff that works well enough, and if we had code which did everything that ICU does, there'd be no point supporting ICU.

alerque:
I follow the logic in that, but the non deterministic behavior bothers me. What happens when I prepare a document and send it to somebody else and they get a different result. I say "It works for me" and they are left floundering until we figure out that they didn't compile with ICU. Not so cool.

What if there was a way to either detect situations where this would make a different and throw a warning or to tag requirements in the document (i.e. a command that requires: graphite, icu, etc.) so that they would trip an error on systems without those features?

simoncozens:
At the moment the ICU/non-ICU output is visually compatible, and in fact I think `libtexpdf` will generate the same output because of the way that `set_string` maintains state and keeps the same PDF text operator open where possible for subsequent calls. Graphite is a bit trickier because when you need it, you really need it, and it will be visually incompatible. It also happens down within Harfbuzz, so isn't directly controllable by SILE. I can't imagine a *command* that requires Graphite/ICU - it'll just shape things different (better) if they are present.

I think the best I can do is add more code to `--debug=versions` to show whether ICU/graphite/etc. are being used. This is becoming our standard way of picking up this kind of thing: e.g. font version differences (which would have the same problem you outline - different output from the same document).

PR #160, opened 2015-08-14
Run Travis tests with and without ICU

alerque:
Default to with for most of the runs but do one run on Lua 5.2 without.

See issue #159.


simoncozens:
Except this isn't actually picking up ICU on build. Investing now...

simoncozens:
So this is why: pkg-config files weren't shipped by Debian https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=687339 or Ubuntu https://bugs.launchpad.net/ubuntu/+source/icu/+bug/1037588

alerque:
Bah.

Issue #162, opened 2015-08-15
Punctuation goes AWOL in language=cs


alerque:
As discussed in #161, there is a bug related to handling text in some languages whereby punctuation marks disappear without reason.

alerque:
Now you see it, now you don't: (I know you know what this looks like at the moment, the screen shot is for the sake of being able to understand what the _failed_ state looked like later when reviewing whether the test suite is behaving properly.)

![selection_045](https://cloud.githubusercontent.com/assets/173595/9288167/aa0dc86e-4341-11e5-8344-c46041adb028.png)

PR #166, opened 2015-08-15
Update expected test output affected by bidi bug

alerque:
See issue #162.

These tests were also effected by the bug fix to chunking and their
expected output changed a little.


alerque:
Looks like these fixes are, at least for the time being, obsolete due to other recent changes.

Issue #175, opened 2015-08-30
Bidi should not depend on frame direction


khaledhosny:
In this example:
```tex
\begin[papersize=a4]{document}
\font[language=ar,family=Amiri]
\thisframedirection[direction=RTL]
خالد
\end{document}
```
If I comment the frame direction line, the Arabic word is left aligned, which is correct, but the characters are reversed which is obviously wrong.

khaledhosny:
Seems like `has_mixed_direction_material()` is the culprit, since bidi algorithm will not run if returns false which it obviously do here.

simoncozens:
Yes, it should also check if there are nodes which are not the same directionality as the frame direction.

khaledhosny:
I think the fact that frame direction has any say in the reversal of characters is a bug in itself.

simoncozens:
OK, my current plan is to get SILE to always run the bidi algorithm, whether it's "needed" or not. This has the advantage of showing some bugs in the `reorder` function...

simoncozens:
OK, we now do bidi everywhere. Let me know of any Arabic/RTL bugs you spot.

PR #176, opened 2015-09-02
Always run the bidi algorithm

khaledhosny:
See #175.

### DIFF
--- a/packages/bidi.lua
+++ b/packages/bidi.lua
@@ -21,27 +21,6 @@ local bidi = require("unicode-bidi-algorithm")
 require("char-def")
 local chardata  = characters.data
 
-local function has_mixed_direction_material(v)
-  local linetext = {}
-  if not v.nodes then return false end
-  for i = 1,#(v.nodes) do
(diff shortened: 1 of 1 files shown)


khaledhosny:
Hmm, I need to check the failing files to see what went wrong.

simoncozens:
It's the crazy code which tries to combine the code back into tokens. These lines:

```lua
      and ( chardata[SU.codepoint(this.text)] and chardata[SU.codepoint(prev.text)] and (
        chardata[SU.codepoint(this.text)].linebreak == chardata[SU.codepoint(prev.text)].linebreak or
        chardata[SU.codepoint(this.text)].linebreak:match("^i") or
        chardata[SU.codepoint(prev.text)].linebreak:match("^i")
```

probably need to be removed. The idea was to paper over the differences between ICU's tokenisation and our non-ICU break iterator, (in `unicode.lua`) but it just makes the code more complicated.

khaledhosny:
I removed that block of code but it made no difference at all AFAICT.

PR #178, opened 2015-09-16
\nofolios should’t suppress showing progress

khaledhosny:
When set no page numbers are printed on the console, which doesn’t seem to a be an intended side effect.

### DIFF
--- a/packages/folio.lua
+++ b/packages/folio.lua
@@ -15,6 +15,7 @@ return {
         if SILE.scratch.counters.folio.off == 2 then
           SILE.scratch.counters.folio.off = false
         end
+        io.write("["..SILE.formatCounter(SILE.scratch.counters.folio).."] ");
         SILE.scratch.counters.folio.value = SILE.scratch.counters.folio.value + 1
         return
       end
--- a/packages/folio.lua
+++ b/packages/folio.lua
@@ -11,26 +11,24 @@ return {
   exports = {
     outputFolio = function (this, frame)
       if not frame then frame = "folio" end
+      io.write("["..SILE.formatCounter(SILE.scratch.counters.folio).."] ");
       if SILE.scratch.counters.folio.off then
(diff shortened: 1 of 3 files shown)


khaledhosny:
No idea why the tests are failing, there should be any layout related changes.

simoncozens:
The tests compare stdout...

khaledhosny:
But all the failures are Mx and My differences, and in files where `\nofolios` is not even used.

khaledhosny:
OK, so there are two kind of failures being shown, things like:
````diff
--- tests/masters.expected	2015-09-16 07:49:51.797254439 +0000
+++ tests/masters.actual	2015-09-16 07:55:25.282999232 +0000
@@ -217 +217 @@
-New page
+[1] New page
@@ -453 +453 @@
-New page
+[2] New page
@@ -679,2 +679,2 @@
-New page
-End page
+[3] New page
+[4] End page
````
Which I think I should just update the expectation, right?
And things like:
````diff
--- tests/process-commands.expected	2015-09-16 07:49:51.797254439 +0000
+++ tests/process-commands.actual	2015-09-16 07:55:26.219953559 +0000
@@ -5 +5 @@
-My 	37.33214
+My 	44.76378
````
Which I have no idea why they are happening.

simoncozens:
Yes, but these are on known-bad tests. They fail because SILE has bugs. :-)

khaledhosny:
I see, that got me confused, but I should have looked at the test results summary.

alerque:
@khaledhosny Yes the "expected" output can be updated with the change in standard output for the folios. For the ones with metric changes those need to stay the way they are. The tests are failing because sile is not behaving properly, but the expected test output has been manually tweaked to have the correct values. It's just a matter of bringing sile in line with what we've manually computed _should_ be expected.

khaledhosny:
I get this now, all good tests pass now.

simoncozens:
Thank you very much!

Issue #186, opened 2015-11-09
Cannot compile against Harfbuzz 1.0.6


alerque:
I just realized that my testing in #181 is faulty because I wasn't actually running the version of SILE I thought I was … because it hadn't actually compiled when I updated the repo … because at some point it stopped compiling against the latest Harfbuzz version. Testing this in Travis shows that bumping from 1.0.4 to 1.0.6 breaks SILE's make step. Relevant make output includes:

    fontmetrics.c: In function ‘get_typographic_extents’:
    fontmetrics.c:41:3: error: unknown type name ‘hb_font_extents_t’
       hb_font_extents_t metrics = {0,0,0};
       ^
    fontmetrics.c:41:34: warning: excess elements in scalar initializer
       hb_font_extents_t metrics = {0,0,0};
                                      ^
    fontmetrics.c:41:34: note: (near initialization for ‘metrics’)
    fontmetrics.c:41:36: warning: excess elements in scalar initializer
       hb_font_extents_t metrics = {0,0,0};
                                        ^
    fontmetrics.c:41:36: note: (near initialization for ‘metrics’)
    fontmetrics.c:44:3: warning: implicit declaration of function ‘hb_font_get_extents’ [-Wimplicit-function-declaration]
       hb_font_get_extents(hbFont, &metrics);
       ^
    fontmetrics.c:45:21: error: request for member ‘ascender’ in something not a structure or union
       ascender = metrics.ascender / (double)upem;
                         ^
    fontmetrics.c:46:22: error: request for member ‘descender’ in something not a structure or union
       descender = metrics.descender / (double)upem;
                          ^

alerque:
Git bisect turns up 9abada1 as the first bad commit for this issue.

behdad:
You probably need this:
https://github.com/behdad/harfbuzz/pull/165

Sile is using a patched HB, and the patch became stale after my changes.  Simon kindly updated the patch.  I'll try to finish and merge it tomorrow.

simoncozens:
Yep, and the ifdef should refer to the *next* version of Harfbuzz.

simoncozens:
OK, that commit should make the font metrics code fall back to FreeType.

alerque:
Confirmed working.

PR #187, opened 2015-11-09
Bump Harfbuzz version in Travis config to 1.0.6

alerque:
If desired merge this after a fix for #186 is in place to avoid Travis flashing red at us.

Also if you like I can split the build up to test across multiple versions of Harfbuzz if the fix for this could involve future breakage but I've left it at one for now pending better knowledge of what we are testing for and why.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ compiler:
 env:
   global:
    - LUAROCKS_BASE=luarocks-2.2.2
-   - HARFBUZZ_BASE=harfbuzz-1.0.4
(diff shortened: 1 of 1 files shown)


simoncozens:
I don't think there's a need for testing multiple versions of Harfbuzz. It's very rare that we'll require features from Harfbuzz that aren't released, and I'll try to supply fallbacks to FreeType where possible. So I don't think this is necessary. Let's wait until HB 1.0.7,  and use that opportunity to make sure the font metrics code is tested.

Issue #184, opened 2015-11-08
Class search path should include location of input file


alerque:
Trying to compile a sil file that is not in CWD is tricky if it references a special document class. The CWD is searched for `classes/<class>.lua`, but this is relative from the CWD. In many cases the desired path will be relative to the target sil file. That should be included in the standard set of paths searched.

deepakjois:
This caused a regression. Please check my comments in #181.

deepakjois:
I have temporarily fixed the regression like this:
https://github.com/deepakjois/sile/commit/5b346a6db7a430591424a8e17983dd0686e88df6

Issue #192, opened 2015-11-10
RTL typesetting broken when pagebreak happens


deepakjois:
Here is some code to typeset two Urdu poems. The first Urdu poem consists of 6 stanzas of 4 lines each. In order to replicate the problem, the last 3 stanzas are exactly the same. The second Urdu poem is short and just spans a single page.

```
\begin[papersize=a5,direction=RTL]{document}
\set[parameter=document.parindent,value=0pt]
\nofolios
\font[family=Scheherazade,style=Bold,language=urd,size=36pt]
\center{لکیریں}
\medskip
\set[parameter=document.baselineskip,value=0.75em]
\set[parameter=document.parskip,value=0.5em]
\font[family=Amiri,style=Regular,language=urd,size=14pt]
\begin{raggedright}
لکیریں کھینچیں ہیں کاغز پہ مگر\break
کہانی کویٔ بنتی نہیں\break 
کصّے تمام یہں میرے پاس مگر\break 
حقیقت کویٔ ملتی نہیں

کھڈے ہوۓ ہیں اس موڑ پر \break
کھو جانے کے لۓ تیّار\break 
نقشے سارے جلا دۓ ہیں لیکن\break 
گُمراہی ملتی نہیں

کب سے ہیں اِسی کوشش میں اب\break 
کہ نیند ہی بچاےگی\break 
سپنوں میں بھی نضارے لیکن\break 
سنہرے ملتے نہیں

کب سے ہیں اِسی کوشش میں اب\break 
کہ نیند ہی بچاےگی\break 
سپنوں میں بھی نضارے لیکن\break 
سنہرے ملتے نہیں

کب سے ہیں اِسی کوشش میں اب\break 
کہ نیند ہی بچاےگی\break 
سپنوں میں بھی نضارے لیکن\break 
سنہرے ملتے نہیں

کب سے ہیں اِسی کوشش میں اب\break 
کہ نیند ہی بچاےگی\break 
سپنوں میں بھی نضارے لیکن\break 
سنہرے ملتے نہیں
\eject\par
\font[family=Scheherazade,style=Bold,language=urd,size=36pt]
\center{سفر}
\medskip
\font[family=Amiri,style=Regular,language=urd,size=14pt]
سفر میں خود ہی گھایل ہو گیا ہوں\break 
میں اپنا راستہ روکے کھڑا ہوں

میرا گھر دور ھوتا جا رہا ہے\break
میں پیچھے کی ترف چلتا رہا ہوں

[...]

PR #194, opened 2015-11-12
Fix regression due to 820e8d9

alerque:
@deepakjois already mentioned this in #184, but this in some scenarios masterFilename is getting called before initialization. This is the same fix from Deepak's original commit (5b346a6) with the comment that goes with the code line moved to match.

I didn't look very hand to see if there is a better way to fix this, but it does need to get fixed.

### DIFF
--- a/core/sile.lua
+++ b/core/sile.lua
@@ -88,6 +88,8 @@ Options:
 
   parser:on ('--', parser.finished)
   _G.unparsed, _G.opts = parser:parse(_G.arg)
+  -- Turn slashes around in the event we get passed a path from a Windows shell
+  SILE.masterFilename = _G.unparsed[1]:gsub("\\", "/")
   SILE.debugFlags = {}
   if opts.backend then
     SILE.backend = opts.backend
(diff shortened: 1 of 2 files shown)


deepakjois:
Please note the comment I made in your code.

Looking at the fix again, I realised I wasn’t accounting for all the cases. Here is the complete fix: efdd055

alerque:
Updated again, this time with the comment fix added back in to your new adjustment.

alerque:
@deepakjois I think we did something wrong. The tests aren't passing. I tried looking at it but my brain is so full of Haskell right now (I've spent the day writing a SILE output plugin for Pandoc) that I can't handle Lua.

simoncozens:
Looks like it's not seeing a filename left at all.

deepakjois:
@alerque Just a typo in my commit. Please not the comment in the diff.

alerque:
Thanks again @deepakjois. I updated this PR again and we got a pass from Travis. (The ominous red × is actually on account of the new test for #192, the part this was failing on actually passed). I think we're good now.

simoncozens:
Thanks both!

PR #197, opened 2015-11-13
Unravel interdependency between packages

alerque:
The `url` package was providing `\url`, which in turn was calling
`\code`, but that wasn't defined (unless your document happened to do
so). The `markdown` package meanwhile did have a working definition for
`\code`. Re-arranging this means that loading any one of the packages
will actually get them each the commands they need to run.


simoncozens:
Thanks!

PR #201, opened 2015-11-14
Improve localization

alerque:
This moves English only output into the EN localization, adds a function for localizing TOC headers and adds localized strings for Turkish. Blank default strings rather than English seems like a better way to encourage properly setting the document language and adding localizations to all languages.

### DIFF
--- a/classes/book.lua
+++ b/classes/book.lua
@@ -87,7 +87,7 @@ end)
 book.registerCommands = function()
   plain.registerCommands()
 SILE.doTexlike([[%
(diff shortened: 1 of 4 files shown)


simoncozens:
I think we also need a patch to \chapter to call `\book:chapter:pre:{appropriate language}`.

alerque:
I have exactly that (`\book:chapter:pre:tr`) in my library right now, I just haven't ported that to be an upstream patch yet. I mixed and matched style and content and I need to split that before contributing it.

PR #205, opened 2015-11-17
Tweak current.parindent setting to reset itself automatically

### LINKED ISSUE #188: document.parindent setting not working at the beginning of document
Using sile at HEAD:

```
\begin[papersize=a4]{document}%
\font[family=Kohinoor Devanagari,language=hi]
\set[parameter=document.parindent,value=0pt]
\font[style=Light]
दीपक जोइस

\font[style=Medium]
दीपक जोइस

\font[style=Bold]
दीपक जोइस
\end{document}
```

Produces output:
screenshot 2015-11-10 10 07 01

I was expecting the first paragraph to not be intended as well.

deepakjois:
* Make `current.parindent` nil by default (this fixes #188)
* Reset it to nil once we have indented the current paragraph
* Remove redundant calls to reset `current.parindent` to zero, because now it is reset to nil immediately after a new paragraph is started.


deepakjois:
Please take a close look at the removal of calls to `current.parindent`. I removed these calls only in the cases that I thought it would be already nil.

deepakjois:
Don’t think the build failure is due to these commits.

simoncozens:
No, that's me messing something else up.

simoncozens:
Thanks very much!

PR #219, opened 2015-11-26
documentation/sile.sil: small fixes

akavel:
Those are attempts at some small fixes I tried to do while reading the manual. I'm not sure if all of them are right, so please review.

Especially, I'm not sure how `SILE.Length` (or `SILE.length`?) and `Glue` (or `SILE.Glue`? `SILE.newGlue`?) are intended to be named on lines ~1677-1678 (even after a cursory look in the code). But naming one in a different way than the other looks inconsequential.


simoncozens:
Thanks very much for this. It should be `SILE.length`, which I will fix. But `Glue` is interesting because you can't instantiate the object directly (except through `SILE.nodefactory.newGlue`) but the object's type is definitely `Glue`, which was the point in the documentation.

Issue #213, opened 2015-11-19
Inconsistent leading in footnotes.


alerque:
For some reason the first line (and first line only, not the first footnote insert) sometimes shows up with a much larger leader than its cousins. Note the extra leading where marked:

![selection_107](https://cloud.githubusercontent.com/assets/173595/11265536/94085106-8ea5-11e5-9d7b-d60e3870a9fc.png)

One is manifesting after the first line in a wrapped comment, the others ore in the inter-insert space. I haven't been able to find what triggers this, but when extra space is added it seems to be always after the first line. Note about 50% of pages are fine, but this is pretty frequent. This is possibly related to #212 but distinct in that it's never squished and always the first line.

My current footnote code (used in the example above) is in [this comment](https://github.com/simoncozens/sile/issues/211#issuecomment-157983989).

simoncozens:
Can you see if this still happens with the latest insertion package? If so, I may need a MWE, unfortunately - I can't reproduce it.

alerque:
I'm out of town right now but I'll re-run this as soon as I get a chance and report back where it's at.

alerque:
I'm getting nothing but an error when typesetting any of my books that have footnotes. Others without seem to work fine.

    Error detected:
    /usr/local/share/sile/core/utilities.lua:111: attempt to get length of a nil value (local 't2')

alerque:
So it's not actually failing on _all_ footnotes, it's actually throwing in the towel on the very ones that, in the previous version I was using, would have ended up overlapping the body text.

alerque:
So with the patch above, this is still not fixed. The first line of footnotes is still not consistently spaced with. But it seems like it's not just the first line any more, sometimes the last one in affected.

![selection_112](https://cloud.githubusercontent.com/assets/173595/11429704/687437a0-9484-11e5-9554-c867389fd5cc.png)

simoncozens:
As per slack, I think the issue is that leading in footnotes is applied according to the usual TeX rules, i.e. lines with descenders will have different spacing to lines without. Even if the body text has fill-font spacing, we reset the settings when we box up footnote content.

Setting inter-insertion-skip to zero shows the problem:

![screen shot 2015-11-27 at 16 09 42](https://cloud.githubusercontent.com/assets/106728/11435752/4bd16dbe-9521-11e5-99e4-1fef876b2a7d.png)

And that's even with

```TeX
[...]

PR #221, opened 2015-11-26
Don't try to iterate over and append nil type to table

alerque:
See issue #213.

I don't know if this is actually the correct fix or if this should be checked more intelligently upstream, but it gets me by to do a successful rendering pass.

### DIFF
--- a/core/utilities.lua
+++ b/core/utilities.lua
@@ -108,6 +108,10 @@ function utilities.sum(array)
 end
 
 function table.append(t1, t2)
+  if type(t1) == "nil" or type(t2) == "nil" then
+    t1[#t1+1] = nil
+    return
+  end
   for i=1,#t2 do
       t1[#t1+1] = t2[i]
   end


alerque:
I just realized that this "fix" was eating my chapter headers and some page breaks. I tweaked it again so that it doesn't destroy my documents, but the same caveat as above still applies.

simoncozens:
Sorry, but I don't like this fix. I think it should actually throw an error with traceback (`SU.error(...,true)`) so we can find out *why* it's being passed nil parameters.

alerque:
I agree this isn't a proper fix, but I don't know enough about what should be passed to this function to debug the upstream part. The PR was more for the sake of seeing test output and being able to show you where the problem was surfacing that because I thought it should be merged.

simoncozens:
Try this instead. If you do get assertion failures, (I don't get any processing your document) then please file them as issues.

alerque:
You don't get any in my document because this hack is in my document. Nuke it from the end of the `avadanlik/viachristus.lua` file and you'll get the error. Depending on your font situation it might vary on what page it bombs.

PR #235, opened 2015-12-03
Rename codepoint to gid

khaledhosny:
The use of codepoint here is confusing since it is actually a glyph id
(HarfBuzz calls it that way probably because the buffer holds actual
code points before shaping).


simoncozens:
Thanks, that's annoyed me for a while but not enough to do anything about. :-)

PR #236, opened 2015-12-04
Add support for COLR/CPAL color fonts

khaledhosny:
Very crude and probably inefficient, but seems to work.

I wasn’t sure where to put the get_table() function, harfbuzz module seemed to be the closest.

Issues I’m not sure how to address:
- [x] Cache the parsed tables instead of parsing them for each shaping call.
- [ ] Add an option to `\font` to select which pallette to use.
- [x] Get glyph names of the layers, but are glyph names used at all?


simoncozens:
I haven't looked at this in detail but is there a way we could also tie this in with the OpenType parsing package? Either make that use get_table or move the colour parsing stuff into there or both?

khaledhosny:
I think moving the parsing part to the OpenType parsing package and making it use `get_table` makes sense, I’ll give it a try.

khaledhosny:
Moved table parsing code code to the OpenType parsing package, this does not handle caching I think, but at least it makes it a geneal issue :)

khaledhosny:
Did a boor caching attempt, seems to save decrease the times the tables are parsed, but does not eliminate all unnecessary ones, probably because the life cycle of the face is short.

khaledhosny:
Taking inspiration (read outright copy paste) from the font fallback package, the color fonts package is now self-contained; no changes needed to node maker.

khaledhosny:
Screenshot from the PDF of the test file, for anyone wondering how it looks like:

simoncozens:
This is fantastic! Thank you!

alerque:
:8ball:

behdad:
At some point in the future I like to add the COLR/CPAL code to HarfBuzz...

khaledhosny:
I’d give it a try, but the API can be tricky, a new buffer flag? and how to return the color to the clients, a new `hb_glyph_info_t` field, would that break the ABI?

behdad:
> I’d give it a try, but the API can be tricky, a new buffer flag? and how to return the color to the clients, a new hb_glyph_info_t field, would that break the ABI?

Yeah, I haven't decided on the API yet.  At the outset, it will be new API to allow querying COLR/CPAL.  The integration with shaping will be trickier.  We can return color in hb_glyph_info_t's var1.  It might be easier to ignore integrating with hb-shape, and have a separate function that takes a shaped buffer and does color substitutions.

PR #238, opened 2015-12-05
Compiler warnings

khaledhosny:
Fix misc. compiler warnings.


simoncozens:
Actually I'm not sure if the libthai stuff has a future. It was an experiment because Thai wasn't working out well, but I got ICU working and it looks a lot better now. I can't find any situations where libthai gives us breakpoints that ICU doesn't - and @mhosken says it may be better to concentrate on ICU because there will be additional minority language support in the future.

khaledhosny:
OK then, I was just annoyed by the warnings.

Issue #240, opened 2015-12-05
ICU nodemaker is fragile


simoncozens:
Our new shaping strategy passes the whole paragraph of text to Harfbuzz to shape, and comes back with an array of glyphs. However, we also need to do language specific processing on that glyph array, which we do as part of turning the list of glyphs into a list of nodes.

In the general case, we need to know where in that array we might find additional line break points (e.g. for south-east Asian languages), so we ask ICU. But ICU doesn't want to answer questions about our array of glyphs, it wants to be handed some Unicode text. So we reconstruct the Unicode text and pass that to ICU, and then we have to match up the breakpoints that ICU gave us with the glyphs that Harfbuzz gave us.

In general it just about works, but sometimes you can have a single Harfbuzz cluster which has two ICU breakpoints - a particularly horrible example in the `color-fonts` tests being ☺🏻.  (Arguably this is an ICU bug, but we need to use the ICU that we have, not the one that we want to have.) So as we advance down the string cluster-by-cluster we skip over one of the ICU breakpoints and SILE can't match the two up any more:

```
{1={subtype=soft,token=,type=line},2={subtype=soft,token=☺,type=line},3={subtype=soft,token=🏻,type=line}}
ptr is 1	index is 1
Adding 	☺🏻

! Couldn't resolve ☺ in input at tests/color-fonts.sil
```

khaledhosny:
It does not even work with `f­ile` (there is a soft hyphen between f and i) when a ligature is formed.

I feel that it working in a needlessly convoluted way. The way I’d have done it is to find the indices of all breaking points in the text, iterate over the items to find items with indices corresponding to the break positions and insert the breaks there.

khaledhosny:
My idea wouldn’t still break inside ligatures, but the current code does not either AFAIK.

simoncozens:
OK, I am almost there with the indexing idea; but hitting another problem. Harfbuzz gives us the segments as indexes into a UTF8 string - which we need for other parts of the system like bidi - but ICU gives us the break points as indexes into the corresponding `UChar` (codepoints) string. So we have to convert the indexes returned by ICU into UTF8 offsets. I don't have the brains to do that now.

khaledhosny:
It is even worse, `UChar` is actually a misnamed UTF-16 character type, so you are converting indices from a variable length encoding to another variable length encoding, fortunately for mere mortal like us we can use ICU to do the mapping. See #246.

simoncozens:
[...]

PR #246, opened 2015-12-07
Map UTF-16 indices to UTF-8 ones

khaledhosny:
Based on suggestion from:
http://userguide.icu-project.org/strings/utf-8

See #240.


khaledhosny:
We can probably optimize this a bit by saving the UTF-8 index in `previous` and using it with the new index to subtract from the original UTF-8 string.

PR #250, opened 2015-12-09
Avoid the extra UTF-16 to UTF-8 conversion

khaledhosny:
Not needed since 5597a766e9a26b6f78cecf5ccbde9fe0ac66fe9b.


simoncozens:
Clever!

PR #256, opened 2016-01-17
Get nastaliq working

mhosken:
This PR is dependent on [the PR updating hb-graphite2 in harfbuzz](https://github.com/behdad/harfbuzz/pull/216).

It works by setting the interspace glue appropriately based on spaceskip, the natural space width and the resulting kerned advance for the space.

### DIFF
--- a/core/typesetter.lua
+++ b/core/typesetter.lua
@@ -305,7 +305,7 @@ SILE.defaultTypesetter = std.object {
 
     if adjustment > 0 then
       if adjustment > gTotal.stretch then
-        if (adjustment - gTotal.stretch) > SILE.settings.get("typesetter.underfulltolerance") then
+        if (adjustment - gTotal.stretch) > SILE.settings.get("typesetter.underfulltolerance").length then
           SU.warn("Underfull frame: ".. adjustment .. " extra space required but "..gTotal.stretch.. " stretchiness available")
         end
         adjustment = gTotal.stretch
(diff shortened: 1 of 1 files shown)


simoncozens:
This is fantastic - thank you! And I'm glad you were able to navigate the code well enough to get this working.

The only thing that gives me pause is that this adds an optional parameter which simply gets passed down through four functions. It just seems a bit inelegant. 

I wonder if the logic would be better placed in `makeSpace`: after all, when you're using Harfbuzz's idea of a space's length in context, you don't care about the generic, contextless width of the space glyph, right? So presumably you shouldn't need to call `measureSpace` in those cases.

mhosken:
Dear Simon,

> The only thing that gives me pause is that this adds an optional parameter which simply gets passed down through four functions. It just seems a bit inelegant. 

Feel free to refactor. It's your code and should try to fit with your style. I feel no real ownership of the patch. My aim is to say how it might be done than to write any code.

> I wonder if the logic would be better placed in `makeSpace`: after all, when you're using Harfbuzz's idea of a space's length in context, you don't care about the generic, contextless width of the space glyph, right? So presumably you shouldn't need to call `measureSpace` in those cases.

The logic goes:

We say that the harfbuzz length is equivalent to the font space and we then add the difference between the font space and the spacefactor to that to get the actual length of the space. The stretch and shrink come either from the font or the spacefactor. The reason for this is so that a user can do text compression or expansion by setting the spacefactor relative to the fonts space width. So for that one needs all 3 values to get a good result.

Again, I'm wide open to your refactoring or even changing what it's doing.

GB,
Martin

simoncozens:
I see that your Harfbuzz PR is now in, so I shall get to refactoring and merging this ASAP.

PR #272, opened 2016-02-20
bootstrap.sh: Exit upon commain failure

mreed1:
If a command, such as git or automake, is missing, there isn't much point in continuing.

`set -e` is [specified by POSIX][1], so we can probably assume it's present on most systems.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#set


mreed1:
Thanks, that was quick :)

PR #275, opened 2016-02-20
README.md: Augment instructions for systems using `/usr/local`

mreed1:
For whatever reason, the configure script fails on OpenBSD if
{C,LD}FLAGS aren't manually set to the values below.  Note that env(1)
is used for compatibility with non-Bourne shells, like csh.

---

With that said, I don't think this is an optimal solution. Perhaps it would be better if this behaviour was fixed (assuming it's considered broken), although I know very little about GNU autotools.

In case we want to go that route, here's the actual error I experience when running `./configure`: https://gist.github.com/mreed1/9823cd0d97a1bb337fcb. I'd be happy to provide more info if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,14 @@ If you are building from a a git clone, start by running the script to setup you
 Once your dependencies are installed, run:
 
     $ ./configure
+
+If your system places 3rd party software in `/usr/local`, you may need to run this instead:
+
+    $ env CFLAGS=-I/usr/local/include LDFLAGS=-L/usr/local/lib ./configure
+
+Now you should be able to build and install SILE:
+
+    $ make
     $ make install
 
 This will place the SILE libraries and executable in a sensible location.


simoncozens:
Huh. You would think looking in `/usr/local/` was such a commonly-needed thing that there would be autotools magic for it. But like everyone else who uses autotools, I am just cutting and pasting snippets of configure files I found on the web.

alerque:
@mreed1 Did you run the bootstrap script first? Were there any errors there? The work that it does  should have covered this eventuality. The situation  you describe makes me think you jumped straight in to `./configure` without having run `./bootstrap.sh` first. If you did run it it would be nice to know what output it gave you so we can figure out what actually went wrong.

mreed1:
@alerque Yes, I ran the bootstrap script first, and there were no errors.

Here's the output:
```
libtoolize: `build-aux/ltmain.sh' is newer: use `--force' to overwrite
libtoolize: `m4/libtool.m4' is newer: use `--force' to overwrite
libtoolize: `m4/ltoptions.m4' is newer: use `--force' to overwrite
libtoolize: `m4/ltversion.m4' is newer: use `--force' to overwrite
libtoolize: Consider adding `-I m4' to ACLOCAL_AMFLAGS in Makefile.am.
libtoolize: putting auxiliary files in `.'.
libtoolize: copying file `./ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIR, `m4'.
libtoolize: copying file `m4/libtool.m4'
libtoolize: copying file `m4/ltoptions.m4'
libtoolize: copying file `m4/ltsugar.m4'
libtoolize: copying file `m4/ltversion.m4'
libtoolize: copying file `m4/lt~obsolete.m4'
configure.ac:6: installing './compile'
configure.ac:6: installing './config.guess'
configure.ac:6: installing './config.sub'
configure.ac:3: installing './install-sh'
configure.ac:3: installing './missing'
src/Makefile.am: installing './depcomp'
```

mreed1:
@simoncozens @alerque Are either of you opposed to merging this PR as-is?

It doesn't seem like anyone is interested in dealing with what is thought to be the underlying problem (autoconf), which is completely understandable. I figure a workaround is better than no solution at all, at least in this case.

alerque:
@mreed1 I do actually have some objection to it. My apologies I thought I penned a reply to your last comment, but apparently it didn't get posted.

Can you confirm that the `./bootstrap.sh` is actually working for you in a fresh clone on the first run? Running it now that you've also done other stuff doesn't really test whether it is working correctly. Also since it may or may not give you an error message you can understand I'd like to know that the return code was actually set. In other words can you get a fresh clone and try `./bootstrap.sh && echo SUCCESS || echo FAIL`?

khaledhosny:
Can you show the relevant parts of the `config.log`, as [`ax_lua.m4`](http://www.gnu.org/software/autoconf-archive/ax_lua.html) says it looks under `/usr/local` as well.

mreed1:
@alerque 

> Can you confirm that the ./bootstrap.sh is actually working for you in a fresh clone on the first run? Running it now that you've also done other stuff doesn't really test whether it is working correctly. Also since it may or may not give you an error message you can understand I'd like to know that the return code was actually set. In other words can you get a fresh clone and try ./bootstrap.sh && echo SUCCESS || echo FAIL?

I just did what you asked and it seems to work fine:
```
$ ./bootstrap.sh && echo SUCCESS || echo FAIL
Submodule 'libtexpdf' (https://github.com/simoncozens/libtexpdf.git) registered for path 'libtexpdf'
Cloning into 'libtexpdf'...
remote: Counting objects: 2688, done.
remote: Total 2688 (delta 0), reused 0 (delta 0), pack-reused 2688
Receiving objects: 100% (2688/2688), 2.49 MiB | 1.76 MiB/s, done.
Resolving deltas: 100% (2124/2124), done.
Checking connectivity... done.
Submodule path 'libtexpdf': checked out '0cb0c20a3ba40057e6902551300630fed7199bf6'
libtoolize: `build-aux/ltmain.sh' is newer: use `--force' to overwrite
libtoolize: `m4/libtool.m4' is newer: use `--force' to overwrite
libtoolize: `m4/ltoptions.m4' is newer: use `--force' to overwrite
libtoolize: `m4/ltversion.m4' is newer: use `--force' to overwrite
libtoolize: Consider adding `-I m4' to ACLOCAL_AMFLAGS in Makefile.am.
libtoolize: putting auxiliary files in `.'.
libtoolize: copying file `./ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIR, `m4'.
libtoolize: copying file `m4/libtool.m4'
libtoolize: copying file `m4/ltoptions.m4'
libtoolize: copying file `m4/ltsugar.m4'
libtoolize: copying file `m4/ltversion.m4'
libtoolize: copying file `m4/lt~obsolete.m4'
configure.ac:6: installing './compile'
configure.ac:6: installing './config.guess'
configure.ac:6: installing './config.sub'
configure.ac:3: installing './install-sh'
configure.ac:3: installing './missing'
src/Makefile.am: installing './depcomp'
SUCCESS
$
```

mreed1:
@khaledhosny in `config.log`, this is where the `./configure` stage seems to fail:
```
configure:14180: gcc -o conftest -g -O2   conftest.c -llua  -lm   >&5
/usr/bin/ld: cannot find -llua
collect2: ld returned 1 exit status
configure:14180: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "SILE"
| #define PACKAGE_TARNAME "sile"
| #define PACKAGE_VERSION "0.9.3"
| #define PACKAGE_STRING "SILE 0.9.3"
| #define PACKAGE_BUGREPORT "simon@simon-cozens.org"
| #define PACKAGE_URL ""
| #define STDC_HEADERS 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_MEMORY_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_UNISTD_H 1
| #define HAVE_DLFCN_H 1
| #define LT_OBJDIR ".libs/"
| #define HAVE_LUA_H 1
| #define HAVE_LUALIB_H 1
| #define HAVE_LAUXLIB_H 1
| #define HAVE_LUACONF_H 1
| /* end confdefs.h.  */
|
| /* Override any GCC internal prototype to avoid an error.
|    Use char because int might match the return type of a GCC
|    builtin and then its argument prototype would still apply.  */
| #ifdef __cplusplus
| extern "C"
| #endif
| char lua_load ();
| int
| main ()
| {
| return lua_load ();
|   ;
|   return 0;
| }
configure:14197: result: no
configure:14219: error: cannot find Lua libs
```

Then again I'm not sure what you mean by relevant parts, as I know very little about autoconf, so I've uploaded the whole [`config.log` here](https://gist.github.com/mreed1/714c4228bc5ab14a73e3).

mreed1:
Anyway, It seems like our version of `ax_lua.m4` is quite outdated; notably, it looks like it's missing [this commit](http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=commit;h=5fce217c22f16fd88b7e13be5ffbf26a8f998180), which supposedly adds OpenBSD support.

Unfortunately, `./configure` still fails after dropping in the newest version and re-bootstrapping (`git clean -fdx && ./bootstrap.sh && ./configure`).

khaledhosny:
I just wonder why ax_lua.m4 is going through pains to not use pkg-config? Don’t all version of Lua we support come with a pkg-config file?

mreed1:
Yes, it is odd, especially since lua on my system (for both 5.1 and 5.2) seem to have valid pkg-config files:
```
$ pkg-config --cflags --libs lua51
-I/usr/local/include/lua-5.1 -L/usr/local/lib -llua5.1 -lm
$ pkg-config --cflags --libs lua52
-I/usr/local/include/lua-5.2 -L/usr/local/lib -llua5.2 -lm
$
```

simoncozens:
This is tricky because Lua is written in a very austere way for portability. Distributions have to do a lot of work to turn the base Lua sources into something usable on a modern Unix system. So lua itself doesn't do anything about pkg-config (or indeed anything about making e.g. shared libraries), but all of those things will be provided by the distributor. I don't know if, e.g. Windows builds of Lua have pkgconfig support. @pkulchenko, could you check this?

pkulchenko:
@simoncozens, I doubt it; also, there is no "official" windows build for lua. I checked LuaBinaries (that's probably the closest you can get to "official") and I don't see any `.pc` files. What else may I be looking for?

Having said that, the ticket is about OpenBSD; why are we looking at Windows? ;)

simoncozens:
The proposal is to rely on pkgconfig for configuring Lua include/library paths. If that's going to make Windows building more difficult I'm not sure it's the right solution.

alerque:
@harrysummer Can you comment on this? It seems like this PR is some combination of obsolete/addressed other ways and with working Windows builds now and relying on pkg-config for Linux I think this would just bring more confusion than clarity.

Am I missing any other reason we can't close this PR?

harrysummer:
Thanks for inviting me into this thread. In my opinion, this PR is more of dealing with a special case of system setup rather than a general guideline to build SILE. So generally speaking, I don't think it a good practice of including this into README. Besides, we are already providing binaries for both Linux and Windows, so manually building would limit to dev or advanced users who have enough knowledge to figure out what is missing in CFLAGS or LDFLAGS. -- You may not agree with the above but nevertheless, looks like the current plan is to introduce pkg-config on Linux so this PR would not be needed in the future. (although I preferred adding an option like --with-lua in configure.ac)

For Windows build, we are now relying on CMake script. Lua and all other dependencies would be built along with SILE and installed to the same folder. So we've got the workaround to the problem on Windows. I think we could document the CMake build part to README for completeness.

In conclusion, I agree that this PR should be closed as it is no longer relevant on Windows. And we plan for alternative solutions on *nix too.

PR #282, opened 2016-03-09
Call fontconfig for mapping CSS weights

brawer:
After this change, [Source Han Sans](https://github.com/adobe-fonts/source-han-sans/)
becomes available in weight=300.

Resolves https://github.com/simoncozens/sile/issues/281


simoncozens:
Looks good, thank you!

PR #285, opened 2016-03-10
Select fonts by width

brawer:
For example, `\font[width=87]` for 87% Condensed, `width=100`
for 100% Regular, or `width=125` for 125% Expanded.

Eventually, it would be nice to support font variations (Multiple
Masters, AAT/GX) in SILE. This would need more work than merely
passing a number to fontconfig.  However, by using fractional numbers
for the width specification, the document format is "future proof".
Fontconfig simply finds the closest instance for the specified
width.


behdad:
Not a huge fan of using a 100-based scale instead of 1.0.

behdad:
> Not a huge fan of using a 100-based scale instead of 1.0.

Since that's neither a CSS thing nor OpenType.

simoncozens:
Two things I'm not sure about: 1) it would also need to work on the Mac, which doesn't use fontconfig; 2) more seriously, as a user I would expect that choosing 80% width would make the characters 80% width - but this chooses an 80% width font if one is available, which it might not be, and so it might not set the characters at 80% width at all. I would find that confusing. I think we need to think more about this.

I also would like to avoid the ballooning of parameters to the font key; the repetition of the key in two files is a red flag that we haven't quite got the architecture right here. (Not your fault obviously, but getting that right may mean that this could profitably be implemented as an extension instead of in the core.)

alerque:
Hey @brawer, I've been working through outstanding PRs and am wondering what should be done with this one. Do still find this change useful? Do you have any plans to address the feedback raised? Were you going to wait to see if we did something else with the `\font{}` command options before proceeding?

core/font.lua
```diff
@@ -40,11 +41,12 @@
         SILE.process(content)
         SILE.settings.popState()
       end
-    end, "Set current font family, size, weight, style, variant, script, direction and language")
+    end, "Set current font family, size, weight, width, style, variant, script, direction and language")
 
 SILE.settings.declare({name = "font.family", type = "string", default = "Gentium"})
 SILE.settings.declare({name = "font.size", type = "number or integer", default = 10})
 SILE.settings.declare({name = "font.weight", type = "integer", default = 200})
+SILE.settings.declare({name = "font.width", type = "number or integer", default = 100})
```
alerque: I would second @behdad's request for this to be a 1.0 based scale factor rather than a percentage.

Omikhleia:
> > Not a huge fan of using a 100-based scale instead of 1.0.
> 
> Since that's neither a CSS thing nor OpenType.

Isn't that the CSS [font-stretch](https://www.w3.org/TR/css-fonts-4/#font-stretch-prop) property? In which case a percentage certainly ok, and "strech" would be more appropriate than "width".

How can we select a Condensed or Expanded font currently?

alerque:
> Isn't that the CSS [font-stretch](https://www.w3.org/TR/css-fonts-4/#font-stretch-prop) property? In which case a percentage certainly ok, and "strech" would be more appropriate than "width".

Re 'width' vs. 'stretch': even the CSS docs linked note that the value being selected by the `font-stretch` property **is** the width axis of a font. To me the term "stretch" sounds more like something *we* might do to distort the output than it does a selector for an existing property, especially when OpenType officially has this in the `wdth` variation and the advent of variable fonts has broadly popularized the recognition of what 'width' is in the context of font faces.

The percent(100) vs. float(1) based approach is a harder choice and I'm a little confused. My memory suggested CSS was going one way (percentages) and the OpenType axis and everything giving direct access to it had gone the other way (float). However it seems my memory is faulty. The [`wdth` attribute in FVAR](https://docs.microsoft.com/en-us/typography/opentype/spec/fvar#example) and also the [`usewidthclass`](https://docs.microsoft.com/en-us/typography/opentype/spec/os2#uswidthclass) are documented as using percentages.

Perhaps @behdad or @simoncozens can shed more light on whether a 1-based float is appropriate here.

Omikhleia:
Whichever name and value type are retained, I _do_ think we should support usual string names (condensed, expanded, etc.).

I would actually favor it too for font weight (bold, semibold, etc.). I find it hard for end-users to remember numeric values rather than common names. It would not be unheard of, in SILE, to have such a convenience - we have it for colors (admittedly, there are far more names to remember, but calling for `weight=700, width=75` rather than `weight=bold, width=condensed` is on the same line as saying we should only have `color=#ff0000` at our disposal and no `color=red`...)

simoncozens:
The floats are only used internally by OpenType: "normalised" locations inside a rendering engine run from -1.0 to 1.0 across any variable font axis. For user facing stuff, you're meant to use whatever (float) range makes sense for your font. Percentage is recommended for width, however be aware that the font's default value for the wdth axis may *not* be set to 100.

In summary, you can expect as a rule of thumb variable fonts with a width axis to have a user-facing range between 75-125 will a default normally at 100. This should be selectable as a float if the user really wants to have the font set to 99.95% regular width.

alerque:
@simoncozens I'm down with that and it roughly matches my own research, I'm just not sure how to square what you're saying with the comment above from Behdad:

> Not a huge fan of using a 100-based scale instead of 1.0. Since that's neither a CSS thing nor OpenType.

Was he just thinking about something else?

simoncozens:
We didn't even have variable fonts when he wrote that.

core/font.lua
alerque: Based on current discussion, this should be changed because the type is not right (could be a percent or a keyword). We will need a new type and matching cast/validate functions to handle this—unless it turns out we can just pass raw values through and get sensible errors back from lower lever handling.

examples/weight-width.sil
```diff
@@ -0,0 +1,13 @@
+\begin[class=book,papersize=129mm x 198mm]{document}
+
+\font[family=DejaVu Sans]{
```
alerque, line 3: I'm okay with this if we don't have a better choice in our current test repertoire already. I wouldn't add a new font to the test downloads for this, but given DejaVu is unmaintained and starting to age badly I'd prefer something else if we have one with widths already.

Also we should include a sample using a variable font in this test.

alerque (changes requested):

alerque:
Hey @brawer I totally understand if you are not around to work on this any more. Given that this PR was opened before GitHub allowed upstream project maintainers to push to branches like this one being submitted to their projects, I can't actually pick up and work on it from this PR. One of us could pick up the work in a new branch and start replacement PR. If you don't plan to fuss with it we'll happily do that, I just though I'd ask before we move on in case you do want to work on it here.

brawer:
@alerque Indeed, sadly I won’t be able to work on this myself anytime soon; please feel free to pick it up!

examples/weight-width.sil
ctrlcctrlv: Is that actually required? We don't support variable fonts at all unless they have instances — right? Or even then we don't I thought.

ctrlcctrlv:
@brawer Would you mind giving Caleb and I push access to [brawer/sile](https://github.com/brawer/sile/tree/width) ? You can do it in the repo settings.

brawer:
> Would you mind giving Caleb and I push access to [brawer/sile](https://github.com/brawer/sile/tree/width) ? You can do it in the repo settings.

Of course! GitHub invite sent to @alerque and @ctrlcctrlv.

Omikhleia:

Issue #203, opened 2015-11-14
What the font? It's all downhill from here…


alerque:
Starting at commit c3fc882 some of my fonts are unable to stand upright. Crimson for example, is perpetually falling over. Yes, it has a Roman style. Furthermore some font weights no longer work. Montserrat for example is always normal no matter what weight you request; 600 would normally load the bold and 800 would load the black faces. No longer.

	\begin[papersize=a7]{document}
	\begin{raggedright}
	\font[family=Montserrat,weight=600]{One two}\break
	\font[family=Libertine Serif]{buckle my shoe}

	\font[family=Montserrat,weight=600]{Three four}\break
	\font[family=Crimson]{buckle my shoe}
	\end{raggedright}
	\end{document}

![selection_092](https://cloud.githubusercontent.com/assets/173595/11164915/5c67cd40-8b07-11e5-93ab-a062404f2cb9.png)

Here is a reference of how this should have rendered:

![selection_091](https://cloud.githubusercontent.com/assets/173595/11164916/6c16b4fe-8b07-11e5-8842-48fe0e7ccc24.png)

My system font list looks something like this:

	❯❯❯ fc-list | pcregrep '(Crimson|Montserrat|Libertine Serif)' | sort
	/usr/share/fonts/OTF/Crimson-BoldItalic.otf: Crimson:style=BoldItalic
	/usr/share/fonts/OTF/Crimson-Bold.otf: Crimson:style=Bold
	/usr/share/fonts/OTF/Crimson-Italic.otf: Crimson:style=Italic
	/usr/share/fonts/OTF/Crimson-Roman.otf: Crimson:style=Roman
	/usr/share/fonts/OTF/Crimson-SemiboldItalic.otf: Crimson:style=SemiboldItalic
	/usr/share/fonts/OTF/Crimson-Semibold.otf: Crimson:style=Semibold
	/usr/share/fonts/OTF/libertineserif-bolditalic.otf: Libertine Serif:style=Bold Italic
	/usr/share/fonts/OTF/libertineserif-bold.otf: Libertine Serif:style=Bold
	/usr/share/fonts/OTF/libertineserifdisplay-regular.otf: Libertine Serif Display:style=Regular
	/usr/share/fonts/OTF/libertineserifinitials-regular.otf: Libertine Serif Initials:style=Initials
	/usr/share/fonts/OTF/libertineserif-italic.otf: Libertine Serif:style=Italic
[...]

PR #286, opened 2016-03-10
Call FcConfigSubstitute and FcDefaultSubstitute before FcFontMatch

brawer:
According to the [FcFontMatch documentation](https://www.freedesktop.org/software/fontconfig/fontconfig-devel/fcfontmatch.html), fontconfig can return
wrong results if these functions have not been called before.

Note: When running the regression tests on my local machine with
this change, I get some diffs for `bug-264b.sil` and `vertical.sil`.
However, when building SILE from git head without any changes,
I get the exact same errors.


behdad:
lgtm

simoncozens:
@alerque can you try this and tell me if it fixes #203 for you?

alerque:
Yes in fact it does!

PR #300, opened 2016-04-01
Don't reset spaceskip inside \justified

alerque:
This is a revert of what seems to be a stay setting added in 20d0617.
Reseting the spaceskip value to the font default value for just one
environment but not all the similar ones is inconsistent at best and I
can't figure out what problem it solves. Either this should be done in
all the justification environments so that users come to expect the
value to be changed or (as in this commit) it should be left alone and
whatever gets set in the parent node should carry through to justified
environments.

### DIFF
--- a/classes/plain.lua
+++ b/classes/plain.lua
@@ -106,7 +106,6 @@ SILE.registerCommand("ragged", function(options,c)
     if options.right then SILE.settings.set("document.rskip", SILE.nodefactory.hfillGlue) end
     SILE.settings.set("typesetter.parfillskip", SILE.nodefactory.zeroGlue)
     SILE.settings.set("document.parindent", SILE.nodefactory.zeroGlue)
-    SILE.settings.set("document.spaceskip", SILE.length.parse("1spc"))
     SILE.process(c)
     SILE.call("par")
   end)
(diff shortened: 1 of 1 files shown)


alerque:
This PR might not be legit, but I couldn't figure out what this was fixing and why the unexpected behavior tripped me up when the value I set wasn't set. If you can remember what this was addressing maybe there is a better solution?

simoncozens:
I think something else must be wrong here.

In normal justified text, spaces are stretchable, but ragged text only works if the space is constant. Therefore, in ragged text, we need to change the spaceskip to a constant, unstretchable value, which is what we're currently doing.

Can you show me a bug?

alerque:
I had a document where I was overriding the font default space value and using a much larger one. Every time I entered a `\ragged` environment (section headers, poetry blocks) the space value reverted to the too-small value. Not resetting the value works fine even if the default value has stretch. I assumed the stretchability wasn't even being considered. If that's not the case, perhaps changing it so that only the base value is inherited and the stretchability is dropped might be a better solution than resetting the base value.

simoncozens:
OK, I understand why leaving the space as-is works in ragged environments: the stretchability is swallowed up by the magnitude of the rskip/lskip glue. But I would have more confidence in this if the `alignment` tests passed, or were even close to passing...

alerque:
I'm officially withdrawing this PR. While it solves some problems for me it creates others and as per comments is not the correct long term solution. With another bug open to track progress on a real fix it seems pointless to keep this open. I've added it manually the one place I need it and gone back to using a SILE build without this as the default.

PR #302, opened 2016-04-01
Fix parameter name missmatch introduced in 8ed8b88c

alerque:
(Sometimes Lua not throwing errors on undefined variables makes it really hard to debug!)

### DIFF
--- a/core/measurements.lua
+++ b/core/measurements.lua
@@ -38,7 +38,7 @@ SILE.registerUnit("mm", {definition = "2.8346457pt"})
 SILE.registerUnit("cm", {definition = "10mm"})
 SILE.registerUnit("in", {definition = "72pt"})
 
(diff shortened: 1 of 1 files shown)


simoncozens:
That's clearly wrong and you have clearly the right fix, but I am concerned that tests didn't pick this up; keeping this open until I've added a test for it.

alerque:
The tests didn't  break  because nothing was actually using that code, there were two separate processors for `%` units and this one wasn't getting hit. My suggestion is that the whole ideas of percentages without being explicitly relative to something is broken and should be abandoned (see #304).

alerque:
Looks like we're moving in favor of the full fix rather than the band-aid, so this will be obsolete. Closing it so it stays out of the noise.

PR #304, opened 2016-04-02
Make relevant percentage lengths explicit and introduce more useful units

alerque:
I've been struggling with some layouts that are highly dependent on the relative size of the page canvas (which comes in multiple sizes for the same content) and finally decided that the limiting factor was access to relative units. Scripting out to find sizes at every turn was cumbersome when I could just add new unit definitions, so I started doing that in my own libraries. I found #302 in the process of reviewing how units are handled.

It turns out that bug not  being caught is symptomatic of the exist `%` unit being entirely ambiguous. There were two completely separate parsers for it and two different calc methods depending on whether the context was frames or not. The not-frames version wasn't even being used anywhere and the code was  full of self-identified "dirty hacks" to disambiguate directions. After tying my brain in knots trying to figure out how to get `\rule` that was based on the height of a document rather than the width, it finally dawned on me that the whole thing was bogus. Percentages shouldn't be ambiguous, they should be a percentage of something.

This:

  - [x] Replaces the ambiguous `%` unit with`%pw` units for percentage of page width and `%ph` units for percentage of page height. You can now specify any length (frame size, rules, fonts, etc.) relative to either the hight or width of something else instead of being railroaded into one direction (previously the active axis for frames or horizontal only for other lengths).
  - [x] Adds assorted other units (similar to [CSS viewport units](https://css-tricks.com/viewport-sized-typography/)) such as `%pmax` for whatever is larger of the page height and width, `%min`, a whole set for relative to the current frame rather than the page, etc.
  - [x] Rips out the percentage calculator from the frames code and replaces it with the main dimensioned unit calculator. Dimensions in frame specs now work the same way as everywhere else.
  - [x] Updates some of the existing code.

Todo:

  - [x] Add tests to show they actually work.
  - [x] Document the new units and breaking change to previous syntax.


alerque:
Any thoughts on this one?

simoncozens:
I love the idea in principle. I have been struggling to get e.g. full-column (100% of frame width) rules to work. I feel slightly sad to lose the Do-What-I-Meanery of the "ambiguous" frame percentages, but two extra characters are not too hard to learn to type. Yes please to tests and docs! I guess it means we can then close 302.

alerque:
I liked the Do-What-I-Meanery too the first time I used it and played with a basic layout. Sadly the luster faded quickly when I started beating my head on it because it wouldn't do what I meant and couldn't be corrected to do what I meant.

It's possible that a raw `%` unit could be shimmed back in with a guessing algorithm, but I feel like this needs to _not_ be done right now because of the confusion it will create. The default needs to be something predictable that behaves the same way across all contexts, then if somebody wants a magic units package that introduces some level of intelligence, that should be separate.

Either way existing documents are going to break, but making it obvious why and how to fix it seems like a better option that having the existing unit behavior change in a way that breaks but confuses everybody about proper use.

alerque:
I added tests and fixed up the `%lw` unit so it subtracts the lskip/rskip values. Is there ever a time where line width as returned by `SILE.typesetter.frame:lineWidth()` is going to be different than the current frame width `SILE.typesetter.frame:width()`?

Documentation might take me a little longer to get to, I'm stretched pretty thin as it is.

alerque:
And there are some docs to float this barge...

alerque:
Rebased to fixed merge problem that came up when I changed the test output format the other day. There will be two tests that break and need some adjustment after this is merged, but fixing them pre-merge would be a lot more complicated and I don't think the necessary commits make sense as part of this feature branch.

alerque:
I've rebased this whole thing to work out issues with tests and updated the bits from master that changed since I put this together.

And now it's GREEN!

PR #311, opened 2016-05-02
SILE: font: add support for HyphenChar parameter

rajeeshknambiar:
Many scripts/languages use different hyphenating character than "-". For instance Malayalam prefers non-visible hyphen. XeTeX has support for specifying the preferred hyphen per font via fontspec. Add similar functionality to SILE.
One question: it seems setting `font.hyphenchar` affects subsequent different font/language text as well unless explicitly set to something else. Is this correct, or is there a better way?

Please note: this is my first attempt to make changes to SILE and in Lua - would be great to get your feedback and whether it could be implemented better.

### DIFF
--- a/core/font.lua
+++ b/core/font.lua
@@ -35,6 +35,10 @@
           SILE.settings.set("font.script", lang.defaultScript)
         end
       end
+      if (options.hyphenchar) then
+        -- must be in the form of, for example, "-" or "U+2010" or "0x2010" (Unicode hex codepoint)
+        SILE.settings.set("font.hyphenchar", SU.utf8charfromcodepoint(options.hyphenchar))
+      end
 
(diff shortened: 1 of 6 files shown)


rajeeshknambiar:
@simoncozens 
Hi Simon,
Would you take a look at this change and share your feedback; or any plans to integrate?

simoncozens:
Sorry for the delay in looking at this; life has been a bit overwhelming recently. This is a wonderful patch - very thorough, thank you very much! I guess the next stage would be for the language support file for Malayalam to change the hyphen character when the Malayalam language is selected. This would need support from the `language` command in `languages.lua`.

rajeeshknambiar:
Thanks for the review, and merging it!
I will try to understand and work on the `language` support part.

PR #314, opened 2016-05-07
Fix change from ee97145d to only be warning so we don't kill existing working docs

alerque:
Even in the SILE source tree we have tons of `font[language=ar]` etc. calls for languages that don't have specific setups (yet). This at  least works to pass some data about the text through to Harfbuzz etc, blocking compilation altogether for languages we don't have yet is worse than failing gracefully.

### DIFF
--- a/core/languages.lua
+++ b/core/languages.lua
@@ -8,7 +8,7 @@ SILE.languageSupport = {
     if ok then return end
     ok, fail = pcall(function () SILE.require("languages/"..language) end)
     if fail then
-      SU.error("Error loading language "..language..": "..fail)
+      SU.warn("Error loading language "..language..": "..fail)
     end
   end,
   compile = function(language)


simoncozens:
Not sure about this. I get the sentiment but would prefer to distinguish between "we don't have a language pack for this language" - warn but proceed - and "the language pack failed to compile." - error and die.

alerque:
That might be a better solution, but the current master branch was so broken it was impeding my workflow. With trying to push half a dozen books through production I really need the bleeding edge fixes going into SILE. Code that makes the program die on purpose when it could actually work fine if left alone brings by builds to a halt. Your solution sounds better, but the current code doesn't do that. This backs it off to a usable state and a correct fix that differentiates between those states could  happen whenever.

simoncozens:
OK, good enough. :-)

alerque:
Oh the joys of using HEAD in production :wink:

PR #315, opened 2016-05-14
Pass objects to TOC

alerque:
This takes a slightly different approach passing content to the TOC.

- The **original** model: Parse the entry down to a single string item, pass it to the TOC, then use it as a string for the PDF index and wrap it it an object again to typeset.

- The **new** model: Pass the original entry as-is to the TOC. Then reduce it to a string (by walking the entire object) for the PDF index and use the whole object minus the top level function to typeset the contents.

This approach allows chapter titles to include extra code (such as emphasized words, or in my first use case discretionary break points) in the chapter title string that actually gets used in the TOC entry. Of course having SILE code passed along to the TOC works fine for typesetting it, but not so much for cramming into the PDF index. For that I wrote a couple utility functions to reduce content structures down to a string or just strip the top level command for various use cases.

### DIFF
--- a/classes/book.lua
+++ b/classes/book.lua
@@ -67,6 +67,7 @@ SILE.registerCommand("right-running-head", function(options, content)
 end, "Text to appear on the top of the right page");
 
 SILE.registerCommand("book:sectioning", function (options, content)
+  local content = SU.subContent(content)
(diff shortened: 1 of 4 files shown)


simoncozens:
Thank you!

Issue #269, opened 2016-02-17
Strechyness of parskips not balanced


alerque:
The default value for `document.parskip` seems to have some stretchyness to it. This is quite reasonable, but the stretch should be balanced between all the paragraph breaks on a given page. As it is, it's _all_ getting applied to just the first break. See for example:

![selection_164](https://cloud.githubusercontent.com/assets/173595/13103189/d7a946c4-d55c-11e5-9167-d868899ccbfb.png)

simoncozens:
I'm not sure why you're seeing what you are seeing, but the parskips are being balanced across the page:

```
\begin[papersize=a5]{document}
\set[parameter=document.parskip,value=0.1pt plus 30pt]
\script{for i = 1,30 do
SILE.typesetter:typeset("Hello world.")
SILE.call("par")
end}
\break% No vfill
\end{document}
```

![screen shot 2016-02-18 at 07 22 32](https://cloud.githubusercontent.com/assets/106728/13123481/67d93444-d610-11e5-9e50-b00180b8a7dd.png)

alerque:
Bah! At least at first blush I can't get a MWE to fail either (or my document to succeed). Maybe it's time to add a "Needs MWE" tag to the issue tracker?

alerque:
I have a hunch this might be related to #262. It only happens on the _first_ paragraph of any given page and might be some glue value carrying over from the previous page rather than anything that would happen on a page by itself. I still can't pin it down though.

alerque:
Alright here's an MWE. That was a royal pain in the butt. In the end it looks like the last thing I couldn't rip out and still get this to reproduce is the footnotes.

```TeX
\begin[papersize=halfletter,class=book]{document}
\font[language=tr]
\nofolios
\script[src=packages/footnotes]
Müjde, Hristiyan bireylere ve kiliseye verilmiş en büyük hazinedir.
Müjde, Hristiyan bireylere ve kiliseye verilmiş en büyük hazinedir.
Müjde, birçok şeyin arasındaki herhangi bir mesaj değildir, her şeyin
üzerindeki bir mesajdır! Tanrı’nın kurtarma gücüdür ve Tanrı’nın sayısız
bilgeliğinin, insanlara ve meleklere verilmesi ile ilgili en büyük
vahiydir.\footnote{Romalılar 1:16; Efesliler 3:10} Bu nedenle elçi Pavlus vaazlarında müjdeye ilk sırayı
[...]

PR #321, opened 2016-05-31
Update expected output of footnote test

alerque:
This test was affected by bug #269, and this is what the output should have been all along.


alerque:
Thanks for dropping  by @coveralls and informing us (through broken Markdown) you don't actually know what just happened. Better luck next time?

Issue #309, opened 2016-04-14
Setting document.parskip with stretch breaks typesetter


alerque:
In trying to come up with an MWE for #269 (which is manifesting itself on nearly every page of one book but still defies being trimmed down to an MWE), I notice the `document.parskip` setting isn't even working right:

```TeX
\begin[papersize=a6]{document}
\set[parameter=document.parskip,value=2em plus 1em]
Test par

Test par
\end{document}
```

> Error detected:  
> /usr/local/share/sile/core/typesetter.lua:326: attempt to perform arithmetic on a table value (field 'stretch')

PR #322, opened 2016-06-02
Add tests for open issues

alerque:
Here are some of the MWE's in open issues adapted for the test suite.


simoncozens:
Thanks for this. I'm pulling it in even though it breaks the tests because it breaks them in a way that it really shouldn't, and maybe getting to the bottom of that will help with the bugs this is preparing us to fix.

alerque:
The strange test failure mode is the fault of #309 which is (as expected given the bug) crashing the typesetter rather than actually comparing the output metrics.

Issue #337, opened 2016-06-14
Using crop marks cause %ph and %pw units to compute bad measures 


gnodet:
That's because the cropmarks package redefines the size.

PR #339, opened 2016-06-14
Using crop marks cause %ph and %pw units to compute bad measures, fix…

gnodet:
…es #337

Also enhance the cropmarks package so that the header can be overridden (useful for testing as it includes the date/hour by default)


gnodet:
Is there any known differences between lua 5.2 and 5.3 to cause such a test to fail ?

simoncozens:
Looks like the "draw line" debug code needs to be fixed to deal with the integer/float change in Lua 5.3.

simoncozens:
Very nice, thank you!

alerque:
Good catch @gnodet!

Issue #336, opened 2016-06-11
Unaligned lines with grid


gnodet:
I'm having trouble with grid alignment.  Note that I've disabled column balancing, so this does not look like a balancer bug.
I'm also not sure why page 5 as empty lines at the top...
See the  following snapshot:
![grid01](https://cloud.githubusercontent.com/assets/84022/15984138/48ee058c-2fc0-11e6-9672-efbdd977e6a8.png)

gnodet:
Any hints would be appreciated.  I'm willing to investigate myself, but I have no real clue where to start ...

simoncozens:
Hi Guillame. Were the black lines in your example created by "debugGrid" or the `\grid:debug` command? If not, that would be the first place I would start looking, as that will tell you where SILE thinks the grid should be.

The next thing after that I would try is to turn off grid mode and try using a fixed linespacing using the (sadly currently undocumented) `linespacing` package. Set `linespacing.method` to `fixed` and `linespacing.fixed.baselinedistance` to the grid size you want. That will tell you if it is a problem with the grid package or core typesetting system.

Next I would want to find out what is going on with those five blank lines. Use the `debug` package, and at verse 28 use `\debug[typesetter=on,pagebuilder=on]`. Turn them off again at the 2nd chapter mark. That will give you copious debugging output about the line and page building decisions. If you can show me those I might be able to work out what is going on.

gnodet:
Thx, I was indeed using the grid package. The `linespacing` one looks promising.  The result is already much better. I'm using the `fixed` method but I still have one problem : the first line is sometimes not aligned.  I think this is because `leadingFor` is never called by the typesetter for the first box of a given paragraph.  I'll try to see if I can fix that somehow.

gnodet:
I'm trying to fix the first line `leadingFor` problem, but any `vglue` is use for the first line is discarded.  I tried setting the field `discardable = false`.  I found a hack, which consist in:
  * a change to the typesetter to call `leadingFor` with a `previous` equals to `nil` for the first line
  * a change to the linespacing package to compute the correct value 

The hack is to return a `vglue` object with overriden methods so that `isVglue` returns false, which tricks the typesetter so that it does not ignore this vglue at the beginning of the frame.  If you know a better alternative, let me know, else I'll try to submit a PR with those 2 changes.

simoncozens:
Have a look at this again, and see where we're up to.

gnodet:
Thx, with the above small PR, this work like a charm !

PR #340, opened 2016-06-21
Fix typesetter and linespacing so that the first line of a frame can …

gnodet:
…be aligned correctly


simoncozens:
I've implemented some of this myself, and I think my recent changes have fixed #336 now.

PR #341, opened 2016-06-22
Fix VKern node name and type, final fix for #336

gnodet:


simoncozens:
Thanks for this (a year on!). Not sure how it got missed.

PR #354, opened 2016-08-06
Supply absolute length where needed in float function

alerque:


simoncozens:
Looks like we're failing the frametricks test for this one...

alerque:
Try that version. The test case specifies the `rightboundry` using `pt` units, which don't have an `:absolute()` method. My document used `em` which is why I ran into this. Calling the method manually accounts for both cases.

simoncozens:
Lovely, thanks!

PR #356, opened 2016-08-10
Restore (and fix) function to multiply lengths

alerque:
SILE may not have a use for these internally, but I keep running into times it would be nice to be able to treat a length as a number that can be multiplied and dived by some increment. Rather than having to manually convert to a number every time, it seems to make sense just to teach Lua how to handle × and ÷ operations on lengths. Given that it already handles + and -, this seem like the most expected behavior from an end user perspective.

### DIFF
--- a/core/length.lua
+++ b/core/length.lua
@@ -75,6 +75,20 @@ _length = std.object {
     return result
   end,
    
+  __div = function(self, other)
+    local result = _length {}
+    result:fromLengthOrNumber(self)
+    result = result:absolute()
+    if type(other) == "table" then
+      SU.error("Attempt to divide two lengths together")
+    else
+      result.length = result.length / other
+      result.stretch = result.stretch / other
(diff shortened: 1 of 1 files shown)


simoncozens:
Thanks, looks good. To give a bit of historical context, the functions were originally removed because there was a time I wanted to make very sure what was a number and what was a length; because adding a number to a length propagates the lengthiness, Length objects were turning up in strange places, like output co-ordinates. So I made the multiplication cause an error, so that length handling had to be done explicitly. I think we've got beyond that stage now and sorted out what needs to be a length and what needs to be a number, so it's fine to add this back in.

alerque:
That's pretty much what I figured from `git  blame` which told pretty much the same story.

Incidentally I don't think we're quite out of the woods yet. There is still some inconsistency in when function arguments take string arguments and parse them, when they take pre-parsed lengths and when they need ready-made glues etc.

PR #357, opened 2016-08-10
Background color package

### LINKED ISSUE #353: Need way to set PDF background color
For print work the hard coded default of a white / clear background is fine, but there are situations where being able to set the page color from SILE would be useful. The ink color can be set pretty universally at the outset:

```Lua
local color = SILE.colorparser("#5a4129")
SILE.outputter:setColor(color)
```
A way to also set the page background would be nice. In my case applying to all pages would be preferred, but a current-page-only function might be useful other scenarios.

alerque:
Fixes #353.

### DIFF
--- a/documentation/sile.sil
+++ b/documentation/sile.sil
@@ -1025,6 +1025,9 @@ a very big letter.
 \section{color}
 \package-documentation[src=packages/color]
 
+\section{background}
+\package-documentation[src=packages/background]
+
 \section{rotate}
 \package-documentation[src=packages/rotate]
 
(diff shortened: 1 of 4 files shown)


alerque:
Judging from the test output, the mere _presence_ of this package is screwing up tests that have nothing to do with it. It looks like some pages are getting offset. This is a potential bug I expected, but the fact that it doesn't fail ALL the time is a mystery and why it is failing for tests that don't load the package is even weirder to me. Any thoughts before I go chasing my tail in Travis?

simoncozens:
The problem looks like an erroneous double close-brace in your documentation: ` \code{\\background}}`

alerque:
I'm not going to pretend I understand how that extra `}` broke the tests quite the way it did, but we're in the green now.

Issue #200, opened 2015-11-14
Can Lua functions be locale enabled?


alerque:
How can a Lua function be made locale aware? I have a string processing function that needs to be aware of the language. Ideally the current document output language value would be passed, but even if the system locale value was being obeyed this would work:

	local inputfilter = SILE.require("packages/inputfilter").exports
	SILE.registerCommand("uppercase", function(options, content)
	SILE.process(inputfilter.transformContent(content, string.upper))
	end, "Typeset the enclosed text as uppercase")

Currently calling this function from my section typesetter:

    SILE.call("uppercase", {}, content)

…on content like this:

    \section{Müjdeyi Bilmek}

…turns out this catastrophe:

![selection_086](https://cloud.githubusercontent.com/assets/173595/11163640/bfe1a26a-8adf-11e5-966e-783e7b03ee47.png)

This is what needs to have happened:

![selection_087](https://cloud.githubusercontent.com/assets/173595/11163641/c36277fc-8adf-11e5-9a38-abeb28589a0a.png)

How can I make a document language (or locale) aware string processing function?

simoncozens:
If you are on lua 5.3, then `utf8 = require("utf8")` and use `utf8.upper` instead of (the dumb, character-by-character, non-UTF8-aware) `string.upper`.

khaledhosny:
Also, being locale dependent in not a good idea for typesetting software (change the locale e.g. process the document on a different machine, and you get different output). If some code needs to be aware of the language, it should use the document language.

alerque:
@khaledhosny Good point.

alerque:
@simoncozens You sure about that? The Lua 5.3 [documentation for the utf8 module](http://www.lua.org/manual/5.3/manual.html#6.5) does not include a `utf8.upper()` command at all, nor was I able to get one.

I installed the luarock for [lua-utf8](https://github.com/starwing/luautf8) and got `utf8.upper()` out of it, but I was unable to figure out a way to set the language. As a result it was able to uppercase `ü` to `Ü` but not knowing the string was Turkish it transformed `i` to `I` instead of `İ`.

@starwing perhaps you have some input into this?

simoncozens:
Sorry, you're right, it's the lua-utf8 module you want. I thought the utf8 library included in 5.3 was compatible, but it's not.

And yes, I don't know of any Lua Unicode library that does full Unicode CLDR-aware case mapping. You would basically need to wrap ICU to get that. The luatex people would love it if you did, too.

[...]

PR #358, opened 2016-08-10
Adapt ICU case function to handle 'title' mode

alerque:
This should enable title casing strings:

```Lua
icu = require("justenoughicu")
icu.case("teST me", "en", "title")
```

   Test Me

See also #200.


alerque:
However the test comes back don't merge this. I clearly have no idea what I'm doing with the iterator thing. It works once, but the next time you call it it seg faults. Any ideas?

alerque:
The test suite is on drugs. It passed the first commit, then I added a test which should fail and did not, but _other_ unrelated tests failed.

src/justenoughicu.c
```diff
@@ -28,24 +29,32 @@ int icu_case(lua_State *L) {
...
 
-  if (strcmp(recase, "upper") == 0) {
-    conversion = u_strToUpper;
-  } else if (strcmp(recase, "lower") == 0) {
-    conversion = u_strToLower;
-  // } else if (strcmp(recase, "title") == 0) {
-    // conversion = u_strToTitle;
+  if (strcmp(recase, "title") == 0) {
+    conversion_function_ti conversion;
+    UBreakIterator *iter;
```
srl295: at least you want `iter=NULL;` here. That's why it blew up. see [docs](http://icu-project.org/apiref/icu4c/ustring_8h.html#a8c6202ad4ac954e225aae061b184e450)

Side note: This is the oddest way to call ICU I've seen yet… you're including `ubrk.h`, why not just call `u_strToTitle` and dispense with the typedef?

src/justenoughicu.c
```diff
@@ -28,24 +29,32 @@ int icu_case(lua_State *L) {
...
+    conversion = u_strToTitle;
+    l2 = conversion(NULL, 0, input_as_uchar, l, iter, locale, &err);
```
srl295: Instead of passing in `0` here, use a small stack buffer if you think the results will fit in some cases. Then you can avoid the malloc and second trip through the conversion function. See [this sample](http://userguide.icu-project.org/conversion/converters#TOC-4.-Pre-flighting) . This goes for the other examples.

src/justenoughicu.c
alerque: I've seen odder ways — I saw my own earlier attempts at this before I sorted out even to this point! :wink:

Thanks for the feedback. I just pushed another commit that takes your suggestions into account as best I understood them.

src/justenoughicu.c
```diff
@@ -33,13 +32,10 @@ int icu_case(lua_State *L) {
   int32_t l2 = 0;
 
   if (strcmp(recase, "title") == 0) {
-    conversion_function_ti conversion;
-    UBreakIterator *iter;
-    conversion = u_strToTitle;
-    l2 = conversion(NULL, 0, input_as_uchar, l, iter, locale, &err);
+    l2 = u_strToTitle(NULL, 0, input_as_uchar, l, NULL, locale, &err);
```
srl295: :+1: the avoid-malloc is an optimization, I'll file a separate issue for that.

alerque: Thanks @srl295. I'm afraid I was out of my depth just hacking this feature in, the optimization might have to come from somebody with better bigger C chops. On the other hand your tip on the iterator did stop it from blowing up in my face, so that's nice.

alerque:
Per [comments on diffs](https://github.com/simoncozens/sile/pull/358#discussion_r74961295) this is actually working instead of crashing for me now.

srl295:
@alerque no problem. I added a new issue. OK, back to actually using sile!

alerque:
When I was just a little lad my favorite color was pink—hot pink. At some point  when I got a little older somebody convinced me of the relative merits of red for my case. I've clung to  red ever since. But lately Travis is starting to dull my enthusiasm for red and subtly pushing the merits of green.

Anybody have any idea why `./configure` is not finding ICU on Travis? That's what's killing tests on this branch. Without detecting ICU justenoughicu doesn't get compiled, which of course means running tests on it fail.

srl295:
@alerque you should be able to find ICU via `pkg-config` … `icu-config` needs to go away…

.travis.yml
```diff
@@ -5,8 +5,10 @@ compiler:
 addons:
  apt:
   packages:
+   - pkg-config
    - libgraphite2-dev
    - libicu-dev
+   - libicu52
```
srl295: no no… please don't put in a specific icu version here…

khaledhosny: These are apt package names for Travis, they always have the version number.

srl295:
so:
```
$ env PKG_CONFIG_PATH=${PKG_CONFIG_PATH}:/usr/local/Cellar/icu4c/57.1/lib/pkgconfig/ pkg-config --exists --print-errors "icu-uc icu-io"
$ bash bootstrap.sh ; ./configure
checking for ICU... yes
```

let me look into this a little more

alerque:
@srl295 That's not what's actually wrong with Travis. That has to do with the git submodule and Travis's build  cache. I've since fixed that, but that was  never the issue I was looking at because I was using Travis on my own branch with caching disabled. I'll restart that job now chat caching is fixed, but you'll see that the config is not finding ICU in the Travis container.

I did actually get it to find icu-config and compile justenoughicu, but _that version_ throws an error about some ICU symbols not being available.

alerque:
Also it looks like the thing to do might be to switch back off the docker images and up to the new Trusty builds. ICU works there! In fact nearly everything works and it's now as fast as the container version was, but it currently introduces [font metric issues](https://travis-ci.org/alerque/sile/builds/156791427). It would be a closer match to my Linux setup but would leave @simoncozens  and others on Mac floundering without being able to run the test suite locally.

.travis.yml
alerque: @srl295 Besides what @khaledhosny said, this was a test to see if I could get Travis to find ICU at all. So far no go on the containerized system, although following @behdad's lead from harfbuzz, the non-containerized Travis images for Ubunty Trusty find ICU out of the box.

alerque:
With the hack using `icu-config` on the outdated Travis images, the current issue with ICU is this:

```
### Regression testing tests/bug-200.sil
This is SILE 0.9.5-unreleased
Error detected:
error loading module 'justenoughicu' from file 'core/justenoughicu.so':
	core/justenoughicu.so: undefined symbol: u_strToUpper_48
```

Obviously the `icu-config` hack worked enough to compile the module, but not enough to actually have the symbols we need. My guess is we need to a) make ICU real dependency not just a if-the-system-has-it thing and b) there should be a minimum version so that we know we get the functions we need.

Beyond that to get Travis to cooperate with the above, we need to get a sanely new ICU.

srl295:
@alerque 
> but not enough to actually have the symbols we need

actually, probably you're using the headers from one ICU and the library from another. 

also, i mentioned above how to build on the mac with icu and homebrew.

So… i'm very new to travis… how can i reproduce an environment locally that's similar to the various travis images?

alerque:
@srl295 Thanks, but I know how to do this on a Mac — and it works just fine on my local Linux box that is a much more modern system. The problem here is is the containers on Travis being out of date and the containerized system not making it easy to install up to date versions of anything. Have a look an the install monkey business we are already doing to get harfbuzz, fonts, and the like.

The closest thing you can get is to run a VM with a similar Ubuntu build (in this case I think it's _precice_ with some updates). Its only a rough approximation, but it helps some. Debuging Travis is a pain in the butt because of the round trip times and not having access to poke around the environment between tries. If you want to play with it the best thing to do is to fork sile to your own Github, turn on Travis builds for your repo including branch builds on push, then check out  this branch and push some change to your fork.

alerque:
> actually, probably you're using the headers from one ICU and the library from another. 

That sounds quite likely by the way ---  and given that pkg-config wasn't finding it right that would be something to look into. So far I haven't come up with a way to make pkg-config work on the containerized version of Ubuntu Travis is running right now.

khaledhosny:
+1 for ICU as a requirement.

.travis.yml
srl295: also from travis:
```
Disallowing packages: libicu52
If you require these packages, please review the package approval process at: https://github.com/travis-ci/apt-package-whitelist#package-approval-process

```

srl295:
OK. so the problem here with `pkg-config` is that while I [provided .pc files in 4.5 - 7 years ago](http://bugs.icu-project.org/trac/ticket/6981), debian for whatever reason didn't actually pick them up from our build until ???.  So yeah, use `icu-config`.

configure.ac
```diff
@@ -83,7 +83,15 @@ PKG_CHECK_MODULES(ICU, icu-uc icu-io, [
...
+  AC_CHECK_TOOL(ICU_CONFIG, icu-config, no)
+  AC_MSG_CHECKING([for ICU by using icu-config fallback])
+  if test "$ICU_CONFIG" != "no" && "$ICU_CONFIG" --version >/dev/null; then
+    with_icu=yes
+    AC_MSG_RESULT([yes])
+  else
+    AC_MSG_RESULT([no])
+    with_icu=no
+  fi
 ])
```
srl295: missing the guts of it here. You need something like:

```
ICU_LIBS=`icu-config  --ldflags-libsonly --ldflags-icuio`
ICU_CFLAGS=`icu-config --cppflags`
```

alerque: You mean I actually have to _do something_ with the information detected. Silly me!

alerque:
Thanks a lot for your help with this @srl295. And yes I agree that this kind of packaging muckup from Debian and propagated through Ubuntu is pretty frustrating. It would be nice to be able to "move on", but its not always that easy to get away from this stuff.

With the guts you supplied for `configure.ac`, detection is now working and the compiled module actually works. The r regression test for this issue just passed. If everything else goes off without a hickup this should have a green light for merging.

configure.ac
srl295: imagine that…

Issue #344, opened 2016-07-21
Pushed material has leading issues


alerque:
Along very similar lines to my other issues with spacing when content gets pushed to new pages (see #301, #262, and #342) I had a problem with a TOC where the first two lines of the second page of the TOC w were not using the document's baseline skip. I saw this problem a long time ago (I have an internal issue report from a book on Nov 21st, 2015) but it went away when I switched to using the linespacing package and I forgot about it. Now it turned up again with linespacing turned on and after some fiddling I've decided it must have been there all along and I just wasn't triggering it because the metrics were different.

I played around with it and came up with an MWE that works with or without the linespacaing package showing the same problem either way.

```TeX
\begin[papersize=a5,class=book]{document}
\nofolios
\define[command=mwe]{One, two. Buckle my shoe.\smallskip}

% Note this is not specific to using non-default linespacing methods,
% this MWE shows the same problem with or without these lines
\script[src=packages/linespacing]
\set[parameter=linespacing.method,value=fit-font]
\set[parameter=linespacing.fit-font.extra-space,value=6pt]

This old man.

% Note: making this skip 37em pushes the same amount of material to the
% next page but the first line no longer has a spacing issue
\skip[height=36em]

\mwe
\mwe
\mwe
\mwe
\mwe
\mwe
\end{document}
```

![selection_232](https://cloud.githubusercontent.com/assets/173595/17021651/0d43774a-4f52-11e6-83df-492045c63104.png)

Note the highlighted line that has the wrong  line spacing.

alerque:
This has to do with the `pushBack()` routine. Of course.  Specifically the  `\<size>skip` commands in the plain document class are getting pushed to the output queue as explicit skips, but for some reason the reboxing  routine is not detecting them as such.

alerque:
Scratch that. This is a much bigger problem.

The real issue is that the `pushBack()` routine is the _ONLY PLACE_ we're handling explicit skips correctly by inhibiting the usual leading for the next vbox.

Issue #301, opened 2016-04-01
Page break fruitcake with a side of ragged


alerque:
I've been trying to isolate an elusive problem with paragraph alignments in a long document and failing for months. I finally pinned it down to an MWE. There are two symptoms that can be reproduced here—a third that I think involves insertion splitting is still too jello-y to even describe.

``` TeX
\begin[papersize=a7,class=book]{document}
\script[src=packages/linespacing]
\set[parameter=linespacing.method,value=fit-font]
\set[parameter=linespacing.fit-font.extra-space,value=1.15ex plus 0.5pt minus 0.5pt]
A line of stuff in a paragraph here that wraps one line.

\hfill The undersigned,

\hfill John Hancock

\skip[height=14em] % Change this to 13 em and the next right align gets applied to the wrapped half of the paragraph preceding it

A line of stuff in a paragraph that wraps on the page following the one it started on.

\rightalign{The undersigned,}

\rightalign{John Hancock}

\skip[height=16em]

The problem with hfills on the next page doesn't repro if you try to do it on this page.

\skip[height=15em] % Change this to 14em and the next hfill breaks

A line of stuff in a paragraph that wraps not just one but two lines on the page following the one it started on.

\hfill The undersigned,

\hfill John Hancock

\end{document}
```

Here is the expected rendering:

![selection_189](https://cloud.githubusercontent.com/assets/173595/14203497/01e8d31a-f807-11e5-9046-094581b24a99.png)

Here it is again with a slightly different document shape that triggers broken renderings.

![selection_190](https://cloud.githubusercontent.com/assets/173595/14203500/0724319e-f807-11e5-848d-0482b308264f.png)

alerque:
Possibly related to #262?

alerque:
This can play out in really bizarre ways. Check out this two page spread:

![selection_196](https://cloud.githubusercontent.com/assets/173595/14487396/f2f260a2-016a-11e6-859c-568d02bcbb67.png)

The content on the left is from the previous chapter and should be left justified. The chapter title, subtitle and quotation at the chapter on the right is centered. At least they are centered for all the other 94 chapters in the book. This one seems to has passed it's center alignment to the wrapped line on the previous page and adopted the left justification for itself. It also picked up the line spacing settings from the previous chapter and applied them to the chapter title which should be in an isolated settings environment—explaining why the reference is overlapping the quote there.

alerque:
[...]

PR #360, opened 2016-08-29
Seek and destroy 262 (and all its friends)

### LINKED ISSUE #262: Page break resets rskip/lskip. Sometimes.
Since the fix for #255 I have an elusive issue with my block quote environments that have lskip/rskip values applied. Sometimes when a page break happens inside the environment, the next page will not be properly typeset. What's funny is the font settings are coming across, but not the adjusted margins.

Even funnier, it only happens sometimes.

Even funnier, if I tweak the `document.spaceskip` value ever-so-slightly so that the whole document lays out with slightly different line breaks and the words break in a different place, which blocks fail and which succeed changes. I have a couple different styles with different fonts and line spacing inside the environoments and they all fail sometimes and succeed sometimes.

Here's a win:

selection_156

Here's a fail:

selection_155

But wait. Here is that las fail — winning:

selection_157

The only difference between this and the previous compile is this:

``` diff
-\set[parameter=document.spaceskip,value=0.5en plus 0.45en minus 0.2en]
+\set[parameter=document.spaceskip,value=0.55en plus 0.45en minus 0.2en]
```

alerque:
Fixes #262. Maybe — cross your fingers, toes, branches, whatever you've got.

This probably needs work, but  I need to see the test suite relative to a merge rather than just my branch. Also if you'd like me to  rebase the cosmetic stuff into a separate PR from the functional changes I can do that.


alerque:
Now also fixes #344 which was a closely related.

alerque:
Now also fixes #301. All these things were pretty closely related.

alerque:
This is progress, but I'm not out of the woods yet. There is still something whacky going on in documents that have use in-document page templates or switch masters. Some of those tests worked, but a few broke pretty badly and there is still at least one regression in that we're not discarding something that should be (and had previously been fixed to be). In any event I'm out of time for now.

alerque:
Hey @simoncozens if you have a few  minutes can we jump in [the Gitter room](https://gitter.im/simoncozens/sile)? This is going to be a bug-bear to finish and merge and I'd like to do it as painlessly as possible, but it would be nice to coordinate.

alerque:
This is down to one issue. Somehow or another I've broken the page builder in the test where it switches masters. Interestingly everything typesets correctly and the pushBack() seems to be working there, but two PDF pages are overlaid on top of each other.  Ideas?

alerque:
It only took 999 tries, but Travis is [giving this a green light](https://travis-ci.org/simoncozens/sile/builds/157015418)!

simoncozens:
Excellent! I want to read them all first and make sure I understand them, but green lights make me happy.

alerque:
There are a couple bits of commits that _don't_ make sense, but I left them in for the history. For example my line leading snafu. When I touch functions I usually look up when they were added, how they were changed and where they are used. The were they are currently used is usually a matter of `git grep <function>`, but the when and why I usually look for in `git log -S <function>`. I left my mistake in messing with `inhibitLeading()` because the change and revert commits document why _not_ to change it: `git log -S inhibitLeading` leaves a pretty solid trail of why that is there and why it is not somewhere else.

simoncozens:
All good! Thank you so much for this!

PR #363, opened 2016-08-31
Normalize Lua coding style

alerque:
PR #360 got out of hand and needed rebasing, so I'm splitting it up. This is most of the style related changes squashed into a few commits. Nothing in this PR should introduce any functional changes, but white space and coding conventions should be a little more consistent (at least for the bits of code I was touching working on the `pushBack()` thing).


alerque:
This can be merged any time. The test shows as failing for the PR but that's only because of master being broken right now. These commits actually don't break anything as can been in [this passing Travis test run for commit 288c5dc](https://travis-ci.org/alerque/sile/builds/156444437).

simoncozens:
Thanks very much for this; looks great!

PR #366, opened 2016-09-04
Allow customization of TOC with a footer to match header

alerque:
We had a command for \tableofcontents:header to make it easy for a class
to typeset some stuff up there. This is sometimes where the command for
opening a new page (if any) is placed. Unfortunately we didn't have a
matching footer one. This adds one and sets it up to add flexibility to
the layouts a class can accomplish without having to modify the document
content and without having to do too many antics with redefining
standard commands.

I have documents that I want to typeset using the same template
(generated by Pandoc) but with different classes and sometimes the TOC
will take up a full page and sometime not. Having to change the body
content of the document is not an option in these cases and the most
logical place to put it is the same place where the control for starting
on a blank page or not goes.

This likely didn't get noticed before because it is often used alongside
the \chapter command in the book class which starts a new page by
defalt.

### DIFF
--- a/packages/tableofcontents.lua
+++ b/packages/tableofcontents.lua
@@ -5,50 +5,53 @@
 --          writeToc (call this in finish)
 --          moveTocNodes (call this in endPage)
 
-SILE.scratch.tableofcontents = { }
+SILE.scratch.tableofcontents = {}
+
 local loadstring = loadstring or load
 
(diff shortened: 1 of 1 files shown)


alerque:
I'm going to go ahead and blame my commit message typo of 'style' as 'stile' on how many times I type this project's name in a day.

PR #380, opened 2016-10-17
Update tests

alerque:
See commit message, but this should bring Travis up to date again.


alerque:
Almost there. This is passing the regression tests now, it just doesn't work with graphite disabled. Any clues?

simoncozens:
I have no idea. @mhosken, why would horizontal metrics of Gentium Plus change with graphite disabled? Either the glyphs or the space character is consistently a fraction thinner.

Harfbuzz version: 1.2.6
GentiumPlus-R.ttf:0	Version 1.508; 2011; Maintenance release

alerque:
Can we merge this and open a separate issue for dealing with the Graphite problem? It's hard to test PRs against master when the test suite is broken, and this PR at  least fixes it so _some_ of the Travis runs are expected to pass.

PR #388, opened 2016-10-25
Assorted niceties

alerque:
I can split this PR up if there are bits you like are don't like, but these are some tweaks I've been using to good effect.


simoncozens:
I like everything apart from d879e5f. It's a bit too magic - what if someone erroneously does `SILE.registerCommand("blah", "blah")`? I would expect that to come to a screeching halt rather than silently do the wrong thing. See `\xmltricks:passthru` for a generic way to do this.

But the rest looks really good!

alerque:
Okay let me try another approach  to that one. I'm working on a way to fix the class require path too which I've been hacking around forever.

alerque:
My class hack thing is going to take some more thought, it doesn't work as a general purpose  loader the way it works for me specifically. I dropped the pass-through function shortcut  pending review and fixed some issues with my variable renaming including switching to `func` instead of `fn` when a function is meant and `filename` when that is meant. I had mixed this up in a couple of cases in the previous version of this commit.

PR #390, opened 2016-10-28
Add Arabic counter styles

khaledhosny:


simoncozens:
I'm in favour of doing this but I'm wary of adding lots of potential special cases. I wonder if there is a way we could use something like ICU to do locale aware number formatting.

alerque:
I have code for handling Turkish numbers too. My implementation is a bit different, but obviously we're trying to solve the same problem and this is a problem SILE should handle (whether it calls out to ICU or does it internally). One other thought is that we might want to split this up the resources (data objects + any handling logic that differs from some standard pattern) to be part of the language support files.

khaledhosny:
ICU seems to have an API for locale-specific number formatting, but I haven’t used it. It does not seem to have an API for Abjad numerals, though.

simoncozens:
See the commit I just referenced. :-)

simoncozens:
I'm going to close this in favour of the solution I just commited. Use `display=arab`.

PR #395, opened 2016-11-04
Don’t write garbage at the start of the path

khaledhosny:
strncat() appends at the terminating null byte, but the very first time
“output” will contain uninitialized garbage.


simoncozens:
Thanks. My C is horrible. :)

PR #396, opened 2016-11-04
Handle open paths

khaledhosny:


khaledhosny:
BTW, since this C code is getting really ugly. What if we made the SVG module just expose nanosvg API to Lua, and handled the SVG to PS conversion in the Lua side?

simoncozens:
It is really ugly and I agree that C is a horrible language for string manipulation, which is what this basically is. The question I have is if we are thinking of contributing SVG font support back to XeTeX. If so, then the code needs to live in libtexpdf (dvipdfmx) and it needs to be in C.

And it terms of neat separation of concerns, having the PDF library handle the font's SVG outlines seems pretty consistent. So I think architecturally that's the best place for it to end up.

But it would be *so* much easier for SILE to handle SVG internally. I'm conflicted.

khaledhosny:
Any comments on this? (the example SVG used in tests/simplesvg.sil has an open path).

simoncozens:
Thanks for this. I am still wondering what to do in the long run. Really it should head to dvipdfmx, but I haven't spoken to those folks yet. I'll send them an email now.

PR #404, opened 2016-11-10
Allow classes to be loaded from non hard-coded paths

alerque:
The current system demands that any custom classes reside in a specific place relative to the current document. This makes it impossible to include a toolkit in a project in just one folder, you have to pollute the namespace with several folders. This makes it possible to use classes that are in user specified locations so they can even be outside the current directory (referencing a common toolkit folder) or whatever.

### DIFF
--- a/classes/bible.lua
+++ b/classes/bible.lua
@@ -1,4 +1,4 @@
-local plain = SILE.require("classes/plain")
+local plain = SILE.require("plain", "classes")
 local bible = plain { id = "bible", base = plain }
 if not(SILE.scratch.headers) then SILE.scratch.headers = {}; end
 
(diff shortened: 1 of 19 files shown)


alerque:
Oh Travis you know love you, why you no love me back?

alerque:
Doh. Wrong branch. Standby.

alerque:
There. I also refactored the `resolveFile()` function to be more DRY because it was very prone to bugs and the order files were getting checked was not what _seemed_ obvious.

alerque:
Um, stand by on this one. It works on Travis and it works on my Arch systems but it **doesn't work** on some other installs I just tested it on. Antsy for this to work I compiled from a branch that has this merged on a CI system of mine and it broke. I think I know what the problem is but let be fix it before you merge.

alerque:
Alright I fixed the blooper, I'm a go on this again.

PR #406, opened 2016-11-10
Normalize language command usage

alerque:


simoncozens:
Well, I want to keep `main` for a good reason. The idea is that later we will have the ability to select subsidiary languages and auto-detect based on script change. I haven't worked out a syntax for this yet, but the idea is that you have e.g. a document which is in English and Japanese. Its main language is English, but whenever you see Japanese script in the input, you flip over to a different font (and typographic convention) automatically.

alerque:
Okay. In that case can I switch this the other way and make "main" the parameter we are setting until such as a time as there is another parameter? I don't care what it's called so much is that there is some usage consistency. The current behavior was mystifying to me as an end user trying to figure out what did what.

simoncozens:
Are we good to go with this one? Looks good to me.

PR #411, opened 2016-11-19
Assorted niceties

alerque:
Here is some more stuff I've found useful. Check commit messages for detailed explanations. As with previous assorted PR's let me know if there is stuff in here you're unsure about and I'll split it up.


simoncozens:
Thanks!

PR #419, opened 2016-12-15
Pdf external links

alerque:
This adds an `\href` command to the *url* package that can be used to make links to external URI's.


simoncozens:
Nice one, thank you!

PR #423, opened 2016-12-21
Assorted convenience functions

alerque:
This is some assorted stuff I've had in my publishing toolkit(s) for a while that I find myself using a lot and re-inventing any time I don't have those libraries handy. It seems like they should just be part of SILE core but in anybody has better ideas  on how they should be implemented I'm listening...

### DIFF
--- a/core/length.lua
+++ b/core/length.lua
@@ -12,6 +12,11 @@ _length = std.object {
     }
   end,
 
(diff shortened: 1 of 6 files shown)


simoncozens:
Very lovely! Thank you!

Issue #414, opened 2016-12-05
SILE cannot read from a named pipe


alerque:
For some reason the  Lua file handler is refusing to process input from a named pipe. This is really quite a useful thing to be able to do and we should probable make whatever adjustments need to be made to allow this. (Note would also allow a document to be passed in on `/dev/stdin` which is also a named pipe.)

Example to pre-process an file with a macro engine.

    sile -o out.pdf <(m4 source.sil)

This throws:

> ! /dev/fd/63 isn't a file, it's a named pipe! at /dev/fd/63

alerque:
@rjmunro since you expressed a flicker of interest in this I would be curious if you had any input on my PR #441 and whether that covers your use cases and expected behavior or not.

rjmunro:
I'm vaguely interested in sile, but don't have any actual use for it at the moment. I've moved further away from any design / publishing stuff in my recent work.

alerque:
@rjmunro Fair enough, thanks for the feedback.

PR #440, opened 2017-01-26
What comes in must go out

alerque:
This resolves my issue with #414 and makes SILE get along a lot better in a *nix environment. You can now use SILE pretty much any way you would use a any standard  POSIX utility to read/write input. This includes acting as part of a pipeline where the document comes from STDIN and the rendered PDF goes to STDOUT. For example you could run

    $ echo "<sile>This is an XML string</sile>" | sile - | lpr

This feeds an XML document string into sile and sends the output straight to the default printer without fussing with intermediate files. This may sound like a party trick, but consider even for this simple trick four commands would have been run:

    $ echo "<sile>This is an XML string</sile>" > doc.xml
    $ sile doc.xml
    $ lpr doc.pdf
    $ rm doc.xml doc.pdf

This quickly gets out of hand when you have a complicated chain of things you want to run either on the input side, the output side, or both.

This works and using "-" as a filename flag to trigger processing STDIN is _relatively_ standard behavior, but personally I would rather see this be the default for when no filename is provided and move the interactive readline interface behind a flag.

### DIFF
--- a/core/break.lua
+++ b/core/break.lua
@@ -551,9 +551,9 @@ end
 
 function lineBreak:dumpActiveRing()
   local p = self.activeListHead
-  io.write("\n")
+  io.stderr:write("\n")
   repeat
-    if p == self.r then io.write("-> ") else io.write("   ") end
+    if p == self.r then io.stderr:write("-> ") else io.stderr:write("   ") end
     print(lineBreak:describeBreakNode(p))
     p = p.next
   until p == self.activeListHead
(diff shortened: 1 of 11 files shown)


alerque:
Um, actually I didn't mean to merge this without discussion first, esp with at least one  hack in place. That was entirely accidental because I had my git remotes listed backwards. I've reset that and will re-submit this PR.

PR #441, opened 2017-01-26
What goes in must come out

### LINKED ISSUE #414: SILE cannot read from a named pipe

alerque:
This fixes #414 and makes SILE get along a lot better in a *nix environment. You can now use SILE pretty much any way you would use a any standard  POSIX utility to read/write input. This includes acting as part of a pipeline where the document comes from STDIN and the rendered PDF goes to STDOUT. For example you could run

    $ echo "<sile>This is an XML string</sile>" | sile - | lpr

This feeds an XML document string into sile and sends the output straight to the default printer without fussing with intermediate files. This may sound like a party trick, but consider even for this simple trick four commands would have been run:

    $ echo "<sile>This is an XML string</sile>" > doc.xml
    $ sile doc.xml
    $ lpr doc.pdf
    $ rm doc.xml doc.pdf

This quickly gets out of hand when you have a complicated chain of things you want to run either on the input side, the output side, or both.

This works and using "-" as a filename flag to trigger processing STDIN is _relatively_ standard behavior, but personally I would rather see this be the default for when no filename is provided and move the interactive readline interface behind a flag.

### DIFF
--- a/core/break.lua
+++ b/core/break.lua
@@ -551,9 +551,9 @@ end
 
 function lineBreak:dumpActiveRing()
   local p = self.activeListHead
-  io.write("\n")
+  io.stderr:write("\n")
   repeat
-    if p == self.r then io.write("-> ") else io.write("   ") end
+    if p == self.r then io.stderr:write("-> ") else io.stderr:write("   ") end
     print(lineBreak:describeBreakNode(p))
     p = p.next
   until p == self.activeListHead
--- a/core/inputs-common.lua
+++ b/core/inputs-common.lua
@@ -1,9 +1,9 @@
 SILE.inputs.common = {
-  init = function(fn, t)
-    local dclass = t.attr.class or "plain"
-    t.attr.papersize = t.attr.papersize or "a4"
+  init = function(doc, tree)
+    local dclass = tree.attr.class or "plain"
+    tree.attr.papersize = tree.attr.papersize or "a4"
     SILE.documentState.documentClass = SILE.require(dclass, "classes")
-    for k,v in pairs(t.attr) do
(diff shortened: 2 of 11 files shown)


alerque:
Stand-by on this, I seem to have broken the `-I` flag usage.

alerque:
I take that back, I didn't break `-I`, it was _never_ setup to do what I was trying to accomplish. So I fixed it. A Lua file can now be included as the preamble.

simoncozens:
Are you happy with this now?

alerque:
Yes. I have this branch as the main version I'm using in production. I have some ideas for improving it  more and the docs are a little dated now but as far as I'm concerned this code is where further work should start from and I know of no regressions.

alerque:
Status check?

simoncozens:
Sorry, I have a lot of work on at the moment so I am getting slower to respond to things. LGTM.

alerque:
Can you open that as an issue? I'm not sure I follow the scenario but obviously it needs review.

alerque:
Is this what #465 is about?

simoncozens:
Yep. Not sure if it was due to this patch but this was the last thing that touched this bit of code.

Issue #410, opened 2016-11-16
Windows binary distribution


simoncozens:
We need one. But I don't have the resources to make one.

To make this happen we will need:

- [ ]  A reproducible (hopefully scriptable) Windows build. See also #82. We just made this harder by requiring ICU.
- [ ]  A C wrapper around the SILE Lua script, so it can be more easily run as an EXE. (And without the Lua runtime.)
- [ ]  To isolate the libraries which are required, so that they can all be shipped as part of a Windows distribution
- [ ] To ship all the needed files as a distribution and test them on a virgin system ie with no ming32/msys2 installed.

indiamcq:
Simon,

I am working in mingw64 on Win 10. I have used your build from source instructions in this environment. I am thinking that this will create a sile.exe file, but as I don't know much about this stuff, I may be off track.

The ./configure is getting stuck on:
```
checking for FONTCONFIG... yes
checking for lua5.3... /mingw64/bin/lua5.3
checking if /mingw64/bin/lua5.3 is a Lua interpreter... no
configure: error: not a Lua interpreter
```
I had modified the configure.ac
from AX_PROG_LUA([5.1]) to AX_PROG_LUA

Also I modified the ax_lua.m4 to
```
  m4_define_default([_AX_LUA_INTERPRETER_LIST],
    [lua5.3 lua53 lua5.2 lua52 lua5.1 lua51 lua50 lua])
```
i.e it now includes lua5.3 that was missing. Though moving lua to the end may be wrong.

Any insights on what may be wrong? If what I am doing is off track then just say so.

Ian

simoncozens:
I've no idea why that's failing. The test it should be performing is:

```
$ lua5.3 -e "print('Hello ' .. _VERSION .. '!')"
Hello Lua 5.3!
```

Can you attach the output of config.log?

simoncozens:
Well, I just managed to replicate that error! Looking into it now.

simoncozens:
It's this: https://github.com/Alexpux/MINGW-packages/issues/949
Not sure how to fix it but I think the end goal for Windows is going to involve a custom SILE binary anyway, so I wonder if we could actually get rid of this test.

simoncozens:
OK, here's the state of SILE building on Windows so far:

* I'm building a Lua from source on cygwin/msys2/mingw64, using `make mingw`. It seems to work.
[...]

PR #445, opened 2017-02-08
Fix loading font data on Windows.

pkulchenko:
Fix for not properly loading font data on Windows as described in #410.


simoncozens:
In fact, probably all reads should be binary, what with UTF8 and everything. But this is a good start, thanks!

pkulchenko:
I briefly did a quick check on `io.open`, but it seems like other calls deal with text reads, which should be fine even with UTF8. This was the only one that was working with binary data after my cursory look.

Issue #444, opened 2017-02-05
Color spaces and ICC profiles.


lprndn:
Hi!

First of all, great work! I'm just starting to use SILE and I feel like it has a lot of potential.
I was just wondering if color spaces and ICC profiles were planned to be implemented? I know it's a complicated feature but I find it mandatory for a print focused tool.
I'm no dev but I started to look at libtexpdf and justenoughofit and it seems a function is not working (for icc loading?) in libtexpdf... does it make icc profiles impossible?

Thanks,

simoncozens:
Hi there! It looks like libtexpdf does have the functionality for loading ICC profiles but I haven't looked into any of these functions yet, let alone plugged them into SILE. (And they're not prefixed with `libtexpdf_` so won't be exported by the binary.) I think right now they're only used for loading JPEG images. What do you want to do with them?

lprndn:
Right now, I have nothing precise to do but I know I often need to include ICC profiles for printing.
If I'm not mistaking, it's also needed for PDF/X PDF/A conformity.
I don't totally understand how profiles affect PDF ( or is it just embedded and that's all?)
I might also try to make `pdf_setcolor` able to deal with with CMYK. It doesn't seem too hard... I'll tell you if I succeed...

alerque:
Just for cross reference PDF validation is an open issue, see #391.

And by the way of commentary, the colors in the cover of the last book I sent to press came back wildly off and I'm pretty sure the lack of color profiles (and some combination of the print shop's bad sloppy calibration or bad defaults) were to blame. This wasn't SILE's problem since the SILE typography PDF was post-processed and layered into an SVG that was then rendered from Inkscape, but the final didn't have a color profile and it got me in trouble.

I also remember from some LaTeX projects that had raster images that were black and white engravings that I was getting wildly different printing results (overall contrast out of whack) across printers until I started embedding ICC profiles in the PDF and the results across different systems got a lot more consistent.

shirat74:
Hi,

Is there any progress on this topics?
I am the one who implemented the most part of dvipdfmx's (thus libtexpdf) font and graphics support.
Maybe I can help you.

simoncozens:
Hi - I asked this on the TeX-live list but I didn't get any response. I have not done anything with ICC because the code in libtexpdf is commented out and described as "not working". In `pdfcolor.h`:

```C
/* Not working */
extern int      pdf_colorspace_load_ICCBased      (const char *ident,
                           const char *profile_filename);
```

[...]

PR #448, opened 2017-02-18
Add support for CMYK and grayscale colors.

lprndn:
Hi!
As said in #444, I tried to implement cmyk and grayscale colors in SILE.
Everything seems to work as you can see in this quick example: [colors.pdf](https://github.com/simoncozens/sile/files/785048/colors.pdf).
Might not be the best implementation as I'm a newbie in lua  but it doesn't break anything either, which is nice, I think.
So now, you can set colors as you used to but you can also specify each channel's value separated by a space (between 0 and 255, decimals separated by a comma `125.65`).

`colorparser` will then choose which type of color it is depending of the number of channels: 
`\color[color = 125]{1 for grayscale}`
`\color[color=0 0 255]{3 for rgb}`
`\color[color=255 255 0 0]{4 for cmyk}`

Should help a lot to use cmyk colors when printing pdfs even if ICC profiles are still needed for consistent colors.

Have a nice day.

### DIFF
--- a/core/colorparser.lua
+++ b/core/colorparser.lua
@@ -154,25 +154,25 @@ SILE.colorparser = function(col)
     local c = colorsRGB[col]
     return { r = c[1] / 255, g = c[2] / 255, b = c[3] / 255 }
   end
-  local r, g, b =  col:match("#(%x%x)(%x%x)(%x%x)")
+  local r, g, b = col:match("#(%x%x)(%x%x)(%x%x)")
   if r then
-    return { r = tonumber("0x"..r)/255, g = tonumber("0x"..g)/255, b = tonumber("0x"..b)/255,}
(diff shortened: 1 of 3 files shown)


alerque:
This looks like a great feature add. I'll leave it for @simoncozens to decide if this is how he wants to implement it, but I took the liberty of cleaning up the lua coding style just a touch and normalizing the way patterns are matched against the way SILE was already doing it. This shouldn't be functionally any different.

lprndn:
Thanks! I'll try to be tidier next time!

simoncozens:
Lovely, thank you very much!

alerque:
I just realized one thing missing from this PR was documentation ;-) We should probably make a note about this in the manual.

PR #451, opened 2017-03-05
Update libtexpdf-output.lua

tcyrus:
Replaced `pdf.colorpush` with `pdf.colorpush_rgb` (`pdf.colorpush` was removed from `src/justenoughlibtexpdf.c` in f0deadb).

This allows `fold_unless_fail 'Compiled documents' 'compile_docs' ./tests/attack.pl` to exit successfully in Travis CI builds.


simoncozens:
Oops! Nice catch, thank you!

PR #484, opened 2017-09-02
readme: fix showoff.pdf link

paulirish:
It was removed from the repo, I linked up the last time it was in there.
Also using rawgit otherwise github will force a download.


pabru:
+1, was about to submit a corrective PR for this as well

simoncozens:
Thanks for this!

PR #488, opened 2017-09-19
replace odd \em{ } in what-is page with markdown equiv

timo:
also a few lines later there's a \supereject that i don't know what to do with


alerque:
Thanks @timo. It looks like Simon must have copied that content from the manual in SILE's TeX like format and didn't get it completely stripped. I've  fixed the other stray command as well.

alerque:
...and the Jekyll build system appears to be broken for sub-pages. I think Github must have done something since the last commit because this was working a minute ago. Arg.

alerque:
For some reason the value for _baseurl_ being set in `_config.yml` is not being output in the template. I can find this documented but looking around at other people's recent solutions it seems like Github must have started running Jekyll with `serve --baseurl=""` for sites with CNAMES. I'll commit something that's less elegant than my previous URL solution but works for the case where they are forcing the variable value while still allowing testing on a local server with an extra path element.

PR #492, opened 2017-10-27
core/typesetter.lua: treat leading tabs as spaces in .sil

escondida:
I like to indent lines within environment blocks, and typically indent using tabs. In LaTeX, leading whitespace is irrelevant. However, the following input results in a document with a slashed square representing a tab, rather than the line with no leading whitespace characters as expected.

```TeX
\begin{document}

\script[src=packages/pullquote]

\begin[author=Doctor Clayton Forrester]{pullquote}
	Ha ha ha! Baseball! Great American pastime.
\end{pullquote}

\end{document}
```


simoncozens:
I like the idea, but doesn't `%s` match a tab character?

```lua
> =(("\t"):match("%s") and "yes" or "no")
yes
```

simoncozens:
OK. I can reproduce this, but it still confuses me! Maybe `pullquote` is doing something odd, because it doesn't happen when you start a normal paragraph with a tab.

But thinking about it, I wonder if, more generally, we should instead replace *all* tabs with spaces, and then if there are spaces at the start of a paragraph, discard them all. I don't know that it's useful to have tabs in a document, and of course it's more TeX-like. What do you think?

escondida:
Hey, thanks for the response!

I think that when it comes to the normal document environment, there's no compelling reason to treat tabs and spaces differently. TeX treats leading ones/multiple ones between words as equally irrelevant, which I'd personally argue is the "Right Thing"™.

I don't know that it's good to do away with them altogether, though; the most important place to keep them intact would be in preformatted environments such as code listings.

I'll see if I can figure out what's happening in pullquote.

escondida:
I'm sorry to say that although it's a short file, I have no idea what in pullquote is causing the problem.

alerque:
This has nothing to do with the pullquote package or _leading_ tabs. SILE is doing this for all tabs appearing _inside_ a line. The existing code already strips them from the beginning of the line, but it just happens that when using the pullquote package the line is prefixed with something else, hence the tab is getting treated the same way one it the middle of a string does.

```TeX
\begin{document}
Tab here>	<.
\end{document}
```

The solution is going to be to substitute tabs for spaces internal to all strings before shaping. Is there any reason we can't go ahead and do that?

alerque:
If merged this will fix #561.

alerque:
This is rightly failing tests because one of our tests has a tab in the body content (used to test escapes). @simoncozens if you think this is the right approach I'll rebase this on master and fix the test so it can be merged.

alerque:
I redid this again because I realized my previous attempt would ruin multiple spaces being used for a purpose with `obeyspaces` set. This should treat tabs as a single space _each_.

I'm still working on making the test pass again. For some reason even using the font file from Ubuntu Xenial is resulting in different metrics for me. *Arggg.*

PR #495, opened 2017-11-29
Pass through Lua code in script

### LINKED ISSUE #92: \script command interprets braces as grouping operators
`\script{ a = { "a"; "b"; "c" } }` will fail because the parser will eat the braces and not pass them to Lua.

### LINKED ISSUE #450: `%` in inline scripts
Hi,

I'm new to SILE and not yet sure if I got it's "TeXlike" syntax right, but I could not find anything about this in SILE's official documentation. I was expecting the following MWE to work:

    \begin{document}
      \script{SILE.typesetter:typeset(tostring(5 % 3))}
    \end{document}

i.e. typesetting a "2". Yet SILE (current git master) fails with:

    Error detected:
    ["\begin{document}\n\sc..."]:3: parse error, } expected
    \end{document}
    ^

Obviously because the `%` is parsed as a comment in TeXLike mode. During ordinary text input you can escape it with a backslash in order to typeset a `%`. I've also tried that in the MWE above, but SILE then fails with `! [string "SILE.typesetter:typeset(tostring(5 "]:1: ')' expected near <eof> at test.sil l.2, col.1`.

I've read in #92 that "TeXlike-mode is just for toy examples and tests" and so this might be considered a wont-fix, but then it might be helpful to mention in the official docs that in TeXlike mode all special characters (`\`, `{`, `}`, and `%`) cannot be used in inline scripts? I'd love to be able to use `%` in inline scripts as you could then use `string.format()` inline as well.

But, perhaps I've just missed some relevant section in the documentation and there is already some special syntax for this?

alerque:
This allows the literal characters `%` and balanced pairs of `{` `}` to appear in either command (`\script{...}`) or environment (`\begin{script}...\end{script}`) forms of the **script** command and be passed through as-is to Lua for evaluation.

Todo before merging:

- [x] Make sure we're passing regression tests on everything that used to build.
- [x] Update documentation to describe new  allowed syntax and limitations.
- [x] ~~Refactor LPEG capture groups or add AST tree cleanup so we're not passing as many arbitrary data types further down the parsing pipeline than we need to be.~~

This will close #450 and closes #92.


alerque:
Hmm it looks like the only reason escape sequences aren't allowed currently is because of something I overlooked in the LPEG grammar. It should be possible to get them working properly so 100% of valid Lua code would be allowed, minus whatever the closing tag (or unmatched closing brace) is.

alerque:
I updated this to fix handling Lua string escape sequences (and remove the caveat about them from the docs).

alerque:
Refactoring the LPEG groups is going to be a different kind of job that this PR, so I'm nixing that item. It looks like the right way to do that is to compress the tree one more level than it is (so what is now passed around as `content[1]` from TeXlike functions becomes just `content`). This would also normalize the use of function  from Lua vs. from TeX-flavor markup, but that's a major change and I don't see any simpler way of cleaning up the grammar. So I cleaned up the coding style and we'll call it good for this PR.

I've tested this on my own machine and the metrics I get for the current _master_ branch. This doesn't seem to introduce any new regressions.

Issue #505, opened 2017-12-25
Allow including TeX-flavor fragments instead of only whole documents


kappa2222:
Hello. I have two `.sil` files

    % include/test.sil
    \define[command=xxx]{expanded thing}

and

    \begin{document}
    % aaa.sil
    \include[src=include/test.sil]
    % the following is line 10
    \xxx       
    \end{document}                                                

When I compile this example I get:

    This is SILE 0.9.4
    <../somefolder/aaa.sil>
    <../somefolder//include/test.sil>

    ! Unknown command xxx at include/test.sil l.10, col.1

I invoke sile like:

     # from the folder where I built sile
    ./sile -o ../somefolder/cv.pdf ../somefolder/aaa.sil

At first I thought Sile was not reading the file to include, but it's not like that because if I pass \\include a wrong path I get a meaningful file-not-found error. Notice also the filename in the error message (which refers to the included) and the line (which refers to the includee).

simoncozens:
Urgh, this is really bad. Sorry.

Note to self: `SILE.inputs.TeXlike.process` does not seem to do anything when a documentclass is already defined. I blame @alerque for 97a4e3e8, but I'm not sure if that's fair or not.

alerque:
Is the `% include/test.sil` bit in your .sil file or was that just for the sake of your Github post? If the former this might be an iteration of #502.

I'll try to have a look today and see if I can figure out what's up.

kappa2222:
> Is the % include/test.sil bit in your .sil file or was that just for the sake of your Github post

For the sake of the github post.

alerque:
Wow yes this is as sticky little cart/horse problem. Our file type detection wants to have its cake and eat it too, but I'm struggling to visualize how we can have both sets of nice things. I suspect the answer is going to be to stop trying to guess file types and just expect one of the other. We can't simultaneously sniff file types, allow valid comment syntaxs, and validate our own parsing tree.

With auto-detecting whether percentage units were vertical or horizontal the "smart" behavior turned out to be both finicky and limiting. Requiring the user to specify the axis both reduced unexpected behavior quirks and enabled more complex  layouts. Likewise I think sniffing every text file and input string and trying to figure out if it's Lua, TeX, or XML will turn into a fiasco. At the very least we'll have to stop pretending the first few bytes actually tell us anything and actually try to feed the whole blob to each parser until one of them doesn't croak.

alerque:
This issue is a duplicate of #465.

[...]

PR #507, opened 2017-12-26
Fix includes of TeX-flavor content

### LINKED ISSUE #465: TeX-like input filter doesn't process when called from another file
Because `SILE.process(tree)` is only called when `SILE.documentState.documentClass == nil`, a file included from another is never processed. See examples:

alerque:
This should fix the issue with included TeX-flavor markup being silently dropped.

- [x] Whip up something that fixes #465 ~~~(and #505 which is essentially the same but subtly different in that the include does not have its own document tag).~~~
- [x] ...while not breaking #414.

### DIFF
--- a/core/inputs-texlike.lua
+++ b/core/inputs-texlike.lua
@@ -142,14 +142,14 @@ end
 function SILE.inputs.TeXlike.process (doc)
   local tree = SILE.inputs.TeXlike.docToTree(doc)
   local root = SILE.documentState.documentClass == nil
-  if root then
-    if tree.tag == "document" then
+  if tree.tag then
+    if root and tree.tag == "document" then
       SILE.inputs.common.init(doc, tree)
-      SILE.process(tree)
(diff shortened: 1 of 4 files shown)


alerque:
The secondary issue in #505 (not having a wrapping document tag at all) is different and I'm going to leave that for another time (see discussion in #508).

alerque:
I think this would be ready to merge, but CI tests didn't pass. Except that if you look at the one that failed it looks like everything passed. Any idea what's up with that @simoncozens?

simoncozens:
Ignore Appveyor. It’s trying to build on Windows.

alerque:
@simoncozens It wasn't Appveyor I was worried about. It was Travis ... one of the three runs was failing for no apparent reason. It appears to have been an issue with the coverage output and not the actual build. I reran it several times when I posted this PR and it kept failing, today re-running it succeeds.

simoncozens:
What a star. Thank you for working on this.

PR #517, opened 2018-02-20
Add a hbox debugging tool

### LINKED ISSUE #515: Question: how can I draw a bounding box around pushHbox?
To improve the prototype implementation of math support (#220), I need to do various tweaks and fixes to layout and rendering of the generated hbox. But I'm now having trouble observing/pinpointing what exactly is wrong. Ability to force SILE to draw a "debugging" bounding box around the hbox (generated with `SILE.typesetter:pushHbox(...)`) would be very useful. Does SILE provide some feature which could let me do something like that?

alerque:
Fixes #515

Use it like this in your hbox’s output routine:

    SILE.outputter.debugHbox(self, self:scaledWidth(line))

### DIFF
--- a/classes/plain.lua
+++ b/classes/plain.lua
@@ -186,6 +186,7 @@ SILE.registerCommand("hbox", function (options, content)
       if typesetter.frame:writingDirection() ~= "RTL" then
         typesetter.frame:advanceWritingDirection(self:scaledWidth(line))
       end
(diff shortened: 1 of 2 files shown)


alerque:
I thought I commented on this PR again, but apparently it didn't send or I was sleep walking or something.

I would hold of merging this until it's documented. I was looking at the debug docs and think not only the flag I added but a couple others that are missing could fit well in this PR.

simoncozens:
Merged to devel.

PR #518, opened 2018-02-20
Stuff that does something?

alerque:
This is stuff from Simon's devel branch, rebased on master and with commits related to other issues removed (since they are in separate PR's with Travis having a look at them now).

I don't really know what this stuff is or weather it's ready for merging, but here it is collected in one place for CI testing. Is there more to come here?


simoncozens:
Yes, these are a few minor fixes suitable for merging to master.

Issue #312, opened 2016-05-04
Is there a way to detect the context of a command from a nested one?


alerque:
Lets say I  have something like this:

```TeX
\begin[papersize=a6]{document}
\define[command=em]{\font[style=Italic]{\process}}
\define[command=em2]{\font[weight=800]{\process}}
\define[command=blk]{\font[style=Italic,size=0.8em]{\process}}

This is a normal paragraph with an \em{emphasised bit} in the middle. Here italics is the right form of emphasis.

\begin{blk}
This is a block quote paragraph with an \em{emphasised bit} in the middle. Lets say it’s already in italics or in a font that does not have italics. So we need some other treatment. 
\end{blk}
\end{document}
```

The issue here the source text is coming from Markdown where there is only one kind of emphasis markup and the transformation to SILE knows little about the context.

What I really need is to be able to use the `\em` command throughout the document but have it behave slightly different depending on context. Ala cascading style sheets. Is there a way I can define the command such that it can pick up on the typsetter's context and switch depending on whether it was in or out of the block quote? Should I be setting scratch variables on begin/end and testing for those or is there a better way conceptually?

simoncozens:
The easiest way would be to override `SILE.process` so that it maintains a context stack, and then query that.

PR #546, opened 2018-12-29
Hansel and Gretel

alerque:
A while ago in #312 I asked about detecting the context a command was executed from. I've experimented with several ways of doing this and this is the most useful one I've come up with so far. Overloading `SILE.process()` tends to turn up far too much noise to be useful. This hijacks the `SILE.registerCommand()` function and creates functions that are wrappers that push and pop their own names off a stack of command names.

I was going to put this tracking behind a flag on some sort, but I did some profiling on an old slow machine compiling the manual and came up with 80.82 seconds with the original sile version and 81.04 with the breadcrumb tracker. In other words the difference is not appreciable. In fact to even come up with the difference that direction I had to run both versions several times. The difference is system noise.

This enables some useful things like getting a list of all the commands in the current stack (i.e. all the open nested function contexts). There is a `dump()` function for quick and dirty debugging, you can turn on `sile -d breadcrumbs` to see every entry and exit as it happens, but perhaps most interestingly you can have a look at the context directly by looking at the name of the `parent()` function (or nth parent `parent(2)`) or query if there are any parents with a given name with `contains(name)`.

```lua
SILE.typesetter.breadcrumbs:dump()
local grandparent = SILE.typesetter.breadcrumbs:parent(2)
local in_ragged = SILE.typesetter.breadcrumbs:contains("ragged")
```

Note due to a race condition where the baseclass is fired up before a typesetter exists, this does not track `define()`, `process()`, or `comment()` — but the noise these produce suggests that might be a good thing anyway.

### DIFF
--- a/core/baseclass.lua
+++ b/core/baseclass.lua
@@ -1,7 +1,7 @@
 SILE.Help = {}
 
 SILE.registerCommand = function (name, func, help, pack)
-  SILE.Commands[name] = func
+  SILE.Commands[name] = SU.breadcrumbs(name, func)
   if not pack then
     local where = debug.getinfo(2).source
     pack = where:match("(%w+).lua")
--- a/core/baseclass.lua
+++ b/core/baseclass.lua
@@ -1,7 +1,11 @@
 SILE.Help = {}
 
 SILE.registerCommand = function (name, func, help, pack)
-  SILE.Commands[name] = SU.breadcrumbs(name, func)
+  if SILE.typesetter and SILE.typesetter.breadcrumbs then
+    SILE.Commands[name] = SILE.typesetter.breadcrumbs(name, func)
+  else
+    SILE.Commands[name] = func
+  end
   if not pack then
     local where = debug.getinfo(2).source
     pack = where:match("(%w+).lua")
--- a/core/baseclass.lua
+++ b/core/baseclass.lua
@@ -1,11 +1,7 @@
 SILE.Help = {}
 
 SILE.registerCommand = function (name, func, help, pack)
-  if SILE.typesetter and SILE.typesetter.breadcrumbs then
-    SILE.Commands[name] = SILE.typesetter.breadcrumbs(name, func)
-  else
-    SILE.Commands[name] = func
-  end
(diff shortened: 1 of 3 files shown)


alerque:
Nevermind my last comment, I've hacked at that a few times and come up with different ideas each time and none that I like. IT shouldn't hold up this feature any more. I use the `parent()` function in several places in production and it would be nice to have this merged.

simoncozens:
A question - the last patch which provides per-typesetter breadcrumbs only sets breadcrumbs on commands defined when there is a typesetter declared. Is this necessary? Are there situations where one can declare a command before `SILE.typesetter` is loaded?

alerque:
That's an interesting thought, but I don't think there is. My first implementation kept a breadcrumb trail independent of the typesetter and I had nothing but trouble with it. It just didn't make sense because sometimes you switch back and forth between typesetters and the chain of things that affect functions don't make any sense outside of the actual chain of things the typesetting is _doing_.

A look at the code shows that a starter typesetter is declared in `core/typesetter`. The `registerCommand()` itself is in `core/baseclass`, which is earlier in the chain. In between there really isn't a place for any user defined functions to get declared, so the only question would be is SILE itself declaring anything in between there. The answer is yes: `hyphenator:add-exceptions()` and `set()` get defined in the interum and thus would be exempted from breadcrumbs. Given that neither process content or have children, I think that actually makes sense.

If you have a better idea how to track chains of command other than attached to the thing that's actually processing them I'd be open to hearing it, but this makes more sense than anything else I can think of.

simoncozens:
Yeah, I think it all makes sense, I was just thinking about this:

```lua
  if SILE.typesetter and SILE.typesetter.breadcrumbs then
     SILE.Commands[name] = SILE.typesetter.breadcrumbs(name, func)
   else
     SILE.Commands[name] = func
   end
```

If that condition is always going to be true, we can scrap the `if` statement.

alerque:
Unfortunately the condition won't always be true. As noted above there are a couple exceptions of functions that get registered before the typesetter does.

What will be true is that if we have a typesetter we will have the breadcrumbs object to go with it, so that part of the test can be simplified to only check for the first bit. I just adjusted that bit of code.

Issue #555, opened 2019-03-05
Empty simpletable


ck3d:
The package _simpletable_ typesets nothing.

# Example
The document developers (file _documentation/developers.sil_) contains a table which is forwarded to the package _simpletable_ in file _documentation/macros.xml_. The PDF shows white paper at the pace where the table should be.

PR #558, opened 2019-03-08
fix empty simpletable #555

ck3d:
I try to learn and understand SILE since a few days. I do not really
know what I am doing, but I could somehow fix the issue #555 .
A problem is that the command noindent is only applyed to the first
cell.

I hope to get some feedback to make the PR ready.


simoncozens:
This looks like a much tidier way of doing things anyway!

What I would do, before entering the table is temporarily set `document.parindent` to zero, to turn off indentation altogether.

ck3d:
Thanks for reviewing. I updated the PR according to your feedback and made the PR ready to merge.

alerque:
Thanks for the contribution @ck3d.

PR #562, opened 2019-03-24
List actual Debian/Ubuntu packages needed

jamadagni:
The list is updated to refer to packages which need to be installed on a system which has never used Lua.

Modified luarocks line because at least on Ubuntu the given `apt-get` command is enough to get `sile` to compile. 

Fixes “serious” > “series” error BTW.


alerque:
Thanks for the contribution. This is an improvement for sure, so I'll go ahead and merge it, but it looks like this section could use an even heavier overhaul. I won't rebuild the PDF version just yet.

PR #563, opened 2019-03-25
WIP: Cleanup installation documentation

alerque:
This is a follow up for #562. How does this look to you @jamadagni? 

I suggest _not_ merging this until closer to the next release. Before the next release comes out this should be rebased and the PDF rebuilt, then merged.


jamadagni:
I have some more corrections for the doc. Will send later.

alerque:
What's your status on this @jamadagni? No pressure, I'm just wondering if I should consider merging this now or hold off until we can combine more related work into one merge.

jamadagni:
@alerque Sorry for the delay. I don't have too much more for now as I got distracted by other work. Please see #594 for what I had on hand.

PR #564, opened 2019-03-29
Add `makedefault` option to set command for changing default setting value

harrysummer:
One of the user case is to replace the default font of the document. Since the default font is hard-coded into the script. Everytime when execute SILE.settings.reset, we will go back to that font. This behavior
is not desired when you are using book class with footnotes and folio.

For example, with the following piece of code, the master frame would be set to Source Han Serif font, while the footnotes and folio frames are still the default Gentium Plus (on my computer, it is fallback to simsun):

```latex
\begin[class=book, papersize=100mm x 100mm]{document}
\script[src=packages/lorem]
\set[parameter=font.family, value=Source Han Serif CN]
\lorem[words=10]\footnote{\lorem} \lorem[words=10]
\end{document}
```

image

After applying this patch, we added a `\set-default` command, which can change the default value for all following codes. So the following code:
```latex
\begin[class=book, papersize=100mm x 100mm]{document}
\script[src=packages/lorem]
\set-default[parameter=font.family, value=Source Han Serif CN]
\lorem[words=10]\footnote{\lorem} \lorem[words=10]
\end{document}
```
will generate this:
image

### DIFF
--- a/core/settings.lua
+++ b/core/settings.lua
@@ -9,27 +9,21 @@ SILE.settings = {
   declarations = {},
   stateQueue = {},
   defaults = {},
-  dirty = {},
-  dirtyQueue = {},
   pushState = function()
     table.insert(SILE.settings.stateQueue, SILE.settings.state)
-    table.insert(SILE.settings.dirtyQueue, SILE.settings.dirty)
     SILE.settings.state = std.table.clone(SILE.settings.state)
   end,
   popState = function()
     SILE.settings.state = table.remove(SILE.settings.stateQueue)
-    SILE.settings.dirty = table.remove(SILE.settings.dirtyQueue)
   end,
   declare = function(t)
     SILE.settings.declarations[t.name] = t
     SILE.settings.set(t.name, t.default)
     SILE.settings.defaults[t.name] = t.default
-    SILE.settings.dirty[t.name] = false
   end,
   reset = function(t)
     for k,_ in pairs(SILE.settings.state) do
       SILE.settings.set(k,SILE.settings.defaults[k])
-      SILE.settings.dirty[k] = false
     end
   end,
   get = function(name)
@@ -48,7 +42,6 @@ SILE.settings = {
       SU.error("Setting "..name.." must be of type "..wantedType..", not "..t.." "..value.."\n"..name..": "..SILE.settings.declarations[name].help)
     end
     SILE.settings.state[name] = value
-    SILE.settings.dirty[name] = true
   end,
   temporarily = function(f)
     SILE.settings.pushState()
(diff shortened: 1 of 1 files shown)


alerque:
Thanks for addressing this issue. I've run into it before. I have some questions about this implementation. Something doesn't feel quite right, but the problem it addresses definitely does need fixing. I'll give it a bit of a review when I get a chance and compare to how I've worked around this in the past.

alerque:
I don't understand why all the smoke and mirrors with a parallel set of "dirty" queues is necessary. Why not just set the value in `SILE.settings.defaults` and be done with it?

harrysummer:
@alerque My original thinking was just adding a command to change the default settings. However, when I tried to implement it, I found that only changing the default value might not be sufficient.

For example, assuming we have a package implementation which uses a new set of settings by calling:
```lua
SILE.settings.pushState()
SILE.settings.reset()
-- typesetting code of the package
SILE.settings.popState()
```
And we assume that in the outer environment, we didn't call any `SILE.setings.set()` to chang e the setting value. Then, what if we call `\set-default` inside the package? (This might not very common though, but I just want to make the setting system robust and complete for all cases) The natural behavior is to change the current value of the outer environment too, because it wan't been detained. This is where is dirty queue come from. When we are poping through the state stack, we will replace the setting value if the value is not detained. (I missed some part in the `pop` function though) Hope this can help you understand my design decisions.

I agree that your idea of implementing as  a flag in existing `\set` command is brilliant. I think I will go with that by adding a `default` boolean option in `\set` command.

alerque:
The scenario you outline is an edge case I hadn't considered, but I'm not sure I agree with your desired output. If I call `popState()` in my package I would _expect_ it to revert to the state that was pushed even if that state was the default at the time but no longer is. I don't think monkeying with that is actually an improvement. If a package monkeys with the defaults it shouldn't be doing so after a reset in a temporary typesetter anyway.

Note if you rework this to be part of `set()`, be sure to use `SU.boolean()` like this to check the flag so _false_ and _nil_ values get handled correctly:

```lua
local makedefault = SU.boolean(options.makedefault, false)
```

simoncozens:
I agree with alerque: temporary changes (even to defaults) should be temporary. Consider the case where you're actually typesetting two documents in parallel. (Yes, this does happen.) The two sets of settings should be isolated.

harrysummer:
Updated. Handled default value changing as `makedefault` boolean option of `\set` command.

Now I can write
```latex
\begin[class=book]{document}
\set[parameter=font.family, value=Source Han Serif CN, makedefault=true]
Hello\footnote{Hello}
\end{document}
```
And both master frame and footnote will be the same font.

PR #565, opened 2019-03-30
Change default font weight

harrysummer:
400 is not a good default value for font weight. According to [fontconfig source](https://gitlab.freedesktop.org/fontconfig/fontconfig/blob/master/fontconfig/fontconfig.h#L147), the weight for regular font should be 80.

### DIFF
--- a/core/font.lua
+++ b/core/font.lua
@@ -52,7 +52,7 @@
 
 SILE.settings.declare({name = "font.family", type = "string", default = "Gentium Plus"})
 SILE.settings.declare({name = "font.size", type = "number or integer", default = 10})
-SILE.settings.declare({name = "font.weight", type = "integer", default = 400})
+SILE.settings.declare({name = "font.weight", type = "integer", default = 80})
 SILE.settings.declare({name = "font.variant", type = "string", default = "normal"})
 SILE.settings.declare({name = "font.script", type = "string", default = ""})
 SILE.settings.declare({name = "font.style", type = "string", default = ""})


alerque:
I don't think this is correct. There must be a scale conversion somewhere because 400 is actually a quite standard number to use for "normal" weight, and this works on my systems to using font families with dozens of weights to select the "normal" one. Dropping down to 300 in some font families will start selecting light or thin variants.

harrysummer:
> I don't think this is correct. There must be a scale conversion somewhere because 400 is actually a quite standard number to use for "normal" weight, and this works on my systems to using font families with dozens of weights to select the "normal" one. Dropping down to 300 in some font families will start selecting light or thin variants.

Thanks for commenting on this! After I read your comment, I digged deeper into the source codes. The conversion from OpenType weight (regular=400) to fontconfig weight (regular=80) is defined in the source code of fontconfig (see [here](https://gitlab.freedesktop.org/fontconfig/fontconfig/blob/master/src/fcweight.c#L35), the `FC_WEIGHT_REGULAR` macro is defined to 80), so that when fontconfig loads a OpenType font, it will convert its font weight value (as defined [here](https://docs.microsoft.com/en-us/typography/opentype/otspec180/fvar#VAT)) to fontconfig's scale. However, an issue in SILE could be, it directly passes the font information in `SILE.settings` to `justenoughfontconfig` to get the font face (from [here](https://github.com/simoncozens/sile/blob/master/core/typesetter.lua#L216) and eventually to [here](https://github.com/simoncozens/sile/blob/4f63bfc7fc8155792c7aeec8653c668f8e050615/core/harfbuzz-shaper.lua#L61)). If we assume the font weight used in SILE is OpenType standard font weight (400 for regular), we must convert the weight into fontconfig weight before send it to `justenoughfontconfig._face`.

Let me correct this pull request to convert the font weight before send it to fontconfig. I think I will do that in `justenoughfontconfig.c`.

harrysummer:
Found out the weight is [already converted](https://github.com/simoncozens/sile/blob/ee610d89620aa6d30978aadac142c85b9ca8d4d3/src/justenoughfontconfig.c#L57) in `justenoughfontconfig.c`. Let me close this PR. Maybe it's my fonts.conf file where something get wrong. Great typesetting software!

PR #566, opened 2019-03-31
Mask font index from fontconfig to lower 16-bit

harrysummer:
Since fontconfig 2.11.95, fontconfig [introduced TrueType GX support](https://github.com/freedesktop/fontconfig/commit/27d61f1ddcda5543e9c6440a0f8794caa0b1eac7), which changes the meaning of `index` property in fontconfig. The field is originally used for specifying the index of the font face in a font file which contains multiple font faces. However, since 2.11.95, fontconfig start to use its higher 16-bit to hold the number of named instances available for the current face if we have a GX or OpenType variation (sub)font (which is quite similar to the [style_flags](https://github.com/aseprite/freetype2/blob/master/include/freetype/freetype.h#L905) option in freetype). However, libtexpdf does not support GX and `index` is only treated as the font face index. When font with GX feature is rendered, libtexpdf could crash because SILE passes a very large font face index to it (when the high 16 bits of variation instance is non-zero). To handle the crashing in SILE, considering libtexpdf does not support GX yet, the only reasonable thing left to us is masking the low 16 bits of fontconfig `index` property.

### DIFF
--- a/core/harfbuzz-shaper.lua
+++ b/core/harfbuzz-shaper.lua
@@ -1,6 +1,7 @@
 
 if not SILE.shapers then SILE.shapers = { } end
 local hb = require("justenoughharfbuzz")
+local bit32 = require("bit32-compat")
 
 SILE.settings.declare({
   name = "harfbuzz.subshapers",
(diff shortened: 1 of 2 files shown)


harrysummer:
A better solution could be, testing the index value at Lua's side, generating a warning message and masking to lower 16 bits.

harrysummer:
Updated according the previous comment.

simoncozens:
Thanks for this; I like this solution, which will tide us over until we get proper variable font support.

PR #570, opened 2019-04-04
Create package to handle everything Pandoc might throw at us

alerque:
I've been using an automated book publishing workflow starting with canonical sources in Markdown, processed through Pandoc, written out to SILE for typesetting, and the result sent directly to press for some time now. A dozen books and dozens of other publications later, this workflow has been through several iterations:

1. I started with a hacked version of Pandoc that butchered the LaTeX output, crammed it in a SILE document wrapper, added some command patches so that it didn't just get thrown back up, and crossed my fingers. Books went to press this way, but it sure wasn't a process I could hand off to anybody else.

2. I then hacked together a new document writer for Pandoc (still modeled after the LaTeX one) that wrote out something a lot more SILE looking. This included ripping out everything it couldn't cope with anyway (i.e. tables, figures, math, etc.) and using simplified "sile-ish" commands. Still the default out of the box SILE didn't have all the necessary commands, so the template it was producing had to include definitions for them. More books went to press this way, but the Pandoc code was still to messy to contribute upstream.

3. More recently I started realizing that there was no reason _not to_ take more radical departures from the LaTeX like output and make the output more consistent and similar to the Pandoc's internal AST.  Trying to target "sile" as an output format is difficult when the format itself is still relatively incomplete. I started re-hashing the output with this in mind and discovered it was a lot easier to implement all the commands necessary in Lua than trying to recreate the expected document behavior from existing commands. I started a Lua library of my own to add SILE commands for everything Pandoc generates. Another book has gone to press this way.

Development got a lot easier once I got well into phase 3. Part of this is I'm much better with Lua than Haskell, but the other part of it is that Pandoc doesn't have to try to cobble together something that works, it can just pass off all the hard work to the typesetter and forget about it. **This has the added benefit of giving end users a way to manipulate the result much easier.** For example rather than trying to teach Pandoc how to output a rule using `\hrule[height=0.2pt,width=100%lw]` which makes the resulting document difficult to re-style if desired, it would just output `\HorizontalRule` (the command name matching the internal AST for the element type), the SILE command would have some defaults for this, and the user could override them in a style sheet if desired.  

As such I would like to propose that — instead of trying to teach Pandoc a bunch of "tricks" telling SILE exactly how to output each element of a document, we add a package that teaches SILE what to do with all the document elements Pandoc might ever throw at it. A template would still be needed on the Pandoc end, but this would be as bare bones as possible and basically just import the one package from SILE. Every element would be named to match the way Pandoc thinks about them, and all attributes would be passed on.

This PR isn't ready for merging yet, but it is a start to this approach and I wouldn't mind hearing some feedback. If this works it would make it relatively easy to finish and contribute the Pandoc end of things upstream.

### DIFF
--- a/packages/pandoc.lua
+++ b/packages/pandoc.lua
@@ -29,6 +29,14 @@ end)
 
 SILE.registerCommand("span", function (options, content)
   handlePandocArgs(options)
+  for _, class in pairs(options.classes) do
+    SU.debug("pandoc", "Add div class "..class)
+  end
+  SILE.process(content)
+end,"Generic inline wrapper")
+
+SILE.registerCommand("div", function (options, content)
+  handlePandocArgs(options)
   SILE.process(content)
 end,"Generic inline wrapper")
 
@@ -104,6 +112,7 @@ Modified from default:
 Provided specifically for Pandoc:
 
 \listitem \code{span}
+\listitem \code{div}
 \listitem \code{emph}
 \listitem \code{smallcaps}
 \listitem \code{strikeout}
--- a/packages/pandoc.lua
+++ b/packages/pandoc.lua
@@ -27,42 +27,53 @@ SILE.registerCommand("rule", function (options, _)
   SILE.call("hrule", options)
 end)
 
-SILE.registerCommand("textem", function (options, content)
+SILE.registerCommand("span", function (options, content)
+  handlePandocArgs(options)
+  SILE.process(content)
+end,"Generic inline wrapper")
+
+SILE.registerCommand("emph", function (options, content)
   handlePandocArgs(options)
   SILE.call("em", {}, content)
 end,"Inline emphasis wrapper")
 
-SILE.registerCommand("textstrong", function (options, content)
+local oldStrong = SILE.Commands["strong"]
+SILE.registerCommand("strong", function (options, content)
   handlePandocArgs(options)
-  SILE.call("strong", {}, content)
+  oldStrong(options, content)
 end,"Inline strong wrapper")
 
-SILE.registerCommand("textsc", function (options, content)
+SILE.registerCommand("smallcaps", function (options, content)
   handlePandocArgs(options)
   SILE.call("font", { features = "+smcp" }, content)
 end,"Inline small caps wrapper")
 
-SILE.registerCommand("textup", function (options, content)
+SILE.registerCommand("csl-no-emph", function (options, content)
   handlePandocArgs(options)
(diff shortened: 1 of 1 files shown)


simoncozens:
I think this is the right approach. I even wonder if it actually obviates the need for a separate SILE format altogether. If we're just passing in something that reflects the AST, something more "generic" - such as an XML output that echoes the pandoc AST - would be more useful (and more acceptable as a pull request to pandoc).

Worst case you can write something in Lua that consumes the output of `pandoc -t native` or `pandoc -t json` - but I think XML makes more sense. I'm vaguely surprised it doesn't already have something which outputs "pandoc XML".

alerque:
There are several advantages to having a dedicated output path. Besides the recognition and things working the way people would expect out of the box without extra monkeying on either end (for example Pandoc's "standalone" output would work without a user needing to pass SILE any special parameters to load a package or document class), it provides us access to things we wouldn't otherwise be able to manipulate about the process.

A major one is syntax extensions. The primary use case for these I've run across so far is allowing inline markup inside _other document formats_ that Pandoc supports. This is already widely used in the Markdown←→LaTeX world, and I found that allowing inline SILE code inside my Markdown freed me up to not ever have to patch the output after conversion, I could always put something in my canonical source text to address whatever problem came up.

Incidentally this is largely what has kept me going down the TeX-like syntax route rather than XML. Of course one could setup inline XML too (much like inline HTML tags are an extension to Markdown already) but the result is much more cumbersome to edit and (and teach my content editors!) than inline TeX-ish markup.

Ultimately probably having both is in order. And either way the Pandoc _reader_ end of things is going to be more troublesome than the _writer_ end of things.

alerque:
This is almost certainly not the final form this package will take, but keeping it on the sidelines is getting silly. Saying it isn't production ready when I have a whole shelf full of published books that have depended on this package is just semantics. Also there is a checken and egg problem because I can't very well submit the changes to Pandoc upstream when anybody who tests it would have to also run my SILE fork to make it work. Finally not having this in anything but my forks is making work on CI based workflows weird.

I'm going to merge this as is to get the party started, and we can iterate on what this does and how later.

alerque:
Note on previous comments for future reference: The inline issue can probably be handled with direct AST parsing as well because the AST includes raw inline and raw block objects that could be processed as TeXish markup. Adding a direct AST reader and skipping the .sil file format entirely is actually a really interesting option, but the input format is also a separate issue from what this package provides which is output functions corresponding to what could come in.

PR #572, opened 2019-04-08
[counters] Fix roman style capitalization

### LINKED ISSUE #571: `roman` counter mode outputs upper case numbers through ICU
[`SILE.formatCounter`](https://github.com/simoncozens/sile/blob/master/packages/counters.lua#L40) attempts to format counter value through ICU, if available. SILE specifies, that `roman` displays lower case roman numerals, while `Roman` displays upper case roman numerals. However, ICU has [different definition](https://www.unicode.org/repos/cldr/tags/latest/common/bcp47/number.xml) with `roman` being upper case and `romanlow` being lowercase. This leads to a bug and inconsistency.

A temporary workaround is to specify `romanlow` explicitly, but this, of course, creates an unneeded dependency on ICU (though I am not sure if it can be missing, but it seems like it).

Darkyenus:
ICU uses different convention for roman numeral capitalization than
SILE. This adds a translation which fixes this issue.

Closes #571

### DIFF
--- a/packages/counters.lua
+++ b/packages/counters.lua
@@ -40,11 +40,22 @@ SILE.formatCounter = function(options)
     local res = SILE.languageSupport.languages[lang].counter(options)
     if res then return res end -- allow them to pass.
   end
+
   -- If we have ICU, try that
   if icu then
-    local ok, res = pcall(function() return icu.format_number(options.value, options.display) end)
+    local display = options.display
+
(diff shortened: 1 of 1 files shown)


alerque:
Thanks for the contribution @Darkyenus! This looks good to me. We seem to have some problem with the Travis setup at the moment, but I'll merge this anyway and try to figure out why we are double timing the tests.

Issue #245, opened 2015-12-06
Text "width" in tate mode wrongly computed?


shirat74:
I am currently trying to implement simple support for ruby annotation.
I noticed a strange behavior in calculating text "width" in TTB writing direction.

I prepared the following files test.lua

```
SILE.registerCommand("test", function(options, content)
   local rb = SILE.Commands["hbox"]({}, {content[1]})
   print("width=",rb.width)
   print("height=", rb.height)
end)
```

and test.sil

```
\begin[papersize=a4,class=jplain,layout="tate"]{document}
\script[src=test]
\test{longlonglonglongword}
\end{document}
```

Processing test.sil with SILE gives the result

```
width=  10pt
height= 10
```

in the "tate" layout. While it gives

```
width=  102.744140625pt
height= 7.6708984375
```

in "yoko" layout.

I tested with SILE after commit 2a36d77d4874d9bdd5818350d3d48de1391c04bf.

simoncozens:
Will look at this later but have you noticed the ruby package?

alerque:
This issue has been fixed, the width & hight reported by this test code is now correct both in tate and yoko layout modes.

Note there is another problem with the jplain class, namely that the content frame is hard coded to a specific size and does not adapt to the paper size. I'll open a separate issue to track that.

PR #574, opened 2019-04-11
Try to make ruby look a bit nicer in Tate layout

s-gilles:
This is my attempt at #245 . The first commit deals with the different
types of lengths; I broke it out because it looks especially clumsy, and
probably indicates that I don't know how the lengths work and that
there's a better solution. The second commit contains all the real
changes: fiddling with width vs. height and cursor movement.

An example which I used for testing is at http://ix.io/1FUg .

The regression tests don't actually pass, even from the master branch,
on my machine, which I think is because I don't have exactly the same
font versions. I apologize, therefore, that I may have broken some
behavior without knowing.

Please feel free to reject or cannibalize if you'd prefer to do this a
different way.


alerque:
Thanks for taking the time to contribute @s-gilles! The regression tests are very much tied to specific versions of fonts and even @simoncozens and I can't always keep the same versions on our systems, so test are hard to pass locally. I usually check using Travis in my fork or from a PR like this, so don't worry on that count.

simoncozens:
These are great; it looks much better now.

alerque:
Oh look @s-gilles got his account back# See #575 for where these contributions actually got merged. Thanks for contacting us offline to clear up what happened and that we could go ahead and use the code!

Issue #220, opened 2015-11-26
Support Math — somehow (general discussion of approaches and status)


akavel:
I'm aware MathJax is listed [on the wiki](https://github.com/simoncozens/sile/wiki/easy-jobs) as ["up-for-grabs"](http://up-for-grabs.net/), but I just wanted to start by asking: have you maybe tried pushing some [stuff from the MathJax's _unpacked/_ subdir](https://github.com/mathjax/MathJax/blob/master/unpacked/jax/input/TeX/jax.js) through some [js-to-lua](https://github.com/tessel/colony-compiler/blob/master/src/colonize.js) translator ([or other](https://github.com/wizzard0/js2lua))?

I'm on Windows, so personally, before I maybe start trying to contribute, I have to first manage to compile the stuff (yes, I've seen the issue tracking this)...

_(By the way, I got here by accidentally watching the 2015 FOSDEM video of yours; very entertaining and interesting :) although I've apparently already +1'ed your project at some point, only this video made me aware of many things about it (both feature demos, architecture, and how it relates and improves over *TeX) and thus more interested.)_

-----

*edit:* A few years later, after some prototyping and experiments, I'm expanding this initial post with more detailed current status of the project. The path I've eventually picked to try and implement this is based on ASCIIMath, not MathJax.

## Current TODO tasks

Roughly in priority/painfulness order. I'd be especially grateful if someone could help with the ones marked "HELP NEEDED", I don't know how to proceed with those unfortunately :(. Jumping in to help with *any* task is enthusiastically encouraged!

   - [x] ~~Translate [ASCIIMathML.js](https://github.com/asciimath/asciimathml) from JS to Lua~~ *([done and tested](https://github.com/akavel/asciimath))*
   - [x] ~~Translate [SVGMath](https://github.com/w3c/svgwg/tree/73855c2f16b36122cfd3e917ceb4c7d8f6314834/tools/math) from Python to Lua~~ *([done](https://github.com/akavel/svgmath) with help of https://github.com/akavel/pylua based on https://github.com/kevinw/pylua)*
   - [x] ~~Write SVG parser and renderer for SILE supporting SVGs generated by SVGMath~~ *([done](https://github.com/simoncozens/sile/issues/468))*
   - [x] ~~Integrate the tools into a pipeline callable from SILE with an `\asciimath` directive~~ *([done](https://github.com/akavel/silemath))*
[...]

PR #578, opened 2019-04-12
Native math typesetting support via ASCIIMath→MathML

harrysummer:
Disclaimer: this PR is work-in-progress and far from complete. It is just to pull up its visibility to the community, which could support and improve the design and possibly also contribute to the code.

Milestones:

- [x] Basic structure and framework for math equation parsing and representation.
- [x] Parse MATH table from OpenType fonts to get layout parameters.
- [ ] Support common MathML tags
    - [x] Subscript, superscript and mixed sub-superscript.
    - [ ] Letter accents (e.g., hat, tilde)
    - [ ] Fraction
    - [ ] Radical
    - [ ] Brackets
    - [ ] Big operators (summation, integration)
    - [ ] Matrix
- [ ] Fully support the mathematical alphanumeric symbols defined in unicode.
- [ ] Support mathematical symbols from LaTeX `amssymb`.
- [ ] Better integration with SILE typesetter
    * Convert top-level horizontally layout math nodes into SILE's `_nnode` to enable line breaking within math.
    * Insert penalty nodes into the `_nnode` sequence to have better line breaking.
- [ ] LaTeX math to MathML converter.
- [ ] Add test cases.
- [ ] Update documentation.

### Motivation
I am aware of the recommendation of the math supporting approach in the [wiki](https://github.com/simoncozens/sile/wiki/faq#does-it-support-mathematics), and the work of integrating math support into SILE here: https://github.com/simoncozens/sile/issues/220. I have to say that solution will have several major drawbacks (big thanks to @akavel for correcting me here):
1. Equations are treated as images. There is no baseline information for the typesetter to align it with the other texts on the same line. And it also has no clue to make the equation to have the same font size as its context. Needless to say line-breaking inside inline math equations. In one word, this MathJax solution would make math equation a second-class citizen in SILE and the typesetter cannot actually typeset an equation.
2. (__edit__ 1) ~~The MathJax source code is converted into Lua, and the machine-generated code will be hard to read and hard to manage. Either we put the machine-generated code in SILE or we add the toolchain itself in the building system. Neither of them sounds like a good idea.~~
3. (__edit__ 1) ~~We are limited to the features that MathJax provides. See [here](http://docs.mathjax.org/en/latest/mathml.html#supported-mathml-commands) and [here](http://docs.mathjax.org/en/latest/tex.html#differences).~~
4. Performance. The performance will be limited because of the intermediate result (~~pdf~~) need to be saved on disk.

Therefore, I start to think. Why don't we start to write math typesetting logic inside SILE typesetter? Also math layout is complex, its logic is quite clear and even easier than line-breaking, or hypernation problems in traditional typesetting. We just need to organize the math equations into a tree of nodes instead of a sequential list in text typesetting. So I started to investigate this. Appendix G of [The TeXbook](https://ctan.org/pkg/texbook) provides an extensive but hard-to-read alrogithm for math typesetting. So I started with [Appendix G illuminated](https://www.tug.org/TUGboat/tb27-1/tb86jackowski.pdf), and also referred to [A Functional Description of TeX's Formula Layoout](https://people.eecs.berkeley.edu/~fateman/temp/neuform.pdf) and [Mathematical Typesetting with LaTeX](https://www.tug.org/~hvoss/PDF/mathmode.pdf) during the initial design. The current code is able to typeset subscript and superscript correctly with a Opentype math font. And the implementation can be served as a boilerplate for implementing other math layouts.

### Architecture
I don't think it is a good practice to re-implement all details of the math typesetting algorithm described in The TeXbook, which reads the input sequence into an `mlist` and converted it into a `hlist` and then did some black magic to make it happen. I just followed the general ideas and here are the most important aspects of the design:
* Math equations can break into a hierarchical structure of nodes (in the code it's `_mbox` and its child classes). `_mbox` is a subclass of `_box` (`_box` is defined in `nodefactory.lua` as the base class of node for general typesetting). Each node, along with its descendants, represents a part of the equation. This part can be treated as an individual equation but with possibly smaller font and cramped style.
* Each node have a atom type as defined in TeX that is used to specify the type of the content in this node. E.g., for binary operators `+` and `-` have the atom type of `Bin` (binary operator), while `=` has the atom type of `Rel` (relational operator). We need the atom type information to correctly handle the layout. E.g., kerning should be added bewteen `Bin` and `Ord` (ordinary) atoms in display or text style. The atom type is determined on the creation of the nodes.
* Each node has a certain style (which is named `mathMode` in the code). which can be `display`, `text`, `script`, `scriptScript` or their cramped forms. After the construction of the node tree, it determines the style of the children according to their parent. We call this process `styleChildren`, which is an important member function of `_mbox` that will be executed recursively before text shaping process. Each type of layout must define their `styleChildren` function.
* During the shaping process, unlike text shaping where we just infer the shape from the glyphs and assemble into words, we need to recursively determine the shape (width, height, depth) and the relative position of nodes. The relative position here means the position regarding to its parent node. As long as the shapes and positions of all children is determined, we can get the shape of the node itself. We call `shape` function recursively to shape the full equation tree. This `shape` function must also be defined in each type of layout.
* After done with shaping, our math layout structure is ready to render. We implemented `outputYourself` interface at the root-level of the structure in class `_stackbox`. Through this interface, SILE can output this math structure just like it calls other layout blocks. Inside this function, we call `output` function recursively to output every node in this math tree.

In conclusion, similar to text typesetting, the math typesetting process contains the following four steps:
1. Parse the input from `.sil` or `.xml` file. Most of this part is already handled by the lpeg parser in SILE.
2. Build math structure tree, determine atom type and math style (from outside in).
3. Shape math tree, determine shape and position recursively (from inside out).
4. Render the math tree by outputting recursively.

### MathML v.s. TeX math

I chose MathML as the representation of math equations because it complies to XML structure and can be seamlessly intergrated into SILE-style document. We may add a TeX-math-to-MathML converter for people to use TeX syntax later. But I think that would be an addon package instead of the core, because LaTeX have a different kind of syntax from SILE and it is not elegant to put that into the SILE core.

Although we are going with MathML without much doubt, we will miss a bunch of useful symbol definitions in TeX, such as `\rightarrow`, `\alpha`, `\neq`. To ease the use of MathML without struggling into typing these characters, they should be implemented in the MathML parser so we can will be using a superset of tags supported by MathML and these "symbol" tags will be translated directly to the corresponding character.

### SILE typesetter integration
SILE has a great implementation (from the even greater TeXbook) of the typesetting logic which supports glue and kerning. And math equations should also be compatible with the main typesetting algorithm. In the implementation of the math structures, we defined the shape and position of nodes as `SILE.length` class, which means every node in the equation can tolerate shrinking/stretch to some certain degrees. When typesetting a scaled line of text, the math equation will also be scaled proportionally. (Better than that, we even have the shrinking and stretching capability on the vertical direction, which will be useful when we want to, say, scale the line height in a page some we can accomodate certain number of lines on one page without overflowing.)

Another possible integration is to tweak the line-breaking penalty. We may prefer the inline equation to break at certain positions (like before or after a binary operator) and refrain from breaking at some other certain positions (like after the integration operator). We can insert penalty values between the math nodes just like what SILE already did in its typesetter.

### How to contribute
You may send PR to the `math` branch in my forked repository so it will be updated here. However, it is still under actively development so the PR should keep as small and as specific as possible. For example, you can implement an unsupported MathML tag, but not changing other logics in the same PR.

To render a sample document, you may need an Opentype math font (I am using [XITS Math](https://github.com/alif-type/xits)). After installing the font, you can try the following code
```tex
\begin[class=plain, papersize=100mm x 100mm]{document}
Display math:

\math[mode=display]{
	\mrow{
		\mi{E}
		\mo{=}
		\msup{
			\mi{mc}
			\mn{2}
		}
	}
}

Text math:

\begin{math}
	\mrow{
		\msubsup{
			\mi{f}
			\mn{1}
			\mrow{
				\mi{a}
				\mo{+}
				\mi{b}
			}
		}
		\mo{-}
		\msup{
			\mi{d}
			\mn{2}
		}
	}
\end{math}
\end{document}
```

And it will be rendered as:
image

### Other Words
SILE is a great typesetting engine and its uniqueness may never be surpassed by other alternatives. I look forward to seeing math support in SILE so that SILE will take larger influences in the scientific world. If you have the passion and skills, please join us to improve SILE!

I was always thinking, LaTeX is great but it is also the barrier and burden for people to understand typesetting. Its obscure manual and steep learning curve stop many people from grasping it. The world really needs a free-of-charge, general-purposed typsetting software. I tried ConTeXt a few years before but it was hard to use. I can't understand what's happening under the hood from its sparsely documented user mannual. And I could hardly find more useful material on the Internet. SILE is the one, and may be the only one that does the job. It is flexible, and its source code is human-readable.

Great thanks for Simon to bring us such a good software and hope it is going to be even better.

### DIFF
--- a/core/math.lua
+++ b/core/math.lua
@@ -135,6 +135,7 @@ local _mbox = _box {
   relX = 0, -- x position relative to its parent box
   relY = 0, -- y position relative to its parent box
   value = {},
+  mode = mathMode.display,
   __tostring = function (s) return s.type end,
   init = function(self)
     local options = {
@@ -219,7 +220,6 @@ local _stackbox = _mbox {
   styleChildren = function(self)
     for i, n in ipairs(self.children) do
       n.mode = self.mode
-      n.options = std.tree.clone(self.options)
     end
   end,
   setChildrenRelXY = function(self)
@@ -304,22 +304,11 @@ local _subscript = _mbox {
   end,
   styleChildren = function(self)
     self.children[1].mode = self.mode
-    if self.children[2] then self.children[2].mode = getSubscriptMode(self.mode) end
-    if self.children[3] then self.children[3].mode = getSuperscriptMode(self.mode) end
-    self.children[1].options = std.tree.clone(self.options)
-
-    local constants = getMathMetrics(self.options).constants
-    for i = 2,3 do
-      if self.children[i] then
-        self.children[i].options = std.tree.clone(self.options)
-        if self.children[i].mode == mathMode.script or self.children[i].mode == math.scriptCramped then
-          local fontSize = SILE.font.loadDefaults(self.options).size * constants.scriptPercentScaleDown
-          self.children[i].options.size = fontSize
-        elseif self.children[i].mode == mathMode.scriptScript or self.children[i].mode == math.scriptScriptCramped then
-          local fontSize = SILE.font.loadDefaults(self.options).size * constants.scriptScriptPercentScaleDown
-          self.children[i].options.size = fontSize
-        end
-      end
+    if self.children[2] then
+      self.children[2].mode = getSubscriptMode(self.mode)
+    end
+    if self.children[3] then
+      self.children[3].mode = getSuperscriptMode(self.mode)
     end
   end,
   setChildrenRelXY = function(self)
@@ -376,7 +365,7 @@ local _terminal = _mbox {
   setChildrenRelXY = function(self) end
 }
 
--- text node
+-- text node. For any actual text output
 local _text = _terminal {
   type = "text",
   _type = "Text",
@@ -404,8 +393,10 @@ local _text = _terminal {
   end,
   shape = function(self)
     local face = SILE.font.cache(self.options, SILE.shaper.getFace)
-    if self.mode == mathMode.script or self.mode == mathMode.scriptCramped then
-      local fontSize = math.floor(self.options.size * 0.7)
+    if self.mode == mathMode.script or self.mode == mathMode.scriptCramped or
+        self.mode == mathMode.scriptScript or self.mode == mathMode.scriptScriptCramped then
+      local constants = getMathMetrics(self.options).constants
+      local fontSize = math.floor(self.options.size * ((self.mode == mathMode.script or self.mode == mathMode.scriptCramped) and 0.7 or 0.5))
       face.size = fontSize
       self.options.size = fontSize
     end
--- a/core/math.lua
+++ b/core/math.lua
@@ -53,10 +53,6 @@ local function isCrampedMode(mode)
   return mode % 2 == 1
 end
 
-local function isItalicScript(script)
-  return script == 3 or script == 4 or script == 12
-end
-
 local mathScriptConversionTable = {
   italicLatinUpper = function(codepoint) return codepoint + 0x1D434 - 0x41 end,
   italicLatinLower = function(codepoint) return codepoint == 0x68 and 0x210E or codepoint + 0x1D44E - 0x61 end
@@ -73,7 +69,7 @@ local function getFont(options)
   return font
 end
 
-local function getMathConstants(options)
+local function getMathMetrics(options)
   local font = getFont(options)
   local face = SILE.font.cache(font, SILE.shaper.getFace)
   if not face then
@@ -143,10 +139,6 @@ local _mbox = _box {
     SU.error("This is a virtual function that need to be overriden by its child classes")
   end,
 
-  shape = function(self)
-    SU.error("This is a virtual function that need to be overriden by its child classes")
-  end,
-
   output = function(self, x, y)
     SU.error("This is a virtual function that need to be overriden by its child classes")
   end,
@@ -169,6 +161,21 @@ local _mbox = _box {
     end
     self:setChildrenRelXY()
     self:shape()
+  end,
+
+  shape = function(self)
+    local minX, maxX, minY, maxY = 0, 0, 0, 0
+    for i, n in ipairs(self.children) do
+      if n then
+        minX = math.min(minX, n.relX)
+        maxX = math.max(maxX, n.relX + n.width)
+        minY = math.min(minY, n.relY - n.height)
+        maxY = math.max(maxY, n.relY + n.depth)
+      end
+    end
+    self.width = maxX - minX
+    self.height = -minY
+    self.depth = maxY
   end,
 
   -- Output the node and all its descendants
@@ -287,8 +294,7 @@ local _subscript = _mbox {
     if self.children[3] then self.children[3].mode = getSuperscriptMode(self.mode) end
   end,
   setChildrenRelXY = function(self)
-    print(self.children)
-    local constants = getMathConstants(self.options)
+    local constants = getMathMetrics(self.options)
     self.children[1].relX = 0
     self.children[1].relY = 0
     if self.children[2] then
@@ -319,21 +325,6 @@ local _subscript = _mbox {
         self.children[2].relY = self.children[2].relY + subShift
       end
     end
-  end,
-  shape = function(self)
-    local height, depth = self.children[1].height, self.children[1].depth
-    local width1, width2 = self.children[1].width, self.children[1].width
-    if self.children[2] then
-      depth = self.children[2].depth + 1
-      width1 = width1 + self.children[2].width
-    end
-    if self.children[3] then
-      height = self.children[3].height + 3.75
-      width2 = width2 + self.children[3].width
-    end
-    self.height = height
-    self.depth = depth
-    self.width = math.max(width1, width2)
   end,
   output = function(self, x, y) end
 }
--- a/core/math.lua
+++ b/core/math.lua
@@ -1,5 +1,6 @@
 local nodefactory = require("core/nodefactory")
 local hb = require("justenoughharfbuzz")
+local ot = require("core/opentype-parser")
 
 local mathMode = {
   display = 0,
@@ -60,26 +61,27 @@ local mathScriptConversionTable = {
 
 SILE.settings.declare({name = "math.font.family", type = "string", default = "XITS Math"})
 
-local function getFont(options)
-  local font = SILE.font.loadDefaults({
-    family=SILE.settings.get("math.font.family")
-  })
-  if options.family then font.family = options.family end
-  if options.size then font.size = options.size end
-  return font
-end
+local mathCache = {}
 
 local function getMathMetrics(options)
-  local font = getFont(options)
-  local face = SILE.font.cache(font, SILE.shaper.getFace)
+  local face = SILE.font.cache(options, SILE.shaper.getFace)
   if not face then
     SU.error("Could not find requested font "..options.." or any suitable substitutes")
   end
-  local constants = hb.get_math_constants(face.data, face.index, font.size)
-  if constants == nil then
+  local mathTable = ot.parseMath(hb.get_table(face.data, face.index, "MATH"))
+  local upem = ot.parseHead(hb.get_table(face.data, face.index, "head")).unitsPerEm
+  if mathTable == nil then
     SU.error("You must use a math font for math rendering.")
   end
-  return constants
+  local constants = {}
+  for k,v in pairs(mathTable.mathConstants) do
+    if type(v) == "table" then v = v.value end
+    if k:sub(-9) == "ScaleDown" then constants[k] = v / 100
+    else constants[k] = v * options.size / upem end
+  end
+  return {
+    constants = constants
+  }
 end
 
 -- Style transition functions for superscript and subscript
@@ -129,7 +131,12 @@ local _mbox = _box {
   relY = 0, -- y position relative to its parent box
   value = {},
   __tostring = function (s) return s.type end,
-  init = function(self) return self end,
+  init = function(self)
+    local options = {
+      family=SILE.settings.get("math.font.family")
+    }
+    self.options = SILE.font.loadDefaults(options)
+  end,
 
   styleChildren = function(self)
     SU.error("This is a virtual function that need to be overriden by its child classes")
@@ -283,6 +290,7 @@ local _subscript = _mbox {
   _type = "Subscript",
   kind = "sub",
   init = function(self)
+    _mbox.init(self)
     if self.kind == "sup" then
       self.children[3] = self.children[2]
       self.children[2] = false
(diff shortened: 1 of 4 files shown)


alerque:
Awesome to see work on this happening, thanks for your contributions. I don't typeset much math myself but there are lots of people who do!

As a side note, Github has a feature for "draft" PRs now. Unfortunately now that you've opened this there is no way to convert it to a draft (at least not yet), but maybe look out for that the next time you have similar feature branch ready for some extra eyeballs but not ready for merge review.

simoncozens:
Holy smokes. This is amazing.

khaledhosny:
FWIW, HarfBuzz has [API that exposes the math table](https://harfbuzz.github.io/harfbuzz-hb-ot-math.html), so you might want to use this instead of parsing the table on your own.

harrysummer:
> FWIW, HarfBuzz has [API that exposes the math table](https://harfbuzz.github.io/harfbuzz-hb-ot-math.html), so you might want to use this instead of parsing the table on your own.

Thanks for he suggestion @khaledhosny. I am aware of the API (actually I had [some initial implementation](https://github.com/simoncozens/sile/blob/ce37c234eecc2088b396fe67d6b99762898bb2ab/src/justenoughharfbuzz.c#L326-L435) previously). However, using that API means we need to read the font file every time when querying per-glyph data, unless we maintain the hb_* objects in Lua's side to cache. I think this could either make the design of Lua-C interface rather complex, or create a considerable performance degradation. So I decide to parse the MATH table purely in Lua. I have fully followed the MATH table spec so all MATH information would be ready in the parsed results. And we could cache that at Lua's side.

simoncozens:
Agree with doing the parsing in Lua. At least in theory, the shaper is replaceable. Anything that ties us further in to one shaper is a move in the wrong direction.

akavel:
Hi! Awesome that you're trying to do this! :D 

Just some quick notes regarding #220: the title there is kinda misleading, actually :( in fact, the approach I tried is not using MathJaX at all.....! I see now I might not have made this clear enough :( The MathJaX in the title was only the initial idea; for the implementation, I chose something completely different... o_O

Trying to summarize in shortest possible words, the pipeline is basically:

```
ASCIIMath → MathML → SVG
```

where, notably:

  1. `ASCIIMath` is a different format than LaTeX. I believe it's easier to use and more readable, but that's obviously arguable. Anyway, it's mostly treated as "finished", so I didn't expect to support any future need for "keeping up with the upstream". Also, I believe the "machine translated" (actually, "computer aided", i.e. more like "by hand, with some help of custom automation") translator is quite readable Lua! You may be interested in taking a look! :)
  2. `SVG`, as used in the pipeline, it's actually not really "a picture"! I myself was surprised to learn it, but it's actually basically a list of characters/glyphs to render, with font info, size, position, etc. Basically, quite close to what SILE itself stores and expects! So, the import into SILE is quite straightforward.
  3. Finally, the `MathML → SVG` step is indeed, machine translated from Python. But, again, quite surprisingly to me, I believe it resulted in code that's *very* readable as Lua! Not perfectly idiomatic, but I believe it could be not so unreasonable to keep it, and hopefully gradually translate to idiomatic Lua. Also again, I didn't intend to have this part kept synchronized with upstream. Given that "the upstream" is rather mostly an abandoned/archived project.

Please note, I don't mean this all to detract you from anything. Quite opposite, I would just like to try and let you know that it might be interesting to you to try and scavenge some parts of my project to boost you in building yours! That was at least kinda my idea when doing this the way I tried: to try and reuse some existing code some other people wrote, to hopefully have to write as little as possible myself.

As a guide to what you could try to have a *quick* look at in my codebase, to try and convince you it's simpler than it sounds like; going roughly "from the end":

  - [justenoughsvg.lua](https://github.com/akavel/silemath/blob/72102bde232f8d3828ad5fee67a3b268dccd872a/justenoughsvg.lua) is a 232-line Lua script, which is basically a quick and dirty XML parser, which can parse text-only SVG looking [like this](https://raw.githubusercontent.com/akavel/silemath/master/sample/test13.out.svg). Please note that although it looks unreadable in this format, when looked at more carefully, it's actually super brain-dead simple, composed totally only of text nodes, which when expanded all look like this:

        <svg:g transform="translate(6.084797, -6.196289)">
            <svg:text fill="black" font-family="DejaVu Serif" font-size="15.000000" text-anchor="middle" x="4.471436" y="4.702148">
                ∂
            </svg:text>
            <svg:g transform="translate(8.942871, -3.083496)">
                <svg:text fill="black" font-family="DejaVu Serif" font-size="10.650000" text-anchor="middle" x="3.387927" y="-0.000000">
                    2
                </svg:text>
            </svg:g>
        </svg:g>

    As you can see, those are regular glyphs (`∂` and `2`, which compose into literal *∂<sup>2</sup>*), plus font info (font family, font size, coordinates), plus extra translation (i.e. move) directives. In other words, purely: glyphs + font info + coordinates.

  - this simple subset of SVG is then rendered in SILE Lua by a [40-lines long tiny function "render"](https://github.com/akavel/silemath/blob/72102bde232f8d3828ad5fee67a3b268dccd872a/main.lua#L48), which basically recurses through the loaded SVG and handles the "translate" operations, calculating resulting coordinates, and ordering SILE to place glyphs at specified places.

  - as to the `MathML → SVG` translator, [here's some example "machine translated" code measuring text size](https://github.com/akavel/silemath/blob/72102bde232f8d3828ad5fee67a3b268dccd872a/svgmath/mathnode.lua#L481) — do you think it's really so unreadable? I don't want to argue with you if you do, but could I just ask you to take a look at this code? personally, I don't think it would look that much different if someone were to write it in Lua! I mean, not completely idiomatic, but if I was told it's written by a newbie Lua coder coming from a very OOP-focused world, I personally would *absolutely* believe this story...

        - the one thing that's important to know here, is that this module is the one that calculates the coordinates that are later put in the SVG that I've shown above. As such, it needs to already know what font will be used, and based on the font metrics + internal layout algorithms, it has to decide on all the dimensions, etc., where exactly to put the glyphs. As such, it has a [full-blown TTF format parser](https://github.com/akavel/silemath/tree/72102bde232f8d3828ad5fee67a3b268dccd872a/svgmath/fonts) for extracting measurement info. As such, I believe this part should eventually be changed to use info from SILE instead.

  - [asciimath.lua](https://github.com/akavel/silemath/blob/72102bde232f8d3828ad5fee67a3b268dccd872a/asciimath/asciimath.lua) is just *a single Lua file* (admittedly, 1100-lines long). Again, if I could ask you for just this only favor, to take a look at its contents...? could you find a place in it where it actually *doesn't* look like perfectly idiomatic Lua? I challenge you! :D 

  - finally, just to give you a feeling for how ASCIIMath looks, [here's some non-trivial expression from the asciimath.lua test-suite, together with the generated MathML](https://github.com/akavel/silemath/blob/72102bde232f8d3828ad5fee67a3b268dccd872a/asciimath/testall.lua#L370):

        f(x)=\sum_{n=0}^\infty\frac{f^{(n)}(a)}{n!}(x-a)^n

    Couldn't it be a reasonable format to use? I mean, LaTeX is totally the "endgame" and long-term goal to aim for (and something I was personally interested in too, as a long-term goal). It's just that I thought this format is actually quite interesting out of itself, but more importantly, *the parser was super easy to translate to Lua* and have *something* that people could already use, before attacking the (much bigger in my opinion) LaTeX format.

So, that's basically what I wanted to try and say at this point... please note I totally only would like to maybe hopefully boost you at some area of your project, *purely and only if you would like to do so*. If you prefer to do this your way anyway, I only wish you best of luck, and plan to be looking at your advances with joy and excitement! :) By the way, if you would have any questions, you're more than welcome to ask them either here, or on #220, or by emailing me at czapkofan@gmail.com.

OlivierNicole:
Hi! Just to avoid duplicate work in case anyone else considers contributing to this: I am currently working on “big operator” support (sum, product, integration…). Things are going well, here is my current output:
Capture d’écran_2019-06-16_23-10-29
Before proposing a PR to @harrysummer's fork, I need to make it support all kinds of big operators (and user-defined ones), notably integration which probably needs a different typesetting.

I enjoy a lot working on your clean and well-designed code, @harrysummer!

harrysummer:
> ASCIIMath → MathML, and MathML → SVG) have rather extensible test suites ([here](https://github.com/akavel/silemath/blob/72102bde232f8d3828ad5fee67a3b268dccd872a/asciimath/testall.lua) and [here](https://github.com/akavel/silemath/tree/72102bde232f8d3828ad5fee67a3b268dccd872a/svgmath/testdata) — in the latter, `*.mml` are inputs, `*.svg` are outputs from the original Python code, `*.out.svg` from the Lua-translated code). That said, they actually also _do_ have some bugs :/ (especially the layouter — but that's, like, the hardest part of this whole process) That said, you can never fully avoid bugs in any software... I personally

Thanks a lot for your rich response regarding the math support of SILE @akavel . I must take back my words on the readability of machine-translated codes after your concrete defense. I have to say, I did misunderstand part of your intention in #220 from the title. After your explanation here, it looks more reasonable to me now. However, one potential drawback of this ASCIIMath -> MathML -> SVG -> SILE pipeline could be the missing of glue and kerning of inline equations. When the original width of a line of text slightly exceeds the width of the frame, we may need to squeeze the elements in the equation a little bit so as to make it fits the frame. The independent font system could also be a problem just like you mentioned.

I also agree that ASCIIMath is a easy and reasonable format to use -- I might add it as a parser in my structure, but I do believe the internal representation should be MathML to make it more SILE-ish. And SILE directly uses the MathML structure for layout and typesetting. End-user may use various formats (say, ASCIIMath, TeX...) in the source file, but everything should be converted to MathML for SILE to typeset.

A great thank you for your interesting and supportive to this PR. Let me take some time to digest the resources and align them with my plan.

OlivierNicole:
I have been doing some work on this PR, but lately the necessities of reproducing my material existence in a modern economy leave me very little time to keep doing so, so I thought I'd leave a note on the state of my fork.

I have added support for fractions and “big operators” to @harrysummer's math engine, as well as an embryo of a vaguely TeX-like language to input math in a more human-friendly way than with MathML. But this language, if extended, will require a proper semantics and many non-trivial usage choice.

The things I'm not happy with: currently the subscripts and superscripts are too far from integral signs and I don't know why. Sample rendering (ours on the left vs LuaLaTeX's on the right):

Capture d’écran_2019-11-01_19-50-24

Other kinds of subtle spacing issues are bugging me because I don't know typography nor font formats. Also, many math fonts I wanted to try cannot be parsed because of some error in the parsing of coverage table. The only ones I could use were XITS Math and Garamond Math.

If someone sees this and feels like adding support for more math constructs or anything else, they should feel free to do so and I would be happy to explain/amend the few lines of code I have written.

OlivierNicole:
Update: I copy here @harrysummer 's milestones with a few new boxes ticked:

* [x] Basic structure and framework for math equation parsing and representation.
* [x] Parse MATH table from OpenType fonts to get layout parameters.
* [ ] Support common MathML tags
  - [x] Subscript, superscript and mixed sub-superscript.
  - [ ] Letter accents (e.g. hat, tilde)
  - [x] Fraction
  - [ ] Radical
  - [ ] Brackets
  - [x] Big operators (summation, integration)
  - [x] Matrix **(very basic support)**
- [ ] Fully support the mathematical alphanumeric symbols defined in Unicode.
- [ ] Support mathematical symbols from LaTeX `amssymb`.
- [ ] Better integration with SILE typesetter
  - [ ] Convert top-level horizontally layout math nodes into SILE's `_nnode` to enable line breaking within math.
  - [ ] Insert penalty nodes into the `_nnode` sequence to have better line breaking.
- [x] LaTeX math to MathML converter **(TeX-like language, but not exactly TeX. Includes a multi-argument macro system)**
- [ ] Add test cases.
- [ ] Update documentation.

To which I propose to add:

- [ ] Support all symbols from LaTeX package `unicode-math` **(can be easily done automatically)**

Small demo:
image

alerque:
Just connecting the dots, anybody working on OpenType Math should probably read comments on [this Harfbuzz issue](https://github.com/harfbuzz/harfbuzz/issues/2585#issuecomment-668621880) and follow references there.

OlivierNicole:
@alerque Interesting read. However, based on previous messages in this thread, I assume you would rather not depend on Harfbuzz-specific APIs.

alerque:
@OlivierNicole Our primary backend (and even our debug backend) already depends on HarfBuzz. We don't have a working alternative shaper used in any backend at this point. While it would be cool to have an implementation that would allow other backends to easily support this too, if the choice is between not having the feature at all and only having it implemented and working on one backend then I'd rather see the implementation happen. We can always expand from there.

OlivierNicole:
@alerque I see. In my understanding however the mentioned API change hasn't happened yet, so my point was quite superfluous.

alerque:
I'm closing this PR because it is obsoleted by #1103, but never fear! None of this work is going to waste. All these commits are included and I'm even cleaning up the commit history of this bit now. Many things got refactored later too, but that's to be expected. To anyone following, I expect this to land in master soon!

alerque:
Also for posterity (and in case I did anything wrong munging commits around), I'm archiving the branch making up this PR at [harrysummer/math-578](https://github.com/alerque/sile/tree/harrysummer/math-578).

PR #581, opened 2019-04-21
Allow setting PDF metadata with \pdf:metadata

s-gilles:
I often find myself organizing PDFs by metadata. LaTeX's
`\usepackage[pdfusetitle]{hyperref}` is useful for that, so here's my
attempt at adding an equivalent to `packages/pdf`. (For verification,
I usually use `exiftool` or the stricter `pdfinfo`.)

It doesn't appear that PDF-specific details can be tested at present,
so there's no test file, but hopefully documentation/sile.sil will serve
as an example.

As before, please feel free to reject, rewrite, or take some parts.

### DIFF
--- a/documentation/sile.sil
+++ b/documentation/sile.sil
@@ -23,6 +23,8 @@
 \par\center{\font[size=50pt,weight=600,filename=documentation/Silson-Condensed.otf]{The}\par
 \center{\img[src=documentation/sile-logo.pdf,height=40mm]}\par
 \font[size=50pt,weight=600,filename=documentation/Silson-Condensed.otf]{Book}}
+\pdf:metadata[key=Title, val=The SILE book]
+\pdf:metadata[key=Author, val=Simon Cozens]
 \bigskip
 \font[size=12pt,filename=documentation/Silson-Condensed.otf]{\hss for SILE version \sileversion \hss \par}
 \skip[height=1in]
--- a/documentation/sile.sil
+++ b/documentation/sile.sil
@@ -23,7 +23,7 @@
 \par\center{\font[size=50pt,weight=600,filename=documentation/Silson-Condensed.otf]{The}\par
 \center{\img[src=documentation/sile-logo.pdf,height=40mm]}\par
 \font[size=50pt,weight=600,filename=documentation/Silson-Condensed.otf]{Book}}
-\pdf:metadata[key=Title, val=The SILE book]
+\pdf:metadata[key=Title, val=The SILE Book]
 \pdf:metadata[key=Author, val=Simon Cozens]
 \bigskip
 \font[size=12pt,filename=documentation/Silson-Condensed.otf]{\hss for SILE version \sileversion \hss \par}
--- a/packages/pdf.lua
+++ b/packages/pdf.lua
@@ -89,14 +89,32 @@ SILE.registerCommand("pdf:link", function (options, content)
   })
 end)
 
+SILE.registerCommand("pdf:metadata", function (options, content)
+  local key = SU.required(options, "key", "pdf:metadata")
+  local val = SU.required(options, "val", "pdf:metadata")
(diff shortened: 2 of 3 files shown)


simoncozens:
Thanks, this looks great - I'm just fighting with the CI at the moment...

alerque:
Especially with the manual (which is turning into a big binary blob with a lot of church) just reverting a commit won't properly keep the repository free of all the data. At some point before this is merged can you rebase to squash them together and make sure the resulting commit doesn't touch the PDF file? I don't think Github's squash/merge will handle this properly.

s-gilles:
Sure thing. I think this rebase gets rid of the pdf change completely,
and I'll rebase again if needed once the CI issues are worked out.

documentation/sile.sil
```diff
@@ -23,6 +23,8 @@
 \par\center{\font[size=50pt,weight=600,filename=documentation/Silson-Condensed.otf]{The}\par
 \center{\img[src=documentation/sile-logo.pdf,height=40mm]}\par
 \font[size=50pt,weight=600,filename=documentation/Silson-Condensed.otf]{Book}}
+\pdf:metadata[key=Title, val=The SILE book]
```
alerque:
```suggestion
\pdf:metadata[key=Title, val=The SILE Book]
```

packages/pdf.lua
```diff
@@ -89,15 +89,33 @@ SILE.registerCommand("pdf:link", function (options, content)
...
+    end
+  })
+end)
+
 return { documentation = [[\begin{document}
 The \code{pdf} package enables (basic) support for PDF links and table-of-contents
-entries. It provides the three commands \command{\\pdf:destination}, \command{\\pdf:link}
-and \command{\\pdf:bookmark}.
+entries. It provides the four commands \command{\\pdf:destination}, \command{\\pdf:link},
+\command{\\pdf:bookmark}, and \command{\\pdf:metadata}.
```
alerque: Thank you for the Oxford comma here, the _only_ sane way to use commas in English. :wink:

packages/pdf.lua
```diff
@@ -89,15 +89,33 @@ SILE.registerCommand("pdf:link", function (options, content)
...
 
 The \command{\\pdf:destination} parameter creates a link target; it expects a
 parameter called \code{name} to uniquely identify the target. To create a link to
 that location in the document, use \code{\\pdf:link[dest=\goodbreak{}name]\{link content\}}.
 
+To set arbitrary key-value metadata, use something like \code{\\pdf:metadata[key=Author,
```
alerque: Is there a list of the most common/canonical values to set that we could tip people off to here?

For reference perhaps some of all of [Pandoc's meta data keys](https://pandoc.org/MANUAL.html#metadata-variables) maps to something useful and we could document those.

s-gilles:
On 2019-04-23T02:35:57-0700, Caleb Maclennan wrote:
> Is there a list of the most common/canonical values to set that we could tip people off to here?
> 
> For reference perhaps some of all of [Pandoc's meta data keys](https://pandoc.org/MANUAL.html#metadata-variables) maps to something useful and we could document those.

I don't own a copy of the most recent PDF spec, but the [1.6](https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdf_reference_archive/PDFReference16.pdf)
spec lists some values in Table 10.2 (page 772). Three of the values
(Creator, Producer, Trapped) don't seem to be intended as
human-editable, and the others are few enough that I just added them.

I wouldn't have any objections to other fields, such as the Pandoc ones,
being mentioned, but I'm not familiar enough with the best practices to
add them myself. (For example, `subtitle` could probably map to
`Subtitle` as another key, but is not currently listed as being
included in PDF metadata. Is the "right" thing to do to simply include
it in the Title field?) If you'd like anything added, I'll certainly do
so, though.

PR #584, opened 2019-04-24
Untangle master/devel dichotomy

alerque:
We've got ourselves tied up in knots again with mix and match workflows. We can't keep two long-running development branches around. On the positive side of things with CI builds and restricted merges, we are getting a lot closer to having a master branch that can be used! **Lets keep going that way.**

_But_ ... having both master and devel used for long running development cycles is the worst of two words. I see two options. Feel free to suggest others.

1. Use `devel` as the primary branch for ongoing development. Use `master` only for actual release cycles, merging in the latest from `devel` and bumping release numbers, submodule SHAs, and rebuilding documentation. This would mean:
    * Mark `devel` as being the primary branch on Github so that clones would use it by default and pull requests would be against it.
    * Add merge restrictions and require CI build on it much like we currently do for `master`.

2. Drop `devel` but keep the restrictions on `master` and **always** use _short running_ feature branches, with everything incrementally merged to `master` and known working points (as proved by passing CI tests).

Either way the "trunk based development" workflow of multiple people hacking on one branch (whether that's called `master` or `devel`) is a bit of a pain for everybody except the first person to hack on it. Given that one thing we do not have yes is reliable local tests (our gold standard has lots of font version specific stuff that is only being properly tested in CI), I don't think the long running development branch workflow is going to work for us.

The primary differences between ① and ② is how the actual release process would go. In ① everything could go on as before in parallel leading up to a release. In ② there would be a bit of disruption in the form of a bit of a feature freeze to make sure everything was hunky-dory. In exchange for this disruption the overall concept is a bit simpler and closer to what people not familiar with release cycles might expect. 

Whether we go with ① or ②, I am strongly recommending that we keep the current restrictions so that nothing gets pushed or merged to master that is not known to pass at least our minimal test suite. Additionally I recommend that everybody working on the project (looking at you @simoncozens!) sticks to a workflow of short lived local feature branches (short being relative of course) that are then vetted using PR and the CI process before remotely merging to whatever branch our "trunk" ends up being. No merging unfinished stuff to trunk locally and moving on to something else out of sequence.

For my own purposes I've been maintaining a `testing` branch that at any given time has several active but as-yet unmerged PRs merged together. This is the build I'm actually using in production (hence things like my `package/pandoc` feature branch is already getting a workout). It is not the one I ever do development work on locally, for that I work on one feature at a time branched from ... whichever of `master` or `devel` looks more promising at that moment.

That's what this PR is for. I would like to merge what's in `devel` now into `master` (assuming it's in working condition) and vise versa so that they get back in sync, then stop using one or the other of them. Either way I'm not too particular, but we can't have our cake and eat it too. Either make `devel` the official trunk or drop it entirely, and only ever merge to whatever trunk is using Github PRs so we have confirmation from CI.


simoncozens:
I'm sorry for (again) causing chaos and confusion! Getting away from a trunk-based development model is definitely worthwhile but is a bit of a cultural shift for me. Thank you for all the work you've done on bringing discipline and order to this.

What I've been unintentionally working towards is `devel` as a development trunk and `master` as a collection of known-good and ready-for-release changes, such that anyone who pulls master (say on homebrew) is guaranteed to get a latest working version, while `devel` is more up to date but without that guarantee. But that sounds a bit like the "two development branches" you are eager to get rid of.

I agree we should absolutely keep the restrictions on master. (A lot of the recent jiggery-pokery has been because of trying to get various CI things working: adding Azure, and fixing the breakage caused by the new version of luasec.)

How does this idea sound:

* Close down `devel` after this PR is merged.
* Protect `master` and do all changes on short-lived feature branches.
* Use release branches to prep and tag releases.

alerque:
Great.

Tell you what. If you ever end up with a jumble of commits in some local hacking branch of yours hat need splitting into feature branches for reviewing/testing separately, just go ahead and post in and I'll help you split it. I'm pretty good at using `git rebase`.

> Use release branches to prep and tag releases.

Everything else sounds good, but I don't think we'll actually need this bit. We might need hotfix branches if we need to do point-releases on against old versions without adding in more recent development work since the release was cut, but otherwise I think we can probably get away with tags right on master.

I'll be working on the `Makefile` end of things to normalized the release process a bit.

Issue #587, opened 2019-04-24
Tests do not work reliably locally


alerque:
Our current test suite is dependent on very specific versions of font files, usually those included with Travis's current Ubuntu packages. This makes it next to impossible to run `make tests` locally, which is inconvenient at best.

In order for the tests to pass they need to work against a known set of font files. I suggest:

1. Paring the fonts used by the tests down to some sort of minimum necessary to actually prove the tests pass/fail. 
2. Assembling this set of font files into a tarball that we can host on [sile-typesetter.org](http://sile-typesetter.org/). Obviously this would require all the fonts used in the test to be freely distributable, but I think that's doable.
3. Setting up the `Makefile` to download and extract that set of fonts as a prerequisite for running any tests.
4. Referencing those local font files by filename in all tests.

alerque:
_The eagle has landed!_ :eagle:

Here is my suggestion of something to run in a terminal while doing any local hacking:

```sh
git ls-files | entr -c make test_previews test
```

This will run the regression tests including PDF previews of any tests in the suite affected by any files you save that are tracked by the repository. If you are hacking on files not tracked by git or doing something outside of Lua your millage will vary, but it should cover most cases of working on the typesetter with a pretty short feedback loop.

If you are interested in only specific test(s):
```sh
make test_previews test TESTSRCS=tests/bug-61.sil
```
That can be an environment variable too, and `TESTSRCS` is a space separated list of tests you want to run so you can quote in any way our shell lets you.

Issue #305, opened 2016-04-04
Return list of files used in compilation of document


alerque:
I have a number of projects that consist of large numbers of documents (in the 1000s) and a collection of supporting classes, function libraries, etc. that are used across them. These projects are of course under version control and, when pushed to a central repo, get built by a CI system that looks through all the documents, re-renders the ones that have been updated and ships the new PDFs off to a shared file server.

Each project has a Makefile which comes up with a list of sources in the project and checks whether they have been updated since the last build. Make is really good at this — provided you can supply it with a good list of dependencies for each target. For my SILE projects this is quickly becoming unwieldy. For example, given a target PDF one might assume that a like-named `.sil` file is a dependency along with the project's main lua library and some cross-project shared class or function library. Optionally if it exists, a like-named lua file could be a dependency:

``` Make
%.pdf: %.sil <project>.lua <toolkit>/class.lua $(wildcard $$*.lua)
```

But now suppose that _some_ of the `.sil` files also pull in custom bits:

``` TeX
\script[src=toolkit/specialfunction]
```

Now `toolkit/specialfunction.lua` should technically be listed as a dependency to build that PDF, but it shouldn't necessarily be listed as a dependency for _all_ the PDFs in a project, just the ones build from that particular source. Regenerating rendered PDFs is expensive.¹ Such a list could be maintained by hand, but that's a royal pain. Instead we must enter the world of auto-depenency-generation.

There are pretty [well established ways](http://make.mad-scientist.net/papers/advanced-auto-dependency-generation/) of doing this in the C/C++ world. For example one can run `gcc -MM` to generate a makefile snippet that lists all the dependencies of a source file such that make can know when to rebuild it or not. Some LaTeX compilers have similar solutions that parse documents for `\usepackage`, `\import`, etc. Others such as latexmk [parse the compilers log](http://tex.stackexchange.com/q/68277/5100) to determine what files were included and use those when checking for dependencies on the next build.² 

A quick hack would be to grep the source and look for script commands, but this would only be a band-aid as it would miss any conditional includes, nested includes from includes, etc. A true list of  dependencies can really only be achieved by parsing the source and processing all the functions. Ergo this can only be done properly by SILE itself.

I propose that:
[...]

Issue #284, opened 2016-03-10
Line break with lskip set carries space to start of next line


alerque:
I just noticed that some of my block-quotes have spaces at the _start_ of lines. It turns out that with

1. a non-zero value for `document.lskip` and
1. a line with em-dashes

it's actually possible to wrap a line such that the space is pushed to the next line. Here is an MWE:

```TeX
\begin[papersize=a6]{document}
\font[family=Libertinus Serif,size=11pt,language=tr]
\set[parameter=document.lskip,value=15pt]
\set[parameter=document.rskip,value=10pt]
\noindent
Çünkü Müjde’den utanmıyorum. Müjde iman eden herkesin –önce Yahudiler’in, sonra Yahudi olmayanların– kurtuluşu için Tanrı gücüdür.
\end{document}
```

![selection_184](https://cloud.githubusercontent.com/assets/173595/13664998/ecac3326-e6b2-11e5-86d4-66e0dd6987dc.png)

P.S. Try not  to break #262 again while fixing this!

PR #592, opened 2019-04-25
Refactor tests so they work reliably locally and in CI

alerque:
Work in progress for #587.


alerque:
By far the most frustrating part of messing with the tests is getting them to _fail_. It's really hard to confirm that the test is still doing anything if you can't reproduce the failure condition. Since a lot of them are very closely tied to exact metrics, doing things like changing fonts requires determining that the new font still fail before setting it up to succeed. For example see #284. The "fix" it (78d8bddc2bb24a7ef803582782025e0a1d796ce2) should be absurdly simple to back out to make the test fail again, but it doesn't fail. This must have gotten fixed another way later, but now we don't know if we are actually testing something that _would_ fail if we rebroke it.

Also my the way @simoncozens this is a good case study for why making separate commits to the typesetter and test suite is a good idea. I can't easily use `git revert <sha>` to test regressions if the commit also changes the test.

alerque:
Hey @simoncozens any chance you or anybody else listening in all this racket could check this branch out and post a dump of a test run for comparison? `make test -B > test.log.txt 2>&1` should do the trick.

Darkyenus:
@alerque I have run tests from this branch on OSX 10.11.6: [test.log.txt](https://github.com/simoncozens/sile/files/3206189/test.log.txt)

alerque:
Thanks @Darkyenus that's quite helpful. The result is not pretty but having it is helpful.

Right off the bat it's apparent that the environment variable `FONTCONFIG_FILE` is not being respected at all on OSX. I wonder what can be done about that. That's kind of the basis of how I was hoping to force font versions for test suite purposes!

alerque:
Hey @Darkyenus one more thing, can you check the output of this command run from your project working directory is as shown?

```sh
$ FONTCONFIG_FILE=/Users/Darkyen/IntelliJProjects/sile/fontconfig.conf fc-match 'Libertinus Serif' file
:file=/Users/Darkyen/IntelliJProjects/sile/.fonts/LibertinusSerif-Regular.otf
```

Note the path should be found relative to the project as a copy in the `.fonts` folder inside the project. If it returns a system path or some other font entirely then fontconfig isn't working as expected. I looked up other examples of people doing this on other projects on OSX so it seems like it should work.

Darkyenus:
Is the `filename` not sufficient for that? I know that it can't handle `style` (and possibly other parameters, `weight`?), but I don't know how necessary is it for the tests.

In my case, I would appreciate a way to specify the font file together with its variants, so that commands like `\font[style=Italic]` work even for `filename`-set fonts. This comes from similar need as the tests have: I want to store whole project in a git repository and I don't want to have to install fonts on each new machine.

There are two ways to achieve this that I can think of (and which I would like):
1. Allow to specify a folder in `filename`. `\font` would then consider all fonts (and only them!) in the directory (possibly recursively) as installed. Font selection would then work through standard `family`/`style`/etc. I am not sure how hard would this be. It could use fontconfig internally or partially reimplement it. Would that be hard?
2. Allow to specify a function, which takes required font parameters (family name, style, weight, maybe even size) and returns loaded font/filename. This would be easier to implement and more flexible, but harder to use.

Both approaches could even get implemented on top of each other.

If you people think that it is a good idea, I can create an issue for further discussion.

Darkyenus:
@alerque The command seems to work:
```bash
bash-3.2$ FONTCONFIG_FILE=/Users/Darkyen/IntelliJProjects/sile/fontconfig.conf fc-match 'Libertinus Serif' file
:file=/Users/Darkyen/IntelliJProjects/sile/.fonts/LibertinusSerif-Regular.otf
bash-3.2$ :file=/Users/Darkyen/IntelliJProjects/sile/.fonts/LibertinusSerif-Regular.otf
bash: :file=/Users/Darkyen/IntelliJProjects/sile/.fonts/LibertinusSerif-Regular.otf: No such file or directory
```

alerque:
> I know that it can't handle style (and possibly other parameters, weight?), but I don't know how necessary is it for the tests.

We could of course hard code all the tests in such a way as to avoid this, but it would mean we can't test things the way they actually get used. For example we couldn't test that `\font[style=Italic]` even works because in our test suite we could only use `\font[filename=foo-Italic.ttf]` or whatever.

My idea was to setup the external environment (fontconfig) to handle this transparently so that `sile` could be expected to function in tests as normally as possible.

alerque:
> The command seems to work:

Yes it does. In other words OSX's fontconfig does support the basic feature, it just isn't getting propagated into the instantiation if it SILE uses internally. I assume we can figure how to make the environment propagate.

simoncozens:
Note that on Mac, SILE does not use fontconfig, preferring to use the system's font libraries instead. That could be why it's not using the fontconfig config. :-)  There's merit in this as Mac users can have fonts installed by third-party applications that fontconfig does not know about:

```
$ fc-list 'Rockwell Nova'
$ ./sile
This is SILE 0.9.5.1
> require("macfonts")._face({family = "Rockwell Nova"})
{
  family = "Rockwell Nova",
  filename = "/Users/simon/Library/Application Support/.C02J89BQDTY4-8cbef5ba/.0d042bb0-3697-4f76-a6e3-735b00a1d504/.bf757e77-3f75-4f03-a38a-f9e0bca1e111.ttf",
  index = 0,
  pointsize = 12.0,
}
```

Not helpful for testing, I appreciate. The problem is in https://github.com/simoncozens/sile/blob/5b5a25331896726c99266588d44c8f0d857b966f/core/harfbuzz-shaper.lua#L13-L19

The *correct* way to fix it would be to move those lines into a new file called `core/fontmanager` which sets `SILE.fontmanager` or similar and have the shaper access that.

A less correct way would be to nobble the `macfonts` library with something like:

```lua
package.loaded["macfonts"] = require("justenoughfontconfig")
```

alerque:
Is there a similar way we can tell `macfonts` to only consider finding font files in a specific path _without_ changing what SILE code is executed during test runs? In other words I'd like the tests to actually test the way the system normally works rather than side-stepping it for the test. Telling the system to restrict it's search to a specific set of files instead of having SILE do that (or switching to fontconfig just for the sake of the tests) allows us to actually test SILE rather than a jury-rigged variant of it.

simoncozens:
Not really; the OS X font manager wants to hide away implementation details like filesystem paths from you as much as possible, so you can't lock it down to a set of paths.

I understand your point about testing reality, but at the same time we are creating a particular test environment by providing the right font versions. I don't think it's a stretch to also add the choice of font manager to that environment, especially since fontconfig should be regarded as the "mainline".

alerque:
Okay maybe we can separately test whether the platform's font manager is working as expected without locking that test in to testing shaped font metrics at the same time, then fall back to using fontconfig for the rest of the test suite. I'll work in that direction.

Makefile.am
```diff
@@ -43,20 +43,33 @@ EXPECTS = $(wildcard tests/*.expected)
...
-tests/%.actual: tests/%.sil | sile
+tests/%.actual: tests/%.sil | sile testprep
 	$(LOCALTESTFONTS) ./sile -b debug $< -o $@
 
 .PHONY: force
 force: ;
 
+.PRECIOUS: .fonts/%
+
+TESTFONTFILES  = NotoSansCJK-Regular.ttc
```
simoncozens: I don't know where you're going with this long term but for me the obvious way forward here is to have a CSV/YAML/JSON/whatever file in the `tests/` directory which lists all the font files required and the URLs to download them from.

alerque: It's not quite that simple. If all we needed was a map of files and URLs that would be one thing, but it isn't a straight map like that. Different sites have different download restrictions, some are in archives, some not. Doing this with `make` recepies allows us to do some pattern matching and script the download/extraction process differently for each font or group of fonts by pattern.

I did look at extracting a list of fonts needed using SILE's debug output itself and will look at doing it again as a last step here, but decided listing them by hand at least during the transition while I went through each test trying to replicate issues was a lot easier. Once they all work this way it should be easy to compare if the auto generated list of required fonts is actually accurate.

simoncozens: OK, before we go much further on this I wonder if the prerequisite is to try to reduce the number of unique fonts used in tests. Currently it looks like this:

| Font name | Number of tests using | Tests using
| ---------- | ---- | --- |
| 	 | 1 | color-fonts* |
| Amiri | 1 | bug-192-full |
| Amiri Quran | 1 | sura-2 |
| Crimson | 1 | bug-307* |
| DejaVu Sans | 1 | bug-76 |
| DejaVu Sans Mono | 2 | bug-132,bug-317 |
| Gentium | 4 | 049EPH,bug-124_upstream,chapterverse,combining |
| Gentium Book Basic | 1 | nondeterminism |
| Gentium Plus | 81 | |
| Kohinoor Devanagari | 1 | bug-75_darwin |
| Lateef | 1 | arabic-scripts |
| Libertinus Serif | 1 | bug-307* |
| Linux Biolinum O | 1 | balanced* |
| Linux Libertine O | 4 | balanced*,bug-283,bug-79,linespacing |
| Montserrat | 1 | bug-307* |
| Noto Naskh Arabic | 4 | bug-124_upstream,bug-192,bug-76,mini-arabic |
| Noto Sans CJK JP | 3 | bug-524,bug-525,vertical |
| Noto Sans Kannada | 1 | kannada |
| Noto Sans Malayalam | 1 | bug-308 |
| SBL Hebrew | 1 | bidi |
| Scheherazade | 2 | bug-124_upstream,bug-192-full |

(* marks knownbad)

It would be nice to see if we can tickle the same bugs without some of these fonts or by substituting other ones.

simoncozens: Some comments on these:

* `arabic-scripts`, `bug-192-full`: Any Arabic would do.
* `bug-124_upstream`: Is these even being tested? Should it be? It's fixed now. So Scheherazade can go.
* `bug-76`: DejaVu Sans can be replaced by any Latin.
* `bug-132`, `bug-317`: Actually any secondary font would do; doesn't even need to be monospaced.
* Gentium only appears where expectations files are old.
* `nondeterminism`: I don't even remember what this is meant to test. I don't think it's valid any more.

alerque: I've been reducing the number of fonts as I go along by trying to reproduce bugs with one of the fonts already required before requiring a new one for the test. I suspect almost half of these will not be needed by the time I'm done.

alerque: I think `nondeterminism` was testing white space trimming options, as in different inputs could end up with the same output because of newline replacements and trimming in the parser.

Most of the others I've already chopped these fonts and more.

simoncozens: Fantastic. Thank you!

alerque:
All the tests that are failing on Travis now are failing with the same ~0.2pt vertical offset starting from the beginning of the document. I don't see any particular common denominator to the handful of tests that are failing this way. I have a vague suspicion that it's line spacing related, but can't pin in on anything. Does this ring any bells @simoncozens? What might be outputting vertical space that is not related to fonts used in the document but varies by system?

simoncozens:
All of the tests are now failing for me locally because paper size is not being correctly set when the test is run.

And you know, I might be being a bit stupid here but... you seem to have deleted the line from `regressions.pl` which actually runs SILE...?

https://github.com/simoncozens/sile/blob/2952b878c98fe8b2bbcf68812e0b14b4bc51b7f9/tests/regressions.pl#L32-L36

alerque:
Use `make test`, the perl script just gathers the diff outputs and presents the findings, it does not actually run the tests. Make takes care of that.

simoncozens:
OK, so this is what I have failing locally:

```
Failed tests:

❌ tests/bidi.sil
❌ tests/bug-192-full.sil
❌ tests/bug-75_darwin.sil
❌ tests/insertions-sidenotes.sil
```

- no overlap with the Travis failing tests, and with other weird stuff, not with a 0.2pt offset. So that was no help.

Are you sure it's not font related? A suspicious number of the failing tests use Libertinus Serif (`bug-61`, `bug-79`, `bug-283`, `linespacing`, `linespacing2`, `alignment`) and Noto Sans CJK (`bug-524`, `bug-525`).

alerque:
I'm reorganizing the code using `SILE.fontManager` as [suggested](https://github.com/simoncozens/sile/pull/592#issuecomment-495206534) and it occurred to me that I could set it up to load _both_ *fontconfig* and *macfonts* (subject to availability) and fallback to the latter if a font is not found. Of course that might require some other complicated back and forth logic since *fontconfig* has internal falbacks ... is this even an issue? Is there ever a case where *macfonts* does not find fonts on a Mac system where *fontconfig* would (other than the obvious case of trying to force a local test font)?

simoncozens:
I guess you could in theory set a custom fontconfig directory where you only want SILE to use fonts in that directory and not the ones on the system... but that's pretty pathological. What you're suggesting sounds like a good idea.

alerque:
> a custom fontconfig directory where you only want SILE to use fonts in that directory and not the ones on the system... but that's pretty pathological

That's exactly what the new test suite font mechanism does. It puts the fonts we need in a directory and uses a `fontconfig.conf` file to make sure it can't fall back to ones on the system. Pathological? Great then I'll carry on...

alerque:
@simoncozens I actually don't believe your last test run worked properly, and that's probably my fault because the test suite is a bit less aggressive at regenerating `*.actual` files. That's actually a bug I know about but haven't fixed yet (I am hoping to slip in a fix to #305 in time to use it for this), but for now you ran run `make test REGRESSIONMODE=true` to get the same behavior as before.

Can you pull the latest commits here and try that? I suspect you'll get a completely different list of test failures (or none!). Then try it with `make test REGRESSIONMODE=true SILEFLAGS="-d versions -f macfonts"` and see if the results come in all different.

simoncozens:
Just pulled it, I have the same test failures as before without the "macfonts" flag. With it, I get

```
Failed tests:

❌ tests/alignment.sil
❌ tests/arabic-scripts.sil
❌ tests/bidi.sil
❌ tests/bug-132.sil
❌ tests/bug-192-full.sil
❌ tests/bug-192.sil
❌ tests/bug-200.sil
❌ tests/bug-214.sil
❌ tests/bug-241.sil
❌ tests/bug-269.sil
❌ tests/bug-283.sil
❌ tests/bug-284.sil
❌ tests/bug-301.sil
❌ tests/bug-317.sil
❌ tests/bug-337.sil
❌ tests/bug-524.sil
❌ tests/bug-525.sil
❌ tests/bug-61.sil
❌ tests/bug-75_darwin.sil
❌ tests/bug-76.sil
❌ tests/combining.sil
❌ tests/footnote.sil
❌ tests/insertions-sidenotes.sil
❌ tests/linespacing2.sil
❌ tests/mini-arabic.sil
❌ tests/vertical.sil
```

alerque:
> OK, so this is what I have failing locally:
> 
> ```
> Failed tests:
> 
> ❌ tests/bidi.sil
> ❌ tests/bug-192-full.sil
> ❌ tests/bug-75_darwin.sil
> ❌ tests/insertions-sidenotes.sil
> ```

YAY! We're almost there! You (on OSX) and I (on Archlinux) have the same test results now.

* `tests/bidi.sil` does not have an expected file commited yet, you must have created one locally. If we are ready to add the output of this test do we really need Hebrew or could we go with Arabic and save ourselves a font in the test suite?
* `tests/bug-192-full.sil` doesn't even exist in the repo, again your local clutter
* `tests/bug-75_darwin.sil` is Darwin only, so I can't test it. It a bit I'll ask you to try to update the expect for this if you can tickle the bug.
* tests/insertions-sidenotes.sil` again does not have an expected file in the repo.

I'm still working on getting Travis to behave.

alerque:
> Are you sure it's not font related? A suspicious number of the failing tests use ...

No at this point I'm not sure at all. Likely you're on to something, I just can't figure out what. In fact fiddling with `bug-61.sil` changing it to *Gentium Plus* does not exhibit the same off-by-a-smidge error *Libertinus Serif* does. On the other hand there are at least a couple tests that use each of the otherwise failing fonts that pass.

* `kannada.sil` and `bug-76.sil` use *Libertinus Serif* but pass fine. Interestingly both use other fonts on the same line.
* `bug-524.sil` uses *Noto Sans CJK JP* and passes fine. Note the test that's failing (`bug-525.sil` is the one that has Latin ruby text)

alerque:
Well well well. Commit e75671a worked to make `bug-524.sil` and `bug-525.sil` layout the same on my system and on Travis while still using the same *Noto Sans CJK* font. Ridiculously in the latter case the *Gentium Plus* based output is _after_ the initial call to the Noto font. I'm baffled at this point, @simoncozens do you have any idea what might being calculated differently in these situations?

alerque:
Just FYI the git history in this branch is buggered right now. I was making alternating changes to two different branches so that I could spend half as long waiting on Travis for results and biffed a rebase across a merge point. I can straighten it out pretty easily by cherry picking everything back in the right order (and I have both branches to work from) but git just released 2.22.0 today which includes a feature that should make fixing my mess even easier. I'm going to wait until the Archlinux package hits (<48 hours) to try to fix this PR just to give the new feature a real life use-case.

alerque:
The all new `git rebase --interactive --rebase-merges` is my hero.

alerque:
Thanks to #350, `make test` is now unreasonably smart about when it needs to rerun a regression test. It knows EVERY source file that would effect the outcome of the test. As a consequence is is able to run very quickly in many cases by only re-executing regression tests that have any chance of being affected by a change.

This has other interesting side effects too, such as being able to see what sources we are or are not testing:

```sh
# What tests load the lorem packages?
$ grep -R packages/lorem.lua .deps/ | cut -d: -f2
tests/alignment.actual
tests/split-footnote.actual
tests/footnote.actual
tests/bug-248.actual
tests/balanced.actual
tests/linespacing.actual

# What tests have Turkish language?
$ grep -R languages/tr.lua .deps/ | cut -d: -f2
tests/counters.actual
tests/bug-284.actual
tests/bug-283.actual
tests/hyph-except.actual
tests/bug-269.actual

# What tests use the URL package?
$ grep -R packages/url.lua .deps/ | cut -d: -f2
```

Oops, nothing is testing that ;-)

Or here are all the files being used in any test from my system rather than from the current directory I'm testing SILE from:

```sh
$ find .deps -name '*.d' -execdir sed -e 's/ /\n/g;' {} \; | pcregrep '^/' | grep -v $(pwd) | sort -u
/usr/share/lua/5.3/ltn12.lua
/usr/share/lua/5.3/mime.lua
/usr/share/lua/5.3/socket/headers.lua
/usr/share/lua/5.3/socket/http.lua
/usr/share/lua/5.3/socket.lua
/usr/share/lua/5.3/socket/url.lua
/usr/share/lua/5.3/ssl/https.lua
/usr/share/lua/5.3/ssl.lua
```

alerque:
Hey @simoncozens (or @khaledhosny or @behdad) — I'm baffled by this one.

I've gotten everything coming out the same across platforms and most library versions, but for some reason when we compute line height metrics for *Libertinus Serif* or *Noto Sans CJK* we are coming up with different answers locally (Arch Linux & OSX) vs. on Travis (Ubuntu). I've tried with multiple versions of Harfbuzz on Travis (and currently using the latest 2.5.1 matching my Arch Linux) and the line height metrics are still different. 

* Same font files (Libertinus 6.9, etc.)
* Same Harfbuzz version (and the Harfbuzz's metrics are used, not Freetype's)
* Both local and Travis using fontconfig to find fonts (not the same version, but all this does is pass the font file off to Harfbuzz right?)
* Of course the same SILE code & shaper

What else is there that could affect our line height calculations only for these fonts and not for all the others? 

Note only lines in which the _tallest_ thing in the line is these fonts matter. The fonts can be included in the document and text rendered, as long as _something_ on the line is larger than these fonts.

By example this works:

```tex
\font[family=Gentium Plus,size=15pt]{Gentium \font[family=Libertinus Serif,size=12pt]{Libertinus}!}
```

But this fails:
```tex
\font[family=Gentium Plus,size=12pt]{Gentium \font[family=Libertinus Serif,size=15pt]{Libertinus}!}
```

...And by fails I man the line metrics are ~0.2pt different in the version where Libertinus sets the pace.

alerque:
Okay with the linespaceing issue issolated and the related tests excluded, the branch this PR is based on is passing Travis with Lua 5.3 and 5.2 — 5.1 is being recalcitrant but that seems to be something in the new make-deps system. Something different about what variables it keeps paths in perhaps.

Almost there…

simoncozens:
Very nice. I’m presuming the deps calculations are automated. Of course, we should be able to tell what is and isn’t tested from coveralls reports. Yes, agree to setting linespace failures to knownbad.

alerque:
> I’m presuming the deps calculations are automated

Of course. Running `sile -m input.sile.d input.sil` will generate a makefile formatted list of all the files encountered while rendering the document. This can be used with `include input.sil.d` inside a Makefile to add all the dependencies to the target.

In the case of the test suite we assume that if the dependency file does not exist (first run) then of course you need to run the test. Thereafter it checks the timestamps of everything (including the sile executable itself, all lua packages, font files, etc.) and runs the test only if any of them are newer than the last test output.

alerque:
**I :green_heart: green.**

alerque:
I now have Travis testing 3 versions of Harfbuzz:

```yaml
  matrix:
   - LUA=lua5.3 COVERAGE=true HARFBUZZ=2.5.1
   - LUA=lua5.2 HARFBUZZ=1.4.2
   - LUA=lua5.1 HARFBUZZ=1.1.3
```

This should be the current stable, the latest Debian, and our minimum supported version. Since they are mix and matched with older versions of Lua it won't be entirely clear if a test breakage was because of older Lua or Harfbuzz, but I think a little intuition and actually looking at the errors should clear that up. The alternative was to either ⓐ only test one version or ⓑ expand the matrix and abuse Travis. This seemed like a reasonable compromise, what do you think @simoncozens?

simoncozens:
Is there anything wrong with expanding the matrix? If we can, let's let the computer do the detective work...

alerque:
> Is there anything wrong with expanding the matrix?

We _can_, and I [just did](https://travis-ci.com/simoncozens/sile/builds/115554604) but Travis seems to only assign up to 3 containers at a time, so any more than that and we have to wait for sequential execution. Given that Azure is currently way slower anyway this might not be the end of the world, but I'm wondering how much we actually gain...

Issue #230, opened 2015-11-30
Three (3) passes required to generate books with correct TOC


alerque:
After printing a draft run of a book on my laser printer and scratching my head for a while trying to figure out how I messed up flipping the pages, I realized my problem was with SILE itself. Based on the feedback given when generating a book from scratch and on experience coming from the LaTeX world, I assumed two passes was enough to render a proper TOC. It turns out it takes SILE 3 passes if the TOC itself takes up more than 1 page.

1. Generates the body of the book and stuffs all the TOC entries in the meta data file.
1. Re-renders the book with the TOC in place, this time stuffing the meta data file with the TOC entries again, but now that the TOC pages themselves have been rendered the page numbers will be right.
1. Re-renders the book with the TOC data that has correct page numbers this time.

This is somewhat less than ideal. At the very least it needs to be properly documented so people don't end up with books that have TOC page numbers that are off-by-one, but it seems like the last two steps should be short circuited by storing the body page number relative to the start of the body content, then doing math and adding the number of (extra) pages the TOC takes up to the page numbers references.

mhosken:
libtexpdf could be extended relatively simply to allow the insertion of a page out of order. There would need to be an insert_page() to parallel LASTPAGE() in pdf_doc_new_page and appropriate parameters.

With that in place one could create the TOC at the end of the document without having to reprocess the main text. The TOC would have to be processed twice. The first to get its length and so any page number offsets due to the TOC insertion and then the final rendering with good page numbers. But that's much quicker than rendering the whole doc 3x :)

alerque:
@mhosken That would work as long as the number of pages taken by the TOC is fixed. If inserting the TOC into  the document bumps all the pages up they would need to be renumbered on every page (unless you do the silly thing most books do starting page numbering with 1 on what is at least the 7th physical page).

In order to get the processing down to 2, the first pass needs to store all the TOC content and then _render_ it to pages and see how long it ended up being, then save that information. On the second pass it needs to block out that many pages in the output, then render the rest of the document updating the TOC data as it goes, then it needs to go back and render the TOC content itself into the pages it blocked out before.

For now my book project `Makefile`s just check to see if the page count changed between iterations and only runs a third pass if the count changed between the 1st and 2nd runs.

simoncozens:
My preference would be for SILE to detect that a rerun is needed and just do it. I don't believe typesetting is a specifically time-critical operation enough to justify doing clever PDF tricks, but at the same time let's automate what can be automated.

[...]

PR #593, opened 2019-04-25
Toc toc toc

alerque:
Work in progress for #230.

It turns out this is going to be harder than I thought. I figured we could just iterate over processing the main file 1 time, plus add an iteration when the `.toc` didn't exist, plus add an iteration for when the total page count changed since the previous iteration. Unfortunately there is a lot of stuff that is not cleanly initialized / finalized (counters, scratch variable space, etc) and the loop process is not doing smoothly yet.


Darkyenus:
Can the TOC package detect that it needs to typeset again? Since doing this in one pass would be pretty hard, and the number of passes may vary, I think that it would be more extensible to *just* allow any code to trigger another pass. (In which case SILE may want to stop after some configurable limit, to produce some output even if a package triggering passes repeatedly does not converge.)

alerque:
Yes, it can. In fact a I already got that bit to work, but we have about five layers of `init()`s and even more scattered `finish()` and the junk that gets created in between is not reset between passes because all code assumes pristine scratch space and that the whole app will just go away at the end of a document.

packages/tableofcontents.lua
```diff
@@ -30,6 +30,7 @@ SILE.registerCommand("tableofcontents", function (options, content)
   local tocfile,_ = io.open(SILE.masterFilename .. '.toc')
   if not tocfile then
     SILE.call("tableofcontents:notocmessage")
+	SILE.passes = SILE.passes + 1
```
Omikhleia, line 33: This is a wrong approach, IMHO, with a real bad separation of concerns if a package (here `tableofcontents`) increases a global counter.
Other packages could have a similar need of "auxiliary files" generated along the way to be updated by multiple passes (e.g. cross-references; lists of figures or tables if they decide not to leverage the ToC system; indexes; alternate not even mentioning alternate implementation of the ToC would allow multiple ToC subsets or ToC per chapters, etc.).

The general case cannot be solved this way. These packages are not all going to do a `+ 1`... They cannot know.

alerque: You're right, this global usage was a wrong approach. Packages could set some global status flag for requesting a rebuild, but they shouldn't be in charge of marking the current pass status.

PR #594, opened 2019-04-26
some minor corrections for now

jamadagni:


documentation/sile.sil
```diff
@@ -464,7 +464,7 @@ culpa qui officia deserunt mollit anim id est laborum.
 \line
 \end{verbatim}
 
-\noindent{}…you might not necessarily get a line break after ‘eiusmod’; you’ll
+\noindent{}…you might not necessarily get a line break after ‘tempor’; you’ll
```
alerque: This mistake appears to be the result of my mistakenly re-wrapping text in 83e1ebb0. Given that this same `verbatim` input example is used several other places in the manual, the best fix for this is probably to revert the wrapping to the way all the other examples are rather than adjusting it here. That also makes it a little more clear in the output that the text in the example isn't just soft-wrapped.

Alternatively we could update _all_ the examples, perhaps with even slightly shorter line width in the input to make the examples even more obvious.

documentation/sile.sil
```diff
@@ -1668,7 +1669,6 @@ Now we can finally complete our implementation of centering:
 \\set[parameter=document.rskip,value=0pt plus 100000pt]
 \\set[parameter=document.spaceskip,value=0.5en]
 \\set[parameter=current.parindent,value=0pt]
-\\set[parameter=document.parindent,value=0pt]
```
alerque: Why did you drop this line? Without it the example will only work for one input paragraph instead of for several.

jamadagni: Oh. This seems to have been naïve of me but I deleted it because the actual usage of this centering formulation in the following lines [doesn't have this line](https://github.com/simoncozens/sile/pull/594/files#diff-58dba0558da425ff2c2d4e38a9c14421R1681). Now I realize this is because it is not needed only because there is only one example paragraph but it would be needed elsewhere. Please feel free to edit my PR.

documentation/sile.sil
alerque: Bah, this is fine. The other examples aren't `verbatium` so they are not comparable. This fix is fine.

documentation/sile.sil
alerque: Fixed. I amended your commit to resolve this and fix the commit message.

alerque:
Sorry it took so long to fix up and merge, but thanks for taking the time to contribute this sort of thing @jamadagni — even minor fixes make a difference.

Issue #588, opened 2019-04-24
Normalize release process


alerque:
Building a source release should be a one step process. The `Makefile` should have everything necessary to build and validate an entire release. The impact on the repository should be one commit.

Things that need to happen as part of a release (not necessarily all done from one command, but all the things that need to happen to cut a release).

## General todo

- [x] Add configure flags to prefer system Lua libraries over bundled ones where 
possible, see #659.
- [x] Make sure `lua_modules` ~~and `.fonts`~~ are populated and included in the source tarball, e.g. see #659.
- [x] Make sure paths in tarball on release tags don't have extra strings

## Things to happen Pre-release

- [x] Version numbers in `core/sile.lua` and `configure.ac` updated, started in #590 and operational in #758.
- [x] ~~Recompile (serialize?) languages with hyphenators, see #633.~~
- [x] Make sure the libtexpdf submodule SHA is current

## Things to happen at Release

- [x] Tag a new release number
- [x] `make docs` to include an up to date manual build in source tarballs (See #585)
- [x] `make examples` to regenerate examples and add PDF/PNG resources referenced in documentation (and/or website), e.g. see #595 
- [x] Tag the commit with the release version number
- [x] Push tags (and start a release on Github?), see #759.
- [x] Validate release: build from tarball and run tests (CI deploy script)
- [x] Check that the changelog (`NEWS.md`) contains an entry for the new version
- [x] Attach binary release artifacts from CI builds (Linux, Windows) to Github release tag (CI deploy script)

## Things to happen Post-release

- [x] Attach latest manual and examples to _____ (Github release? Release blog post?)
- [x] Add a make target to summarize changes since last release for `NEWS.md` editor (See below)  and post to gh-pages branch.
- [x] The manual also to be updated in the gh-pages branch
- [x] Ping distro packagers
- [x] Update Github milestones and bump any remaining issue to next

simoncozens:
The version number in `core/sile.lua` should probably be filled in by autoconf anyway, so it's not manually repeated in different places.

alerque:
I plan to generate the release number from `git describe --tags` anyway, so it will always be current any time `make` is run. We can add it to PDF meta data too ;-)

simoncozens:
I still want the changelog written by hand, but I found it useful in the past to have a summary of the commits for the changelog editor to use for reference, e.g.:

```
git log --pretty=format:'%h %d %s <%an>' --abbrev-commit c36bddc...HEAD  > changes-0.9.5
[...]

PR #595, opened 2019-04-26
Cleanup examples/test.sil

DavidLRowe:
This pull request includes really minor corrections to typos in test.sil, plus changes to the first and last lines of the text to emphasize "the" (rather than the current "the woman" in the first line, or nothing in the last line), in order to match the text as found on: http://www.gutenberg.org/files/48320/48320-h/48320-h.htm

test.pdf and test.png will need to be regenerated from this revised text.


examples/test.sil
```diff
@@ -290,14 +290,14 @@ At three o'clock precisely I was at Baker Street, but Holmes had not yet returne
...
 
      "I am following you closely," I answered.
 
-     "I was still balancing the matter in my mind when a hansom cab drove up to Briony Lodge, and a gentleman sprang out. He was a remarkably handsome man, dark, aquiline, and moustached-evidently the man of whom I had heard. He appeared to be in a great hurry, shouted to the cabman to wait, and brushed past the maid who opened the door with the air of a man who was thoroughly at home.
+     "I was still balancing the matter in my mind when a hansom cab drove up to Briony Lodge, and a gentleman sprang out. He was a remarkably handsome man, dark, aquiline, and moustached -- evidently the man of whom I had heard. He appeared to be in a great hurry, shouted to the cabman to wait, and brushed past the maid who opened the door with the air of a man who was thoroughly at home.
```
Darkyenus: SILE does not support `--` for en-dash, IIRC. The unicode character should be used directly instead.

DavidLRowe: I was following the usage elsewhere in the document. (See, for example, `They were admirable things for the observer -- excellent for drawing` halfway through the first paragraph.)

Certainly all the `--` occurrences could be changed. Should it be an en-dash or an em-dash?

alerque: It looks like most of the instances in that file should be _em-dashes_ (`—`). But of course that depends on whose style guide you're following.

simoncozens:
OK. So, this file dates back from the time when SILE had virtually no packages and classes. The aim was to have a "nice" piece of text that could be quickly and easily typeset so that users could feel confident both that their installation was working and that SILE could do its job of doing decent typesetting of the kind of documents they expect. At the same time, it's a kind of easy-to-get-your-head-around example to base your own documents on.

The indenting and wrapping and everything else just came from cutting and pasting from Gutenberg. It's not significant. We should probably use `\chapter`, yes, but I am not sure about only providing a one-page sample; the point is to show off the book class a bit.

Perhaps it could be renamed, the indentation/em dashes/etc. can certainly be improved, but if there are concerns about the size and problems sticking binary PDFs in git then let's deal with that issue separately. (Maybe as part of the build process?)

alerque:
> ... the point is to show off the book class a bit.

Okay knowing this I'm sure we can make that happen. Let's make it a nice example of using the book class. Not a full length book, but clean simple coding showing off the easiest, cleanest way to do something.

I'll keep thinking about what to do about raster resources and yes, probably do something with the build system — perhaps only regenerating them as one step right before a tagged release is the answer. For now lets just work on the `*.sil` source and worry about when to regenerate the build version in another issue (I'm adding it to the checklist in #588).

alerque:
I started fixing this up, but quickly realized I'm in over my head — I'm so accustomed to using my own book classes that I don't even know how to use `class=book` any more. There is to `\title` or `\part` or various other things I'm used to using. I fixed up the typography a bit, but haven't cleaned up the class usage yet. I might suggest some changes to *book* first!

alerque:
See #597 for some stuff I started working on for the *book* class.

alerque:
This has been sitting around too long, sorry @DavidLRowe. The book class is ... a bit lackluster. Having published many books using my own classes it's kind of harry coming along and having to do everything from scratch! But this PR wasn't about the book class, it was fixing some obvious issues with the content and there isn't any reason to hold it up. I've fleshed out the example a bit more including a title page, a TOC, use of the `\chapter` function, and a custom Lua function for the embedded letter. Nothing super fancy, and improvements *should* be applied here when the book class gets its long needed overhaul.

DavidLRowe:
No worries, @alerque, I was using it to test my SILE installation, then ended up reading the story and noting some typos (which I then checked against Gutenburg). Just trying to reduce the distraction of typos in the example.

alerque:
>  Just trying to reduce the distraction of typos in the example.

It was a good call. Those very typos and typographic issues (especially the double hyphen `--` instead of an em-dash `–`) caught us a lot of flack on Hacker News. People totally blamed the engine for the example content.

> The test example literally has a double hyphen "--" in place of an em dash [...]

> The faux em dash definitely caught my eye. [...] If these are the examples to pitch the system... it doesn't seem ready for primetime.

Never-mind that it handles Unicode em-dashes perfectly, people just rag on it without using it for anything or seeing how it works. Good examples matter. We're still a long way from being there...

follower:
FYI: The change of file base name from `test` to `book` in this PR appears to have broken the "A Story" example here: https://sile-typesetter.org/examples/

I've created a PR for a hopeful fix here: https://github.com/sile-typesetter/sile-typesetter.github.io/pull/15

alerque:
Ouch, I didn't realize the website was hotlinking stuff from this repository. Thanks for that catch.

See also https://github.com/sile-typesetter/sile-typesetter.github.io/issues/16

PR #597, opened 2019-04-29
Cleanup book class, set sane defaults, expand arsinal of tools

### LINKED ISSUE #598: Book gutter on wrong side by default
The default language & layout of the *book* class is LTR, but the gutter margin (and assumed binding) is on the right. This should be reversed.

### LINKED ISSUE #599: Book class needs some default way to handle titles
The book class currently has no default way of handling book titles.

I would suggest `\title` function, much like `\chapter`, with some default output styling as a "title page".

Along with this change I would suggest setting the default running header to the title & chapter instead of chapter & section.

### LINKED ISSUE #600: Book class incorrectly styles running headers
The current book cleass has some self defeating code resulting in unwanted indents in the header. The indents are removed, but outside of the closure that brings back in settings from elsewhere. The order of operations needs to be reviewed.

Concurrent to that specific issue, it seems there is some conflict of interest between functions which set content and functions that define style. These interests should be separated so that more than just indenting doesn't get messed up.

### LINKED ISSUE #601: Book class could use a \part command
Not all books use `\chapter` as their top level division, and having a `\part` command that behaved as expected would ease transition from LaTeX. :stuck_out_tongue_winking_eye:

alerque:
The book class is missing a few basic things I would expect (like `\part` and `\title` (or `\titlepage`?) handling. It also has a few things backwards (like gutter margin on the right in what is otherwise an LTR default layout).

This is a work in progress to fix those things.

- ~~[x] Flip gutter margin side, fix #598~~
- [ ] Handle book titles, fix #599
- [ ] Remove header margin and content vs. style issues, fix #600
- [x] Add `\part` command, fix #601
- [ ] ???


alerque:
No idea if that merge resolution makes any sense, this branch was only 2748 commits behind *master*! I couldn't even reason about what it was doing any more the code was so old. Some bits still seem to make sense, some probably are obsolete and should be backed out.

alerque:
Nope merge is borked. I'll probably just look through the commits again and cherry pick instead of trying to fix the whole branch.

Omikhleia:
This 5+ yer old PR now has conflicts. It isn't linked to any discussion regarding how the issues at stake would be tackled in a grander scheme, and tries to dabble into too many unrelated things at once (titles, headers, sectioning, margins, and even a "???" item)... It's unlikely it will happen as-is.

Issue #549, opened 2019-01-07
Enforce good commit messages


simoncozens:
Working on a 0.9.5 release changelog and it's horrible because I am undisciplined.

I would like a git commit hook which checks the commit messages is formatted as `[area] ...` where area is taken from a given list of areas.

alerque:
You're telling me! If it's hard for you to figure out what you mean reading your own commit messages, imagine the rest of us out here.

I don't think a commit hook is the right solution, because there are times when hacking on stuff and you just want to cobble together a commit as a rollback point in case something goes sideways while hacking. The issue here is that such development sessions should get cleaned up using rebase before they are pushed to any permanent branch. Some needlessly incremental commits can be combined (debounced in hardware lingo), others just reworded to properly describe what they do. Requiring proper commit messages should happen before  anything hits master or devel branches, but I wouldn't want my local machines slowing up the hacking process before the churn dies down.

Maybe just a reminder message in the commit hook, and an actual requirement on master/devel merges.

alerque:
Also just a reminder, besides the full commit change log the [closed issues in a milestone](https://github.com/simoncozens/sile/milestone/4?closed=1) should be a pretty good summary of what goes into a release — especially if feature branches are used and put through the PR workflow even for  little things we hack on independent of any listed issues. Besides having benefits for testing like CI checks, the merge workflow also groups commits by topic and the commit graph shape starts telling more of a story.

rjmunro:
I would start by blocking commits to master and making all commits go though a branch and a PR. That way other people can review them, and the detail of the PR. You can list all the merges into master with `git log --first-parent` - this will correspond to all the pull requests, which should be a better summary of the changes since the last release.

simoncozens:
We did that. What I'm looking to do is ensure that commits are categorised by bug fix, new feature, code cleanup, etc.

alerque:
I'm not sure how much sense it really makes to have subject line tags for `[bugfix]`, `[feature]`, etc. In my experience following the Linux kernel's (and Progit's) guidelines on always starting commit messages with imperative present tense verbs pretty much solves that while still leaving some room for things that don't neatly fit into one box or the other.

> Describe your changes in imperative mood, e.g. "make xyzzy do frotz" instead of "[This patch] makes xyzzy do frotz" or "[I] changed xyzzy to do frotz", as if you are giving orders to the codebase to change its behavior.

[...]

PR #603, opened 2019-04-29
Improve error and warning outputs with document location information

### LINKED ISSUE #602: inputs-common seems to be polluting global namespace
I am not sure if this is or isn't intentional, and I might be completely wrong as I am not a good lua programmer, but some code in `inputs-common.lua` seems to mistakenly using [global `content`](https://github.com/simoncozens/sile/blob/master/core/inputs-common.lua#L60). ([Two places](https://github.com/simoncozens/sile/blob/master/core/inputs-common.lua#L36) actually.)

I am writing a PR in which I can fix it, but I am not sure if it isn't intentional. I haven't find anything indicating that it is though.

Darkyenus:
This PR has two related but distinct commits.

First commit adds a pretty useful system of "location stack traces", for `SU.error` and `warning`. It expands on the idea of `SILE.currentCommand` with `SILE.currentCommandStack`. (I have kept them separate for backwards compatibility.) This allows to accurately pinpoint where in the document an error or warning happened (with --trace flag enabled). Example output:
```
[1] [2] [i] [ii] [iii] [iv] 
! Error loading language en-u-va-posix: no support for this language
  at font on line: 25, column: 23
  at code on line: 339, column: 47
  at process on line: 140, column: 80
  at font on line: 140, column: 63
  at process on line: 28, column: 78
  at font on line: 28, column: 27
  at boldfont on line: 140, column: 53
  at tableofcontents on line: 143, column: 1
```

In some cases it improves the output even without --trace flag, as `SILE.currentCommand` was sometimes set to string and had no location information.

The second commit slightly improves error messages in `core/harfbuzz-shaper.lua`. It is motivated by this line: "`Couldn't find face  at main.sil`" appearing in my error output. (I am still not sure what caused it, but the error will at least have a nicer formatting.)

This closes #602 .

Commits use the new style suggested in #549, but I can change it to anything else.


Darkyenus:
I see. I haven't noticed that.

If I understand correctly, the breadcrumb system is somewhat more elegant, as it needs to inject itself only in one place. On the other hand, it doesn't track the AST on which the command works, so it can't be used for location stack traces right now. Also, I'm not sure why is it exposed through `SILE.typesetter`, instead of staying in `utilities`, where it is defined.

However, this seems fixable. Right now, [`breadcrumbs.__call`](https://github.com/simoncozens/sile/blob/master/core/utilities.lua#L393-L404) does not look at its arguments and passes them as they are. However, since it is only used for commands, the arguments in `...` should only ever be `options` and `content`. It would be easy to extract and store the content alongside the command name to get location information as well. This could break some backward compatibility with external code, but it would not be hard to do this in a different method or only when optional parameter is specified.

This should cover all cases where this PR's `push`/`pop` was used, with the exception of directly typesetting strings through `SILE.process` (which is arguably not useful as they don't carry location information) and for `texlike_stuff` in the same place, which could be done manually (that does carry location information, right?).

Thoughts?

alerque:
> If I understand correctly, the breadcrumb system is somewhat more elegant, as it needs to inject itself only in one place. On the other hand, it doesn't track the AST on which the command works, so it can't be used for location stack traces right now.

What part of the AST would it be able to track that it isn't right now? I haven't reviewed your implementation yet (which I will do because I'm sure we can improve mine based on it, having two implementations and reviewing the differences is a great way to learn) so this is a question from ignorance, but I am not aware of what I could have been tracking there and wasn't.

> Also, I'm not sure why is it exposed through `SILE.typesetter`, instead of staying in `utilities`, where it is defined.

The rational for that is that SILE is able to process multiple documents (or document fragments, it doesn't really matter) in parallel. Where each one is in processing the content passed to it needs to be tracked separately for each typesetter.

> Both seems fixable. Right now, `breadcrumbs.__call` does not look at its arguments and passes them as they are. However, since it is only used for commands, the arguments in ... should only ever be options and content. It would be easy to extract and store the content alongside the command name to get location information as well.

How does extracting `content` get you location information? I passed `...` because I wanted to be able to trace function calls even if a document tampers with SILE's internals and introduces more arguments or changes their type or something like that. I didn't think the breadcrumbs utility itself had any business assuming what it would be passed even if SILE's internals only did one thing, the function should still be useful if used in unexpected ways.

> This could break some backward compatibility with external code, but it would not be hard to do this in a different method or only when optional parameter is specified.

At this point I'm probably the only one using `SU.breadcrumbs` in production. I would have to change a number of bits of code, but I'm not too worried about that. If we come up with something better for SILE's end I'll adapt.

> This should cover all cases where this PR's push/pop was used, with the exception of directly typesetting strings through SILE.process (which is arguably not useful as they don't carry location information) and for texlike_stuff in the same place, which could be done manually (that does carry location information, right?).

Why is content passed to `SILE.process` not useful to track? It still tells you something about the command chain, so wouldn't all the commands that came across in `content` be just as relevant as any other? Again I haven't actually reviewed your code in detail so I'm speaking from ignorance of your implementation here, but I don't understand how some chains of commands are considered not useful and others are. Or for that matter what you even are counting as "location information".

I guess I'd better go review now...

core/inputs-common.lua
```diff
@@ -33,7 +33,7 @@ local function debugAST(ast,level)
   if type(ast) == "function" then SU.debug("ast",out.."(function)") end
   if not ast then SU.error("SILE.process called with nil", 1) end
   for i=1, #ast do
-    content = ast[i]
+    local content = ast[i]
```
alerque: If you are ever rebasing you might throw these scope bugfixes into a different commit in case we cherry pick other pars of the PR without taking the whole thing.

core/sile.lua
```diff
@@ -240,10 +240,59 @@ function SILE.resolveFile(filename, pathprefix)
...
+    result[#result + 1] = SILE.currentCommand
+    lastIsExtra = true
+  end
+
+  return result, lastIsExtra
+end
+
 function SILE.call(cmd, options, content)
-  SILE.currentCommand = content
+  SILE.currentCommandStack:push(content)
```
alerque: Unfortunately `SILE.call()` is not a reliable way to hook into functions being called. They can (and even in SILE's own code, do) get called directly from `SILE.Commands["cmd"]()`.

core/sile.lua
```diff
@@ -240,10 +240,59 @@ function SILE.resolveFile(filename, pathprefix)
   return nil
 end
 
+SILE.currentCommandStack = {}
+--- Push currently processed command on to the stack to record the execution trace for debugging.
+function SILE.currentCommandStack.push(stack, command)
+  if command == nil then return end
+
+  stack[#stack + 1] = command
+  SILE.currentCommand = command
```
alerque: Why track this separately, won't it always be the last one in the stack anyway?

core/sile.lua
```diff
@@ -240,10 +240,59 @@ function SILE.resolveFile(filename, pathprefix)
...
+end
+--- Pop previously pushed command from the stack.
+--- Command is provided to check that pushes and pops are balanced.
+function SILE.currentCommandStack.pop(stack, command)
+  if command == nil then return end
+
+  local popped = stack[#stack]
+  if popped ~= command then
+    SU.warn("Unbalanced command push/pop. Expected '" .. popped .. "', got '" .. command .. "'.")
```
alerque: See overall review notes, but this fault condition will be reached in a lot of documents because it doesn't allow for there being multiple typesetters.

core/inputs-common.lua
Darkyenus: Ok, I'll create a smaller PR only with this and HB warning improvements.

core/sile.lua
Darkyenus: I agree, I haven't realized that at the time of writing it.

core/sile.lua
Darkyenus: Not necessarily. I thought so at first, but consider this sequence of events:
```
foo
\command{bar}
baz % <- error happens here
```
This will produce something like:
```
PUSH(whole document)
PUSH(command{bar})
POP(command{bar})
ERROR
```
Now, if the trace was based only on the stack, the location shown would be the start of the document. Now it will be somewhat closer - at the `command{bar}`. It isn't completely correct and I forgot to complete this feature, but the idea is, that it would be able to print something like: `error, after line 2 in block starting at line 1`. (Where "line 2" comes from last command, which is no longer on stack, and "line 1" is the top of the stack.)

This could be useful for accurately pinpointing, for example, where the frame is under/overfull, etc. (Arguably, that should have its own, even better diagnostic outputs, but that is currently out of scope for this PR.)

core/sile.lua
Darkyenus: Can you please give an example, where multiple typesetters are used in parallel?

Darkyenus:
> What part of the AST would it be able to track that it isn't right now?
> How does extracting content get you location information?

I might be completely wrong, but I will explain my reasoning. This is the implementation of `registerCommand`, the place where breadcrumb system injects its entrypoint:
```lua
SILE.registerCommand = function (name, func, help, pack)
  SILE.Commands[name] = SILE.typesetter and SILE.typesetter.breadcrumbs(name, func) or func
  if not pack then
    local where = debug.getinfo(2).source
    pack = where:match("(%w+).lua")
  end
  --if not help and not pack:match(".sil") then SU.error("Could not define command '"..name.."' (in package "..pack..") - no help text" ) end
  SILE.Help[name] = {
    description = help,
    where = pack
  }
end
```
`SILE.typesetter.breadcrumbs(name, func)` returns a function, which works exactly like `func` itself, i.e. takes two parameters, `options` and `content`. In addition, while it executes real `func`, it ensures that the command's `name` is pushed on to the `breadcrumb` stack. Using this name, it is possible to determine the location where (in code) the command was *defined*. It could be interesting to show that information as well, but this PR does not care about that.

When the wrappe command is called, the breadcrumb wrapper function is invoked. It takes whatever it was given (`options` and `content`, in this case) and passes it into the real `func`.  If an error happens during the execution of this command, it is possible to check the breadcrumb stack. It will contain the names of the commands which are being executed (which is great), but it doesn't contain locations from where they are called (e.g. on what line of my `.sil` the offending `\font[]{}` call which can't  find the font file is). Arguably, this is the most useful information we could present to the user in that case.

But this location is available to the breadcrumb wrapper. If it looked inside the varargs, it would find, that the second argument, the `content`, looks something like this:
```
{
 1 = "Text to be typeset with custom font.",
 tag = "font,
 line = 15,
 col = 3
}
```
(This answers: *How does extracting content get you location information?*)
And this is what my system is collecting and where it gets its location information. Breadcrumb system could do it as well, but it would have to inspect its varargs, which is can't, because it is designed to be generic. But, of course, there could be two variants, one generic, which wouldn't collect information, and one for commands, which would.

---

> Why is content passed to SILE.process not useful to track?

Right now (on `master`), anything that goes into `SILE.process` is stored inside `SILE.currentCommand`. Very often, that `command` is actually a string to be typeset. If an error occurs during the typesetting, it checks `currentCommand`, finds that there is no location information (= `line` and `col` properties of the AST), and prints an unhelpful message, without location information.

Since the whole system of this PR is designed to help with printing of `.sil` location information, and plain strings don't contain any, it is not useful to track. I still track it, because it could be useful to someone else, but it does not appear in the error output.

---

> You just can't hook in where this does and expect a 1 dimensional stack, SILE doesn't assume that about itself.

I didn't know that SILE was multithreaded, as I haven't encountered anything that suggested it. My bad.

core/sile.lua
simoncozens: I think that’s a bug (or at least a historical artefact) if they do, though. We should be able to assume SILE.call, especially if it makes this kind of hooking easier.

alerque: Working on fixing this, see #605.

simoncozens:
It’s not so much multithreaded as re-entrant. It’s possible to e.g. start typesetting one document, load up another and typeset that, and go back to the first one.

core/sile.lua
alerque: Checkout packages such as footnotes: https://github.com/simoncozens/sile/blob/c6a0c07911e50448acea660fcaf9a7d1a8bdaa55/packages/footnotes.lua#L40-L53 or [parallel](https://github.com/simoncozens/sile/blob/master/packages/parallel.lua).

Darkyenus:
I am not convinced that it would produce wrong results, even for re-entrant code.

For example in the case of error being thrown during typesetting a footnote:
```tex
Blah blah.\footnote{Blah.\undefinedcommand}
```

This produces something like (--trace enabled):
```
! Unknown command undefinedcommand at main.sil
  at undefinedcommand on line: 1, column: 11
  at footnote on line: 1, column: 26
stack traceback:
	.../local/Cellar/sile/0.9.5.1/share/sile/core/utilities.lua:53: in function 'core/utilities.error'
	...al/Cellar/sile/0.9.5.1/share/sile/core/inputs-common.lua:69: in function 'core/sile.process'
	...al/Cellar/sile/0.9.5.1/share/sile/packages/footnotes.lua:49: in function <...al/Cellar/sile/0.9.5.1/share/sile/packages/footnotes.lua:46>
	(...tail calls...)
	...al/Cellar/sile/0.9.5.1/share/sile/packages/footnotes.lua:60: in function <...al/Cellar/sile/0.9.5.1/share/sile/packages/footnotes.lua:59>
	(...tail calls...)
	/usr/local/Cellar/sile/0.9.5.1/share/sile/core/font.lua:48: in upvalue 'func'
	.../local/Cellar/sile/0.9.5.1/share/sile/core/utilities.lua:434: in field 'font'
	...al/Cellar/sile/0.9.5.1/share/sile/packages/footnotes.lua:59: in upvalue 'func'
	.../local/Cellar/sile/0.9.5.1/share/sile/core/utilities.lua:434: in field 'footnote:font'
	...al/Cellar/sile/0.9.5.1/share/sile/packages/footnotes.lua:46: in function <...al/Cellar/sile/0.9.5.1/share/sile/packages/footnotes.lua:45>
	(...tail calls...)
	/usr/local/Cellar/sile/0.9.5.1/share/sile/classes/plain.lua:203: in local 'f'
	/usr/local/Cellar/sile/0.9.5.1/share/sile/core/settings.lua:48: in field 'temporarily'
	/usr/local/Cellar/sile/0.9.5.1/share/sile/classes/plain.lua:200: in upvalue 'func'
	.../local/Cellar/sile/0.9.5.1/share/sile/core/utilities.lua:434: in field 'vbox'
	...al/Cellar/sile/0.9.5.1/share/sile/packages/footnotes.lua:45: in upvalue 'func'
	.../local/Cellar/sile/0.9.5.1/share/sile/core/utilities.lua:434: in field '?'
	...al/Cellar/sile/0.9.5.1/share/sile/core/inputs-common.lua:65: in function 'core/sile.process'
	...l/Cellar/sile/0.9.5.1/share/sile/core/inputs-texlike.lua:149: in field 'process'
	/usr/local/Cellar/sile/0.9.5.1/share/sile/core/sile.lua:204: in function 'core/sile.readFile'
	/usr/local/Cellar/sile/0.9.5.1/libexec/bin/sile:52: in function </usr/local/Cellar/sile/0.9.5.1/libexec/bin/sile:52>
	[C]: in function 'xpcall'
	/usr/local/Cellar/sile/0.9.5.1/libexec/bin/sile:52: in main chunk
	[C]: in ?
```

As you can see in the traceback, this happens while the used typesetter is changed by the `footnote` package. And the location trace contains both commands leading up to the crash, which is in my opinion correct and useful. If the stack was bound to the typesetter, it would contain only the first line of location trace, which is somewhat less useful.

alerque:
I'm starting to see what you are saying, sorry if I'm a bit dense. The fact is the implementations differ so much because their purposes were pretty different. I was trying to track nesting — I wanted to know what the _scope_ of each command was when it was run. You want to know the _sequence_ in which it was run, in other words the order it was processed not the position in the command tree.

I can see the utility in doing both, but I still think we should unify the implementations a bit. Note that in #605 I've cleaned up all the SILE core code so that `SILE.call()` is consistently used (and hence a reliable place to hook into. That should give us a little better place to start.

Darkyenus:
I am wondering how to merge them, since they track slightly different things. What is the main use case of the breadcrumb system?

I'll probably keep the unified stack from this PR, but record the typesetter for each entry. That way it will be easy to extract the breadcrumb stack for backward compatibility.

Darkyenus:
I have rewritten this PR, now it will generate even better error reports:

With --trace
```
! Couldn't find face ''
  at fontbug.sil "␤x␤x␤x␤x␤x␤x␤x␤x␤x…" after "x "
  at fontbug.sil l.6 col.1 \verbatim
```
Without --trace
```
! Couldn't find face '' at fontbug.sil "␤x␤x␤x␤x␤x␤x␤x␤x␤x…" near l.6 col.1 \verbatim after "x "
```
And similarly for warnings.

---

Notes about some decisions:
- Using `SILE.currentCommand` for extra context turned out to be somewhat confusing. However, the idea of tracking last popped command wasn't bad, so I implemented it with `SILE.currentCommandStack.lastCommand`. `currentCommand` is kept, but only for backward compatibility. If this information is present, it is appended with `"after"` prefix.
- As you can see in the examples above, strings are tracked and are shown (truncated) in error traces. I am not sure if it will be that useful, but it could be. (Newlines are replaced with `␤`, so it doesn't break formatting. I experimented with `\n`, but that was even more noisy.) When string is shown, it always looks at the first non-string on the stack to provide information as well. (This is how previous version worked.) When this happens, the real location is appended with `"near"` prefix.
- The stack tracks active `typesetter` and allows to filter by current one, mimicking how breadcrumb system works. I have not removed/unified this with the breadcrumb system yet, because I don't know what the most important features of that system are or how is it used in practice (@alerque ?).
    - Difference between this and the breadcrumb system is, that the breadcrumb system allows arbitrary items on the stack, while this currently doesn't (but it would be easy to add it)
- I have added (will in next push) ability to print out `PUSH/POP` stream via `SU.debugging("commandStack")`. This has already proven useful when debugging #604.

Darkyenus:
*NOTE: I have edited previous command slightly*

I am pretty happy with how this turned out. The traces it generates are very useful, especially with `--debug=commandStack`. I have added one more commit, which improves error messages for `\define`s from `doTexlike`:
```tex
\center{\undefinedcommand{}}
```
```
! Unknown command undefinedcommand
  at fontbug.sil l.1 col.9 \undefinedcommand
  at </usr/local/Cellar/sile/0.9.5.1/share/sile/classes/plain.lua:92> l.27 col.55 \process
  at </usr/local/Cellar/sile/0.9.5.1/share/sile/classes/plain.lua:92> l.27 col.25 \ragged
  at fontbug.sil l.6 col.1 \center
stack traceback: [...]
```
It does it by annotating the `content` AST with `file` property, because otherwise there is no easy way to access that information.

The only two items that should be noted before merging this are:
- Note that I have removed `SILE.currentCommand`, as it wasn't used anymore. This could be a breaking change, if somebody was using it.
- I have not touched breadcrumb system, but this one is ready to replace it. I just don't know what API should be provided to do that.

(I haven't found any real usages of neither feature in code on GitHub, but I don't know how much can that be trusted.)

alerque:
Thanks for the work on this, the output is looking really good. I'm pretty busy this weekend but I'll give this a shake down early next week and look over my own implementation and see what I can come up with as far as meeting both sets of needs.

core/utilities.lua
```diff
@@ -231,6 +253,18 @@ utilities.subContent = function (content)
   return out
 end
 
+-- Call `action` on each content node, recursively, including `content` itself.
+-- Not called on leaves, i.e. strings.
+utilities.forEachContentNode = function (content, action)
```
alerque: I think there is a naming problem here, "content" is not itself a "node" in the sense we talk about nodes elsewhere in the code. Also this seems more like a recursion function (depth) than an iteration function as named. The name of the function should reflect that.

core/baseclass.lua
```diff
@@ -24,10 +24,21 @@ SILE.setCommandDefaults = function (command, defaults)
 end
 
 SILE.doTexlike = function (doc)
-  doc = "\\begin{document}" .. doc .. "\\end{document}"
+  -- Setup new "fake" file in which the doc exists
   local cpf = SILE.currentlyProcessingFile
-  SILE.currentlyProcessingFile = "<"..debug.getinfo(2).short_src..":"..debug.getinfo(2).linedefined..">"
-  SILE.process(SILE.inputs.TeXlike.docToTree(doc))
+  local caller = debug.getinfo(2, "Sl")
+  local temporaryFile = "<"..caller.short_src..":"..caller.currentline..">"
```
alerque: Why overload the concept of "files" instead of just having a status flag of some kind to track when we are in what processing mode? The concept of faking out the file name doesn't quite sit right with me.

core/sile.lua
```diff
@@ -132,6 +132,7 @@ Options:
   -- http://lua-users.org/wiki/VarargTheSecondClassCitizen
   local identity = function (...) return unpack({...}, 1, select('#', ...)) end
   SILE.errorHandler = opts.traceback and debug.traceback or identity
+  SILE.detailedErrors = not not opts.traceback
```
alerque: `not not`? Are you trying to force a boolean? Maybe use SU.boolean() instead. Or just use it as is if you assume Lua knows how to cast it already.

core/sile.lua
```diff
@@ -240,10 +241,251 @@ function SILE.resolveFile(filename, pathprefix)
   return nil
 end
 
+SILE.currentCommandStack = {
+  --- Internal: push-pop balance checking ID
+  _lastPushId = 0,
+  --- Internal: Function assigned to frames to convert them to string
+  _frameToString = function(self, skipFile, withAttrs)
```
alerque: This should probably be added to the frame prototype as a `__tostring()` function just like our other internal objects have dump functions.

core/sile.lua
```diff
@@ -240,10 +241,251 @@ function SILE.resolveFile(filename, pathprefix)
...
+
+    --str = str .. "\n\n" .. self.pushTrace .. "\n"
+
+    return str
+  end
+}
+--- Push a command frame on to the stack to record the execution trace for debugging.
+--- Carries information about the command call, not the command itself.
+--- Must be popped with `pop(returnOfPush)`.
+function SILE.currentCommandStack:pushCommand(tag, line, column, options, file)
```
alerque: Is there any time we push anything to the command stack that is not a command?

core/sile.lua
```diff
@@ -240,10 +241,251 @@ function SILE.resolveFile(filename, pathprefix)
...
+    column = column,
+    options = options or {}
+  })
+end
+--- Push a command frame on to the stack to record the execution trace for debugging.
+--- Command arguments are inferred from AST content, any item may be overridden.
+--- Must be popped with `pop(returnOfPush)`.
+function SILE.currentCommandStack:pushContent(content, tag, line, column, options, file)
+  return self:_pushFrame({
+    tag = tag or (type(content) == "table" and content.tag) or "???",
```
alerque:
```suggestion
    tag = tag or content.tag or "???",
```
I don't think you have to do all the type testing here, if content isn't a table `.tag` won't exist an evaluate to `nil` anyway, allowing you to keep moving down the flow control chain. 

Also should this particular instance throw a `SU.warn()`? What scenario could through the third option here?

core/sile.lua
```diff
@@ -240,10 +241,251 @@ function SILE.resolveFile(filename, pathprefix)
...
+    file = file or (type(content) == "table" and content.file) or SILE.currentlyProcessingFile,
```
alerque:
```suggestion
    file = file or content.file or SILE.currentlyProcessingFile,
```

core/sile.lua
```diff
@@ -240,10 +241,251 @@ function SILE.resolveFile(filename, pathprefix)
...
+    line = line or (type(content) == "table" and content.line),
+    column = column or (type(content) == "table" and content.col),
+    options = options or (type(content) == "table" and content.attr) or {}
+  })
+end
+--- Push a text that is going to get typeset on to the stack to record the execution trace for debugging.
+--- Must be popped with `pop(returnOfPush)`.
+function SILE.currentCommandStack:pushText(text, line, column, file)
```
alerque: If we're pushing text on it, then this isn't exactly a command stack is it? Also if it's a stack then it is more than just "current". Maybe it should be renamed `SILE.contentStack` or something generic like that.

core/sile.lua
```diff
@@ -240,10 +241,251 @@ function SILE.resolveFile(filename, pathprefix)
...
+    self[#self] = nil
+    if SU.debugging("commandStack") then
+      print(string.rep(" ", #self) .. "POP(" .. popped:toString(false, true) .. ")")
+    end
+  end
+end
+--- Returns a stack of commands, which can be used to show information about location in document.
+--- Most relevant object is in the last index.
+--- Second return: recently popped value which could provide extra location information
+function SILE.currentCommandStack:locationStack(currentTypesetterOnly)
```
alerque: Function names like this should probably be verbs, maybe `:dumpStack()` or something like that.

core/sile.lua
```diff
@@ -240,10 +241,251 @@ function SILE.resolveFile(filename, pathprefix)
...
+  local result = {}
+
+  for i = 1, #self do
+    if not currentTypesetterOnly or (self[i].typesetter == SILE.typesetter) then
+      result[#result + 1] = self[i]
+    end
+  end
+
+  local extra = self.lastCommand
```
alerque: This block is _really_ confusing. Don't test for something just to skip other tests, lets rearrange this entirely. Perhaps make it two tests, one `if currentTyesetterOnly and ...` and another for `if extra == ...`. But some of those checks seem to be done in previous blocks, so I'm really confused not just because the flow control here is weird but because it's unclear why it would be necessary at all.

core/sile.lua
```diff
@@ -240,10 +241,251 @@ function SILE.resolveFile(filename, pathprefix)
...
+--- Internal: Format top of the stack from locationStack(), may return nil
+function SILE.currentCommandStack:_locationTraceEntry(stack, after)
+  -- Stack may contain mix of tables and strings.
+  -- We hope to get a table, because that contains position information.
+  -- A string may also be useful, but preferably with a table as well.
+  local top = stack[#stack]
+
+  if not top then
+    -- Stack is empty, there is not much we can do
+    if after then
```
alerque:
```suggestion
    return after and "after " .. after or nil
```

Why don't you give `after` a `__tostring()` function (if it does't already have the default one, and simplify the flow control with `return testval and a or b`.

core/sile.lua
```diff
@@ -240,10 +241,251 @@ function SILE.resolveFile(filename, pathprefix)
...
+  return self:_pushFrame({
+    text = text,
+    file = file,
+    line = line,
+    column = column
+  })
+end
+--- Internal: Push complete frame
+function SILE.currentCommandStack:_pushFrame(frame)
+  frame.toString = self._frameToString
```
alerque: Lua already has a `tostring()` function as well as `__tostring()` methods on tables. Naming this `toString()` is confusing — if you need to pass arguments then at least call it `.dump()` or something like that, but I would try to standardize on a `__tostring()` that takes no arguments and if you need a special output beyond that build it on top of that.

core/sile.lua
```diff
@@ -240,10 +241,251 @@ function SILE.resolveFile(filename, pathprefix)
...
+          or (currentTypesetterOnly and (extra.typesetter ~= SILE.typesetter)) -- no extra if wrong typesetter
+  then
+    extra = nil
+  end
+
+  return result, extra
+end
+
+--- Internal: Format top of the stack from locationStack(), may return nil
+function SILE.currentCommandStack:_locationTraceEntry(stack, after)
```
simoncozens: after what?

core/sile.lua
simoncozens: No, disagree. This is returning a value, which is fine. A later function actually dumps it.

alerque: Okay right dump was the wrong term, but still what does the function _do_? See my other comment about the naming of this object, but along with that maybe this should be `getLocations()` ... or especially with the argument it should be called `filterSomething()`.

core/utilities.lua
Darkyenus: I haven't soaked up many naming conventions yet, how else would you call it? The `forEach` part could be renamed to `traverse`. (I personally think that even though it is implemented recursively, it essentially still iterates. So `forEach` isn't a bad name, but I have no strong opinion one way or the other.)

core/baseclass.lua
Darkyenus: I am not sure what you mean by this. The `currentlyProcessingFile` change has been there already, but that system can't be reliably used for reporting in which file something was `\defined`, see my [previous comment](https://github.com/simoncozens/sile/pull/603#issuecomment-489264488).

If you refer to the `file` field added to the AST, I don't think that is overloading, as it literally is the file in which the AST node has been defined.

Can you please elaborate?

core/sile.lua
Darkyenus: Yes, this is just to force a boolean. I didn't know about `SU.boolean`, but I don't really need that logic here (I think). And I didn't want to keep it as it is, because then it creates a weird API, which does not communicate intent. (And that mindset leads to problems in the long run, e.g. with GC.)

Cast to boolean would be better, but Lua doesn't have that (or at least I couldn't find it).

core/sile.lua
Darkyenus: I had implemented it that way at first, but I couldn't find in documentation whether `__tostring` supports additional parameters. (And even if some implementation does, unless it is documented it could just be a bug.)

core/sile.lua
Darkyenus: Yes, text and possibly other stuff, after merging with breadcrumbs or if such need arises.

core/sile.lua
Darkyenus: I do believe it is needed, though only in edge cases:
```lua
> content=false
> content or content.tag
stdin:1: attempt to index a boolean value (global 'content')
stack traceback:
	stdin:1: in main chunk
	[C]: in ?
```

Warning seems like a sensible idea.

core/sile.lua
Darkyenus: If you consider literal text to be a command to typeset that text, then it can be a command. But then, if we allow to push any information onto the stack after merging with breadcrumb system, it won't even be a `contentStack`. Naming things is hard.

I would however still consider the `current` to be correct, as the items lower down on the stack are still being currently processed.

Then the name would be `currentStack`, which is still weird. `traceStack`?

core/sile.lua
Darkyenus: Names of these things have evolved over the time, so they don't make perfect sense right now. The names have grown on me, so I didn't realize it. I'll think of a better names once the name for this whole structure is decided.

core/sile.lua
Darkyenus: The second check is not needed anymore, as the `lastCommand` logic has changed, I'll remove that. The typesetter check is done a few lines earlier, but on different objects. It is still needed here, to properly filter the `extra` value.

That said, I have just now realized, that this whole method is no longer needed, except for filtering by typesetter. (Which probably seems obvious to you.) So the logic could be greatly simplified anyway. If the whole method isn't removed in the breadcrumb merge...

core/sile.lua
Darkyenus: See above comment about `__tostring`. But this would be possible even without it, I just arbitrarily wrote it this way.

core/sile.lua
Darkyenus: I don't want to call it `dump`, because that would imply (at least to me) that it is intended as a debug info. But this actually formats an user-facing string. I can't think of a better way to call this, as it is a method on frame, which converts it into a string. But it could be confused with `tostring`, so would `toFrameString` be better?

core/sile.lua
Darkyenus: This is a second return value of `locationStack()` (bad name, I know). 
It is a frame which was popped recently and its description will be appended to the result. It is called "after", because the SILE's position in the document is somewhere *after* that command.

It is hard to think up names for these things.

core/utilities.lua
alerque: How about `utilities.mapContentRecursive()`, or better yet `utilities.walkContent()`? Really what you're doing here is recursing through a deep tree, not looping through a flat array (we have that and call it `utilities.map()` already).

core/sile.lua
alerque:
```suggestion
  SILE.traceback = opts.traceback
```

I would avoid calling it one thing in the user facing opts and a different thing internally, if the user facing name is a good pick then using that as the internal representation should be a good pick too. As for forcing the variable type, I don't think you need to at all. Everywhere it's used it is tested as a boolean and if the casting works here it will work everywhere else.

core/sile.lua
alerque: It looks to me like `skipFile` is _always_ `false` and `withAttrs` is _always_ `true`, I can't see a code path where they are set differently. It's very difficult to evaluate what something should be called if there is no context for when it might be used. In this case I think I'd drop the arguments and just make that those the only behavior until such a time as the potential use for alternatives is clear.

In reviewing this, I also discovered the frame prototype _already has_ a `toString()` function from @simoncozens.

https://github.com/simoncozens/sile/blob/master/core/frame.lua#L74-L82

This should get re-used and extended if necessary, but I'd still argue we should make it a `__tostring()` meta property so it doesn't even have to be explicitly called (as we do for debugging other internal types).

core/sile.lua
alerque: I understand the potential Lua edge case but I don't see such an edge case in our usage. Looking at the code that _calls_ this command it is **already assuming** `content` is a table. If it's not a table this code shouldn't get called at all. If we are supposed to be passed a content table (even an empty one) then we should assume we can treat it accordingly. If we get the wrong data type we should raise an error about it and die anyway rather than silently suppressing an error condition.

core/sile.lua
alerque: I hope you don't feel like I'm nit-picking here. I realize we have quite a few dead code paths already in the repository, but these have grown up over time as things got refactored and dead code paths don't get caught right away. In a case like this of adding new code though I don't want to see stuff added that is effectively dead out of the gate.

core/sile.lua
alerque: Why do you call it "user facing" and contrast that with "debug info"? The only place this function is even called currently is in the context of `SU.debugging("commandStack")`, so that seems like a false dichotomy at this point. Did you have another use case in mind?

core/utilities.lua
Darkyenus: `walkContent` it is then.

core/sile.lua
Darkyenus: Ok. But I still think that keeping this as a boolean leads to a more predictable and cleaner code.

core/sile.lua
Darkyenus: The parameters are used:
- `skipFile` [here](https://github.com/simoncozens/sile/pull/603/files#diff-8e3ebbe41b16928909ff53e84982f1d1R430) and a few more other places
- `withAttrs` is usually unspecified and thus `false` (`nil`)

So it would be incorrect to drop them.

That `frame.lua` refers to a content frame. What I refer to as a frame here is a [stack trace frame](https://en.wikipedia.org/wiki/Call_stack#STACK-FRAME).

core/sile.lua
Darkyenus: The philosophy of this code is, that it is non-essential to the typesetting process and that it should never cause a crash. I can't predict how will the code evolve and I'd rather print a nonsencical trace than crash an otherwise valid program. This would be a lot easier to enforce in a statically typed language, but alas.

core/sile.lua
Darkyenus: I agree with that, it was an oversight.

core/sile.lua
Darkyenus: No, the `frame.toString` is the main source of the error trace information. The internal architecture is:
- There are stack frames on `currentCommandStack`. Those contain information about some call/command/text which is being processed.
- Each stack frame also has a `toString` method, for convenience.
- When a short or long location message is generated through `locationInfo` or `locationTrace`, `_locationTraceEntry` is called for relevant stack traces. This looks at single stack frame (+ some other frames to generate "after" and "near" contextual info), converts it to string via `toString` and pieces it together with other contextual info.
- For `locationInfo`, the output of `_locationTraceEntry` is enough.
- For `locationTrace`, `toString` is also called for other frames on the stack, to capture all information.

`toString` is what converts collected information (for example from `content` AST), into something like: "fontbug.sil l.6 col.1 \verbatim".

Darkyenus:
I have cleaned up what I thought needed cleaning up. As I am nearing my deadlines, I won't be able to work on this further for some time.

simoncozens:
Thanks so much for all your work on this. It looks great! I’ll let @alerque look over the implications for the other breadcrumbs system, and hopefully we can resolve the two ideas, tidy up anything else that needs tidying and get it merged. Good luck with those deadlines and thanks again!

core/sile.lua
alerque: I understand that goal, but I think in some ways your code (was) working against that. Unfortunately with Lua not being statically typed you can't enforce good practice or easily catch mistakes at compile time. I can understand not wanting to crash inside the tracer, but I think unnecessary code complexity is almost a worse enemy. I'd rather start causing a crash early in the development process if we change something and start passing around the the wrong type of content than silently hide the fact that the content type changed and potentially be messed up for a long time before anybody catches the miss-match.

If you really want to be robust, test to make sure it is a table right out of the gate and `SU.error()` if not. Then after that you can code assuming it is a table. In this case I don't even really recommend that because we don't have any way to test the error condition because we would be crashing elsewhere in the code before even getting to that point, so confirming we coded the error condition would correctly would be tricky.

core/sile.lua
alerque: Much better naming...

core/sile.lua
alerque: Thanks for the explanation. This makes a _LOT_ more sense if your most recent code, and I'm following that a lot better now.

The one concern I have left (and based on this explanation I now realize was a large contributor to by confusion) is that `frame` might be perfectly appropriate for stack tracer **except** in the content of a typesetting application that handles everything with in frames. I think we should universally stick to "frame" meaning a content frame on a page across the code base and use something else for whatever unit of content you are tracing here.

core/baseclass.lua
alerque: I see this is a more systemic issue in SILE already and not really introduced my this PR. I'll consider addressing it another time, this PR is already hard to follow and shouldn't be tasked with fixing unrelated problems.

alerque:
@Darkyenus Did you turn off contributors being able to add commits to this branch? I have some Lua coding style cleanup I'd like to add onto this if you don't mind sa that this feature lands in one piece.

Darkyenus:
@alerque I did turn it off, because I have seen this:
Screen Shot 2019-05-06 at 09 50 31

But I couldn't find the commit anywhere in the remote branch (and I still can't), so I have assumed that it is buggy and disabled it just in case. I'll re-enable it, as you probably know what you are doing.

Darkyenus:
Btw. I'd advise against modifying the commits. There are three of them and amending any other than the last one may lead to conflicts or hard to spot mistakes. (That is why my last fix is a separate commit and not an amend of the first one.)

alerque:
I accidentally pushed a commit based on top of your work to your branch instead of my own (I wanted to run it through Travis). I backed it out again when I realized I'd pushed it to the wrong remote. Sorry for the confusion on that.

I won't amend your commits, just do a little bit of follow up work.

core/sile.lua
alerque: ...I can't come up with anything better so I'm actually going to tie off this issue too unless somebody has a brilliant suggestion.

alerque:
See #606 for where I'm starting work on a Lua coding style guide. The blank line thing is definitely a matter of personal preference: I use `vim`'s text objects to skip around over "paragraphs", and it's nice when those correspond to functions. Having a function with 3 lines of code and 4 comments that takes 3 motions to jump through instead of 1 is obnoxious. But of course that is pretty specific to my editor.

If functions become so large that they take up more than a screen full then breaks do help, but at that point I'd also argue the function itself probably needs splitting.

Note I also dropped some comments, such as in 8ec5167. In general code comments are a good thing, but not when they explain something that is otherwise obvious. For example your comments explain the scope of a lexical structure became quite redundant when the lexical structure became so simple that the code was simpler than the comment.

core/tracestack.lua
```diff
@@ -0,0 +1,214 @@
...
+      if SU.debugging("commandStack") then
+        print(string.rep(" ", #self) .. "POP(" .. popped:toLocationString(false, true) .. ")")
+      end
+    end
+  end,
+
+  -- Returns frame stack represented by SILE.traceStack and self.afterFrame, but without frames whose typesetter
+  -- is different than what is given as a first parameter (nil = use current typesetter for filtering).
+  -- Most relevant object is in the last index.
+  filteredByTypesetter = function(self, typesetter)
```
alerque: As far as I can see, this function is no longer called at all anywhere. Does it still have any utility? If not I'll remove it, I'm already refactoring other functions a little so don't worry about doing it @Darkyenus, I'm just looking for feedback to make sure I'm not missing something.

core/tracestack.lua
```diff
@@ -0,0 +1,214 @@
...
+  -- Internal: Function assigned to stack frames to convert them to human readable location
+  _frameToLocationString = function(self, skipFile, withAttrs)
+    local str
+    if skipFile or not self.file then
+      str = ""
+    else
+      str = self.file .. " "
+    end
+    if self.line then
+      str = str .. "l." .. self.line .. " "
```
alerque: Would there by any objection if I edited this to roughly match the output format of `debug.stacktrace()` which uses `filename:N` instead of `filename l.N`, and so forth?

core/tracestack.lua
```diff
@@ -0,0 +1,214 @@
...
+        text = text,
+        file = file,
+        line = line,
+        column = column
+      })
+  end,
+
+  -- Internal: Push complete frame
+  _pushFrame = function (self, frame)
+    frame.toLocationString = self._frameToLocationString
```
alerque: This seems like a bug. Given that there is only ever one instance of `SILE.traceStack`, the function assignment here is always going to return the same thing. As such there is so point in saving a pointer to it with each "frame" on the stack, and calling it can be a bit simpler too.

Am I missing something or can I just factor this out and use `self:toLocationString()` instead of `popped:toLocationString()`.

core/tracestack.lua
Darkyenus: It is not used. I have kept it to ease possible future merger with breadcrumb system, as I understand that separate traces per typesetter are an important feature of it.

If you decide that it is no longer necessary, feel free to remove it. You can then also remove [`frame.typesetter = SILE.typesetter` in `_pushFrame`](https://github.com/simoncozens/sile/pull/603/commits/edafe1e9ea16ba743b1fde139a06d38546874ee0#diff-8e3ebbe41b16928909ff53e84982f1d1R331) as that exists only to support this function.

core/tracestack.lua
Darkyenus: I wanted to do that myself, but I wanted to conform to the existing style of previous location information.

However note, that the file is not always printed - it may not be known or (more often) it is suppressed (for example when two frame infos are on the same line and they both refer to the same file). So rewriting this naively to `str = str .. ":" .. self.line .. " "` **will** lead to rather cryptic outputs.

core/tracestack.lua
alerque: Roger that. Working on that now.

core/tracestack.lua
alerque: Hm. Also it leaves the column info with no good home. I might not touch that, I was just wondering...

core/tracestack.lua
Darkyenus: No, this is correct. Think of it as "inheritance lite". All stack frame objects (erm, "tables") contain a function which can print them. In a more OOP style, I'd somewhere create a "base class" with this method and let further frames be instances of it. But that seems like overkill.

Should this line be removed, all instances of `<frame>:toLocationString(<anything>)` would have to be rewritten to `SILE.traceStack._frameToLocationString(<frame>, <anything>)`. As you can see, that is very verbose, so I did it this way.

Additional benefit of this for the future would be that different frames could have different printing methods. Right now, there are two types of frames: command (created by `pushCommand` and `pushContent`) and text (`pushText`). Both have slightly different fields, so they are printed differently, but the difference is tiny compared to the amount of shared code, so the `_frameToLocationString` deals with that internally. However, if in the future we add something like `pushGeneric`, then it might be useful to allow users to provide their own `toLocationString` function.

Your suggested change would currently break this, as:
1. You are calling it without arguments. It expects at least the `self` argument (a frame), to know what it should print.
2. Other code expects `frame.toLocationString` to be a function, not a pre-formatted string (which it returns).

I have thought about storing the formatted frame information in the frame, but it would be slow and not flexible.

core/tracestack.lua
Darkyenus: The column is then usually formatted similarly: "`file.sile:12:34`"

core/tracestack.lua
alerque: My sample code was just to let you know where I was talking about, it wasn't meant to be a complete replacement. I see what's going no now, thanks for the explanation.

core/tracestack.lua
```diff
@@ -0,0 +1,214 @@
...
+    local afterFrame
+    if self.afterFrame and self.afterFrame.typesetter == typesetter then
+      afterFrame = self.afterFrame
+    end
+    return result, afterFrame
+  end,
+
+  -- Returns short string with most relevant location information for user messages.
+  locationInfo = function(self)
+    return self:_formatTraceHead(self.afterFrame) or SILE.currentlyProcessingFile or "<nowhere>"
```
alerque: Why is this function always operating on the _last item_ popped from the stack? Wouldn't it be more relevant to operate on the top of the stack? As in the last thing pushed instead of the last thing popped?

Darkyenus: It is called with first parameter being the stack itself (notice the `:` instead of `.`), so it primarily does operate on the top of the stack.

core/sile.lua
```diff
@@ -286,15 +291,50 @@ SILE.currentCommandStack = {
...
+        if traceStack[i].line then
+          -- Found a frame which does carry some relevant information.
+          info = info .. " near " .. traceStack[i]:toLocationString(--[[skipFile=]] traceStack[i].file == top.file)
+          break
+        end
+      end
+    end
+
+    -- Print after, if it is in a relevant file
+    if afterFrame and (not base or afterFrame.file == base.file) then
```
alerque: I just noticed `base` is never set anywhere in the scope of this function, what condition was this supposed to catch?

Darkyenus: That was supposed to be `local` with the frame carrying the location information which I removed in the last refactor. I blame Lua's lack of compiler, will fix ASAP.

alerque: I can do it, don't worry about it. Just checking for feedback here...

core/tracestack.lua
alerque: Yet the function it is calling doesn't name the variable as `self`, it names it something else which starts to get confusing. Basically you're passing it something it already has access to just to tip it off to use what it knows rather than ignore it. It's like _Inception_ ... don't worry I'm experimenting with ways around this.

Darkyenus: It is not called `self`, because it is not always called with `:`. In this case I am simply using the `:` as a syntactic shortcut to not have to write `self` twice.

core/tracestack.lua
```diff
@@ -61,20 +61,22 @@ local traceStack = {
...
-          info = info .. " near " .. traceStack[i]:toLocationString(--[[skipFile=]] traceStack[i].file == top.file)
+          locationFrame = traceStack[i]
+          info = info .. " near " .. locationFrame:toLocationString(--[[skipFile=]] locationFrame.file == top.file)
           break
         end
       end
     end
     -- Print after, if it is in a relevant file
-    if afterFrame and (not base or afterFrame.file == base.file) then
+    if afterFrame and (not locationFrame or afterFrame.file == locationFrame.file) then
```
alerque: Here `locationFrame` will never be anything other than `top`, and we already return out of the function if `top` is not set. Testing for `not x` would always be false, so the `or` construct here is more convoluted than in needs to be to get the job done.

Darkyenus:
With the last commit https://github.com/simoncozens/sile/pull/603/commits/e749406b59b9eb31373e9221402f4e356d4f56a3, it should be trivial to support new frame types:

```lua
SILE.traceStack.pushBreadcrumb = function (self, name)
    self:_pushFrame({
        breadcrumb = name,
        toLocationString = function (frame)
            return "breadcrumb("..frame.name..")"
        end
    })
end
```

alerque:
> it should be trivial to support new frame types:

That's quite interesting, but I feel like you are designing for a theoretical need. The two use cases I am trying to actually make this work for do not involve different frame types at all, they involve iterating over the stack (without popping anything) to deduce a different set of information that just dumping the trace would provide.

As such having different frame types with non-normalized values actually makes it harder, not easier, to use.

Darkyenus:
Well, that is why I repeatedly asked for an use case. (Though I wanted this change anyway.)

What kind of information are you iterating for? And how would this approach make it hard? Most of the keys are shared between all frame types, if they support it. And if they don't support it or don't have it, you should probably ignore that frame anyway.

Darkyenus:
I don't like where this is going. This PR used to be self contained, now it is a mess of unrelated changes. I am going to close this and open a new PR only with relevant changes which I believe are good. You can merge it or leave it - I need it to be stable and work, so I am going to continue in my fork, for the time being.

alerque:
@Darkyenus I'm sorry about the mess, again I got by branches crossed and pushed right before leaving the office — and didn't notice I'd pushed to this PR instead of my own experimental branch of it. Those commits were not meant to be on here yet.

That being said I would ask that you be patient for a little while while I work through this. I don't like the other PR with both of our work flattened and changes to other parts of the system not being in their own commits. You are of course welcome to run that version for now (I run my own fork for production work too!) but one of the reasons I've held off merging this is because I don't think we have agreement on how the interface is going to work yet or what the output format is, and given that this is something other tooling will quickly build up around I don't want it to come out of the gate half baked.

Please don't worry about my experimental work until I at least have it working on par with yours. Then we can compare how they integrate with other tools.

simoncozens:
I'd say leave that for now and file a separate bug. It is, in a sense, not related to the command stack code - it is about where nodes get shaped. So we'll deal with that another time; it's not a blocker for this to be merged.

What is a blocker in my mind is the existence of two separate command stack implementations. (This one and breadcrumbs.) It still looks like we have the breadcrumbs one in this PR. Any problem with adding a patch to just take out the breadcrumbs system? If not, I suggest we do that and then merge this.

alerque:
@simoncozens I agree the issue with the sharper not activated until the following command starts (and hence the trace being inaccurate) is not a blocker.

I am still working on this and would like to hold on on merging for the moment until I can make it a functional replacement for breadcrumbs. At the moment it isn't quite there, but the latest data structure and function arrangement brings it close. At the moment I rely on `master + <pandoc>` for production work, and just removing the breadcrumb system would leave that broken for me. I will merge this as soon as I can get it working as a replacement.

alerque:
I've updated this PR with my latest work on it. Thanks again @Darkyenus for getting the ball rolling on this. Even without it being in production, having some trace data made my life easier on a rush weekend project!

Some comments:

1. I kept the API for _using_ breadcrumbs essentially intact as I think the flat list of command names greatly simplifies the use cases I have for it over parsing whole objects full of data; but the breadcrumbs _backend_ is just patched straight through to the trace stack. This way there is only one set of histories kept around. There are a few shortcomings of this current implementation.

    * It doesn't handle multiple parallel typesetters quite right. Then again the original didn't either, it just kind of plastered over the differences. Since this isn't a regression any more, I think we can work on that later. How exactly the yield system works and where to patch into it is a bit mysterious to me, and it's not altogether apparent we have the right parallel column approach yet anyway so it might change.

    * Introducing other kinds of frames other than the ones in this iteration  _might_ break breadcrumbs because it is not filtering for ones it knows how to handle yet, but I think we can cross that bridge when we come to it.

2. I did end up making frames into stdlib objects, added metatables, and make some use of `__tostring()`. This allowed me the flexibility to patch on other function to frame types, even not types added by third party code. It also allows things like having frames that don't output any location information at all. I think this flexibility will be useful down the road. The default interface is a bit simpler too and I think it's a bit more readily apparent what each of the remaining functions do and when/why they have arguments.

3. Turing on `-d commandStack` highlights a few things that are probably bugs in the SILE code. For example the frequency with which we ask the typesetter to typeset completely empty strings.

4. In order to make breadcrumbs work as expected had to add a new top level item to the stack (currently hardcoded as "document"). This should also help when processing multiple inputs, and we can probably adapt the currently kludged `texlike_input` "command" frame to be a document frame instead, but I'm not quite sure how that will play out yet. I have it excluded from the stack output as it currently doesn't have anything that isn't already in the second entry anyway (filename).

   Looking at Lua's debug trace, there is a top level item in there, so it might be more appropriate to output the document frame too. 

P.S. Just because I beat my head on it for a while maybe this reminder will save you all grief someday: Note that using `#table` works for getting the length only if using sequential numerical keys (e.g. the stack itself) but does not work reliably when using names keys (e.g. a frame on the stack). It took me forever to figure out why I was getting inconsistent results! You should remember to use `table.nitems(tbl)` from SILE's utilities in cases like that.

core/sile.lua
```diff
@@ -245,7 +245,7 @@ end
 function SILE.call(command, options, content)
   options = options or {}
   content = content or {}
-  if SILE.traceback and not content.lno then
+  if SILE.traceback and not (type(content) == "table" and content.lno) then
```
alerque: Hey @Darkyenus this is a case where I thought your code was overly complex and I turned out to be wrong and had to backtrack. I'm sorry I had to work through all this  and get to where you got the hard way --- but I hope I did improve things along the way too.

alerque:
Oh brother. So much for my attempt to normalize the usage of `cmd`/`command`/`tag`, `l`/`line`/`lno`, and so on that all refered to the same thing. It worked pretty well for the TeXlike syntax, but now there is a mismatch between the TeXlike and the XML variants. The XML variant uses "tag", and that comes from an external library so we probably shouldn't mess with it.

I'll ... um ... let me think about that.

alerque:
This is quite the rabbit hole.

My addition of a top level `<document>` frame has somehow broken the `\include` statement, and we're throwing a warning about unballanced push/pop on _tests/bug-465-a.sil_.

That would be fine, but throwing a warning for something we have flagged as a bug also triggers the Lua debugger.

And _that_ would be fine except the Lua debugger doesn't work the same way in Lua 5.1.

I can fix the unbalanced push/pop, but we'll still need to think about the fact that our stack trace mode won't work properly on Lua 5.1.

Darkyenus:
Good to see some progress on this. The amount of changes that were added on top would amount for at least two more PRs, but as long as I am not the one that will have to review it... ¯\_(ツ)_/¯

By the way, `-d commandStack` should probably get renamed to `-d traceStack`, for consistency.

alerque:
Large rebase just now wast to fix commitlint issues (missing module and too-long headers).

PR #605, opened 2019-05-01
Normalize lua coding

alerque:
- [x] Cleanup how functions are called.

    The SILE code has used a mix of `SILE.Commands["cmd"](...)` and `SILE.call("cmd", ...)`, probably just an artifact of history except for one use case: the former way can be used to retrieve a return value and the latter cannot. This normalizes all these calls and including making `SILE.call()` pass down return values.

- [x] Cleanup Lua coding style.

   The original code base has a ton of one letter variables and otherwise cryptic content. I started cleaning these up as I had reason to touch the code, but it would probably be worth a systematic replacement of at least some of the most common ones like `options` instead of `o`, `content` instead of `c`, etc.


alerque:
Here is a little shell script to make it easier to play with various Lua code formatters. It should be used as a filter that takes STDIN and writes STDOUT, and can be called with a single argument of which backend to use.

```bash
#!/usr/bin/env zsh

case $1 in
	3|formatter)
		luaformatter -s 2 -t 0 -d unix /dev/stdin
		;;
	2|reformat)
		lua.reformat /dev/stdin >(cat) 1>/dev/null
		;;
	1|fmt|*)
		luafmt -i 2 --stdin
		;;
esac
```

alerque:
@simoncozens What the heck is `pi` (besides not being )?

https://github.com/simoncozens/sile/blob/c6a0c07911e50448acea660fcaf9a7d1a8bdaa55/core/break.lua#L95

https://github.com/simoncozens/sile/blob/c6a0c07911e50448acea660fcaf9a7d1a8bdaa55/core/break.lua#L291

https://github.com/simoncozens/sile/blob/c6a0c07911e50448acea660fcaf9a7d1a8bdaa55/core/break.lua#L224

From an obsolete commit:

https://github.com/simoncozens/sile/blob/c20ca67541f0c670b432ca1e14108ad41881fff4/classes/book.lua#L40


It seems to me this variable has an identity crisis, but I can't figure out what it is to rename it.

simoncozens:
`pi` in break.lua is so called because that's what Knuth called it, and although he's really bad at naming things it's useful to use the same variable names in our implementation for checking against the original.

```
The parameter |pi| to |try_break| is the penalty associated
with a break at |cur_p|; we have |pi=eject_penalty| if the break is forced,
and |pi=inf_penalty| if the break is illegal.
```

alerque:
I have an idea, why don't we just name all our variables after common programming idioms? We want our code to be accessible and extensible right? I'll just start naming variables from most frequent to least frequent occurrences, `foo`, `bar`, `baz`, etc. Maybe we can call the typesetter object `Alice` and the pagebreaker `Bob`. Lets make this as easy for programmers and non programmers alike to get started with.

simoncozens:
I'm *just* talking about the break algorithm, which is a complete *nightmare* to debug. I have no problem renaming stuff elsewhere. And yes, under normal circumstances, giving things more sensible names *should* make them easier to debug, but if you're copying someone else's code and you want to remain compatible, it's a relief to know that if you can inspect variable `d` here and know that it's going to be the same as variable `d` there, and you don't have to mentally translate the variable names between the implementations.

alerque:
@simoncozens Sorry I should have been more explicit that my last message was just sarcasm. I get what you're saying. I already _did_ rename some stuff in break. I didn't touch `pi` yet, but some others that I was able to figure out a meaning for got names that make more sense to me. If there are any you'd like me to backtrack let me know, I can revert  individual renames fairly easily. However I'd like to comment on the ones that are kept for compatibility to some place in the LaTeX code or docs where they are described so that they aren't so mysterious to future code browsers.

alerque:
I think I got all the 1 letter variables refactored with sane replacements. I left the typical Lua `i` and `j` iterators, `k` and `v` key value pairs and a few others like `a` and `b` for lambda style functions with obvious purpose. I also left `x` and `y` when they were actually coordinates (not when they stood for various other possible values) and of course things like `c`, `m`, `y`, and `k` when they were CMYK color values and that sort of thing.

alerque:
@simoncozens If you merge this one be sure to use the **Create a merge commit** option, not squash. These commits are related, but if I did break something it will be very helpful to be able to `git bisect` across these in detail.

core/typesetter.lua
```diff
@@ -566,7 +570,8 @@ SILE.defaultTypesetter = std.object {
     local lines = {}
     local nodes = self.state.nodes
 
-    for i = 1,#bp do local point = bp[i]
+    for i = 1, #bp do
```
simoncozens: We could even rename bp to breakpoints.

alerque: Done.

How about `nl` → `nodes`? I think you meant "nodes from this line" or something like that, but it's pretty opaque.

simoncozens: Maybe next time. :-) I’m not at a computer until Monday so let’s just get this one in so we can merge a few others too.

PR #609, opened 2019-05-02
Fix rare crash when font is specified by filename

### LINKED ISSUE #604: [verbatim] Couldn't find face with no name
This is a very strange bug. Minimal file required for replication is included [here](https://github.com/simoncozens/sile/files/3129437/fontbug.sil.zip), but the main part is here:

```tex
\begin[class=book]{document}
\script[src=packages/verbatim]
\font[filename=libertine-font/LinLibertine_R.otf]%Probably any font file
\define[command=verbatim:font]{\font[family="Menlo"]}%Other fonts may probably need more or less verbatim lines
\chapter{x }%Required, no idea why
\begin{verbatim}
x
x
x
[... and many more x's]
\end{verbatim}
\end{document}
```

This leads to error `! Couldn't find face  at fontbug.sil`, or with --trace:
```
! Couldn't find face '' at fontbug.sil
  at verbatim on line: 6, column: 1
stack traceback:
	.../local/Cellar/sile/0.9.5.1/share/sile/core/utilities.lua:53: in function 'core/utilities.error'
	.../Cellar/sile/0.9.5.1/share/sile/core/harfbuzz-shaper.lua:64: in local 'callback'
	/usr/local/Cellar/sile/0.9.5.1/share/sile/core/font.lua:98: in field 'cache'
	.../Cellar/sile/0.9.5.1/share/sile/core/harfbuzz-shaper.lua:35: in method 'shapeToken'
	...ocal/Cellar/sile/0.9.5.1/share/sile/core/base-shaper.lua:53: in method 'measureChar'
	...cal/Cellar/sile/0.9.5.1/share/sile/core/measurements.lua:111: in function <...cal/Cellar/sile/0.9.5.1/share/sile/core/measurements.lua:110>
	(...tail calls...)
	/usr/local/Cellar/sile/0.9.5.1/share/sile/core/length.lua:9: in method 'absolute'
	...ocal/Cellar/sile/0.9.5.1/share/sile/core/base-shaper.lua:33: in local 'f'
	/usr/local/Cellar/sile/0.9.5.1/share/sile/core/settings.lua:48: in field 'temporarily'
	...ocal/Cellar/sile/0.9.5.1/share/sile/core/base-shaper.lua:29: in method 'measureSpace'
	...ocal/Cellar/sile/0.9.5.1/share/sile/core/base-shaper.lua:143: in method 'makeSpaceNode'
	...cal/Cellar/sile/0.9.5.1/share/sile/languages/unicode.lua:28: in method 'makeGlue'
	...cal/Cellar/sile/0.9.5.1/share/sile/languages/unicode.lua:109: in function <...cal/Cellar/sile/0.9.5.1/share/sile/languages/unicode.lua:84>
```

In generated pdf, last ~5 characters are missing (X's span two whole pages and then the document ends). I don't think that different fonts will fix it, but they may require different amount of verbatim characters. The fact that `filename` with `family` font specification is mixed seems significant, as adding 

```lua
  measureSpace = function(self, options)
    local ss = SILE.settings.get("document.spaceskip")
    if ss then
      SILE.settings.temporarily(function()
        SILE.settings.set("font.size", options.size)
        SILE.settings.set("font.family", options.family)
        -- FOLLOWING LINE
        if options.filename then SILE.settings.set("font.filename", options.filename) end
        ss = ss:absolute()
      end)
      return ss
    end
```

to `base-shaper.lua:32` seems to fix it. But I have no idea why, because the replication criteria seem extremely arbitrary, so it almost seems like there is a different underlying issue. But I may be mistaken, so instead of opening PR, I'd like someone else, with deeper knowledge of the internals, to look at it.

Darkyenus:
Fixes #604, though I am not sure if this is a correct/complete fix, as discussed in the issue.
In any case, it shouldn't break anything.

Even if this doesn't get merged, it could be useful for others facing the same issue.

### DIFF
--- a/core/base-shaper.lua
+++ b/core/base-shaper.lua
@@ -30,6 +30,7 @@ SILE.shapers.base = pl.class({
         SILE.settings.temporarily(function ()
           SILE.settings.set("font.size", options.size)
           SILE.settings.set("font.family", options.family)
+          SILE.settings.set("font.filename", options.filename)
           ss = ss:absolute()
         end)
         return ss


core/base-shaper.lua
```diff
@@ -29,6 +29,7 @@ SILE.shapers.base = std.object {
       SILE.settings.temporarily(function()
         SILE.settings.set("font.size", options.size)
         SILE.settings.set("font.family", options.family)
+        if options.filename then SILE.settings.set("font.filename", options.filename) end
```
alerque: Why does this like need to be wrapped in an `if` statement and the preceding ones do not?

Darkyenus: Short answer: Because it crashed without it.

Long answer: I believe it has to do with the fact, that when the font is specified by filename here, the font family is specified by empty string. But when it is specified by the family, the filename is `nil`, which crashes the `set` function.

As to why that is, I don't know. I don't have the time right now to understand this part of the SILE, so I stopped working on this when I got this fix/workaround.

alerque: I wonder if #607 will fix the issue with `set()`.

alerque: I have verified both that your fix is needed and that the changes in the earlier PR fix the settings so that a nil value does not cause a crash. This can be simply:

```suggestion
        SILE.settings.set("font.filename", options.filename)
```

alerque:
I'm sorry it took so long to get around to reviewing this properly. In the mean time the PR needs an update to resolve the merge conflict in addition to the tweak that I mentioned in the review comment. If you can address those I'd be happy to get this merged. If for some reason you don't have time I'm also happy to fix it up if you check off the box allowing maintainers to push to this branch, or I can cherry pick the fix in another PR. Just let me know. And thanks again...

Darkyenus:
Thanks for getting around to this. I have rebased and removed the `if`, as you suggested. Since I haven't touched sile for over half a year, I haven't re-tested the fix, but it should be ok.

alerque:
Thanks for cleaning it up for merge, even if you haven't been using it! By the way I make good use of the stack trace features you added about 6 days a week.

PR #610, opened 2019-05-02
HarfBuzz-shaper error messages

Darkyenus:
*...and two missing `local`s*

Previously part of #603.


simoncozens:
Thanks, this looks good!

PR #611, opened 2019-05-07
feat(debug): Add tracestack system for better error outputs

Darkyenus:
Based on #603.

### DIFF
--- a/core/baseclass.lua
+++ b/core/baseclass.lua
@@ -24,10 +24,19 @@ SILE.setCommandDefaults = function (command, defaults)
 end
 
 SILE.doTexlike = function (doc)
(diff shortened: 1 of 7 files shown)


alerque:
Can we please not split this discussion _again_. It's already spread out enough. It would be a lot easier to follow when you refactor things yourself if you  kept going on top of your original branch for now so that we can review what's changed instead of the lump-sum result. For example this has my changes to the error/warn function, but a different print mechanism for the traces (that takes some but not all of my experiments). It's not clear what else changes except by diffing branches. I think it would be helpful if you made the changes you thought to make on your other branch / PR so we can use the review review what changes since our last discussion rather than starting all over here.

PR #613, opened 2019-05-10
feat(book): Add option to disable ToC entry for book section

Darkyenus:
It may sometimes be useful to exclude some sections from table of contents, this adds the ability to do it through: `\subsection[toc=false]{Not in ToC}`.

Another two proposed & related changes:
1. In my fork, I don't increment section counters when `numbering` is `false`. This works quite well in combination with `toc=false`, so I propose adding it. The current behavior could be preserved with `numbering=hidden` (or some variant which preserves backwards compatibility).
2. [Current check for `numbering` in book.lua](https://github.com/simoncozens/sile/blob/master/classes/book.lua#L102), "`if options.numbering == nil or options.numbering == "yes"`", looks like it could benefit from `SU.boolean()`. However, that function does not check for `yes/no`, only for `true/false`. I propose extending `SU.boolean()` with this wider matching (and possibly even more options, `y/n`, `1/0`, ...) and using it there. This change should not lead to any any broken code (at least inside SILE), because all usages of `SU.boolean` parse user input and don't depend on its current strict behavior.

### DIFF
--- a/classes/book.lua
+++ b/classes/book.lua
@@ -94,10 +94,11 @@ SILE.registerCommand("right-running-head", function (options, content)
 end, "Text to appear on the top of the right page")
 
 SILE.registerCommand("book:sectioning", function (options, content)
-  local content = SU.subContent(content)
   local level = SU.required(options, "level", "book:sectioning")
-  SILE.call("increment-multilevel-counter", {id = "sectioning", level = options.level})
-  SILE.call("tocentry", {level = options.level}, content)
+  SILE.call("increment-multilevel-counter", {id = "sectioning", level = level})
+  if SU.boolean(options.toc, true) then
+    SILE.call("tocentry", {level = level}, SU.subContent(content))
+  end
   local lang = SILE.settings.get("document.language")
   if options.numbering == nil or options.numbering == "yes" then
     if options.prenumber then
@@ -136,6 +137,7 @@ SILE.registerCommand("chapter", function (options, content)
   SILE.call("book:chapterfont", {}, function ()
     SILE.call("book:sectioning", {
       numbering = options.numbering,
+      toc = options.toc,
       level = 1,
       prenumber = "book:chapter:pre",
       postnumber = "book:chapter:post"
@@ -160,6 +162,7 @@ SILE.registerCommand("section", function (options, content)
   SILE.call("book:sectionfont", {}, function ()
     SILE.call("book:sectioning", {
       numbering = options.numbering,
+      toc = options.toc,
       level = 2,
       postnumber = "book:section:post"
     }, content)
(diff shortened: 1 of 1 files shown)


alerque:
1. I don't think a one to one correlation between counters and display of numbering is a good idea. I actually do make use of them being separate. Somehow the display option should be uncoupled entirely from whether the counter increments. I'd prefer two separate options over a hard coded link or assumed logic.

2. How about getting rid of "yes" altogether? Why have a bunch of user facing synonyms for the same thing? The `SU.boolean` has to handle the boolean `false` vs. the string `"false"` because of the way our parser works, but I think I'd rather purge any other value possibilities rather than adding more.

Darkyenus:
Thanks for the merge.

#### 1
I don't mind it being separate. I believe you, but I'm curious: in which situation would you want to show the same sectioning counter value (i.e. chapter/section number) on multiple sections  without incrementing it? I don't have much experience with booksetting, but I don't recall seeing such scheme anywhere. Is or should that be the default?

#### 2
I would be against it in a programmer's API, but since the tex-like commands are presumably used by not-necessarily-programmer users, I don't see it as a bad thing to provide some alternatives. We have `true`/`false` ingrained in memory, but I can easily imagine someone struggling to recall whether it was "true", "yes" or "on", not to mention that someone may prefer shorter variants. And since it is always used only on boolean parameters, it wouldn't introduce any ambiguity. 

(Speaking of ambiguity: Passing invalid value currently does not generate any error, but silently fallbacks to default. This could lead to some nasty oversights.)

Typesetting can be frustrating even when it works without bugs, so I think that any UX improvement can help. It's the little things that make software a joy to use.

alerque:
1. I was actually imagining the opposite senario: wanting the counter to keep incrementing even on headings that didn't necessarily display the counter value.

2. On the contrary, I think providing a bunch of alternatives would just confuse people. They wouldn't even get it in their heads with parameters were booleans and which were string values. They would see other people's document code and it would be done differently. I use "true", you use "yes", Simon uses "y" (he's terse like that, then we all share our examples around and anybody reading them has to understand they all mean the same thing. It also creates problems for other programs. I am working on a Pandoc reader/writer for SILE and if I have to support the logic of a bunch of different syntax options that's more work for me and others in that position. The same goes for syntax highlighters, IDE support, etc. Imagine a system like Overleaf for SILE. Think of the fact that round-tripping a document though some IDE function might replace arguments with a different syntax.

    Keep it simple. Limit the options and make it consistent so it round trips through format conversion and multiple users without changing.

Darkyenus:
1. Then I probably just misunderstood. What I would like to do is `numbering=`:
    - `false` -> no number shown, number NOT incremented
    - `hidden` -> no number shown, number incremented (current behavior of `false`)
    - `true` -> number shown and incremented
    - Last option, number shown and not incremented is probably not very useful, but could also be added.

2. I guess that it makes sense when you look at it from this point of view. What I had in mind was that the parameters are just some strings, which could be interpreted somehow and that we could be more lenient in interpreting them, since they are a part of user interface. You consider them a part of grammar, which is also fine.

PR #614, opened 2019-05-10
Add list package for creating lists

Darkyenus:
Partially based on [casile/lists.lua](https://github.com/alerque/casile/blob/master/lists.lua).

### Supports:
- numbered/ordered lists with arbitrary numbering (based on counters)
- non-numbered lists with different bullets for each nesting level
- nesting
- text does not flow under the bullet/number (what happens currently in the SILE book)

### Not done:
- Not tested for RTL
- Gaps for bullets/numbers should probably adapt to their real width, now they are sometimes too large. This is a concern especially if it should allow something like [this](https://en.wikibooks.org/wiki/LaTeX/List_Structures#Customizing_lists) (text "bullets" with custom alignment). I don't know how to do that.
- Think more about the system of bullet styling. Current system does not allow for e.g. numbers in parentheses, alternating bullets by row, etc. One idea would be to define global object which holds all styles by their name. Styles would then be represented by a function, which would return a texlike string for given nest level and item number. (The function could be replaced with tables and/or strings for less complex styles.) Example:
```lua
ListStyle.bullet = { {"•"}, {"◦"} }

ListStyle.arabic = function (nestLevel) 
    return function (itemNumber) return itemNumber .. "." end
end
```

### Current output:
```tex
\begin[class=plain]{document}
\script[src=packages/list]
\font[family=Arial, size=15pt]
Quis ipsa omnis aut nihil sed sit. Quaerat minus aut est et. Laudantium aut corporis in rem animi repudiandae. Et facere aut beatae.
\begin{list}
\list:item{First}
\list:item{Second}
\begin[numbered=true]{list}
\list:item{Uno}
\list:item{Dos}
\begin{list}
\list:item{Three times nested}
\end{list}
\end{list}     %
\begin{list}     %
\list:item{Two times nested}
\end{list}
\end{list}
Facere dolores eligendi a reiciendis alias quaerat atque esse. Ut quia assumenda nesciunt alias illum. Perspiciatis et dolorum tempore doloremque.
\end{document}
```

list.pdf

### DIFF
--- a/packages/list.lua
+++ b/packages/list.lua
@@ -1,0 +1,107 @@
+SILE.scratch.listStack = {}
+
+SILE.settings.declare({
+  name = "list.bullet.numbered",
+  type = "string",
+  default = ".",
+  help = "Bullet used after a list item number"
+})
+
+-- TODO Not ready to handle RTL
+SILE.settings.declare({
+  name = "list.bullet.inset",
+  type = "Length",
+  default = SILE.length.parse("1.3em"),
+  help = "The amount of space by which each level of the list is shifted"
+})
+
+local defaultBullets = {"•", "◦"}
+for i = 1, #defaultBullets do
+  SILE.settings.declare({
+    name = "list.bullet."..i,
+    type = "string",
+    default = defaultBullets[i],
+    help = "Bullet used for unordered list items at level "..i
+  })
+end
+
+SILE.registerCommand("list", function (options, content)
+  local level = #SILE.scratch.listStack + 1
+  local topLevel = level == 1
+  if topLevel then SILE.call("smallskip") else SILE.typesetter:leaveHmode() end
+
+  local numbered = SU.boolean(options.numbered, SU.boolean(options.ordered, options.numbering ~= nil))
+  local bullet = options.bullet
+  if not bullet then
+    if numbered then
+      bullet = SILE.settings.get("list.bullet.numbered")
+    else
+      local setting = "list.bullet."..level
+      -- If the level is too large, cycle bullet types
+      if not SILE.settings.declarations[setting] then
+        local mod = math.mod or math.fmod
+        setting = "list.bullet." .. (1 + mod(level - 1, #defaultBullets))
+      end
+
+      bullet = SILE.settings.get(setting)
+    end
+  end
+  local inset = options.inset and SILE.length.parse(options.inset) or SILE.settings.get("list.bullet.inset")
+
+  SILE.scratch.listStack[level] = {
+    nextNumber = 1,
+    numbered = numbered,
+    numbering = options.numbering or "arabic",
+    bullet = bullet,
+    skip = inset
+  }
+  local baseLSkip = SILE.settings.get("document.lskip")
+  baseLSkip = baseLSkip and baseLSkip.width or SILE.length.new()
+  SILE.settings.temporarily(function ()
+    SILE.settings.set("document.lskip", SILE.nodefactory.newGlue({width = baseLSkip + inset}))
+    SILE.settings.set("document.rskip", options.rskip and SILE.nodefactory.newGlue(options.rskip) or SILE.nodefactory.zeroGlue)
+    SILE.settings.set("document.parindent", SILE.nodefactory.zeroGlue)
+    SILE.process(content)
+  end)
+  SILE.scratch.listStack[#SILE.scratch.listStack] = nil
+
+  if topLevel then SILE.call("smallskip") else SILE.typesetter:leaveHmode() end
+  SILE.call("noindent") -- list does not create a new paragraph
+end)
+
+SILE.registerCommand("list:item", function (options, content)
+  local settings = SILE.scratch.listStack[#SILE.scratch.listStack]
+  local number = options.number or settings.nextNumber
+  settings.nextNumber = number + 1
+
+  local markerwidth = settings.skip
+  SILE.call("kern", { width = tostring(markerwidth:negate()) })
+  local box = SILE.call("hbox", {}, function ()
+    if settings.numbered then
+      SILE.typesetter:setpar(SILE.formatCounter({ value = number, display = settings.numbering or "arabic" }))
+    end
+    SILE.typesetter:setpar(settings.bullet)
+  end)
+  box.width = markerwidth
+
+  SILE.process(content)
+  if #SILE.scratch.listStack == 1 then SILE.call("smallskip") else SILE.call("par") end
+end)
+
+return { documentation = [[\begin{document}
+Lists of items can be created with the \code{list} package. It consists of two main commands:
+\command{\\list} and \command{\\list:item}.
+
+\command{\\list} has following optional options:
+\code{numbered} (or its alias \code{ordered}), a boolean determining whether numbers will be printed
+before the bullet or not. Lists are unordered by default.
+\code{bullet} a string with explicit bullet character to use.
+\code{numbering} a string denoting the numbering system to use for ordered lists.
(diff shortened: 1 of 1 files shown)


simoncozens:
This is much needed, thank you. “Two times nested” should start with a 3, not an open bullet, shouldn’t it?

Darkyenus:
No, that is intentional, but a bad example. Imagine another top level bullet "Third" right before it, then it will make more sense. What you suggest is of course also possible, just without the two lines that I have marked with comment in above code.

packages/list.lua
```diff
@@ -0,0 +1,107 @@
...
+    help = "Bullet used for unordered list items at level "..i
+  })
+end
+
+SILE.registerCommand("list", function (options, content)
+  local level = #SILE.scratch.listStack + 1
+  local topLevel = level == 1
+  if topLevel then SILE.call("smallskip") else SILE.typesetter:leaveHmode() end
+
+  local numbered = SU.boolean(options.numbered, SU.boolean(options.ordered, options.numbering ~= nil))
```
alerque: The mashup of variable names "options.numbered", "options.ordered", and "options.numbering" here combined with nested boolean tests is a bit confusing.

Darkyenus: I didn't know which one is better. I would pick based on precedent (html - unordered/ordered, tex - ?), but allowing both as an alias could be also fine, users wouldn't have to remember it.

In the grand scheme of things, it could be useful to implement a system which warns for unused options and possibly knows and suggests valid options. But right now, if I specify for example `\list[unordered=false]`, it won't work and I will get no warning. So if I don't notice incorrect output, I am out of luck. But it wouldn't be hard to do warning like: 
> ! Option `unordered` is not used. Did you mean: `ordered`, `numbering` or `bullet`?

That is obviously completely out of scope for this PR, but I could contribute it sometime in the future, if you want.

packages/list.lua
```diff
@@ -0,0 +1,107 @@
...
+    SILE.process(content)
+  end)
+  SILE.scratch.listStack[#SILE.scratch.listStack] = nil
+
+  if topLevel then SILE.call("smallskip") else SILE.typesetter:leaveHmode() end
+  SILE.call("noindent") -- list does not create a new paragraph
+end)
+
+SILE.registerCommand("list:item", function (options, content)
+  local settings = SILE.scratch.listStack[#SILE.scratch.listStack]
```
Omikhleia, line 73: Using a global. What happens if one has, says, for the mere example a footnote in list, and a list in the footnote? A mess.

packages/list.lua
Omikhleia: Obscure, not understandable.

packages/list.lua
```diff
@@ -0,0 +1,107 @@
...
+  type = "string",
+  default = ".",
+  help = "Bullet used after a list item number"
+})
+
+-- TODO Not ready to handle RTL
+SILE.settings.declare({
+  name = "list.bullet.inset",
+  type = "Length",
+  default = SILE.length.parse("1.3em"),
```
Omikhleia, line 14: Same for bullets and numbering. Ever tries it with roman numerals?

packages/list.lua
```diff
@@ -0,0 +1,107 @@
...
+  baseLSkip = baseLSkip and baseLSkip.width or SILE.length.new()
+  SILE.settings.temporarily(function ()
+    SILE.settings.set("document.lskip", SILE.nodefactory.newGlue({width = baseLSkip + inset}))
+    SILE.settings.set("document.rskip", options.rskip and SILE.nodefactory.newGlue(options.rskip) or SILE.nodefactory.zeroGlue)
+    SILE.settings.set("document.parindent", SILE.nodefactory.zeroGlue)
+    SILE.process(content)
+  end)
+  SILE.scratch.listStack[#SILE.scratch.listStack] = nil
+
+  if topLevel then SILE.call("smallskip") else SILE.typesetter:leaveHmode() end
```
Omikhleia, line 68: Oh, a hard-coded smallskip. Explain. Now set a parskip and explain the expected behavior.

packages/list.lua
```diff
@@ -0,0 +1,107 @@
...
+  SILE.call("noindent") -- list does not create a new paragraph
```
Omikhleia, line 69: Ah? They do not create a new paragraph after them? Explain why.

packages/list.lua
```diff
@@ -0,0 +1,107 @@
...
+before the bullet or not. Lists are unordered by default.
+\code{bullet} a string with explicit bullet character to use.
+\code{numbering} a string denoting the numbering system to use for ordered lists.
+See the documentation of counters for possible values.
+\code{lskip} width reserved for the bullets.
+
+\command{\\list:item} has option \code{number}, to override the number of the item in case of numbered lists.
+By default, the numbered list starts at 1 and each item has a number that is one bigger than of the previous one.
+
+Lists can be arbitrarily nested.
```
Omikhleia, line 106: Proof? When the documentation doesn't show it, and there are not a single test case, am I to believe it?

packages/list.lua
```diff
@@ -0,0 +1,107 @@
...
+Lists of items can be created with the \code{list} package. It consists of two main commands:
+\command{\\list} and \command{\\list:item}.
+
+\command{\\list} has following optional options:
+\code{numbered} (or its alias \code{ordered}), a boolean determining whether numbers will be printed
+before the bullet or not. Lists are unordered by default.
+\code{bullet} a string with explicit bullet character to use.
+\code{numbering} a string denoting the numbering system to use for ordered lists.
+See the documentation of counters for possible values.
+\code{lskip} width reserved for the bullets.
```
Omikhleia, line 101: The list seems to use `options.rskip` (but should it?). Maybe I missed it above, but I didn't see any any `options.lskip`...

packages/list.lua
```diff
@@ -0,0 +1,107 @@
+SILE.scratch.listStack = {}
+
+SILE.settings.declare({
+  name = "list.bullet.numbered",
+  type = "string",
+  default = ".",
```
Omikhleia, line 6: So it's a dot. Weird name then, `list.bullet.numbered` (you are numbering bullets as dots? Heh.)

packages/list.lua
```diff
@@ -0,0 +1,107 @@
...
+    nextNumber = 1,
+    numbered = numbered,
+    numbering = options.numbering or "arabic",
+    bullet = bullet,
+    skip = inset
+  }
+  local baseLSkip = SILE.settings.get("document.lskip")
+  baseLSkip = baseLSkip and baseLSkip.width or SILE.length.new()
+  SILE.settings.temporarily(function ()
+    SILE.settings.set("document.lskip", SILE.nodefactory.newGlue({width = baseLSkip + inset}))
```
Omikhleia, line 61: How does it look if the paragraph before the list has a parindent >  inset ? ;)

packages/list.lua
Omikhleia: And what is expected if one mistakenly calls this list:item outside a list.

Omikhleia:
Some good points, but I think the implementation is not robust enough. I added a few comments, but I don't ask for them to be fixed here. I wanted to see how my own implementation (which I will propose at one point) compare to this, and if there were ideas I had possibly missed. Overall, I think we can do better.

Omikhleia:
I think this could now be closed without merge, since we got #1365 instead - With obviously many thanks to @alerque for the early casile attempt mentioned above and to @Darkyenus for this very PR - You were, how do we say, pioneers or precursors here! Looking at this code obviously helped a lot in reaching #1365 targets.

PR #615, opened 2019-05-11
counters: Improve external API and cleanup internal code

Darkyenus:
Consists of two commits, I suggest to review them separately.

The main motivation for this was to expose a way to obtain a multilevel counter as string. While doing that, I have noticed an opportunity to cleanup some code duplication.

NOTE: It is pretty weird that `\show-(multilevel-)counter` commands can change the `display` of the counter permanently. Is that intentional or just a hack? It would be pretty easy to fix now.

Also, I am not sure about the exports and the `format` field on counters, as it duplicates the API. But it is IMO much nicer and easier to use. What do you think?

### DIFF
--- a/packages/counters.lua
+++ b/packages/counters.lua
@@ -66,7 +66,7 @@ local function getCounter(id)
   local counter = SILE.scratch.counters[id]
   if not counter then
     counter = {}
-    SILE.scratch.counters[id] = { value= 0, display= "arabic" }
+    SILE.scratch.counters[id] = { value= 0, display= "arabic", format = SILE.formatCounter }
   end
   return counter
 end
@@ -94,10 +94,20 @@ SILE.registerCommand("show-counter", function (options, _)
   SILE.typesetter:setpar(SILE.formatCounter(counter))
 end, "Outputs the value of counter <id>, optionally displaying it with the <display> format.")
 
+SILE.formatMultilevelCounter = function(counter, options)
+  local maxlevel = options and options.level or #counter.value
+  local minlevel = options and options.minlevel or 1
+  local out = {}
+  for x = minlevel, maxlevel do
+    out[x - minlevel + 1] = SILE.formatCounter({ display = counter.display[x], value = counter.value[x] })
+  end
+  return table.concat( out, "." )
+end
+
 local function getMultilevelCounter(id)
   local counter = SILE.scratch.counters[id]
   if not counter then
-    counter = { value= {0}, display= {"arabic"} }
+    counter = { value= {0}, display= {"arabic"}, format = SILE.formatMultilevelCounter }
     SILE.scratch.counters[id] = counter
   end
   return counter
@@ -128,18 +138,17 @@ end)
 
 SILE.registerCommand("show-multilevel-counter", function (options, _)
   local counter = getMultilevelCounter(options.id)
-  local currentLevel = #counter.value
-  local maxlevel = options.level or currentLevel
-  local minlevel = options.minlevel or 1
-  if options.display then counter.display[currentLevel] = options.display end
-  local out = {}
-  for x = minlevel, maxlevel do
-    out[x - minlevel + 1] = SILE.formatCounter({ display = counter.display[x], value = counter.value[x] })
-  end
-  SILE.typesetter:typeset(table.concat( out, "." ))
+  if options.display then counter.display[#counter.value] = options.display end
+
+  SILE.typesetter:typeset(SILE.formatMultilevelCounter(counter, options))
 end, "Outputs the value of the multilevel counter <id>, optionally displaying it with the <display> format.")
 
-return { documentation = [[\begin{document}
+return {
+  exports = {
+    getCounter = getCounter,
+    getMultilevelCounter = getMultilevelCounter
+  },
+  documentation = [[\begin{document}
 
 Various parts of SILE such as the \code{footnotes} package and the
 sectioning commands keep a counter of things going on: the current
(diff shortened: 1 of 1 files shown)


packages/counters.lua
```diff
@@ -52,91 +52,103 @@ SILE.formatCounter = function(options)
...
   end
-  if options.display then SILE.scratch.counters[c].display = options.display end
+  if options.display then counter.display = options.display end
 end, "Increments the counter named by the <id> option")
 
-SILE.registerCommand("set-counter", function (options, content)
-  local c = options.id; _init(c)
-  if options.value then SILE.scratch.counters[c].value = tonumber(options.value) end
-  if options.display then SILE.scratch.counters[c].display = options.display end
+SILE.registerCommand("set-counter", function (options, _)
```
alerque:
```suggestion
SILE.registerCommand("set-counter", function (options, content)
```

Please don't discard `content` here. Even though it doesn't get used by this command itself, we have ⓐ a stack trace that can inspect it and ⓑ a hook system that can do things with it. It is useful to always have all `SILE.registerCommand` registered functions have these two variables intact.

packages/counters.lua
```diff
@@ -52,91 +52,103 @@ SILE.formatCounter = function(options)
...
+  if not counter then
+    counter = {}
+    SILE.scratch.counters[id] = { value= 0, display= "arabic", format = SILE.formatCounter }
   end
+  return counter
 end
 
-SILE.registerCommand("increment-counter", function (options,content)
-  local c = options.id; _init(c)
+SILE.registerCommand("increment-counter", function (options, _)
```
alerque:
```suggestion
SILE.registerCommand("increment-counter", function (options, content)
```

packages/counters.lua
```diff
@@ -52,91 +52,103 @@ SILE.formatCounter = function(options)
...
+  if options.value then counter.value = tonumber(options.value) end
+  if options.display then counter.display = options.display end
 end, "Sets the counter named by the <id> option to <value>; sets its display type (roman/Roman/arabic) to type <display>.")
 
 
-SILE.registerCommand("show-counter", function (options, content)
-  local c = options.id; _init(c)
-  if options.display then SILE.scratch.counters[c].display = options.display end
-  SILE.typesetter:setpar(SILE.formatCounter(SILE.scratch.counters[c]))
+SILE.registerCommand("show-counter", function (options, _)
```
alerque:
```suggestion
SILE.registerCommand("show-counter", function (options, content)
```

packages/counters.lua
```diff
@@ -52,91 +52,103 @@ SILE.formatCounter = function(options)
...
+  local counter = SILE.scratch.counters[id]
+  if not counter then
+    counter = { value= {0}, display= {"arabic"}, format = SILE.formatMultilevelCounter }
+    SILE.scratch.counters[id] = counter
+  end
+  return counter
+end
 
-  local currentLevel = #this.value
+SILE.registerCommand("increment-multilevel-counter", function (options, _)
```
alerque:
```suggestion
SILE.registerCommand("increment-multilevel-counter", function (options, content)
```

packages/counters.lua
```diff
@@ -52,91 +52,103 @@ SILE.formatCounter = function(options)
...
-  local currentLevel = #this.value
-  local maxlevel = options.level or currentLevel
-  local minlevel = options.minlevel or 1
-  if options.display then this.display[currentLevel] = options.display end
-  local out = {}
-  for x = minlevel, maxlevel do
-    out[x - minlevel + 1] = SILE.formatCounter({ display = this.display[x], value = this.value[x] })
-  end
-  SILE.typesetter:typeset(table.concat( out, "." ))
+SILE.registerCommand("show-multilevel-counter", function (options, _)
```
alerque:
```suggestion
SILE.registerCommand("show-multilevel-counter", function (options, content)
```

packages/counters.lua
```diff
@@ -52,91 +52,103 @@ SILE.formatCounter = function(options)
...
-local _init = function (c)
-  if not(SILE.scratch.counters[c]) then
-    SILE.scratch.counters[c] = { value= 0, display= "arabic" }
+local function getCounter(id)
+  local counter = SILE.scratch.counters[id]
+  if not counter then
+    counter = {}
+    SILE.scratch.counters[id] = { value= 0, display= "arabic", format = SILE.formatCounter }
   end
+  return counter
```
alerque: Why return an empty table in the case of a new counter instead of the actual counter that just got created? The scoping of `counter` in this bit of code is either confusing or the logic is wrong.

packages/counters.lua
```diff
@@ -52,91 +52,103 @@ SILE.formatCounter = function(options)
...
+  for x = minlevel, maxlevel do
+    out[x - minlevel + 1] = SILE.formatCounter({ display = counter.display[x], value = counter.value[x] })
   end
+  return table.concat( out, "." )
 end
 
-SILE.registerCommand("increment-multilevel-counter", function (options, content)
-  local c = options.id; _initml(c)
-  local this = SILE.scratch.counters[c]
+local function getMultilevelCounter(id)
```
alerque: This isn't the fault of this commit, but it's always bugged me how much duplication there was between counters and multilevel counters. I think we need to redo the backend for all counters to be the same (multilevel) and default everything to acting on the top (or only) level, and regular counters could just format that level only.

Darkyenus: Sounds good. I propose to do that in a different PR, to keep it lean. I can do it after this gets merged.

alerque: Okay.

alerque:
The implementation of multi level counters also needs to  be discussed in light of #601.

PR #617, opened 2019-05-21
Get commitlint-bot working on PRS

alerque:
This is an attempt to get the commitlint-bot to _flunk_ a PR. See #549.

**DO NOT MERGE**.


alerque:
It looks like the thing to do is going to be setup a custom commitlint bot. This looks pretty easy but the bot was configured for now.sh API v1 and they are forcing new free tier apps to use v2. Until [that is figured out](https://github.com/z0al/commitlint-bot/issues/15) it doesn't seem possible to have a custom configuration with the existing bot.

I tried the Travis approach ... and it works but it takes time and seems like a clumsy solution to wait for our test suite just to find out a commit message is bad.

.travis.yml
```diff
@@ -86,6 +86,14 @@ script:
...
  - fold_unless_fail 'Compile documentation and examples' 'compile_docs' make docs examples
  - fold_unless_fail 'Versions' 'versions' ./sile --debug=fonts,versions tests/bug-117.sil
+stages:
+  - test
+  - lint
+jobs:
+  include:
+    - stage: lint
+      name: "Lint commint messages"
+      script: fold_unless_fail 'Commitlint' 'commitlint' npx commitlint-travis
```
rjmunro: I think you need:

```suggestion
      script: fold_unless_fail 'Commitlint' 'commitlint' @commitlint/travis-cli
```

Personally, I would specify the version, so that builds don't start failing if the linter improves:

```suggestion
      script: fold_unless_fail 'Commitlint' 'commitlint' @commitlint/travis-cli@8.0.0
```

alerque: I appreciate the suggestion @rjmunro and I was honestly puzzled by how to execute the check, but your suggestion didn't work either.

rjmunro: Err, whoops. I didn't mean to drop the `npx` part.

```suggestion
      script: fold_unless_fail 'Commitlint' 'commitlint' npx @commitlint/travis-cli@8.0.0
```

alerque:
Thanks for the help on this @rjmunro — but I just checked out the Github Action for Commitlint and it is about 2 orders of magnitude easier to setup and use. I think we'll go with that for now and keep it out of the way of our already troublesome Travis setup.

Issue #474, opened 2017-06-22
Drop cap float causes layout problems under a certain size


kstenerud:
Given the following content:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<sile class="book" papersize="6in x 9in">
  <script src="packages/frametricks"/>
  <font size="12pt">
    <chapter numbering="no">The Title</chapter>
    <float bottomboundary="3pt">
      <font size="50pt">L</font>
    </float>
    orem ipsum dolor sit amet, nonumes interpretaris sit id. At inermis deterruisset vix. Pro ex minim expetenda dignissim, ex viris altera contentiones pri. Ullum laudem an mel, in nemore essent mea. Amet reque pro eu, rebum doctus abhorreant qui ex, legendos consequat eu mei.

    Est ea case graeci. Vero scripserit his no, an meliore noluisse vix. Mea viris fastidii liberavisse ei. Putent veritus eloquentiam ex vis. Semper pericula has in, aperiam laoreet eu has, fastidii placerat eam ea.
  </font>
</sile>
```

This produces (at 50pt):
![screen shot 2017-06-22 at 7 55 35 am](https://user-images.githubusercontent.com/245857/27440719-762f62f6-5720-11e7-8624-39f4543f5349.png)

At 40pt, it produces:
![screen shot 2017-06-22 at 7 55 09 am](https://user-images.githubusercontent.com/245857/27440731-80527200-5720-11e7-9cac-d6f83e73d3aa.png)

At 30pt, things break:

```
This is SILE 0.9.4
<test.xml>
[1] [2] 
! Overfull frame: 10.27265625 extra space required but 0 shrink available
[3] 
```
![screen shot 2017-06-22 at 7 54 38 am](https://user-images.githubusercontent.com/245857/27440782-9905deb8-5720-11e7-893a-ed648ca105a6.png)

alerque:
Actually the 50pt and 40pt versions are broken here too. If you look carefully the leading for the first two lines is different than for the rest of the page. This has driven me bonkers too because I've had to tailor the size and boundaries to get a visual match for every size and layout I've used by hand. 

In fact I think this issue is largely a duplicate of #394.

alerque:
The test for this has a hard coded `1ex` bottombountry, and `frametricks.lua` has a hard coded "1ex" length being chopped off the baselineskip whether that's a relevant amount or not. This should use the parameter instead.

[...]

Issue #394, opened 2016-11-03
Floats need improvement, esp in drop cap use case


alerque:
### Intro

I've had a couple books that use drop-caps at the start of chapters or sections. So far I've hacked around it a couple ways to get the projects out the door, but the result has been brittle and it seems like some functionality is missing.

### Finding the right place to even apply drop-cap code

Right out of the gate, figuring out *when* to apply such a style and getting the right content or letter wrapped is hard. This isn't something I want encoded in my documents which are in Markdown and get converted to other formats as well. I need to do it automatically.

One way has been to overload the `chapter()` or similar command to trigger the style of the *next* content to come along. In the case of a drop cap letter I'm doing something like this:

```lua
local inputfilter = SILE.require("packages/inputfilter").exports
local originalTypesetter = SILE.typesetter.typeset
dropcapNextLetter = function ()
  SILE.typesetter.typeset = function (self, text)
    local first, rest = text:match("([^%w]*%w)(.*)")
    if load and first and rest then
      SILE.typesetter.typeset = originalTypesetter
      SILE.call("dropcap", {}, {first})
      SILE.typesetter.typeset(self, rest)
    else
      originalTypesetter(self, text)
    end
  end
end

SILE.registerCommand("chapter", function (options, content)
   ...
    dropcapNextLetter()
end)
```

This hijacks the typesetter the text time it is asked to output something, strips off the first character, sends it to the dropcap function, then goes on with the rest.

This _works_ but it feels dirty and shouldn't be something SILE users should be required to come up with. If it is the right way of doing this let me know and I'll work it up and submit it as a package. If it's the wrong way to go about this I'd love to hear it.

The other way I've tackled this is when something decorative is being placed in leu of a drop-cap. In this case I've hijacked the typesetter functions a little differently:

```Lua
SILE.scratch.lastWasRef = false

SILE.registerCommand("call-for-chaptercap", function (options, content)
  SILE.scratch.lastWasRef = true
  SILE.call("noindent")
end)

local push = SILE.typesetter.pushHorizontal
SILE.typesetter.pushHorizontal = function (self, node)
  if SILE.scratch.lastWasRef then
    SILE.call("output-chaptercap")
  end
  return push(self, node)
end

SILE.registerCommand("output-chaptercap", function (options, content)
  SILE.scratch.lastWasRef = false
  ...
end)
```

[...]

PR #624, opened 2019-05-31
Cleanup showoff document

### LINKED ISSUE #623: Odd spacing in show-off document
I love the ideas behind SILE so I was trying to evaluate its core typesetting features. When I looked at the [show-off document linked from the readme](https://rawgit.com/simoncozens/sile/b66b979a6dca0c60bd4aa2cbad3da36ae2073672/examples/showoff.pdf) I noticed it has a title that looks squished. It almost looks like there aren't any spaces between the words.

Screen Shot 2019-05-30 at 9 34 02 AM

I tried running Sile on my local machine (SILE 0.9.5.1 on macOS, installed from Homebrew) and I got better results for the title spacing, but the space between "Show-off" and "Document" looks a little too big compared to the other spaces. And there are odd spaces in the paragraph, like after "InDesign" and before "Let's". Also the title doesn't appear to be properly centered relative to either the page or the text boxes below (it is too far to the right), but maybe this is an intended effect of how the frames are set up, I can't tell.

Screen Shot 2019-05-30 at 9 46 13 AM

Are these spacing problems a bug? Can this be improved somehow?

### LINKED ISSUE #625: Center environment should eschew paragraph indents
Using `\center{}` and having the first line not actually centered because there is a paragraph indent in always a surprise to me and feels wrong. The _center_ environment should probably set `parindent` to zero by default.

See #623 and #624.

alerque:
Closes #623.
Closes #625.

Currently the showoff document shows off a lot of issues as noted in #623. Ideally it should be cleaned up to have really good output as well as keeping the input as free from dirty tricks as possible.

- [x] Center heading
- [x] Use distributable fonts so we can reproduce it reliably in CI and on compile/install
- [ ] Fix line spacing following drop cap (see #394 and #474)
- [x] Fix massive stretch spaces in lines
- [x] Fix missing title in the TOC frame and set it apart more visually


examples/showoff.sil
```diff
@@ -4,6 +4,7 @@
 \script[src=packages/image]
 \script[src=packages/bidi]
 \set[parameter=linebreak.tolerance,value=10000]
+\noindent
```
alerque: @simoncozens I expected this to fix the title not being centered ... and indeed it did move it closer to center (see #625). However to my surprise it did not _entirely_ fix the issue. The title is s till not quite centered and I'm not sure where the extra leading space is coming from now. Any ideas?

alerque: This was because of the leading whitespace on lines like this:
```
\begin{pagetemplate}
    \frame{}
    \frame{}
\end{pagetemplate}
```
The spaces were putting stuff in the node queue before starting to typeset the centered content.

I've mitigated this two ways:

1. Discard all content inside pagetemplates. The `\frame` commands still work, not much else.
2. Throw warnings if `\center` or `\nodindent` are used after things are in the horizontal node queue.

alerque:
Not perfect but not quite so egregiously bad. The drop-cap is still killing us. It *happens* to be okay here but it is supper brittle and mostly kills the line breaking in the first paragraph.

image

By the way when working on this remember to render it with `make examples/showoff.pdf` rather than manually with `./sile examples/showoff.sil` to get all the right fonts!

simoncozens:
Looks good apart from the "foo" at the start of the second page. :-)

alerque:
@simoncozens Thanks. The "foo" leaked out in that quick screen capture because I‌ was testing the warning system on `\noindent` locally while updating the issue, but it is not in the source.

PR #626, opened 2019-05-31
Add formats and Turkish language support to counters

alerque:
I'm working my way through things I wrote for CaSILE's `cabook` class and ran across the fact that I completely scrapped and replaced `SILE.formatCounter()`. These commits port most of the changes I made that still looked to me like improvements. My original only handled up to 4 digit numbers correctly, this has been extended up to 15 digits (Nine Hundred Ninety Nine Trillion...). I didn't know off the top of my head what comes after that in Turkish so I had to stop there. Both cardinal and ordinal output is supported.

```TeX
\set-counter[id=test,value=17264256543217]
\language[main=tr]{\show-counter[id=test,display=Ordinal]}
```
image

I will leave English and other languages as an exercsize for somebody else.

### DIFF
--- a/core/utilities.lua
+++ b/core/utilities.lua
@@ -427,6 +427,83 @@ utilities.utf8_to_utf16le = function (str)
   return ustr
 end
 
+local romans = {
+  {1000, "M"},
+  {900, "CM"}, {500, "D"}, {400, "CD"}, {100, "C"},
+  {90, "XC"}, {50, "L"}, {40, "XL"}, {10, "X"},
+  {9, "IX"}, {5, "V"}, {4, "IV"}, {1, "I"}
+}
+
+local icu = require("justenoughicu")
+
+local icuFormats = function (format)
+  if format == "roman" then return "romanlow" end
+  if format == "Roman" then return "roman" end
+  return format
+end
+
+local icuFormat = function (num, format)
+  local ok, result  = pcall(function() return icu.format_number(num, format) end)
+  return tostring(ok and result or num)
+end
+
+-- Language specific number formatters add functions to this table,
+-- see e.g. languages/tr.lua
+utilities.formatNumber = {
+  und = {
+
+    roman = function (num)
+      local out = ""
+      num = num + 0
+      for _, v in ipairs(romans) do
+        val, let = unpack(v)
+        while num >= val do
+          num = num - val
+          out = out .. let
+        end
+      end
+      return out
+    end,
+
+    alpha = function (num)
+      local out = ""
+      local a = string.byte("a")
+      repeat
+        num = num - 1
+        out = string.char(num % 26 + a) .. out
+        num = (num - num % 26) / 26
+      until num < 1
+      return out
+    end
+
+  }
+}
+
+setmetatable (utilities.formatNumber, {
+    __call = function (self, num, format, case)
+      if not case then
+        if format:match("^%l") then
+          case = "lower"
+        elseif format:match("^.%l") then
+          case = "title"
+        else
+          case = "upper"
+        end
+      end
+      local lang = SILE.settings.get("document.language")
+      local format = format:lower()
+      local result
+      if self[lang] and type(self[lang][format]) == "function" then
+        result = self[lang][format](num)
+      elseif type(self["und"][format]) == "function" then
+        result = self.und[format](num)
+      else
+        result = icuFormat(num, format)
+      end
+      return icu.case(result, lang, case)
+    end
+})
+
 utilities.breadcrumbs = function ()
   local breadcrumbs = {}
 
--- a/core/utilities.lua
+++ b/core/utilities.lua
@@ -427,13 +427,6 @@ utilities.utf8_to_utf16le = function (str)
   return ustr
 end
 
-local romans = {
-  {1000, "M"},
-  {900, "CM"}, {500, "D"}, {400, "CD"}, {100, "C"},
-  {90, "XC"}, {50, "L"}, {40, "XL"}, {10, "X"},
-  {9, "IX"}, {5, "V"}, {4, "IV"}, {1, "I"}
-}
-
 local icu = require("justenoughicu")
 
 local icuFormats = function (format)
@@ -451,19 +444,6 @@ end
 -- see e.g. languages/tr.lua
 utilities.formatNumber = {
   und = {
-
-    roman = function (num)
-      local out = ""
-      num = num + 0
-      for _, v in ipairs(romans) do
-        val, let = unpack(v)
-        while num >= val do
-          num = num - val
-          out = out .. let
-        end
-      end
-      return out
-    end,
 
     alpha = function (num)
       local out = ""
--- a/packages/counters.lua
+++ b/packages/counters.lua
@@ -1,5 +1,7 @@
 if not SILE.scratch.counters then SILE.scratch.counters = {} end
 
+local textcase = SILE.require("packages/textcase").exports
+
 local romans = {
   {1000, "M"},
   {900, "CM"}, {500, "D"}, {400, "CD"}, {100, "C"},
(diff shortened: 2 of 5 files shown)


alerque:
I just tweaked the related test here in alignment with the changes I've been making in #592 (less unrelated content on the page to trip up metrics, fewer unique required fonts, smaller paper size, etc.). The test does not have an expected version yet, but as soon as that PR is merged it could safely get one. Until then adding one will trip us up on CI.

image

alerque:
Gah! I just found an set of number for which this doesn't work. Don't merge until I figure out how to handle zero value hundredths places in the middle of longer numbers, they are different that the plain hundredth place.

alerque:
Fixed. For a language with simple number verbalization (e.g. no irregular 11, 12, 13, etc.) this is remarkably complex. I checked all the cases I can think of. That doesn't mean it's all right, but here is some testing code to make it easier to throw a bunch of number cases at this:

```tex
\begin{script}
	SILE.registerCommand("t", function(options, content)
		SILE.call("set-counter", { id = "test", value = options.val })
		SILE.call("show-counter", { id = "test" })
		SILE.typesetter:typeset(": ")
		SILE.call("font", { language = "tr" }, function ()
			SILE.call("show-counter", { id = "test", display = "Ordinal" })
		end)
		SILE.call("par")
	end)
\end{script}

\t[val=0]
\t[val=100]
\t[val=200]
\t[val=1000]
\t[val=1020]
\t[val=2000]
\t[val=2020]
\t[val=20000]
\t[val=200000]
\t[val=2200000]
\t[val=3000000]
\t[val=300011000]
\t[val=456789123456789]
```

packages/counters.lua
```diff
@@ -1,64 +1,77 @@
...
   {1000, "M"},
   {900, "CM"}, {500, "D"}, {400, "CD"}, {100, "C"},
   {90, "XC"}, {50, "L"}, {40, "XL"}, {10, "X"},
   {9, "IX"}, {5, "V"}, {4, "IV"}, {1, "I"}
 }
 
-local function romanize(k)
-  str = ""
-  k = k + 0
+local function num2roman(num)
```
simoncozens: Seems like we're standardizing on `local x = function(y)`

packages/counters.lua
```diff
@@ -1,64 +1,77 @@
...
-    local ok, res = pcall(function() return icu.format_number(counter.value, display) end)
-    if ok then return res end
+    local ok, result = pcall(function() return icu.format_number(counter.value, display) end)
+    if ok then return result end
   end
 
-  if (counter.display == "roman") then return romanize(counter.value):lower() end
-  if (counter.display == "Roman") then return romanize(counter.value) end
-  if (counter.display == "alpha") then return alpha(counter.value) end
+  if (counter.display == "string") and num2string then return textcase.lowercase(num2string(counter.value)) end
```
simoncozens: How can `num2string` etc. be false?

simoncozens: You know what, towers of endless (and inextensible) if-statements feel like an opportunity for a dispatch table. I'm leaning towards putting all these display options into a Lua table and then calling (something like) `SILE.counters.formattingFunction[counter.display](counter.value)`.

If we could do that on a per-language basis (*handwave*) then stuff like "ordinal" doesn't need to be baked in, but could be provided by language support files.

alerque: `num2string` etc. can be false if the language does not support that format.

I thought about hoisting more of this off on the language support module, but at least my first implementation had a LOT more code that would end up being duplicated between language. Passing the smallest items of work off to per-language support and keeping a lot of the logic in core at least had the advantage of not repeating code and bugs in each language and knowing what to expect.

I'm not saying it couldn't be done better though. Your idea might work.

alerque: A dispatch table is just a _little_ bit harder in the case of a language not having a particular display method. We can handle it with meta tables I think, but the method I used falls back gracefully in the event that you change the language and the new language doesn't support a method.

simoncozens: > `num2string` etc. can be false if the language does not support that format. Those functions are instantiated as nil and then replaced with the per-language functions if they exist.

I'm not sure that's true. First you have:

```lua
num2string = LS.num2string and LS.num2string or num2string
```

which gives either the language-specific version or falls back to a "default" implementation if there isn't one. So `num2string` is guaranteed defined here. But then you test it again...

```lua
if (counter.display == "string") and num2string
```

> A dispatch table is just a little bit harder in the case of a language not having a particular display method.

The way I thought about it is similar to the above. You start with a "default" dispatch table in `packages/counters.lua`:

```
{
"string" = num2string,
"String" = function(n) return textcase.titlecase(num2string(counter.value) end,
...
}

SILE.formatCounter = function(counter)
...
return SILE.counters.formattingRoutines[counter.display](counter.value)
```

and then to get the language-specific variants, you overlay its entries on top of the default ones.

simoncozens: Incidentally, I'm sure `X and X or Y` is just another way of saying `X or Y`...

alerque:
Okay I'm trying another implementation. I'm moving it to Utilities and calling it numberFormat, and giving it a hook for the languages to add specific support.

packages/counters.lua
alerque: > So num2string is guaranteed defined here.

The variable is defined, but it is not guarantied to be a function. It's default value is nil. Hence testing it for truthiness works to see if we have either a language specific or default implementation. For example for roman numerals there is a real default implementation, because those are not language specific, but for strings there is not even a default, languages HAVE to define it.

core/utilities.lua
```diff
@@ -427,6 +427,63 @@ utilities.utf8_to_utf16le = function (str)
   return ustr
 end
 
+local icu = require("justenoughicu")
```
alerque: Now that ICU is just a requirement rather than an optional add-on, I think we should probably load one instance of it as a global or into SU or somewhere rather than having to load it in every module separately. Thoughts?

simoncozens: My understanding is that it doesn't matter. Lua caches loaded packages (into a table called `packages.loaded`) so you're not really "loading" it again. I think this way is actually more declarative and easy to follow.

core/utilities.lua
```diff
@@ -427,6 +427,63 @@ utilities.utf8_to_utf16le = function (str)
...
+        else
+          case = "upper"
+        end
+      end
+      local lang = SILE.settings.get("document.language")
+      local format = format:lower()
+      local result
+      if self[lang] and type(self[lang][format]) == "function" then
+        result = self[lang][format](num)
+      elseif type(self["und"][format]) == "fuction" then
```
simoncozens: `function`. Otherwise, I love it.

alerque:
Check out this refactor @simoncozens. I went with your idea of a dispatch table but not quite the way you described. Instead of calling the function directly in this design you still call a top level function (actually a table with a __call meta event) with an argument of the desired format and it figures out from the language support table what functions are available, cases the result, and returns it. This is easily extensible by stuffing functions into the correct language table. It falls back to any functions in the "und" language table, so for example there is a default `alpha()` function that will work by default in any language but that can be overridden on a per language basis.

core/utilities.lua
alerque: Savvy.

core/utilities.lua
alerque: Oopsie. Corrected, good catch.

simoncozens: Also, if your tests passed with that misspelling then either the tests are wrong or there’s a case that was not tested...

alerque: The test does not currently have an `.expected` file at all. I didn't add one to this PR either because I don't have the same font versions as Travis and the local testing PR isn't quite fully baked yet.

alerque: Added `alpha` to test, and used `ROMAN` so that we test all three case modes too. Still no expect for reasons noted.

alerque:
Soooooo, my new system was cool but it only worked in Lua 5.3. I'll have to research what the older versions don't support.

Why are we supporting Lua 5.1 again?

alerque:
The `//` floor division operator was added in 5.3. Ugg. Hopefully that was all...

PR #631, opened 2019-06-06
Handle more number formats (including some English)

alerque:
Add negative and scientific (big) number handling, add some English formats, etc.

Note the English formats here won't work properly out of the box until #633 is fixed. Comment out Lines 6 and 7 of core/languages.lua to test.


alerque:
With #636 the English numbering stuff here could be used, and I'm trying to update the test for it in #592 so I'll be merging soon if there are no objections.

PR #632, opened 2019-06-06
Autogenerate makefile dependencies

### LINKED ISSUE #305: Return list of files used in compilation of document

alerque:
Fixes #305.

This keeps track of every file used in the creation of a document including

* [x] system Lua libraries
* [x] SILE's internal vendor Lua libraies
* [x] SILE itself
* [x] SILE core and packages files
* [x] Font files when using harfbuzz (most backends)
* [ ] Font files when using pango
* [x] Document scripts and includes
* [x] Images

... and outputs them in a format that can be included into a GNU Makefile using the `-m` flag (with at optional filename).

Note this can be used with or without the dummy `-b dummy` backend that avoids writing any actual final output so dependencies can be generated on their own or as a side effect of the last rendering pass.

### DIFF
--- a/core/dummy-output.lua
+++ b/core/dummy-output.lua
@@ -1,0 +1,22 @@
+if (not SILE.outputters) then SILE.outputters = {} end
+
+local dummy = function() end
+
+SILE.outputters.dummy = {
+  init = dummy,
+  newPage = dummy,
+  finish = dummy,
+  setColor = dummy,
+  pushColor = dummy,
+  popColor = dummy,
+  outputHbox = dummy,
+  setFont = dummy,
+  drawImage = dummy,
+  imageSize = dummy,
+  moveTo = dummy,
+  rule = dummy,
+  debugFrame = dummy,
+  debugHbox = dummy
+}
+
+SILE.outputter = SILE.outputters.dummy
--- a/core/harfbuzz-shaper.lua
+++ b/core/harfbuzz-shaper.lua
@@ -63,6 +63,7 @@ SILE.shapers.harfbuzz = SILE.shapers.base {
     local face = fontconfig._face(opts)
     SU.debug("fonts", "Resolved font family '"..opts.family.."' -> "..(face and face.filename))
     if not face.filename then SU.error("Couldn't find face '"..opts.family.."'") end
+    if SILE.makeDeps then SILE.makeDeps:add(face.filename) end
     if bit32.rshift(face.index, 16) ~= 0 then
       SU.warn("GX feature in '"..opts.family.."' is not supported, fallback to regular font face.")
       face.index = bit32.band(face.index, 0xff)
--- a/core/makedeps.lua
+++ b/core/makedeps.lua
@@ -1,0 +1,34 @@
+local makeDeps = {
+  _deps = {},
+  add = function (self, file)
+    if not file and file ~= "nil" then return end
+    self._deps[file] = true
+  end,
+  write = function (self)
+    if type(self.filename) ~= "string" then
+      self.filename = SILE.masterFilename .. ".d"
+    end
+    for dep, _ in pairs(package.loaded) do
+      if dep ~= "_G" then
+        self:add(dep:gsub("%.", "/"))
+      end
+    end
+    local deps = {}
+    for dep, _ in pairs(self._deps) do
+      local resolvedFile = package.searchpath(dep, package.path, "/")
+      if not resolvedFile then resolvedFile = SILE.resolveFile(dep) end
+      if resolvedFile then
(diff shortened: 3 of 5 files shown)


alerque:
Off the top of your head @simoncozens do you know if the Pango backend can tell us what font files it is actually using? If not it's probably not important enough to hold up this PR.

alerque:
In light of #634 (I was going to try to figure out how to get fonts out of Pango) I'm going to say that is a can of worms bigger than this issue, and I need this auto-deps stuff yesterday!

Issue #634, opened 2019-06-07
Cairo backend / Pango shaper needs love


alerque:
I can't get the Cairo / Pango backend and shaper pair to work at all. For starters it looks like the shaper isn't even being loaded, but even with that fixed it still doesn't work.

At least a little bit of love should be  show this just to make sure it handles the basics. Once it works, we should figure out some way to test each backend (see #592) just to make sure they each run and generate a valid file of the type expected. 

See also #104.

PR #635, opened 2019-06-07
Fix up Cairo / Pango pair

alerque:
See #634.


alerque:
This is hardly a complete fix so I'll leave the related issue open, but it is one obvious thing that needs doing and a better starting point for anybody that does try working with alternative backends. Since I'm not actively working on it and this is objectively better than not having this fix, I'm going to merge this and let myself or whoever picks it up next re-branch to keep going

PR #640, opened 2019-06-14
Add warnings to mitigate bug 639

simoncozens:


alerque:
Okay on top of this I need to now make the test suite ignore tests that throw this warning right?

simoncozens:
Yes, but hold up a moment. I want to make it use OT functions for non-CFF fonts even on older HN versions, so that the damage is as limited as possible.

alerque:
I don't have to hold up, that's what rebase is for. I'm just working on how to handle when the warning is thrown.

Issue #644, opened 2019-06-14
Unnecessarily modified upstream Lua XML library


alerque:
I modified the XML parsing library to make it easier to deal with data, but that also makes it harder to track upstream changes. I think I should redo that hack a different way to make it possible to update the library with re-applying monkey patches.

Perhaps adding an `__INDEX()` meta method to the returned result to translate `.tag` keys  to `.command`s.

See also #643.

alerque:
This is bogus. We already _ARE_ wrapping the lua-expat (lxp) module, the wrapper is part of SILE not a potentially external vendor script.

I cleaned up the Lua coding anyway, but unless lomwithpos ever gets made into a Lua Rock and promoted to its own project there is nothing to do here.

Issue #638, opened 2019-06-13
Need new Ubuntu package solution


alerque:
Complaints keep showing up in #296 and elsewhere that there is not an easy Ubuntu install solution. Given that Ubuntu users are (on average) less savvy than other distro users because of it's target audience and given that it's one of the largest distros around, we really need to have a packaged solution for it (whether this is a PPA, or a Flatpack, or a Snap or something else).
 
See also #38.

n7s:
I have been packaging up sile snapshots for Ubuntu (SIL uses sile for font testing and our toolchain is Ubuntu-based). Currently it's a manual process and the package is uploaded to a PPA. I'm looking at version v0.10.0 and its new depency chain targetting bionic and focal.

alerque:
@n7s It's been a lot of years since I messed around with Ubuntu. What would we need to to to either ⓐ get official packages included upstream or ⓑ host an official PPA distributions channel for up to date SILE packages?

n7s:
From what I can see there are a few more dependencies on new lua libs which are not packaged yet but I'm too short on spare time for this right now. I hope to get to in in a few weeks (in the meantime we'll likely deploy and use SILE with a manual script recipe). I really appreciate all the amazing work done on SILE! Immensely useful!!

You can look at the current packaging (from January 2019) in our PPA by doing a dget https://launchpad.net/~silnrsi/+archive/ubuntu/smith-py3/+sourcefiles/sile/0.9.5.1~20190131-2sil1/sile_0.9.5.1~20190131-2sil1.dsc

alerque:
@n7s there _are_ more Lua dependencies now, but SILE's build system also has a way to fetch and bundle all of them if that makes things easier, thus not requiring any system installed Luarocks packages.

I'm be happy to help work through this process if you catch me on Gitter sometime. It would nice to be able to handle Ubuntu updates upstream.

alerque:
I've started trying to figure this out, starting with launching [a SILE team](https://launchpad.net/~sile-typesetter) on Launchpad (@simoncozens shoot me your member info and I'll add you). @n7s I added you too, feel free to much around and tell me if I'm doing something wrong.

Next step, add a PPA to the team? And a project to the PPA?

devosb:
@alerque When you write that the build system can fetch the Lua dependencies, does that mean fetch over the Internet? If so, I think Debian (and therefore Ubuntu) packaging policy would prohibit that, as building a package cannot require Internet access, IIRC.

[...]

Issue #536, opened 2018-07-09
use luajit if it is present


mhosken:
Is there any way that sile can use luajit if it is present. sile is pretty pedestrian when run under lua but flies much like tex when run under luajit.

alerque:
I'm not sure this is possible, executing scripts under a given interpreter is kind of up to the system. Perhaps we could setup a wrapper than checks for other interpreters and re-launches itself if possible. Also perhaps this should be a compile/install time option. We could easily have this as a compile time flag and it would get installed based on what is available on the host system.

Out of curiosity, what kind of benchmark are you getting? For example what is your compile time of the documentation PDF under vanilla `lua` vs. `luajit`?

mhosken:
41 page font test took 18s under lua and 7s under luajit. Not as much speed up as I was expecting, but still significant when you have 20 to run.

alerque:
A speedup of 2× runtime is nothing to turn up your nose at. I'd be all for taking any gains where we can get them. However...

Luajit appears to be only very sporadically maintained, and development is only up to par with Lua 5.1. This in itself isn't a blocker because so far we've managed to maintain compatibility with Lua 5.1 (I can't say as much for my personal project libraries, but that's my own problem). The issue I have is that the Lua 5.1 libraries on my system are nowhere near as maintained as the 5.3 ones, and as such even trying out `luajit` will require some system work. Right off the bat even though I have a `lua51-luafilesystem`  package installed both `lua5.1 $(which sile)` and `luajit $(which sile)` are unable to find the `lfs` library:

> luajit: core/sile.lua:14: module 'lfs' not found:

Were you able to run the latest HEAD version of SILE under `luajit` or just the last released version?

mhosken:
I would try if I could get configure past not finding lua_load of all things.

PR #648, opened 2019-06-16
Cleanup Lua library situation

### LINKED ISSUE #643: Not prefering local Lua libraries over system ones
We have modified versions of some Lua libraries that may exist on some systems. Per [this chat message](https://gitter.im/simoncozens/sile?at=5d03a3b0a8d9871b32875f6f) we may not be preferring our local ones properly.

alerque:
This will eventually...

* [x] Fixes #643 by always including the directory SILE was executed from early in the load path.
* [x] Reduce Lua 5.[12] specific hacks using Penlight's compatibility shims.
* [x] ~~Address #644 by manipulating the results provided by the standard library rather than changing the base library.~~ Original issue was bogus, just cleanup Lua coding.
* [x] Ease work on #588, #143, #638, and #410 by bundling libraries used from LuaRocks as part of build process (for all platforms) rather than our current mix-and-match combination of including them in our repo (update and maintenance nightmare) and requiring them on the system (compatibility and packaging nightmare).
* [x] Start addressing #536 by ~~detecting possible Lua executables at build time~~ starting the process of testing on LuaJIT.


alerque:
I got through a few speed bumps out of the way towards testing with LuaJIT, but not all of them and ran out of time to throw things at Travis. I don't think that issue has to hold up this PR though, so I've opened it up for review.

@harrysummer There is now a rockspec that lists all the Lua dependencies. This should make it a lot easier to maintain the CMake recipies too because one call to `luarocks` should be able to get you all the deps you need.

simoncozens:
If we're using Penlight now, then we should probably revisit our use of `std` as well.

harrysummer:
> @harrysummer There is now a rockspec that lists all the Lua dependencies. This should make it a lot easier to maintain the CMake recipies too because one call to `luarocks` should be able to get you all the deps you need.

Nice work! I will try to integrate the change into windows build after this PR is done.

simoncozens:
I'm having a crack at replacing the std object system with penlight's. It's something I've wanted to do for a long time. It's clear penlight has a much better claim to being the Lua "standard library" than std has these days, and I think moving to pl may have both speed and forward-compatibility advantages. I like the idea of this PR but have problems with us using two "standard library" solutions.

alerque:
I agree that phasing out `std.*` is probably a good idea in the long run, but I don't think sourcing bits from a couple different libraries is really a problem. There are _no_ libraries that really have a claim to being the "Lua standard library". The upstream project seems pretty resistant to such a thing forming anyway. We're already using little bits here, little bits there from several Lua rocks, there is no particular reason I've found not to take a pragmatic approach to using the bits we need.

I would say that Penlight's implementations of various programming idioms seem to be generally better implemented, maintained, and resilient that some others and would favor going ahead and taking advantage of what it offers. I just don't think it has to be all or nothing out of the gate.

Before doing so, I think we should probably work on some sort of benchmark stress test so we can compare performance. I think that should be an important factor when substituting bits of tooling.

simoncozens:
I've just been doing some work on cassowary, replacing std with pl there. Running the test suite 30 times under luabench, I get the following results:

```
$ git checkout master
$ lua cassobench.lua
       system: 4.280	user: 0.000	total: 4.280	real: 4.296
$ git checkout penlight
Switched to branch 'penlight'
$ lua cassobench.lua
       system: 0.710	user: 0.000	total: 0.710	real: 0.713
```

So about a 7x speedup in Cassowary by dropping std for penlight. Of course std might not be the hotspot in SILE, but... I think that tells me that std's object creation is heavy.

configure.ac
```diff
@@ -89,13 +89,6 @@ esac
 
 AX_LUA_HEADERS
 AX_LUA_LIBS
-# Check for Lua modules
-AX_LUA_MODULE([lpeg], [lpeg])
```
simoncozens: I'm not sure about this. I don't like the idea that someone can run ./configure && make and get a non-working product at the end. Why not leave these checks in and people can at least get told if they don't have the dependencies installed?

simoncozens:
I like a lot of this but there are two things I'm not sure about:

* Removing module checks from autoconf (as I commented on)
* Moving everything to a rockspec and keeping rocks updated

I've been bitten too many times by non-backwards-compatible updates (particularly of std) which is why I wanted to ship as many of the Lua modules as we could - that way we can lock down the versions. I suppose with continuous integration we get early warning of breakages, but I want to keep things as predictable for the user as possible.

configure.ac
alerque: Because `pkg-config` does not understand where to find the packages. We're doing a lot of manipulation to the lua path including re-exporting it with `luarocks`. At least in my cursory testing was that it was finding the system libraries, but not our manually supplied ones. That is, after all, what the tool is supposed to do: find stuff on the system not lint the current project. We don't need it to find the lua modules because we do that ourselves as part of the next step: `make`.

alerque:
1. Autoconf helps us interface with different host systems, not with linting the current project. If our sources are not in order in the current directory that's not `./configure`'s problem...

2. If backwards comparability is a problem we can hard code the module version we want in the rockspec.

Running `./configure` will tell you if you _host system_ has what it needs, but it is not supposed to give you a functioning product yet. That happens during `make`. If fetching the lua sources (as specified in the rockspec) fails then `make` should fail. It make succeeds, then you should have a working product.

simoncozens:
OK, so I tried this out on a spare server. I see where you're going with this and I think I like the idea, but there are still some hiccups. For example, luasec is often a pain to build because the Lua module is terrible at working out where openssl is installed. Now on the spare server I happen to have the `lua-ssl` package installed, which came with the (Ubuntu) distribution and so I should be able to skip the whole process of configuring luasec correctly. Maintaining our own luarocks tree has some benefits, but it means that we don't take advantage of stuff installed locally.  I had to manually run

```
luarocks --tree lua_modules install luasec OPENSSL_LIBDIR=/usr/lib/aarch64-linux-gnu
```

after the make failed, and then run it again. Not a massive deal, but since I already had a working luasec module installed, a bit annoying.

alerque:
Setting that env variable via `pkg-config` by exporting it during `./configure` or just looking it up as part of the build process is probably a way to smooth that.

Also we might be able to tell luarocks not to duplicate dependencies already on the system. I'll look into that.

simoncozens:
I'm not sure the code which loads the local locks tree is working properly:

```
simon@jetson ~/sile .*$ ./sile examples/simple
This is SILE 0.9.5.1
<examples/simple.sil>

! Font family 'Gentium Plus' not available, falling back to 'FreeMono' at in "␤I have seldom hea…"
[1]

simon@jetson ~/sile .*$ make install
[...]
simon@jetson ~/sile .*$ cd examples
simon@jetson ~/sile/examples .*$ sile simple.sil
/usr/bin/lua: /usr/local/share/sile/core/sile.lua:18: module 'bit32' not found:
	no field package.preload['bit32']
	no file './lua_modules/share/lua/5.1/bit32/init.lua'
	no file './lua_modules/share/lua/5.1/bit32.lua'
	no file './lua-libraries/bit32/init.lua'
	no file './lua-libraries/bit32.lua'
	no file './bit32/init.lua'
	no file './bit32.lua'
	no file '/usr/local/bin//lua_modules/share/lua/5.1/bit32/init.lua'
	no file '/usr/local/bin//lua_modules/share/lua/5.1/bit32.lua'
	no file '/usr/local/bin//lua-libraries/bit32/init.lua'
	no file '/usr/local/bin//lua-libraries/bit32.lua'
	no file '/usr/local/bin//bit32/init.lua'
	no file '/usr/local/bin//bit32.lua'
	no file '/usr/local/lib/sile/lua_modules/share/lua/5.1/bit32/init.lua'
	no file '/usr/local/lib/sile/lua_modules/share/lua/5.1/bit32.lua'
	no file '/usr/local/lib/sile/lua-libraries/bit32/init.lua'
	no file '/usr/local/lib/sile/lua-libraries/bit32.lua'
	no file '/usr/local/lib/sile/bit32/init.lua'
	no file '/usr/local/lib/sile/bit32.lua'
	no file '/usr/local/share/sile/lua_modules/share/lua/5.1/bit32/init.lua'
	no file '/usr/local/share/sile/lua_modules/share/lua/5.1/bit32.lua'
	no file '/usr/local/share/sile/lua-libraries/bit32/init.lua'
	no file '/usr/local/share/sile/lua-libraries/bit32.lua'
	no file '/usr/local/share/sile/bit32/init.lua'
	no file '/usr/local/share/sile/bit32.lua'
	no file './bit32.lua'
	no file '/usr/local/share/lua/5.1/bit32.lua'
	no file '/usr/local/share/lua/5.1/bit32/init.lua'
	no file '/usr/local/lib/lua/5.1/bit32.lua'
	no file '/usr/local/lib/lua/5.1/bit32/init.lua'
	no file '/usr/share/lua/5.1/bit32.lua'
	no file '/usr/share/lua/5.1/bit32/init.lua'
	no file './core/bit32.so'
	no file './bit32.so'
	no file './lua_modules/lib/lua/5.1/bit32.so'
	no file '/usr/local/bin//core/bit32.so'
	no file '/usr/local/bin//bit32.so'
	no file '/usr/local/bin//lua_modules/lib/lua/5.1/bit32.so'
	no file '/usr/local/lib/sile/core/bit32.so'
	no file '/usr/local/lib/sile/bit32.so'
	no file '/usr/local/lib/sile/lua_modules/lib/lua/5.1/bit32.so'
	no file '/usr/local/share/sile/core/bit32.so'
	no file '/usr/local/share/sile/bit32.so'
	no file '/usr/local/share/sile/lua_modules/lib/lua/5.1/bit32.so'
	no file './bit32.so'
	no file '/usr/local/lib/lua/5.1/bit32.so'
	no file '/usr/lib/aarch64-linux-gnu/lua/5.1/bit32.so'
	no file '/usr/lib/lua/5.1/bit32.so'
	no file '/usr/local/lib/lua/5.1/loadall.so'
stack traceback:
	[C]: in function 'require'
	/usr/local/share/sile/core/sile.lua:18: in main chunk
	[C]: in function 'require'
	/usr/local/bin/sile:38: in main chunk
	[C]: ?
```

simoncozens:
Or maybe the tree didn't get installed.

alerque:
Bit32 is an edge case, and yes it may not be handled right yet. Newer versions of Lua don't need that rock at all, and Luarocks knows about that. But if your Lua doesn't have it it should have fetched it...

simoncozens:
It did fetch it and it works when uninstalled. But it doesn't look like `make install` puts those downloaded luarocks (which it installed into a local tree called `lua_modules` in the development directory) anywhere globally visible on the system, which means running `/usr/local/bin/sile` doesn't find those modules. There are no `/usr/local/share/sile/lua_modules/` or `/usr/local/share/lib/lua_modules/` directories, and I think we should be copying the downloaded rocks from the local tree into one.

alerque:
Hmm, yes you do. We need to tell `autoconf` that `lua_modules` is something we need installed.

alerque:
Gah, that won't work. Hang on.

simoncozens:
Seems to have done...

alerque:
It will work if you've run `make` and gotten the Lua Rocks before you run `./configure`, but not if you do it in a clean chekcout.

alerque:
It's hard for me to test because I _NEVER_ let anything install itself on my system apart from my package manager, and our `./configure --prefix=` is broken. It works for the top level but not for the stuff under `src`.

simoncozens:
OK. We're going to need something *like* this in soon, because `master` doesn't install properly either:

```
simon@upo ~/hacks/sile/catalogue .$ sile test
/usr/local/bin/lua: /usr/local/bin/sile:30: module 'core/sile' not found:
	no field package.preload['core/sile']
	no file 'core/sile'
	no file 'core/sile.lua/usr/local/share/sile/core/sile.lua'
                             ^ looks like there should be a semicolon in the path here
```

alerque:
I suspect that's a problem with your environment, because I'm running the current master in production via the Arch Linux sile-git package, which creates it's install using a `./configure --prefix=/usr; make install DESTDIR=<package path>`. That's working okay right now.

That being said, what is holding up this PR from being merged at this point? As noted above actually weaning ourselves from `std` can be an extended process. What else is going wrong that I can fix?

simoncozens:
Yeah, I'm not worried about std for this PR. But Travis is clearly unhappy about something. Once that's clear I'm happy to merge.

alerque:
Of course it would fail only _after_ succeeding. It looks like it works from a clean slate but after the first success it starts caching things and our scripts don't expect that. I'll fix...

alerque:
@simoncozens Does `pkg-config --libs openssl` find your OpenSSL installation?

simoncozens:
Yes; it gives you the linker flags (`-lssl -lcrypto`) but not the directory. The luasec rock seems to assume (on Linux) that all the world is `x86_64`, when I happen to be on `aarch64`.

```
$ luarocks --tree lua_modules install luasec
Installing https://luarocks.org/luasec-0.8-1.src.rock

Error: Could not find library file for OPENSSL
  No file libssl.a in /usr/lib
  No file libssl.a in /usr/lib/x86_64-linux-gnu
  No file libssl.so in /usr/lib
  No file libssl.so in /usr/lib/x86_64-linux-gnu
  No file matching libssl.so.* in /usr/lib
  No file matching libssl.so.* in /usr/lib/x86_64-linux-gnu
You may have to install OPENSSL in your system and/or pass OPENSSL_DIR or OPENSSL_LIBDIR to the luarocks command.
Example: luarocks install luasec OPENSSL_DIR=/usr/local
```

(However, I can't see where it's getting the `x86_64` bit from.)

alerque:
Are your OpenSSL headers from Homebrew? If so have you run `brew link openssl`?

alerque:
Path issues aren't just you. I just discovered there is a difference between by office test system installed with the AUR package built and tested locally and the same AUR package installed remotely. :fearful:

alerque:
Since my last comment I have not been able to reproduce the badly concatenated path problem as discussed. I've installed/uninstalled/repackaged dozens of times over a couple machines and just haven't seen the same problem.

In the process I did determine there was another bug the install process that only became a problem with this PR's handling of lua modules. The list of files to include in `make install` was being generated _before_ make was run instead of after. Due to an oddity in the way make regenerates anything in includes and the fact that in effect `./configure` was getting run twice, we almost got a free pass on this. However that wasn't enough to catch the files which were being placed by lua rocks but not directly generated by make.

The last commit added here moves the brute force `find` hack being used to generate a list of files to include in the installation into the `Makefile` itself so that it will be run when actually used (by which time all files should be present).

Long term a better way of listing files should be considered as this one will only work consistently if always being run from a clean source directory, might contain artifacts if ever run from a directory used for development, and will certainly include sources that are _not_ actually needed in the installation. I'm leaving that as an exercise for another PR though.

alerque:
@simoncozens re [this comment](https://github.com/simoncozens/sile/pull/648#issuecomment-505331889), would you might checking from a clean repository (at this point `make clean` is not enough because it only removes build artifacts understood by autoconf which does not include our lua rocks situation, `git clean -dxf` can be used in a git repository to get a clean start) with the usual `./bootstrap.sh && ./configure && make && make install` on (you can add a prefix dir of course as required) on your Mac setup to make sure it is putting all the files required to actually run onto your system? It's working out well for me on Linux now including packaged for Arch Linux, but it would be nice to get confirmation on another platform before merging this. And that's one thing we don't have Travis doing yet.

alerque:
There is still some bad mojo going on here, but it is not entirely this PR's fault. The issue dates all the way back to 820e8d9 that introduced a subtle Lua anti-pattern: the `require()` function should _not_ be passed a full path, only ever a module name. Amazingly we managed to side step this manifesting itself by having other path issues, but in expanding the path handling in other places to systematically cover the things it should be doing, I seem to have opened a couple situations where the `SILE.resolveFile()` function finds things where before it was missing them. When it _fails to find anything_ then we were calling `require()` properly, but now if it finds one `require()` gets passed a full path, which it doesn't like.

alerque:
Okay getting better but not quite there. Lua 5.1 needs a shim and there might be a small issue with `-I` from PWD. Or that might be a CaSILE issue.

alerque:
Okay the CaSILE issue is because it creates a special include to be used with  `-I` to pass environment variables into Lua, but it does it as a dotfile by default. This is an issue for `require()` which assumes the leading dot is a module path break character. Standby.

alerque:
Wow. My cleanup papered over the fact that `-I` has _always_ had an identity crisis, sometimes processing includes with the SILE input reader, sometimes processing them with Lua's `require()`. I don't know how I missed that in my own usage, but the same flag being used to load files two different ways (and hence having their variables in different scopes) is confusing. I think we're going to need two flags so the user can specify the purpose.

alerque:
For cross reference, here is the main reason why I switched from `pl.compat` to `compat53`:

https://github.com/stevedonovan/Penlight/issues/303

It looks like Penlight only tried to add shims for a few functions it uses internally itself and didn't try to more comprehensively bring Lua 5.1 up to speed. There are quite a few more cases where we could make use of Lua 5.3 idioms with this set of shims instead.

PR #650, opened 2019-06-19
Update Windows build script

harrysummer:
* [x] Use VS2019 instead of VS2017 (currently use personal fork as workaround. Should change back to official branch after https://github.com/luarocks/luarocks/pull/1040 get merged)
* [x] Update the version of the dependencies in CMakeLists.txt
* [x] Update the patch of fontconfig to make it adapt to future version (previously it is a dirty patch which introduces a lot of intermediate files generated by scripts. Now it's pure CMake. Note there is a [bug](https://gitlab.freedesktop.org/fontconfig/fontconfig/issues/166) introduced in the new version of fontconfig that affect the build. It is currently patched in the diff.)
* [x] Merge the update to libtexpdf here: https://github.com/simoncozens/libtexpdf/pull/9
* [x] Windows installer integration as initially proposed here: https://github.com/simoncozens/sile/issues/410#issuecomment-502793713 (Note: eventually we may change to use [CPack](https://cmake.org/cmake/help/latest/module/CPack.html) to make the script portable to MacOS)
* [x] Test suite integration: https://github.com/simoncozens/sile/issues/646


alerque:
Note the errors by @commitlint aren't very accurate, you can get much better ones on the command line my running `npm install` in your local working directory. That should setup the commit message linter that actually uses the SILE configuration and a git commit hook to warn about issues. It also has a commit message authoring helper script (`npx git-cz`).

harrysummer:
> Note the errors spewed by @commitlint in Github comments aren't very accurate because they don't actually read SILE's commitlint configuration file. You can get much better ones on the command line my running `npm install` in your local working directory. That should setup the commit message linter that actually uses the SILE configuration and a git commit hook to warn about issues. It also has a commit message authoring helper script (`npx git-cz`).

Thanks! I took a look at the configuration file directly and the commentlint problems should be fixed now. First time hearing about commentlint. Cool stuff.

harrysummer:
Note: currently a lot of test cases fail in comparing with the "expected" results. So I temporarily make the test step in Azure Pipeline CI as "continueOnError: true". We should remove it after we fix all failure cases.

harrysummer:
The coverage test result seems unstable. Sometimes it could just range +/- 10% from the previous number. No idea why.

harrysummer:
Just want to add a note here for: https://github.com/microsoft/vcpkg. It could possibly ease the Windows build process.

alerque:
I'm not entirely sure where `vcpkg` would help us, it's a lot of machinery considering we only have a couple of tiny project-internal C dependencies.

On the Travis side of things builds are sped up by caching the external dependencies that are not directly related, like Harfbuzz. Is there a way you can cache those on Azure too and only build SILE itself?

harrysummer:
Current Azure Pipeline does not support caching. But I believe the team is actively developing this feature (see the [blog](https://devblogs.microsoft.com/devops/adding-caching-to-azure-pipelines/)).

Although we can leverage caching in CI, it could still be a performance issue for my local build. For some unknown reason, the local clone of the repo always build the dependencies even if I didn't modify anything.

I think `vcpkg` at least has the following advantages:
1. For Windows build on Azure Pipeline, it could speed up the building process for now by decoupling dependency building and SILE building.
2. Unifying the versions of the dependency libraries (e.g., harfbuzz), which could resolve the unit test parity issues on different Linux distributions.
3. If we are transiting to CMake in the future, this is a good building block towards unifying the build script on different platforms.

What I am playing with at the moment:
1. Fork `vcpkg`.
2. Create a build pipeline on `vcpkg` repo, generating artifact consisting pre-built `icu`, `harfbuzz`, `fontconfig`, `luajit` libraries, as well as supporting CMake scripts.
3. Create a release pipeline to deploy the artifact to Github release page weekly.

After these are done, on SILE's side, building will be as simple as downloading the release from Github,  including the cmake script in it, and adding the dependencies in CMake targets.

harrysummer:
I figured out to create a release feed here: https://github.com/harrysummer/vcpkg/releases

It will automatically fetch updates from the official repo and create weekly release. (It's still stable version for the libraries. Just vcpkg itself is from its master.) Let me try to refine Windows CI build with this pre-built dependency. I am sure it will speed up the build process a lot and clear out the mess of patches in cmake folder...

ras0219-msft:
@harrysummer Thanks for trying out vcpkg!

While we are a community project, we heavily emphasize avoiding regressions. If you can help contribute your CMake patches and fixes to us as PRs, we can help keep them up to date moving forward and enable your single tool dream! :)

alerque:
I've done a bit of maintenance on this PR hoping to get it ship-shape for merging. I hope you don't mind I amended a lot of the commits to comply with our project specific commitlint rules (instead of just the generic ones you used), and then I merged _from_ the current master to resolve the merge conflicts that would have happened when trying to merge this _to_ master.

Is there anything else outstanding on this that should hold up a merge?

harrysummer:
Looks like Windows CI build is not running now. Shall we least make it an optional requirement so we can check this PR against Windows build to make sure everything still works fine?

alerque:
It already is optional, but since it isn't running … it got stuck when we migrated to this org and @simoncozens couldn't login to his account. If that state is permanent I guess we should just add it from scratch on the new org url.

alerque:
I've resurrected the Azure CI runner (by setting up a new one with a matching org namespace). Remarkably it isn't very broken. It seems to run and build okay, although I tried out the artifacts on a Windows VM and there seems to be a problem with the Lua include paths. That's with the v0.10.0 artifact though, not this PR. The runner should start working on this PR again shortly and we'll see how that goes.

alerque:
@harrysummer The Azure build for this is running now, but [seems to have failed](https://dev.azure.com/sile-typesetter/sile/_build/results?buildId=8&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=f33089cd-d3fe-578c-2c5e-4ee560e10a0d). It works on `master` though so the problem should trace to something in this PR, not to the overall setup. It looks like it might be a simple as checksums changing, I'll look into that.

If (now or in the future) adding you to this project or the Azure DevOps user list will help with setting up or maintaining Windows builds just let me know.

alerque:
With some URL fixes, the "Build solution" phase completes; but the whole thing times out on "Run tests". It looks like its waiting for input.

alerque:
Per #1237 I'm disabling the Windows CI runner until this is actually fixed. I'd love to see it working but don't personally have the time or expertise to invest in making it so right now. I'll be happy to facilitate this being merged if anybody picks up on it and works out the kinks so it runs.

alerque:
See [this comment](https://github.com/sile-typesetter/libtexpdf/commit/909618e7fea8a58cce74edd643a97c13ca95d778#commitcomment-56935990) from @be5invis identifying (at least one of) the current problems.

alerque:
Somewhat to my surprise, this seems to be building still, it is just the test segment that fails. That probably needs font updates for current tests, but it seems to be failing even before that. I'd be happy to get the build working first and fix the tests later if that's an option here.

Also probably worth noting, the fontconfig patch in this change set is for <= 2.13.91. It seems the fix is merged in >= 2.13.92. We should probably get this updated to the current release, 2.13.94.

be5invis:
Hmm the AzP message says... timeout???
Maybe you can try increase the timeout and see whether it is just because of _slowness_, or something like deadlocking happened. Ref:
 - https://stackoverflow.com/questions/56015919/azure-devops-build-pipeline-reduce-the-timeout-to-30-minutes
 - https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts

be5invis:
(ps. Github Actions supports Windows too)

Omikhleia:
Due to inactivity, and as part of backlog cleaning, I would have been tempted to close it, but the topic on having SILE on Windows remains interesting, even though the approach might be different with today's SILE.

PR #499, opened 2017-12-04
Add lilypond package

alerque:
This PR is to track progress on adding a package that will provide support for embedding Lilypond music in SILE documents.

- [x] Merge #495 (changes to input grammar needed for this to work)
- [x] Handle Lilypond file includes
- [x] Handle inline Lilypond code
- [ ] Pass directory tracking info so Lilypond includes can be relative to the include file
- [ ] Add fragment mode to handle bare notes
- [ ] Pass information about current position in line so inline mode works as well as paragraph mode
- [ ] Make sure vertical space is available and reserved for each system before rendering (this will likely turn out to be a bug in the `\img` command that needs fixing first).
- [ ] Document usage

Will close #435.

PR #652, opened 2019-07-18
Generalize determining if tag is a passthrough one or not

alerque:
This is an extraction (and slight cleanup) of what I did for #499 that may be useful to other packages.

Note discussion in Gitter chat [starting here](https://gitter.im/simoncozens/sile?at=5d2f77fc7dbfff3242c7241e).

This does not solve the "cart before the horse" issue where external packages will not be able to add commands to the pass-through white-list, but it does generalize the current solution for packages included in the SILE source tree.

If anybody has feedback and we decide on an approach to the catch 22, I'm happy to add it on here. Options I've thought of so far:

 * command line flag to add 'enabled parsers', e.g. `sile -p lilypond,python file.sil`
 * a tag syntax such as `\raw:lilypond{}` where `\raw:*` could be parsed ahead of time no matter what it was
 * a tag option `\raw[passthrough=true]{}` to trigger this behavior on any tag.
 * parse the data both ways and make the raw data available somewhere for commands to get at.
 * **edit:** bonus option: only ever use one tag: `\script`, that defaults to handling Lua code but the command for it looks for other parsers based on a language option: `\script[language=lilypond]` or `\script[language=python]`.

Also note this does not cover the XML syntax input option. For that we should probably use standard XML embedding options like CDATA.

### DIFF
--- a/core/inputs-texlike.lua
+++ b/core/inputs-texlike.lua
@@ -4,7 +4,18 @@ local epnf = require("epnf")
 local ID = lpeg.C( SILE.parserBits.letter * (SILE.parserBits.letter+SILE.parserBits.digit)^0 )
 SILE.inputs.TeXlike.identifier = (ID + lpeg.P"-" + lpeg.P":")^1
 
+SILE.inputs.TeXlike.passthroughCommands = {
+  script = true
+}
+setmetatable(SILE.inputs.TeXlike.passthroughCommands, {
+    __call = function(self, command)
+      return self[command]
+    end
+  })
+
 SILE.inputs.TeXlike.parser = function (_ENV)
+  local isPassthrough = function (_, _, command) return SILE.inputs.TeXlike.passthroughCommands(command) end
+  local isNotPassThrough = function (...) return not isPassthrough(...) end
   local _ = WS^0
   local sep = S",;" * _
   local eol = S"\r\n"
@@ -52,8 +63,8 @@ SILE.inputs.TeXlike.parser = function (_ENV)
       Cg(myID, "command") *
       Cg(parameters,"options") *
       (
-        (Cmt(Cb"command", function(_, _, command) return command == "script" end) * V"passthrough_bracketed_stuff") +
-        (Cmt(Cb"command", function(_, _, command) return command ~= "script" end) * V"texlike_bracketed_stuff")
+        (Cmt(Cb"command", isPassthrough) * V"passthrough_bracketed_stuff") +
+        (Cmt(Cb"command", isNotPassThrough) * V"texlike_bracketed_stuff")
       )^0
     ) - P("\\end{")
   environment =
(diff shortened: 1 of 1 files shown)


OlivierNicole:
Thank you for taking the time to factorise this! One issue I have is that I would need the decision to parse or not to depend not only on the tag, but also on the attributes.

The reason is that I imagined that `\math{...}` would contain MathML (which can be parsed), while `\math[type=texlike]{...}` would contain TeX-like equations (which will have their own parser in `math.lua`).

Is that reasonable? If not, I imagine having distinct commands like `\math` and `\texmath` would work as well. Eventually, I think we will want to have a shortcut in the TeX-like parser such as `\( ... \)` to input math, so maybe I'm splitting hairs here.

alerque:
Hey @OlivierNicole, I've been thinking since I posted this that maybe a better option is one I didn't mention (just added): only ever using one tag that has raw data to be parsed by a different parser. This would mean `\script` would be used for everything that needed to be handled in "raw" format without the PEG grammar descending into it, and the `script()` function would look for extra argument for who to pass the data too. This would mean packages could add their own data parsers.

For your use case I would envision this looking like `\math{}` for your MathML mode that can be parsed and `\script[language=texmath]{}` for the equation language that needed to be passed to an external parser.

What do you think (this is still basically a brainstorm!)?

OlivierNicole:
The `\script` option has the advantage of supporting per-package parsers without modifying SILE's core. So that's a great point.

However, and not completely unrelatedly, I keep worrying that without a very short synonym à la LaTeX to input math, many in the scientific community will find using SILE too verbose (and I understand them). But that seems to require a hardcoded change to the parser…

Or maybe there could be a generic shortcut `\( ... \)` for \script[language=xxx]{...}` in the parser, where `xxx` can be changed in the document and by packages.

alerque:
I think all of the ideas I posted at the top of this PR would support per-package parsers without changes to core (with various usage differences for how to accomplish that). The question here is which one do we want to use?

I wouldn't worry too much about the shortcut syntax for now, there may be several ways to implement that down the road, but we need a general purpose solution for all the other languages that might get embedded in a document with the intention of being run by some external process (whatever sort of thing that is).

OlivierNicole:
Well my arguments, without much order:

- I'm not a big fan of requiring command-line options, the idea of setting these things in the document source (by loading packages or explicit options) seems user-friendlier without disadvantages I can see.
- The `\raw:*` tag or `\script[language=*]` are both nice, I guess my tiny dissatisfaction with them is that they emphasize the raw-ness of their contents and not its semantics. I.e. what `\raw:texmath` primarily says is “this is raw TeX math” and not “this is an equation to be typeset”. But that's not terribly important.
- having a special `passthrough` attribute is interesting, although I don't quite see yet what could be the uses.

OlivierNicole:
> Also note this does not cover the XML syntax input option. For that we should probably use standard XML embedding options like CDATA.

@alerque Would you happen to have a link with details on how this work and what it is for?

alerque:
No, but there is Google. The first result I got was [this SO Q&A](https://stackoverflow.com/questions/2784183/what-does-cdata-in-xml-mean) which seems to cover the basics. Basically XML as a language already has a way to include strings of data in a tag that don't have to follow the syntax rules of XML because they are content is some other format. That's basically what we're inventing for SILE's TeX like syntax.

OlivierNicole:
Thanks. That is an option that we didn't consider yet: a special tag that makes its contents passthrough, but does not have any effect on its own. It would probably be a pain to use, though.

alerque:
I'm going to go ahead and merge this PR, not because it solves the bigger problem of what do do with more tags, but because it is much better code for the what we are currently doing and a _lot_ easier to tinker with. I'm experimenting with several ways of using SILE (embedding bits in another app, wrapping the whole thing for a specific purpose, etc.) and even though the overall effect is the same this code is a lot easier to coerce into doing something else.

Until such a time as we settle on a permanent solution it should at least provide an easier base for branches/PRs to work from that implement other-language parsing inside SILE.

PR #654, opened 2019-07-31
Un-bust busted

alerque:
For the life of me I can't figure out why `make busted` is  failing for me and not anybody else (including Travis). This is my adventure in figuring out why and attempting to fix it. Feedback welcome!

```
No patterns for language 
table: 0x5564919313d0
●●◼◼◼●●◼◼●●●●●●●●●●●●●●●●●●●●●●●●●●●●●
33 successes / 5 failures / 0 errors / 0 pending : 0.601402 seconds

Failure → tests/complex_frame_spec.lua @ 17
Overlapping frame definitions Frame B should have top
tests/complex_frame_spec.lua:17: Expected objects to be equal.
Passed in:
(number) 0
Expected:
(number) 8

Failure → tests/complex_frame_spec.lua @ 24
Overlapping frame definitions Frame A should have bottom
tests/complex_frame_spec.lua:24: Expected objects to be equal.
Passed in:
(number) 0
Expected:
(number) 8

Failure → tests/complex_frame_spec.lua @ 25
Overlapping frame definitions Frame A should have height
tests/complex_frame_spec.lua:25: Expected objects to be equal.
Passed in:
(number) 0
Expected:
(number) 7

Failure → tests/frame_spec.lua @ 14
The frame factory Simple should have width
tests/frame_spec.lua:14: Expected objects to be equal.
Passed in:
(number) 0
Expected:
(number) 270

Failure → tests/frame_spec.lua @ 15
The frame factory Simple should have height
tests/frame_spec.lua:15: Expected objects to be equal.
Passed in:
(number) 0
Expected:
(number) 180
Makefile:984: *** [busted] Error 1

#0  busted at /home/caleb/projects/sile/Makefile:984
```

Busted version 2.0.0.


alerque:
@simoncozens What you you have for `busted --version`?

alerque:
The commits to date here have fixes to the build system that have at least one forward step: plain Lua tests and Busted tests are giving me the same results now. It appears that minor version differences between what Busted was insisting on finding on my system vs. what we are fetching locally when configured `--without-system-luarocks` was enough to produce the inconsistency.

Now the problem is that the condition I've managed to normalize on is the failing one. I'm still not sure where the breakage actually happened.

I think this very problem itself is a strong argument for adding full version numbers to our rockspec. System packages can do whatever they want and in generally we can call out our "minimum" support versions, but for the purposes of tracking down when external dependencies change our internal behavior it would be nice to have a dependency versions in the spec so that `git bisect` could track down when they changed.

PR #663, opened 2019-09-02
Non infinite pages

### LINKED ISSUE #662: Infinite pages
I must have broke something. After updating my build servers using the latest master compiled with `--with-system-luarocks`, some of my book publishing CI jobs started spewing infinite pages (until the jobs timed out). Not all jobs fail, but many of them do. At the same time `make test` is passing, so whatever the fail condition is we're not testing for it.

alerque:
Fixes #662.

Also fixes what #654 was trying to address and failed.


alerque:
Travis why do you hate me?

Issue #665, opened 2019-09-12
Use Fluent for i18n


alerque:
A while back I started playing around with [Project Fluent](https://projectfluent.org/). It is hands down *the* most flexible localization system I've run across (and I've used quite a few). Since first using it for a project everything else has felt like a clunky kludge, sort of like `cvs` & `svn` felt like a limitation after `git` came along with all its lightweight branching and distributed.

While localizing web interfaces is all the rage and even at this early stage Fluent has good Javascript tooling, a lot of my work is more long-form document based than user interface based. I've been doing multi-lingual document build systems for a while, including of course using having [CaSILE](https://github.com/alerque/casile/) spit out books where both the original and a translation get fed through the same publishing pipeline.

While SILE itself handles content in in lots of languages, it doesn't actually itself have very much in the way of content that needs localization. But not very much is not none. A simple example is the "Table of Contents" title string hard coded into ... well actually it's spread out between the relevant languages files, the toc package, the book class, and any user injected code. More than once in writing my own packages either for CaSILE or custom to specific projects I've run into surprise cases of either accidentally overriding SILE's default strings when what I wanted to customize was something about the code logic or wanting to customize a string and needing to re-invent the language selector logic all over again.

I'm propose that SILE officially adopt Fluent for all its internationalization. This includes (in order of priority):

1. [x] All strings in the SILE source that end up output to rendered documents.
2. [x] All strings output by packages.
3. [ ] The interface of SILE itself, such as error messages.

Obviously the tooling isn't there to make this happen yet. As of yet there are no Lua bindings for Project Fluent. I'm sure if I wanted to be famous I could be the one to write an implementation of Fluent for Kotlin and get the Android i18n ecosystem fixed up (or Swift or .Net or whatever), but I only have smaller delusions of grandeur: I've started a project for [`fluent-lua`](https://github.com/alerque/fluent-lua). I'm sure this won't take longer than 10 ~~days~~ ~~weeks~~ years.

I'm opening an issue here to get feedback on the implementation and how it will be useful to this and other Lua projects. I don't want to build it just for SILE (and I'll be using it elsewhere too) but this is one use-case I'd like it to play nicely with.

For SILE's purposes I'm envisioning:

* Keeping parsed resource definitions in a stop level table (`SILE.fluent`?), organized by language, but with all keys kept in the same table. This would give any package –and even document– the ability to provide their own localization either for their own internal use or to reach into SILE's internals and override or extend the core's default localization.
[...]

Issue #437, opened 2017-01-20
Add TikZ support package


alerque:
This was [mentioned](https://github.com/simoncozens/sile/wiki/easy-jobs#tikz) on the Wiki but an issue was never opened, so here goes. For me the need for this is greatly reduced by having SVG support, but it would still be nice to have for bringing documents over from TeX and would probably still be a huge boon for some markets.

Implement a package that generates graphs using TikZ and embeds the output. Since there is a Lua interface this should be easier than (but conceptually similar) to how MathJax (#220) and Lilypond (#435) support might work. It's probably also worth cross-linking my idea for pre-processing raster images with ImageMagick (#436).

akavel:
(Side note: [dpic](https://ece.uwaterloo.ca/~aplevich/dpic/) uses different syntax than TikZ and is not written in Lua, but fills a similar niche (see e.g. page 8 of [dpicdoc.pdf](https://ece.uwaterloo.ca/~aplevich/dpic/dpicdoc.pdf)), is BSD-licensed and written in C. This, or the GNU `pic` installed IIUC on most Linuxes as part of *groff* package, could possibly be used already to render PDF diagrams in SILE based on a language similar in spirit to TikZ.)

baubleb:
I want to add that it is not just graphs with axis and plots that would be useful to be able to do, but other diagramming such as state machines and transition graphs for example. Underneath graphics primatives abstracted up to markup environments for special diagramming.

alerque:
If you have a tool that takes markup in and spits the plots you want out it is pretty trivial to write a SILE package to deal with them. Passing contextual information like the current document font is just a little more work. If you have a language/tool you want to implement feel free to open an issue for it and I'll get you started. We well have to resolve #1092, so it might look like this:

```sile
\script[src=packages/statemachine]
\begin[handler=statemachine]{raw}
% state machine markup here
\end{raw}
```

... or you may need to specify on the CLI what handlers you'll use in advance, something like:

```
$ sile --raw statemachine foo.sil
```

Either way it is quite doable and should be easy to get rolling.

TikZ in the other hand is a special snowflake because it's technically an input language with deep tries to LaTeX itself and no stand alone handler that generates outputs. It expects to reach deep into TeX internals. As such it's not really in the same category as, say, some stand alone tool that converted UML markup to SVG.

[...]

Issue #436, opened 2017-01-20
Add image package that filters raster resources through ImageMagick


alerque:
Including an image in a document is all well and good. Being able to set the size in the layout is all well and good. But this is often the very final step in including an image in a document. First the image has to be manipulated, perhaps cropped or resized, perhaps recolored to fit the context. Then and only then are they placed on a page.

Some desktop publishing programs have non-destructive ways to do basic image manipulation such as cropping and adding borders when inserting an image. They keep the original image internally and process it in place on output. Currently the only way to handle this in SILE is to edit the image externally and include the finished image. This often means doing destructive raster edits on a source image. Sure you can keep this in version control, but that's awkward.

My solution to date has been to keep a directory of raster resources used in a book at the highest quality and original format. Then in each book project's `Makefile` I have target rules to generate versions of those to be included in the document using [ImageMagick](http://imagemagick.org/). Those are then appended to the prerequisites list for the book so they are generated whenever SILE runs.

```make
PREREQUISITES += image.jpg
image.jpg: media/image.png
    magick $< -normalize -crop 600x150+0+0 -resize 400x -bordercolor black -border 5 $@

%.pdf: %.sil $(PREREQUISITES)
    sile $< -o $@
```

Now every time I run `make  my_book.pdf` a cropped, resized, normalized, and bordered version of the image will be generated and I can use it as the source file in my document. The **major advantage** of this method is that the editing is "non destructive". Since the source is untouched and the file will regenerate whenever necessary I can always tweak it with different parameters.

This gets the job done, but it occurred to be that a package could allow SILE to handle this internally. Not everybody is going to build their documents using GNU-Make  and handling this in the source of the document would make it a lot for context sensitive.

I don't know how this should be implemented. An easy hack would be simply to read the file through a pipe and pass some command line arguments along the way. 

```tex
\image[src=image.png,magick="-normalize -crop 600x150+0+0 -resize 400x -bordercolor black -border 5",width=6em]
```

A more complicated way might be to use one of [ImageMagick's APIs](http://www.imagemagick.org/script/api.php) and develop a command that processes the options and handles this image processing internally without having to call out to the command line.

Thoughts?

simoncozens:
Start with the simple way and then try the complex way. :-)

Omikhleia:
Rejected (Wontfix): It does not seem relevant to introduce an ImageMagick dependency of that kind in the core distribution. 
[...]

Issue #435, opened 2017-01-20
Add Lilypond support


alerque:
Add package with commands to embed Lilypond code either inline or as external files that is engraved externally before being included in SILE for layout.

alerque:
Note that this will be conceptually similar to adding MathJax support (#220) in that a block of content will be passed out to some other program to be rendered and the result embedded in the output.

simoncozens:
Can't you do this yourself? ;-) I think there's an example of using the inputfilter package for this kind of thing.

alerque:
Yes I think I can wire this up, I just haven't had the necessary combination of time and motivation. I'm sure the next time I have to fix something in one of my PDF build pipelines that uses LaTeX I'll get in gear and make this happen. Note I'd already self-assigned this issue.

alerque:
Progress report: I have basic systems rendering in a sile document. Cutting through the hack jobs surrounding the TeX implementations of this is crazy hard but this is going to be a lot easier it SILE than it ever was in TeX!

![Lilypond in SILE]()

alerque:
Progress report: now detecting some basic layout parameters like font size, line width, indent, ragged vs. flush formatting, etc. from the SILE context and passing them  to Lilypond.

![selection_030](https://user-images.githubusercontent.com/173595/33514770-c41df55c-d76a-11e7-89f1-9f8c808c0078.png)

alerque:
It's probably worth reviewing the [lyluatex](https://github.com/jperon/lyluatex) project for pointers in implementing this.

Omikhleia:
This issue (and related PR #499 should be moved to [lilypond.sile](https://github.com/sile-typesetter/lilypond.sile) and closed here as "rejected - to be handled in a 3rd-party package]. 
(Noting in passing: I'd be interested in Lilypond support to be resurrected there...)

ctrlcctrlv:
"rejected - to be handled in a 3rd-party package"

PR #675, opened 2019-10-12
Leverage Fluent and Unicode CLDR for localization

alerque:
Early work in relation to #665.


alerque:
This won't start passing CI until I do another release of `fluent-lua` because wrote this against my git HEAD version and tweaked a couple things along the way to make the downstream implementation easier. But the gist of this works already.

[fluent.pdf](https://github.com/sile-typesetter/sile/files/3721085/fluent.pdf)

Noticeably missing at this stage are locale fallbacks for missing messages, but that's coming.

alerque:
@Omikhleia / @OlivierNicole / @simoncozens 

Have any of you by chanced taken a peak at this? I know it's not passing tests yet, the issue with that is what to do when inserting a translation that doesn't exist. The addition of an empty node in the current scheme is changing some tests, but I‌ can fix that.

What I'm really interested in is whether the overall usage / API looks ergonomic as a SILE user or even makes a bean of sense to somebody who hasn't been mucking around waste-deep in localization files for the better part of a couple years.

I have some more SILE stuff coming up that could really benefit to having this merged and I'm interested in finalizing the direction soon.

Omikhleia:
I haven't looked in detail yet, I have to admit.

For the interface of SILE itself (error messages, etc.), it could make some sense, though it's certainly a lot of work to do properly (and not necessarily to maintain when updating/fixing messages).

For strings output to the rendered document, I have to admit that while I admire the effort, I am not much convinced of the general usefulness. I use to say I do not believe much in localized strings in these cases... The only cases currently with the default class and packages (if I am not mistaken) mostly concern the TOC header and the chaptering.
- Regarding the former, if I may quote what I wrote in own revisited TOC package is that
  > such a package should only do one thing well: typesetting the table of contents, period. Any title (if one is even desired) should be left to the sole decision of the user, e.g. explicitly defined with a `\chapter` command or any other appropriate sectioning command, and with whatever additional content one may want in between. Even if LaTeX has a default title for the table of contents, there is no strong reason to do the same. It cannot be general: One could want “Table of Contents”, “Contents”, “Summary”, “Topics”, etc. depending of the type of book. It feels wrong and cumbersome to always get a default title and have to override it, while it is so simple to just add a consistently-styled section above the table…
- As for chapters, parts, tables and figures (which I have in my book class)... My line of thinking is that
  > I have no idea what a “part” would be in your book. “Part I”, are you sure? What if you typeset, say, _The Lord of the Rings_, wouldn’t that rather be “Book I”?

  And a chapter could as well be a "Canto" of typesetting Dante, etc. Likewise, say, for "Figure". Could as well be "Illustration" in some book context, etc.

In other terms, I somewhat feel I will always have to redefine these things in a specific book context, defeating the purpose of a general translation file... And I never liked having to dabble into Babel's file in LaTeX do know what to override.

(As an aside note regarding the more general i18n logic in pattern-based generic tools such as Fluent, I am not sure it addresses in an easy way many language-specific features. Take elision and contraction rules for articles that many Latin-derived language have, and which may depend on grammatical gender or etymology... The rules to get "The {season}, the {tool}" (where _season_ and _tool_ would be the replaceable strings) are quite unpractical. -- e.g. in French, winter = l'hiver (elided), axe = la hache (not elided)... Something one cannot get right without human intervention, or a good bit of NLP if one wants to automate it. Of course, we are hardly in such a case in our context - It's just to say my opinion on such tools is somewhat mitigated.)

alerque:
@Omikhleia Well good morning wet blanket! (Don't worry I'm not offended, I asked for feedback.)

I get some of your points, but I guess I have different needs. I suppose I should rephrase my question to ask whether you think this is actively worse than the current ad-hock localization using redefining SILE commands that output content. It brings 2 more Lua dependencies into the mix, that's one downside — but I don't think it's a blocker.

On the other hand I do think its easier to use than what we have now. Lets say you want to override the default TOC title from Lua:

```lua
-- old way
SILE.registerCommand("tableofcontents:title", function (_, _)
  SILE.process({ "Konu Haritası" }
end)
SILE.call("tableofcontents")

-- new way
SILE.fluent:add_messages("toc-title = Konu Haritası")
SILE.call("tableofcontents")
```

Of course Lua is verbose, but it isn't much different from SILE syntax:

```sile
% old way
\define[command=tableofcontents:title]{Konu Haritası}
\tableofcontents

% new way
\ftl{toc-title = Konu Haritası}
\tableofcontents
```

Yup that's it.

> For the interface of SILE itself (error messages, etc.), it could make some sense, though it's certainly a lot of work to do properly (and not necessarily easy to maintain when updating/fixing messages).

True story. Having this tooling would certainly enable one to localize the SILE interface itself, but that wasn't my goal. I wouldn't object if somebody wants to contribute localizations, but the upkeep on those would be up to the user base. I would note that we've had some extraordinary efforts in this direction already with a full Japanese translation of the manual, but again this wasn't my main objective.

> For strings output to the rendered document, I have to admit that while I admire the effort, I am not much convinced of the general usefulness. I use to say I do not believe much in localized strings in these cases.

I totally see your point here and would be happy to make SILE easy to use as a platform in this mode, but I don't think we're ever going to get away from some people wanting ready to use templates including things like TOC headers. Again the question is is this way actively worse than what we have? And is providing tooling that other packages could use for their own use cases a bad thing rather than making them implement their own content functions or Lua routines?

> And a chapter could as well be a "Canto" if typesetting Dante, etc. Likewise, say, for "Figure". Could as well be "Illustration" in some book context, etc.

Sure. So is providing an easy way to use a templates that includes strings (as opposed to always rolling your own) while providing your own strings a bad thing? My books don't just have a TOC, they also have a TOV (Table of Verses) and a TOR (Table of References) and a TOM (Table of Maps) and a BALROG and so forth and so on. Being able to define all the strings in one place rather than having to have them inline with the content (which is all auto generated and can change order, etc.) is a lot easier for me.

```sile
\begin{ftl}
toc-title = Outline of Topics
tom-title = Maps of the Lands
toi-title = Illustrations and Illuminations
\end{ftl}
```

Lets talk about a real use case: one of the projects I'm working on with SILE is typesetting some workbook material in a whole host of languages. There are a *lot* of boiler plate strings, often longer than a couple of words. Managing the content and the typesetting template separately is the only way that makes sense, having the template strings appear over and over in the content would be hugely repetitive and make managing the actual content harder. Of course I could write a library of commands with localized outputs for each language, but this tooling allows me to write 1 template, then write one localization file for each language, then typeset the content in any language.

Also just in case this was missed, this currently *has to be* in SILE core not a package because enabling raw pass-through commands in the parser (for parsing FTL content in SILE files) not yet supported from a package.

> Take for instance elision and contraction rules for articles that many Latin-derived language have, and which may depend on grammatical gender or etymology... The rules to get "The {season}, the {tool}" (where season and tool would be the replaceable strings) are quite unpractical.

I'm fully aware that localization systems are hard and some languages are much harder to do well than others. That being said I have used a lot of systems and Fluent is *far and away* the most versatile with enough features to make complex replacements possible in natural contexts. Replacements are not just variable expansions but can include meta data about gender or other grammatical forms so that localizers can cover all the possible scenarios. There are a couple of examples on the [Project Fluent](https://projectfluent.org/) site, but those are just the tip of the iceberg. Handily the *code end* doesn't need to understand a given complexity, it is really up to the translator to provide the localization. This helps side step issues like programmers assuming 1 is singular and 2+ is plural where a language might actually have a few/many break at 5 instead. The localization defines this, not the code, so you don't get stuck as often as a translator is some situation where you can't make a natural translation because some coder didn't think about your language's special issues.

Again, do you think it makes SILE worse for providing an advanced (but in my view easy to use) way to localize strings? I'm not saying we have to bake a lot of strings in and we can make the TOC title output optional (for example), but if we do output it do you have an objection to the commands shown above for doing it?

Omikhleia:
> Well good morning wet blanket! (Don't worry I'm not offended, I asked for feedback.)

I feel apologetic, I had no intention to be offending at all :( - I just tried to expressed my concerns that there might not be a general solution to the problem.

Regarding the syntax, 
```
% 1. "Old" command redefinition hook
\define[command=book:chapter:pre:ja]{第}
\define[command=book:chapter:post:ja]{章}
% 2. Fluent way (if I got it correctly !)
\ftl{book-chapter-title-pre = 第}
\ftl{book-chapter-title-post = 章}
%3. My "styling" approach, kind of.
% (eventually - the current progress is not at this point yet and low-level syntax is currently a tad more verbose than that...)
\style:extend[name=sectioning:chapter:number]{\numbering[before=第, after=章]}
```
All do seem more or less equivalent to me. (Well, I could advocate that styling also covers the font selection, which is done currently via another command, and all spacing, skips, alignments etc. with a loose inheritance system).

This said, I don't have strong objections either.

But you make good points too.

>  this currently has to be in SILE core not a package because enabling raw pass-through commands in the parser
> (for parsing FTL content in SILE files) not yet supported from a package.

We might actually need more flexible ways for extending the parser, to be able to have package declare some hooks -  true verbatim, inline scripts with syntax highlighting, inline SVG or any other text-based graphics (e.g. PlantUML and charts), etc. come to mind.

alerque:
> > Well good morning wet blanket! (Don't worry I'm not offended, I asked for feedback.)
> 
> I feel apologetic, I had no intention to be offending at all :( - I just tried to expressed my concerns that there might not be a general solution to the problem.

No no you didn't offend at all, that was good feedback and I appreciate it. Sorry if my pre-coffee joke came off wrong.

> (Well, I could advocate that styling also covers the font selection, which is done currently via another command, and all spacing, skips, alignments etc. with a loose inheritance system).

This is actually a coupling I'm trying to get *away* from. Not that it isn't good for some scenarios, but the mix of content and presentation style is bad new for my projects. I have different people typesetting vs. translating and having Lua commands that have both presentation styles like fonts and spacing and the related content strings is just a disaster. I can code my way out of it with Lua, but adding Fluent libraries to SILE core takes care of that abstraction for me.

> We might actually need more flexible ways for extending the parser

True, but as of yet I haven't figured out how to do it with a one-pass LPEG parser. By the time we get to the package load commands in the document the parsing grammar is already defined and it is too late for a callback to change it. I haven't wanted to go down the multi-pass rabbit hole yet.

> to be able to have package declare some hooks - true verbatim, inline scripts with syntax highlighting, inline SVG or any other text-based graphics (e.g. PlantUML and charts), etc. come to mind.

Yes, but at this point the only way we have to do that is with a whitelist of tags that are processed as raw content instead of SILE syntax. This is already how we deal with `\script` and `\math`, both of which allow full embedded languages.

PlanetUML (I can't find the issue, maybe this was in my head) and SVG (#1091) have come up before, as have MathJax (#220), Lilypond (#435), ImageMagick (#436), and TixZ (#437)

The only alternatives I found to that problem to allow packages to require package loading on the command line before loading the document or do to multi-pass parsing, both of which have huge downsides. See #1092.

core/inputs-texlike.lua
```diff
@@ -7,8 +7,9 @@ local ID = lpeg.C(SILE.parserBits.letter * (SILE.parserBits.letter + SILE.parser
 SILE.inputs.TeXlike.identifier = (ID + lpeg.S":-")^1
 
 SILE.inputs.TeXlike.passthroughCommands = {
+  ftl = true,
```
Omikhleia, line 10: Since these keywords are defined at parsing level, does it mean they are kind of reserved and we cannot afterwards define commands by that name? It does seem its the case (e.g. we cannot seem to `define` a command named math or script and expect it to work correctly... If this is right, it sounds like code smell (if each time we want to define pass-through commands, we make them impossible to exist in input files and be redefined). We are eventually gonna cram the core parser with plenty of path-through commands, no? (if we start adding pass-through this way for tikz, svg, lilypond, etc.)...

alerque: Yes, it does make those commands reserved.

The issue here is that the parser digests a full document into an AST tree before it starts walking through it running commands. Nothing inside the commands (e.g. you redefining `math` in a macro or Lua code can change how the AST was built because it is already built before any processing gets done.

You don't just need to redefine the `math`/`script`/`ftl` commands than run when such a node is hit in the AST, you need to redefine the grammar of the language that parses and builds the AST.

If you have a way to solve that cart/horse problem I'm all ears.

Omikhleia: Indeed. I have commented there on the general parsing issue: https://github.com/sile-typesetter/sile/issues/1092#issuecomment-1065583740 - very rough idea out of my (no so) fresh mind, scrap it if I got it wrong !

alerque: I'm marking this thread resolved only for the purposes of *this PR*, the topic we got on here is very much still in discussion and I dropped [some clips and links in the relevant issue](https://github.com/sile-typesetter/sile/issues/1092#issuecomment-1065826926) so they don't get lost.

alerque:
I completely agree the whitelist of passthrough commands is less than ideal, but I'm also willing to let it stand until such a time as we agree on a better replacement. As such I think I'm going to move ahead with this even knowing we might refactor it later if we change the parser mechanism.

alerque:
When rebasing your feature branch involved 2313 commits and multiple merges from master...

Yes I'm moving a head with this. The current hold up is I discovered my CLDR library's database only has modern languages at the exclusion of historic ones like Latin (which we make extensive use of for things like `lorem` text). Go figure. Once that is complete I'll work through the remaining test differences.

alerque:
Gah, the manual is a language disaster. With localization updates in place it becomes clear we're leaking language settings all over the place. Our code samples set all following content to `und`, chapter 7 leaves the typesetter in Turkish, chapter 8 leaves it in Japanese, etc.

alerque:
In case I forget by the next time I pick this up, the current test failures are because of load paths. We have a lot of things stuffed in the `package.path` for loading lua, but we need to add a loader for FTL files so the path handling is similarly robust. The current code works when installed, or works uninstalled if configured with the install location as the SILE source directory, but not otherwise. Even our CI setup is a case of otherwise.

Review [loadkit](https://github.com/leafo/loadkit) for how to add custom loaders for an extension using the same path handling...

PR #677, opened 2019-10-14
Update cli library

alerque:
3.x update to lua_cliargs breaks SILE. This WIP is a reminder to find and to fix the culprit.


alerque:
This is long since obsoleted, we got this updated in 140ff4dc and it looks like 56c6209 and 56c6209 were probably to work around whatever issues this was raising.

PR #681, opened 2019-10-16
Lint Lua code using Luacheck

alerque:
~~**NOTE:** Please don't merge yet, I meant this to be a draft! Unfortunately GitHub doesn't yet allow PR type conversion to draft mode.~~ *I think I'm done with everything that will go into this PR, as soon as it is green it can go. The remaining lint errors all have separate issues to track them.*

----

The 1236 warnings and 3 errors currently being given might take a while to work through and parse what's meaningful and what's not, but among other things getting rid of unintended global variables will be a good end result.


alerque:
Less that 400 to go, over 2/3 done!

Its kind of scary to me how much code I can mash around *without* tripping up the test suite. I don't think we're effectively testing half the actual code in SILE yet.

Note some stuff getting fixed along the way here like French and Greek hyphenation would *never* have worked right.

alerque:
That was a marathon! Whew.

PR #682, opened 2019-10-16
Drop stdlib usages in favor of Penlight equivalents

alerque:
WIP until after #681 lands, will need to be rebased.


alerque:
The `setfenv()` function seems to come from *std* as well. Will need to look into the three usages of that.

languages/unicode.lua
```diff
@@ -72,14 +72,13 @@ SILE.nodeMakers.unicode = pl.class({
...
-      elseif self.lasttype and (self.thistype and thistype ~= lasttype and not self.isWordType[thistype]) then
+      elseif self.lasttype and self.thistype and thistype ~= self.lasttype and not self.isWordType[thistype] then
         self:makeToken()
         self:addToken(char, item)
       else
         self:letterspace()
         self:addToken(char, item)
       end
-      if not self.isWordType[thistype] then lasttype = chardata[cp] and chardata[cp].linebreak end
       self.lasttype = thistype
```
alerque: @simoncozens If you have any idea what this bit of code was ever supposed to please jump in here. It's obviously been broken for a very long time. I traced it back far enough in history to know my recent refactoring isn't the problem, this has had several kinds of logic issues for a long time. The variable scope is so wrong that Lua would never pass the value anyway, and even if it did the next if statement would obliterate it. Given that none of the tests break no matter what I do to this I don't know that what I did here is actually a correct fix.

alerque: Promoted to issue #699.

alerque:
I think this is going to have to split into 2 parts. One can land any time, but the second part I think should land _after_ the next release. For the next release I'm really anxious to shake out the release and installation process and I want everybody to be eager to upgrade to it. If it works, their documents should keep working.

Unfortunately completely dropping `std` is going to cause some breakage to some people's stuff. Notable its object inheritance stuff was used for classes. This probably was never a good idea, but that's neither here nor there. The point is that if people have written their own classes, swapping out that invisible machinery will change how their classes need to be written.

I don't want to even think about how to write a transition phase that deprecates the non-Lua behavior that people might have just copied into their own classes until the next release is out the door.

alerque:
I think I figured out how to gracefully deprecate the document classing inheritance, but I'll be saving that for another round of torture. This one won't quite live up to its name in that it won't drop stdlib entirely yet.

simoncozens:
Hooray! I shall take a good look at this and hope to be able to merge it as soon as I understand it! :-)

alerque:
@simoncozens Good luck, this is a monster. I'll talk you though anything that comes up. There is more work to be done along these lines, but this gets the ball rolling and unblocks most other work that I didn't want to tackle while I had the core all torn apart.

The most important user facing changes in this batch are to nodes and measurements. Nodes have cleaner interfaces and a lot more smarts.

* Make nodes from the factory by calling them as functions: `SILE.nodefactory.glue()`
* Never check `node.type` yourself, use methods like `node:isGlue()` instead
* The spec can always be blank for zero whatever. It can also take anything that can be parsed into a length (see next bit).

Even bigger is the change to measurements.

* **NOTHING should internally pass around raw Lua numbers when a distance of any kind is meant!** Everything should be a proper instance of the desired type.
* There are two kinds of distances: `SILE.length()` creates a length with stretch, shrink, and other special properties. `SILE.measurement()` creates a lower level object with an amount and a unit. Lengths are compose of 3 measurements (length, stretch, shrink).
* Instantiate both with any interface you prefer:
    ```lua
    SILE.length("16pt plus 1em minus 6en") -- 16pt plus 1em minus 6en
    local dim = SILE.measurement("5%fw")
    SILE.length(dim, "6pt", 7) -- 5%fw plus 6pt minus 7pt
    SILE.length({ length = 6, stretch = 2 }) -- 6pt plus 2pt
    SILE.measurement("2.5em") * 2 -- 5em
    ```
* SILE commands that accept user input should accept numbers, strings, or tables and they should cast to the desired object type as soon as possible.

alerque:
Looks like I must have used some Lua 5.3 specific syntax somewhere, I'll root that out. I tried to be careful but...

alerque:
I just backtracked this PR to about 1/3 through the changeset. For background see [Gitter chat](https://gitter.im/simoncozens/sile?at=5e0f45da4e9ea12051a9fa65) but the short version is there is a serious speed regression in the latter bit.

I think the PR is still ready to review, although it will start with a much smaller set of changes to start with. I'll still be working on the other half and hopefully have something workable early next week. The new interfaces are _so_ much nicer to work with, but the speed hit from instantiating new objects on each math operation is pretty big.

classes/plain.lua
```diff
@@ -174,22 +184,16 @@ SILE.registerCommand("hbox", function (_, content)
       depth = d,
       value = recentContribution,
       outputYourself = function (self, typesetter, line)
-        -- Yuck!
-        if typesetter.frame:writingDirection() == "RTL" then
-          typesetter.frame:advanceWritingDirection(self:scaledWidth(line))
-        end
+        local _post = _rtl_pre_post(self, typesetter, line)
```
simoncozens: This assignment is (conditionally!) being used for its side effect, which is arguably more ugly than the original method! Just a note to remind to try to think of a way to make this clearer.

spec/break_spec.lua
```diff
@@ -1,6 +1,78 @@
...
+  glue({ width = SILE.length.new({ length = 2.20215, stretch = 1.10107, shrink = 0.73404 }) })
+  nnode({ text ="whole", height =  7.56836, depth = 0.14647, width = 24.93652 })
+  glue({ width = SILE.length.new({ length = 2.20215, stretch = 1.10107, shrink = 0.73404 }) })
+  nnode({ text ="of", height =  7.56836, depth = 0.14647, width = 8.13965 })
+  glue({ width = SILE.length.new({ length = 2.20215, stretch = 1.10107, shrink = 0.73404 }) })
+  nnode({ text ="her", height =  7.56836, depth = 0.14647, width = 14.09668 })
+  glue({ width = SILE.length.new({ length = 2.20215, stretch = 1.10107, shrink = 0.73404 }) })
+  nnode({ text ="sex.", height =  4.6875, depth = 0.1953, width = 15.6543 })
+
+  it("should sleauth the right break point", function()
```
simoncozens: Should what?

classes/plain.lua
alerque: I'm not sure I would call this a side effect. It's stashing the result of an operation in a callback to that it can be checked later without redoing the operation. Perhaps better variable names would work. I actually tried 3 or 4 things before I settled on this, it really wasn't clear to me what was happening at first, so I was probably cross-eyed by the time I came up with this.

spec/break_spec.lua
alerque: Minus the extraneous 'a':

> [sleuth](https://www.dictionary.com/browse/sleuth)
>
> &nbsp;&nbsp;verb (used with or without object)
> &nbsp;&nbsp;&nbsp;&nbsp;to track or trail, as a detective.

Living long term in a non-English speaking country and speaking a second non-English language in the home has ruined my English spelling.

alerque: I tacked this fix onto a marginally related commit in #754.

PR #688, opened 2019-10-17
chore(tests): Not all systems support grep -P, use Perl instead.

simoncozens:
Rewrite regression handler to examine the expectations file itself
rather than shelling out to (possibly unsupported) grep magic.

### DIFF
--- a/tests/regressions.pl
+++ b/tests/regressions.pl
@@ -14,8 +14,9 @@ for (@specifics ? @specifics : <tests/*.sil>) {
 	my $actual = $_; $actual =~ s/\.sil$/\.actual/;
   my ($unsupported, $knownbad);
   if (-f $expectation) {
-		# Entirely skip tests designed for an OS that is not us
-		if (system("head -n1 $_ | grep -q -P -v 'OS=(?!$^O)'")) {
+        open my $exp, $expectation or die $!;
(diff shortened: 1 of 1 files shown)


tests/regressions.pl
```diff
@@ -14,8 +14,9 @@
 	my $actual = $_; $actual =~ s/\.sil$/\.actual/;
   my ($unsupported, $knownbad);
   if (-f $expectation) {
-		# Entirely skip tests designed for an OS that is not us
-		if (system("head -n1 $_ | grep -q -P -v 'OS=(?!$^O)'")) {
+        open my $exp, $expectation or die $!;
```
rjmunro: Indentation looks wrong - it's gone from tabs to 4 spaces.

alerque: Oops, I should have caught that.

@simoncozens I don't care for `perl` and you are the expert, tabs or spaces? 2 or 4? Also do you have an EditorConfig extension installed in your editor of choice?

simoncozens: 4 spaces is normal for Perl. And yes, I have Editorconfig now.

alerque: Cool beans. This code review comment is tackled in #697.

Issue #606, opened 2019-05-01
Normalize Lua code formatting


alerque:
I can't find a code beautifier that formats Lua the way I personally prefer, but any kind of normalization is going to be better than none. After reviewing about half a dozen options the two easiest to implement and most complete seem to be [lua_code_formatter](https://github.com/martin-eden/lua_code_formatter) and [lua-fmt](https://github.com/trixnz/lua-fmt), while [luaformatter](https://github.com/LuaDevelopmentTools/luaformatter) is interesting in that it is a let less aggressive but still works alright (with honorable mention to [FormatLua](https://github.com/denglf/FormatLua)). We should probably start assuming one of those is our go-to preference for code formatting and run it over most of our existing code base.

I can't find any really comprehensive Lua coding style guides. If anybody knows of one (especially if they like it) let me know.

Worth mentioning, the auto-indent for Lua in NeoVim does not match any of these, but I actually perfer it. None of these tools are customization without forking them.

Here is a little shell script to make it easier to play with various Lua code formatters. It should be used as a filter that takes STDIN and writes STDOUT, and can be called with a single argument of which backend to use.

```bash
#!/usr/bin/env zsh

case $1 in
	3|formatter)
		luaformatter -s 2 -t 0 -d unix /dev/stdin
		;;
	2|reformat)
		lua.reformat /dev/stdin >(cat) 1>/dev/null
		;;
	1|fmt|*)
		luafmt -i 2 --stdin
		;;
esac
```

pkulchenko:
@alerque, I've implemented Lua re-indenter in ZeroBrane Studio and it satisfies my (picky) requirements quite well. Note that it won't change the line content, only the indentation. You'll need to trigger the re-indentation from the IDE, but if you want to run it as an external command, then something like this may work: `zbstudio -cfg "package{onAppLoad = function(self)ide:GetMainFrame():ProcessEvent(wx.wxCommandEvent(wx.wxEVT_COMMAND_MENU_SELECTED, ID.REINDENT))ide:GetDocument(ide:GetEditor()):Save()ide:Exit()end}" %1`

alerque:
@pkulchenko Thanks, I'll have a look at your implementation too — when I can get ZBS [to run](https://github.com/pkulchenko/ZeroBraneStudio/issues/995).

alerque:
I did a bunch of format normalization along the way in #681, but it was hardly exhaustive, just basic stuff in files that had lint warning.

alerque:
[...]

PR #692, opened 2019-10-17
Experimental: Lint files with editorconfig-checker

rjmunro:
I fully expect this to not pass CI, but it might point to things to be fixed.

Partly inspired by https://github.com/sile-typesetter/sile/issues/606

### DIFF
--- a/lua-libraries/.editorconfig
+++ b/lua-libraries/.editorconfig
@@ -1,0 +1,1 @@
+root = true
--- a/.github/workflows/editorconfig-checker.yml
+++ b/.github/workflows/editorconfig-checker.yml
@@ -1,0 +1,18 @@
+name: EditorConfig checker
+
+on: [push, pull_request]
+
+jobs:
+
+  editorconfig-checker:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install EditorConfig checker
+        uses: editorconfig-checker/action-editorconfig-checker@d4fca16fc71adef10fbe101903b654449fa9570c
+
+      - name: Check all files
+        run: editorconfig-checker
+


alerque:
Thanks for this, we'll get it added sometime.

Note that I'm moving some of the CI work to GitHub Actions, in particular the Lua linting (#681) and possibly the `busted` tests. Also the commitlint things look like they will work better there. This might be another one better suited somewhere other than the middle of our Travis mega-jobs.

.travis.yml
```diff
@@ -22,6 +22,7 @@ addons:
    - libexpat1-dev
    - libipc-run-perl
    - poppler-utils
+   - nodejs
```
alerque, line 25: Since Travis no longer supports OSS projects, this will need to be a GitHub Action or something else. I'm still happy to support linting against our editorconfig setup, just not in Travis. Since I use `package.json`'s `devDependencies` for commitlint stuff anyway this could also be added there. I bet there is a ready made GH Action too.

alerque (changes requested):

.github/workflows/luacheck.yml
```diff
@@ -11,3 +11,11 @@ jobs:
         uses: actions/checkout@v3
       - name: Luacheck
         uses: lunarmodules/luacheck@v0
+
```
alerque, line 14: Lets move this to it's own workflow, it shouldn't be part of the `luacheck` workflow it should be a standalone one (you can copy the rest of Luacheck workflow as boilerplate to start off with).

.github/workflows/luacheck.yml
```diff
@@ -11,3 +11,11 @@ jobs:
...
+      - name: Install EditorConfig Lint
+        uses: editorconfig-checker/action-editorconfig-checker@d4fca16fc71adef10fbe101903b654449fa9570c
```
alerque, line 16: Why the commit hash? Don't they have a usable version tag? Preferably a GHA style major version tracking tag?

.github/workflows/luacheck.yml
```diff
@@ -11,3 +11,11 @@ jobs:
...
+      - name: Check all files for linting issues (not fatally)
+        run: editorconfig-checker || true
```
alerque, line 18: This needs to fail if it finds problems. We can mark the entire CI job as not required, but we do want it to throw the right status code.

.github/workflows/luacheck.yml
```diff
@@ -11,3 +11,11 @@ jobs:
...
+      - name: Check for linting issues in edited files only
+        run: git diff --cached --name-only --diff-filter=d origin/master | xargs --no-run-if-empty editorconfig-checker
```
alerque, line 20: I could see this being useful, but my experience it it then means that lint issues show up in unrelated PRs because they touch a file and get show a bunch of errors. I think I'd rather just fix everything project wide or add the necessary ignores, then do a fill project check all the time.

lua-libraries/.editorconfig
```diff
@@ -0,0 +1 @@
+root = true
```
alerque, line 1: Is there not another way to configure ignoring a directory as part of the run? Having run this locally and looked at the output a bit I think there are going to be quite a few files we want this linter to ignore entirely.

alerque (changes requested):
Thanks for the contribution! This is looking much more promising now. Some thoughts in comments...

lua-libraries/.editorconfig
rjmunro: You can pass a regex to the `editorconfig-checker` command.

```text
  -exclude string
    	a regex which files should be excluded from checking - needs to be a valid regular expression
```

But to exclude a whole directory I think this is better because it also applies to other tools.

.github/workflows/luacheck.yml
rjmunro: They didn't seem to. `v1` was radically different to `main` which is what their documentation said to use.

lua-libraries/.editorconfig
alerque: Hmm, I can see the use in that.

Omikhleia:
This 5+ yer old PR never came to fruition as-is.
SILE now has other linting, styling and in-code documentation strategies, so it's unlikely the above will make it as-is.

alerque:
I tried to revive this when I was working on `stylua` in #1931 and once more just a couple weeks ago, but I agree it is a no-go: `editoryconfig-linter` itself is just too touchy. It doesn't support any of the language specific scenarios where the settings *cannot* reasonably be applied, and trying to add exceptions for it is too verbose for too little gain. The language specific linters (rust-fmt, stylua) used directly are much more useful in practice.

PR #700, opened 2019-10-18
Tidy up variable scope in languages/unicode.lua

### LINKED ISSUE #699: Logic bug in languages/unicode
I have no idea what this bit of code was ever supposed to do. It's obviously been broken for a very long time.

https://github.com/sile-typesetter/sile/blob/69cfffb48898b26f65d786411f930244306f386f/languages/unicode.lua#L72-L80

I traced it back far enough in history to know my recent refactoring isn't the problem, this has had several kinds of logic issues for a long time. The variable scope is so wrong that Lua would never pass the value anyway, and even if it did the next if statement would obliterate it. I've hacked in a guess to keep the linter happy in #682 (143a3cc), but there is probably still a logic bug here. If not we need some way to test this to prove it breaks when the code is changed.

simoncozens:
Fixes #699


languages/unicode.lua
```diff
@@ -66,17 +67,15 @@ SILE.nodeMakers.unicode = SILE.nodeMakers.base {
...
-      self:makeToken()
+    elseif self.lasttype and (thistype and thistype ~= self.lasttype and not self.isWordType[thistype]) then
       self:addToken(char, item)
     else
       self:letterspace()
       self:addToken(char, item)
     end
-    if not self.isWordType[thistype] then lasttype = chardata[cp] and chardata[cp].linebreak end
+    if not self.isWordType[thistype] then self.lasttype = chardata[cp] and chardata[cp].linebreak end
     self.lasttype = thistype
```
alerque: This line will still clobber any possibility of the previous if statement doing anything at all.

simoncozens: You're right. It's superfluous.

alerque:
I've made an attempt to modernize this PR with a merge from master to resolve conflicts, but since I never had a good understanding at the time of what was being fixed or why and there were no tests, I'm not 100% confident in my conflict resolutions. It no longer passes current tests and at first blush suggest either this change was wrong or my merge is wrong.

Either way I don't know where to go next with this.

alerque:
Okay the tests not passing was my fault, I reversed a line in the merge conflict resolution. With that fixed I'm more confident the merge is good and this represents the original patch intent. And we're back to passing tests.

My original complaint that we don't have a test for what this actually *does* still stands.

alerque:
As I understand it the reason this doesn't make a difference in or out to our tests is that it would only matter in the event of a subclassed tokenizer that needed the info, and that isn't something we have currently implemented. I'm tired of resolving merge conflicts on this and the code smell is a lot less on this than on the commented/broken version, I'm just going to put it through.

ctrlcctrlv:
Hooray! Very Fred-pilled move ;-)

PR #701, opened 2019-10-18
Add deprecation notices to retired length functions

alerque:
I've wanted to do this since literally the first day I installed SILE.

* `length.new(spec)` → `length(spec)`
* `length.make(input)` → `length(input)`
* `length.parse(str)` → `length(str)`
* `length.zero` → `length()` or `length(0)`
* `length:<method>()` unchanged

We're doing type checking on the input anyway so this won't be overhead to parse different input types (table, number, string, etc.), and Lua is smart enough to understand the difference between _calling_ and referencing. Call to create, thereafter you have a length instance.

Since lots of documents and user code rely on these, it will be important to follow a multi-stage deprecation procedure, possibly not even throwing the warning (as current) on the old function access in the first iteration.

~~(This will be more manageable to review after #682 lands!)~~

~~This will only be relevant after #751 lands.~~

1. [x] enable the deprecation notices.
1. [x] replace any remaining deprecated usages in core left after other refactoring.
~~1. [ ] schedule when the warning will become an error.~~
~~1. [ ] schedule when the functions can be dropped.~~


alerque:
Update: the actual change in functionality is being baked into #682 but with the deprecation notice skipped. This PR will need to be rebased to just enable the deprecation notices.

alerque:
No I don't know why this started failing on Lua 5.1.

PR #703, opened 2019-10-25
Fix French input by allowing things that are Lua special characters

### LINKED ISSUE #702: Unfinished capture for French language
Having special characters in a French language document including `^$()%.[]*+-?` will throw this error:

```
Error detected:
        ./core/base-shaper.lua:84: languages/fr.lua:26: unfinished capture
```

It is because of the following code in `languages/fr.lua`:

```
return string.match(SILE.settings.get("languages.fr.highpunctuation"), text)
```

This method requires the `text` variable to be escaped on special characters. I quickly tested this code before the preceding call on a simple document with some special characters and it worked:

```
if text == "(" or text == ")" or text == "?" or text == "." then
  text = "%" .. text
end
```

I don't know a proper way to fix it so I prefer to open an issue rather than making a PR.

alerque:
Fixes #702.


alerque:
This adds another test for French when we already have bug-601.sil, but it tests something different and it's very complicated to keep track of what each test is supposed to test for. In fact I think we will need another one for the guillemots issue that still isn't working as expected.

alerque:
Whew. Note to self and anybody listening, never try to update tests with a dirty working tree. Even that little thing in some file you think isn't related needs to be `git stash`ed out of the way before testing.

PR #705, opened 2019-10-25
Handle space around guillemots

### LINKED ISSUE #704: Handle guillemots properly in French
As turned up in #702 (and should have turned up in #621 but didn't) we're still not handling guillemots correctly in French texts.

* [ ] Characters `«`, `‹`, `›`, and `»` should all get thin non-breaking spaces before and after them respectively. Adding the latter two to the `fr.highpuctuation` setting would work for the trailing ones but the whole system needs to be adapted for the leading ones.
* [x] Figure out what to do with non-Unicode variants in input, `<<`, `<`, `>`, and `>>`. My thought is to ignore them, but I don't type French either so I don't know how practical that is.
* [ ] Add or adapt a test to show input with and without manual spaces (similar to 621) but that also catches the marks at a line break so we can prove the breaks aren't happening at the wrong place.

alerque:
Request for comment, eventually meant to fix #704. This implementation is starting to look unwieldy and perhaps resource intensive. Also it will mean copying a lot of boilerplate code logic to other languages with similar issues. Even though it sort of works I'm to set it aside for now until we think of something better or I'm convinced this is the way to keep going.

Needs to handle the case where ICU _does not_ think there is a break anyway, such as `word»`. For some reason the break is showing up in `«word` but not `word»`.

Also needs test coverage.

### DIFF
--- a/languages/fr.lua
+++ b/languages/fr.lua
@@ -13,14 +13,23 @@ SILE.settings.declare({
 SILE.settings.declare({
     name = "languages.fr.highpunctuation",
     type = "string",
-    default = "?:;!",
-    help = "A list of punctuation marks which should be preceded by a punctuationspace"
+    default = "?:;!»›",
+    help = "A list of punctuation marks which should be preceded by a punctuation space"
   })
 
+SILE.settings.declare({
+    name = "languages.fr.highprepunctuation",
+    type = "string",
+    default = "«‹",
+    help = "A list of punctuation marks which should be followed by a punctuation space"
+  })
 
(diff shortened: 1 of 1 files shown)


Omikhleia:
For the record, I commented on #704, providing rules and definitions - Current proposal  here cannot work well (but I've to admit French punctuation typography rules are hard... Or seeing it on a positive side : it's good a good test case for SILE to pass through fire ;) -- Let's be honest, even the first versions of the "babel" package for LaTeX failed short, and this was some decades ago.

alerque:
This was completely superseded by #1221.

Issue #711, opened 2019-11-05
Failed to build on macOS


hajimehoshi:
```
$ make
find: lua_modules: No such file or directory
find: lua_modules: No such file or directory
touch: illegal option -- d
usage:
touch [-A [-][[hh]mm]SS] [-acfhm] [-r file] [-t [[CC]YY]MMDDhhmm[.SS]] file ...
make: *** [.lastcommit] Error 1
```

I'm not familiar with Lua system. I am not sure what I am missing.

simoncozens:
It looks like we need the GNU core utils to be installed (and in the path) for the `touch` command to work. One workaround is to run `brew install coreutils` and add them to your path:

```
PATH="$(brew --prefix)/opt/coreutils/libexec/gnubin:$PATH"
```

Perhaps a better approach would be to rewrite the offneding Linuxisms in something more portable: Perl, or perhaps even Lua itself.

alerque:
Did you run `./bootstrap` and `./configure` before this? Are you sure both of those returned no errors? Simon is right I was using a GNU specific option to touch (I'll just remove that and use a POSIX compliant one instead) but there is another problem before that I think.

hajimehoshi:
> Did you run ./bootstrap and ./configure before this? Are you sure both of those returned no errors? 

Yes, and they didn't have errors.

alerque:
I just fixed the issue with touch by reading the man page for BSD's version. These arguments work fine on GNU as well, they are just a bit less convenient to write. I don't see any problem with just fixing the utilities to work on either POSIX or GNU versions by using a lowest common denominator approach, in this case we're not going to need anything too fancy.

That being  said I am curious whether we're anywhere close to working on BSD make. I kind of doubt it. @simoncozens I think you've mentioned before some of my fancy-pants stuff only worked when you ran it in `gmake`. I may need to dig up a system to test on to fix that -- I wonder if I can get Github Actions or Travis to run an OSX check for us...

hajimehoshi:
The building started with #712 (thank you!), but I found another errors:

```
Making all in src
  CCLD     justenoughicu.la
ld: library not found for -licui18n
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [justenoughicu.la] Error 1
make: *** [all-recursive] Error 1
```

I think this is due to environment variables. Let me think...

alerque:
[...]

PR #712, opened 2019-11-05
Use only POSIX compatable shell arguments

alerque:
Helps #711.

I don't actually have a non-GNU system to test this on, so @hajimehoshi I would love feedback on whether this actually works for you.


hajimehoshi:
Yes, I'll test this asap :)

PR #720, opened 2019-11-08
Fix issue number

burki:


alerque:
Thanks for the fix @burki.

PR #726, opened 2019-11-30
Fix justification of alternates

### LINKED ISSUE #724: Justification alternates system is broken
Demos worked purely by accident. There is a problem with the line length computation when the alternates are considered - by that stage we are looking at the next line, not the current one. I need this for some Arabic magic, so I'm working on it.

alerque:
Fixes #724.

Opening PR because I don't know what the status is beyond what the commit messages say, but its a lot easier to track with a PR than just a branch. For example this way I can get notifications when commits are added.

### DIFF
--- a/core/break.lua
+++ b/core/break.lua
@@ -92,6 +92,7 @@ end
 
 function lineBreak:tryBreak() -- 855
   local pi, breakType
+  if debugging then SU.debug("break", "-> tryBreak") end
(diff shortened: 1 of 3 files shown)


simoncozens:
Thanks. I think this is more or less on the right track but I am working on more tests.

Omikhleia:

PR #728, opened 2019-12-02
Fix and test parallel package

### LINKED ISSUE #727: Parallel package tries to do arithmetic on function
In spite of tests passing, something recently has broken `packages/parallel` on master. Looking at the code it's not clear to me how it ever worked, sometimes `mark` is a number and sometimes it's a function. Besides fixing the bug, we should probably have a test that actually fails if things don't work as expected.

See also #457.

alerque:
Fixes #727.

### DIFF
--- a/packages/parallel.lua
+++ b/packages/parallel.lua
@@ -68,15 +68,14 @@ local setupParallel = function (klass, options)
   else
     folioOrder = options.folios -- As usual we trust the user knows what they're doing
   end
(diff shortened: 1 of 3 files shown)


alerque:
Buggers. The `parallel` class is a bit hard to test because the output order is non-deterministic. It is using a Lua loop to iterate over a list of frames, and that doesn't always happen in a predictable order because ... Lua tables are wierd.

PR #729, opened 2019-12-04
Support for OpenType JSTF table

simoncozens:
This is an experimental package which uses hints within the font to improve the justification algorithm. See https://docs.microsoft.com/en-us/typography/opentype/spec/jstf for more about the JSTF table.


packages/bidi.lua
```diff
@@ -128,7 +128,7 @@ local reorder = function (n, self)
           reverse_each_node(nl[i].postbreak[j].nodes)
         end
       end
-    elseif nl[i]:isAlternative() and levels[i].level %2 ~= base_level then
+    elseif nl[i]:is_alternative and levels[i].level %2 ~= base_level then
```
alerque: The `:isAlternative()` method should still work, of you need a faster test `.is_alternative` should be a preset boolean.

simoncozens:
This is now failing in *fascinating* ways.

simoncozens:
The fascinating failure was an interesting problem. Because bidi was syntactically invalid, `core/plain` didn't load (nor any class based on it). But the error was reported in `SILE.require`. This isn't properly propagating errors:

```
    local status, lib = pcall(require, std.io.catfile(pathprefix, dependency))
    if status then return lib end
```

alerque:
> But the error was reported in `SILE.require`.

The error _was_ in `SILE.require` ... or rather it was encountered while trying to run that command. If the Lua parser borks we don't really know anything further about what happened other than that the Lua loader didn't give us anything back while we were trying to require something.

I guess we do need to check for the difference between a "file not found" type of error and a parsing error. You might pop open an new issue to look at parsing error types there.

simoncozens:
I seem to have thoroughly messed something up here. I'm going to split this PR into smaller bits, have one focused on reading the JSTF table, then fix the alternates thing separately.

alerque:
You don't need to close a PR to do that. Just start a second branch based on this one, then rebase this one to drop irrelevant commits, then rebase the new branch on the pruned one to keep the bits relevant to that branch.

Issue #620, opened 2019-05-22
Frame overflows too common


alerque:
I think something might be wrong with our default penalties. At the very least we need to throw more warnings when we overflow the frame, but out of the box there are a surprising number of cases where lines just run into the margin and that's undesirable.

One case is the test for `tests/bug-269.sil`. We're overflowing the frame at the top of the second page and there is no reason in the world the default settings with these words and languages shouldn't be able to do better than that. Does this have to do with the language switch or is it something to do with that line having come from pushback when the footnote insertions happened?

![image](https://user-images.githubusercontent.com/173595/58108507-a27f5000-7bf4-11e9-9245-acc6ff38db19.png)

----

Another one I ran into is `tests/bug-301.sil`. In this case insertions are not involved and neither is pushback. Admittedly the line is tight and might not actually squish enough, but why didn't we get hyphenation at least?

![image](https://user-images.githubusercontent.com/173595/58149494-e06b8b00-7c6b-11e9-8090-c4e78024c235.png)

alerque:
I might have found at least one cause of this. While working on replacing stdlib with Penlight in #682 I started to fail some tests that looked good to me. In fact some instances of the problem show here were actually working _correctly_ in the version where I'd refactored the object model. Working backwards it looks like there were some insidious bugs in the stdlib code (both caused by stdlib itself and in the usage) causing side effects.

```lua
local ObjA = ObjB + ObjC
```

A reasonably correct value was achieved for ObjA, but with a side effect of the values of ObjB also being changed. Most of the time this didn't matter since only ObjA was used in the end, but sometimes (for example in `pagebuilder.findBestBreak()` it would occasionally decide not to use ObjA and go back to using ObjB — which now has bad properties.

Proper object models and methods (addition of objects returning a new object rather than returning a modified original) will at least mitigate this. I don't know that all instances of this bug were due to that though, so this should be reviewed after that work is complete to see if it is still an issue.

alerque:
Fixing the side effect math bugs did fix a number of cases that looked like iterations of this bug, but overall the measurement system refactoring has not made the bug go away. If anything it has made worse. But it has also made it more apparent what's happening.

[...]

PR #742, opened 2019-12-31
Fix margin and shrink factor bugs

alerque:
This is stuff I've turned up and fixed while working on my #682 and #751. I'm trying to get that finished up but it's _really_ hard to tell what I've broken when there are unrelated bug fixes baked in. I'm going to split existing commits and move the ones that are bug fixes here so the changes to the test suite are more explainable.

This will have some effect on #620.


alerque:
This "bug fix" does something, but I can't make out if it is actually a fix or just moving the milestone on a bug or maybe even bogus. In any case it looks like it should be evaluated in light of #620 again.

PR #743, opened 2019-12-31
Tweak test tooling

alerque:
These are some minor tweaks to the test facilities that I've stuck in other feature branches I'm working on, but they are really unrelated and not having them in _all_ my feature branches is making it hard to know when the tests are broken and when I'm trying to use a feature that's not in the test tooling. This should go through soon, and I'll merge into or rebase my feature branches for easier development.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -161,6 +161,19 @@ clean-benchmarks:
 	git worktree prune
 	git branch --list | grep benchmark- | xargs git branch -D ||:
 
+clean-recursive: clean-tests
+
+.PHONY: clean-tests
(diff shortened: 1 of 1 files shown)


alerque:
I documented most of these changes [in the wiki](https://github.com/sile-typesetter/sile/wiki/Test-Suite). There is a bit more in here like `make compare` for visually comparing test output from any arbitrary commits, but for the moment it is pretty hard coded to my setup (*zathura* and *pdf-diff* on *nix). I'll have to figure out how to generalize that for others to use before documenting it will be useful.

PR #751, opened 2020-01-03
Refactor nodes and lengths

alerque:
This is a follow on to #682 with the second half of what I was working on for there. It's split up so I can think about how to fix the performance compromise while the other changes are reviewed.

The speed hit here isn't directly related it stdlib→Penlight, it's a matter of introducing objects at all in a place than was using numbers. The math logic is a lot prettier this way and it helps to root out a lot of bugs, but it did come with a performance penalty.


alerque:
Speed is back up to muster. In fact it's now 25% faster! The optimizations are pretty rudimentary, we could do a lot more. A lot of the time right now is being taken up with type checking just to make sure we have sane inputs. If I drop those checks and just assume we're passing the right stuff at the right places it is even faster, but I don't think that's a good idea. Maybe with a `--dangerous-fast` mode flag we could skip some of that gunk, but in most cases it's helpful to be warned when you did something wrong.

alerque:
Another 10-15% back. That's probably all from me for a couple days. Sill a long way to go but it's definitely going to be possible to get the speed back withouth compromising too much on the interface.

alerque:
The fix for 5.1 just now also removed a whole bevy of un-necessary object instantiations (redundant action in `SILE.length` that was being done in `SILE.measurement` anyway). That alone chopped another 45% of the speed hit back off. Combined with the various other incremental improvements the speed hit for this refactoring has come down from 260% when I first got it all working to about 40%.

That's still a lot, but –given that **this is holding up almost all other development work** and I'm still convinced that universally handling lengths as special purpose objects (with some exceptions purely for optimization of heavily iterated loops) is much more comprehensible, flexible, less error prone, etc. than the mix of lengths and number values we had before– my vote is to go ahead with merging this now and keep iterating on performance.

alerque:
Buggers. I got Lua 5.1 working locally but somehow broke 5.2 again on Travis! I'll fix that up. You can still review function on 5.3 in the mean time.

alerque:
It turns out, [proper support in Lua 5.1 is going to be harder than I thought](https://stackoverflow.com/q/59664403/313192). At least given what I can see for now we're going to have to require a little bit of obnoxious gymnastics to compare types.

alerque:
...and Lua 5.1 is back on board. The compatibility is kind of ugly in a few places and exactly what I was trying to get away from. But it's all in an isolated commit that we can revert or work backwards from if we ever come up with a more elegant way to shim 5.1 or drop support altogether (note LuaJIT would not be affected even thought it is based on 5.1).

PR #752, opened 2020-01-03
chore(ci): Setup Github Action to run commitlint

alerque:
The way to do this on Travis is too complicated (see #617), I think this will be simpler.


alerque:
Pending [upstream issue](https://github.com/wagoid/commitlint-github-action/issues/21) or figuring out what the correct usage is...

PR #779, opened 2020-01-16
fix(build): Allow building against Lua 5.2 without shim for <= 5.1

alerque:


alerque:
Never mind this was a false alarm. It's true we don't need to test for this module on newer Luas, but the way the test works doesn't actually look for the rock files directly, it just looks to see if it can be loaded in Lua (via `lua -e "require ('$1')"`. In newer Lua versions loading this a noop but returns true. This is, in fact, a better test because it will also cover the case of LuaJIT (which is 5.1 but has a built in module for this).

PR #784, opened 2020-01-18
Docker touchups

### LINKED ISSUE #782: Docker missing ICU lib
In spite of the commit explicitly asking ICU to stick around it's gone in the `master` Docker builds. It works in `v0.10.0` and `latest` which used the old build process from the external repository.

### LINKED ISSUE #783: Docker version string not set right
The version detection isn't working right in the Docker `master` builds, which identify SILE as `vUNKNOWN`.

alerque:
Closes #782.
Closes #783.


alerque:
Docker is a little hard to test because the images form CI test builds don't seem to be cached so you can't actually use and test the results (at least now without a separate test phase), and running it locally is different for several reasons. I think I got it now though....

PR #785, opened 2020-01-18
Fix Coveralls test coverage reporting

alerque:
Coveralls is not getting any data for some reason. It gets pinged after the test run but it doesn't get the payload correctly. This is an exploration into why and hopefully a fix.


alerque:
Travis time limits exceded. Our test suite has grown and running coverage on on everything is slower than mud.

simoncozens:
Would a parallel test runner be helpful?

alerque:
@simoncozens The test runner *already is* using 4 make job threads to run coverage tests in parallel. It does help a lot but not enough. I tested it locally and the `luacov` stuff just slows Lua down an obscene amount.

Some ideas:

1. Not use Travis, we might get more millage out of Actions or another CI runner for this.
1. Not mix and match compiling SILE and setting up the image with running coverage tests, that is eating into out time allowance. If we could use one of the ready-made states from a previous job ...
1. Split up the coverage run into a couple jobs, each with only a subset of the test suite. Coveralls aggregates results anyway, so we don't have to worry about that end.

alerque:
I have no idea why Travis started hanging and returning no output for 10 minutes at a point in a script I didn't touch (in the middle of the LuaRocks `configure` script!):

image

alerque:
Code coverage stats are back baby.

Only some small Travises were harmed in the making of this PR.

PR #796, opened 2020-01-27
ci(cirrus): Add CI testing on FreeBSD with system luarocks

simoncozens:


simoncozens:
Next steps:
* Write up CI script in the form of installation instructions.
* Get similar thing working without system luarocks (and document).
* Put a nice badge on the home page...

alerque:
Someday it would be interesting to also do a `make dist`, then extract the tarball and try installing that, but we should probably do that for _all_ platforms and with a different set of CI stages so lets not muck with that in this PR.

alerque:
Not quite sure on the task and script names, but badge will be something like this:

```markdown
https://img.shields.io/cirrus/github/sile-typesetter/sile?label=Cirrus%20Build%20Status&logo=freebsd&script=main&task=test
```
https://img.shields.io/cirrus/github/sile-typesetter/sile?label=Cirrus%20Build%20Status&logo=freebsd&script=main&task=test

.cirrus.yml
```diff
@@ -0,0 +1,21 @@
...
+    - luarocks-5.2 install cassowary
+    - luarocks-5.2 install compat53
+    - luarocks-5.2 install lua-zlib
+    - luarocks-5.2 install lua_cliargs
+    - luarocks-5.2 install luarepl
+    - luarocks-5.2 install penlight
+    - luarocks-5.2 install stdlib
+    - luarocks-5.2 install vstruct
+    - sh bootstrap.sh
+    - ./configure CFLAGS="-I/usr/local/include" LDFLAGS="-L/usr/local/lib" LUA_LIB="-llua-5.2" --with-system-luarocks
```
alerque: Are these flags and lib path _really_ required? If pkg-config is working properly and this is the default version of Lua on the system the bootstrap / configure process should take care of setting those.

simoncozens: Good point, everything should be coming in through packages, rather than /usr/local/. I will try again, and squish it if it works.

alerque: It didn't like that and I'm out of my depth. Is `pkg-config` even installed on the system properly? It like like this might be a Lua packaging bug on the FreeBSD side of things, in which case I guess it's okay for us to include the minimum configuration to work around it.

simoncozens: OK, it looks like it does put Lua stuff in `/usr/local/` - which *should* be fine. I wonder if the problem is in `m4/ax_lua.m4`. This explicitly tests for Lua include files in various directories underneath `/usr/local/include/`, but when it comes to test for Lua libs, it just looks in "LIBS, LIBRARY_PATH, and LUA_LIB" and I guess `/usr/local/` is not included there. So... we just need the `LUA_LIB` directory?

alerque: That sounds reasonable. As for as those macros are concerned we should be looking _everywhere_ it might be on _any_ system (in some sort of sane order of course).

simoncozens: Oh, now libpng is failing, so we also need the /usr/local/ include path. Apparently this has been a FreeBSD annoyance ["for literally decades"](https://wiki.freebsd.org/WarnerLosh/UsrLocal) ("Other systems don't require adding -I/usr/local/include for packages that are installed. Ours is an odd-man out in requiring it.")

PR #801, opened 2020-01-27
Add more Pandoc goodies

alerque:
In working on a SILE Writer for Pandoc ([discussion](https://github.com/jgm/pandoc/issues/6087), [PR](https://github.com/jgm/pandoc/pull/6088)) I keep bumping into things to tweak from the SILE side to match Pondoc's output. This is a running list of those tweaks, to be periodically added to `master` (so Pandoc testers can easily use it from Docker's `siletypesetter/sile:master` image), and at the very least merged when the writer is upstreamed into Pandoc master (and released to a stable build version before Pandoc support lands).

### DIFF
--- a/packages/pandoc.lua
+++ b/packages/pandoc.lua
@@ -13,32 +13,48 @@ local handlePandocArgs = function (options)
     SU.debug("pandoc", "Set ID on tag")
   end
   if options.lang then
-    SU.debug("pandoc", "Set lang tag: "..options.lang)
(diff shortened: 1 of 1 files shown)


alerque:
Merging "early" because the current SILE `master` branch does not match my current Pandoc fork output, and this branch does. I have a project going to press _now_, and the latest Pandoc + this matching module work together for it fine, older ones do not. I can build the project locally but I want it to build in CI before going to press.

PR #805, opened 2020-01-30
Add configure option for LuaJIT

### LINKED ISSUE #536: use luajit if it is present

alerque:
Closes #536.


alerque:
Automatic selection is not going to work well. Some have luajit but not necessary all our other Luarock deps available for it.There are too many distros that handle this too differently. I'm going to default to newest plain Lua detected plus add `--with-luajit` flag to force that route.

simoncozens:
Agree. luajit is "special" in sufficiently many ways I think it should be treated as a special case, not a default.

ctrlcctrlv:
This idea is awesome. Do you guys want help testing it, or is that not the main blocker?

alerque:
Thanks for the offer @ctrlcctrlv, but at this point testing it isn't the main hold up. I'm ditching the previous finicky attempt to detect and pick between multiple Lua installations in favor of a config flag that just forces one or the other. That switch isn't complete yet so this doesn't work at all yet.

alerque:
Status report: As of now I can build `--with-luajit` or `--without-luajit` (default) and it will find and pick the newest Lua version of the respective class. This process will need a bit of refining for other distros (I'm on Arch) because LuaJIT install locations are not very standardized (even by comparison with Lua, which itself is a fiasco because of the way multiple versions are kept around).

The resulting `sile` is easily 2× faster for things like building the manual (18s vs. 9s on my current system). It will output a PDF too, but *some* functions in the libtexpdf API are broken and hence some PDFS are broken.

I'll work on cleaning this up and making any changes than need to go into *libtexpdf* (or rather *justenoughlibtexpdf*) to make it fully compatible. Next week...

alerque:
Drum roll please.... :drum: :drum:

Note for anybody wanting to test this, earlier fixes to the build system make GNU Automake's program prefix/suffix system work, so you can install multiple versions of SILE on the same system:

```console
$ ./configure --with-luajit --program-suffix=-luajit
```

This can also be handy if you need multiple versions or want to keep a stable version on your system but still be able to mess around with at installed version from GIT or something like that:

```console
$ ./configure --program-suffix=-stable
```

Issue #772, opened 2020-01-13
Keep build without tests from downloading all fonts


alerque:
Building from source tarball includes some batteries. For example there is a manual built to match the source. Lua modules are included already, as is the git libtexpdf submodule.

Building from git is a little more involved. Sure it has to fetch the submodule and so forth, but it also has to go get fonts to build the manual. This isn't too bad except that it downloads all the fonts needed for the test suite for just building the manual, which is obnoxious when doing upgrades from packages that track git head but don't cache builds.

We should probably split the deps from `testprep` into the things they are actually required for.

alerque:
I started trying to fix this, and it is way more buggered than I thought. The manual uses quite a few fonts that aren't even in our test suite, so it is a whole different kettle of fish to do this properly. I'm going to sit on it a while longer while I stew on the pros and cons of various packaging options.

alerque:
Related issues: #853, #854.

PR #818, opened 2020-02-03
fix(build): Fix version detection in sparse git checkouts (#803)

simoncozens:
Previous fix got us into the case statement but didn't provide any useful information once we got there. This provides *something*, at least.

### DIFF
--- a/build-aux/git-version-gen
+++ b/build-aux/git-version-gen
@@ -152,7 +152,8 @@ then
 # derive a version string.
 elif test "`git log -1 --pretty=format:x . 2>/dev/null`" = x \
     && v=`git describe --tags --abbrev=7 --match="$prefix*" HEAD 2>/dev/null \
-          || git describe --tags --abbrev=7 HEAD 2>/dev/null` \
+          || git describe --tags --abbrev=7 HEAD 2>/dev/null \
+          || git log -1 --pretty=format:'v0-HEAD-%h' 2>/dev/null` \
     && v=`printf '%s\n' "$v" | sed "$tag_sed_script"` \
     && case $v in
          $prefix[0-9]*) ;;


build-aux/git-version-gen
```diff
@@ -152,7 +152,8 @@ then
 # derive a version string.
 elif test "`git log -1 --pretty=format:x . 2>/dev/null`" = x \
     && v=`git describe --tags --abbrev=7 --match="$prefix*" HEAD 2>/dev/null \
-          || git describe --tags --abbrev=7 HEAD 2>/dev/null` \
+          || git describe --tags --abbrev=7 HEAD 2>/dev/null \
+          || git log -1 --pretty=format:'v0-HEAD-%h' 2>/dev/null` \
```
alerque: This won't work on some kinds of checkouts. I'll think about this one a bit...

rjmunro: `git describe` has an `always` flag that returns a bare SHA if there is nothing else available:

```bash
$ git describe
fatal: No names found, cannot describe anything.
$ git describe --always
12d7610
```

rjmunro: Also you may be able do a `git fetch --depth 100` or some number that is likely to be more than the time since the last release to try to get the tag.

alerque: Doing a fetch in the build system is problematic because some environments don't have access to the credentials or network access for it. Also I'm not sure `--always` is what we want here because this is _supposed_ to fail if we aren't sure what we are.

alerque: Okay I've done a bit of poking and `--always` does break what this is supposed to do in some scenarios, and ding a `git fetch` from inside our build system is out of the question, but this change as presented doesn't do any _harm_. It's slightly more information about a broken build environment than we would have otherwise. I still think the real solution is to make sure git based builds are never shallow clones, but this is all we can do from inside the SILE source.

rjmunro:
@alerque
> Doing a fetch in the build system is problematic because some environments don't have access to the credentials or network access for it.

It's a public project. No credentials are needed (as long as you have used the http url to clone it out).

Some CI environments don't have network access? That sounds very odd - how do they get the code in the first place? Also you could use a `||` in that case...

alerque:
> It's a public project. No credentials are needed (as long as you have used the http url to clone it).

@rjmunro That's not the issue. Some build systems insist that builds work offline. The build bot downloads the sources, puts them inside a chroot or container, then executes the build scripts. More and more build systems are insisting that the process work offline as a way of ensuring isolation, than only the approved checksummed code is being used, etc.

We currently fall short of this in another area: the test suite tries to download fonts that are not in the source tarball (see #772) but it is possible to build offline. Doing a `git fetch` (even if from a public HTTPS url) as part of the `make` system is not an option.

PR #821, opened 2020-02-04
chore(tooling): Bounce non-GNU make at the door

alerque:
@simoncozens Does this error out with a useful message in BSD make or does it have syntax problems before it processes this far?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -20,7 +20,10 @@ SHELL = bash
 .SECONDEXPANSION:
 .DELETE_ON_ERROR:
 
-ISGNUMAKE=$(shell echo true)
+ALLOWTHISMAKE=$(shell echo true)
+ifneq ($(ALLOWTHISMAKE),true)
+.error This Makefile only works with GNU Make.
+endif
 
 if !SYSTEM_LUAROCKS
 nobase_nodist_pkgdata_DATA = $(shell find lua_modules -type f -print 2>/dev/null ||:)
@@ -274,12 +277,6 @@ compare: $(call compare_prerequisites)
 
 .PHONY: force
 force: ;
-
-.PHONY: gnumake
-gnumake:
-	@[ "$(strip $(ISGNUMAKE))" = "true" ] || { echo GNU make required ; exit 70; }
-
-default $(MAKECMDGOALS): gnumake
 
 .PHONY: update_expecteds
 update_expecteds: $(EXPECTEDS)
--- a/Makefile.am
+++ b/Makefile.am
@@ -20,6 +20,8 @@ SHELL = bash
 .SECONDEXPANSION:
 .DELETE_ON_ERROR:
 
+ISGNUMAKE=$(shell echo true)
+
 if !SYSTEM_LUAROCKS
 nobase_nodist_pkgdata_DATA = $(shell find lua_modules -type f -print 2>/dev/null ||:)
 extra_dist = $(shell find lua_modules_dist -type f -print 2>/dev/null ||:)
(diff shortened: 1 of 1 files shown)


simoncozens:
Neither - doesn't seem to do anything.

alerque:
Hm, I wonder if the test itself is a Gnuism, try this as `make` and `make debug`:

```makefile
.PHONY: default debug

STATUS = $(shell echo gnuish)

default $(MAKECMDGOALS): debug

debug:
	@echo Test ran, explicit goals: $(MAKECMDGOALS)
	@echo Status var set as: $(STATUS)
```

As I understand it non GNU Make should have an empty STATUS variable here, but maybe the MAKECMDGOALS isn't triggering the test. I get:

image

I would expect BSD make to be the same except to never output anything but an empty STATUS, not `gnuish`.

simoncozens:
With BSD make:

```
simon@upo ~ $ make
Test ran, explicit goals:
Status var set as: gnuish
simon@upo ~ $ make debug
make: Circular debug <- debug dependency dropped.
Test ran, explicit goals: debug
Status var set as: gnuish
```

alerque:
So you're saying [this SO answer](https://stackoverflow.com/a/16550052/313192) is no longer valid?

alerque:
>  I added the line `echo "You have GNU make"` to the `gnumake` target

That doesn't sound like you tested the right thing. Try this:

```makefile
MYVAR=$(shell echo foo)

default:
	@echo RESULT: $(MYVAR)
```

If BSD's make tells you `RESULT: foo` then I guess they've fixed that issue and brought it closer to feature parity with GNU-Make, making it harder to detect

alerque:
image

simoncozens:
Oh for goodness sake. Mac OS X make is now (as of I don't know when) gnu make.

alerque:
@simoncozens Now I've forgotten what the original thing was you tried to do that `make` on your system failed for. There was something `gmake` worked when `make` didn't. Can you remind me what that was and shoot me any relevant version differences so I can debug the original problem?

simoncozens:
Turns out to be nothing to do with GNU make, but this is what it was:

```
$ gnumake lua_modules
Makefile:1124: *** unterminated call to function `foreach': missing `)'.  Stop.
```

Still seeing it today. Here's the generated makefile:

```Make
lua_modules: $(LUAMODSPEC) $(shell $(rocksmatch) || echo force)
    if test -e .git; then
        $(LUAROCKS) $(LUAROCKSARGS) install --only-deps $<
    else
        rm -rf lua_modules
        cp -a lua_modules_dist lua_modules
        $(foreach LUA,$(filter-out $(LUA_VERSION),$(SUPPORTED_LUAS)),
        find lua_modules -maxdepth 3 -type d -name "*$(LUA)" -execdir rm -rf {} \;)
    fi
```

alerque:
That `$(foreach ...)` is _not_ un-terminated according to my GNU Make. I wonder if autoconf is actually doing some monkey business, detecting something on the platform, and trying to downgrade syntax even though you don't need it too. That or you have a really old GNU Make or something like that.

I'll try another approach to that specific problem, but its also a pattern I used quite a lot in CaSILE. For example is it just having trouble with the newline or something else. Can you try this with semi-colon instead of line break command separators:

```diff
--- a/Makefile.am
+++ b/Makefile.am
@@ -90,8 +90,7 @@ lua_modules: $(LUAMODSPEC) $(shell $(rocksmatch) || echo force)
    else
        rm -rf lua_modules
        cp -a lua_modules_dist lua_modules
-       $(foreach LUA,$(filter-out $(LUA_VERSION),$(SUPPORTED_LUAS)),
-       find lua_modules -maxdepth 3 -type d -name "*$(LUA)" -execdir rm -rf {} \;)
+       $(foreach LUA,$(filter-out $(LUA_VERSION),$(SUPPORTED_LUAS)),find lua_modules -maxdepth 3 -type d -name "*$(LUA)" -execdir rm -rf {} \; ;)
    fi
 
 $(LUAMODLOCK): lua_modules $(LUAMODSPEC)
```

alerque:
Oh bah. I'm testing from git, this _has to be_ tested from the dist bundle.

simoncozens:
The one gives me:

```
if false; then
bash: -c: line 1: syntax error: unexpected end of file
```

PR #827, opened 2020-02-11
Fix grid frame and floating frame interactions

### LINKED ISSUE #533: Drop cap breaks grid alignment
Barbara Beeton just noticed that in my example of setting a drop cap on a grid, the grid spacing is not respected. Big oops.

### LINKED ISSUE #826: Measurement error (Drop Caps on a grid layout)
From the mailing list:

> when I try to create a floating frame for drop caps on a grid I receive the following error message:
>
>    ...local/Cellar/sile/0.10.1/share/sile/core/measurement.lua:9: attempt to index a nil value (local 'a')
>
> The resulting PDF file is also corrupted.
>
> Here is an example file:
>
> ```tex
> \begin[papersize=a5]{document}
> \script[src=packages/lorem]
> \script[src=packages/grid]
> \script[src=packages/frametricks]
> \makecolumns[columns=2]
> \grid[spacing=15pt]
> 
> \float[rightboundary=5pt,bottomboundary=1ex]{\font[size=36pt]{A}}
> \noindent \lorem[words=50]
> 
> \end{document} 
> ```
> 
> Am I missing something? I am very new to Sile.

alerque:
Fixes #826.
Fixes #533.

### DIFF
--- a/core/baseclass.lua
+++ b/core/baseclass.lua
@@ -181,7 +181,7 @@ SILE.baseClass = std.object {
     self:registerCommands()
     -- Call all stored package init routines
     for i = 1, #(SILE.baseClass.deferredInit) do (SILE.baseClass.deferredInit[i])() end
-    SILE.typesetter:registerPageEndHook(function ()
+    SILE.typesetter:registerHook("pageend", function ()
       if SU.debugging("frames") then
         for _, v in pairs(SILE.frames) do SILE.outputter:debugFrame(v) end
       end
--- a/core/typesetter.lua
+++ b/core/typesetter.lua
@@ -109,8 +109,9 @@ SILE.defaultTypesetter = std.object {
 
   initFrame = function (self, frame)
     self.frame = frame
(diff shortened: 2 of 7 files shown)


alerque:
Hey @simoncozens you might have a gander and this. I'm not sure I'm even conceptually on the right track, but I seem to be getting somewhere. For example is broken in several ways but it gets one thing right that's never been right before: the content in frames after the float are back on track with where the grid should be.

```tex
\script[src=packages/lorem]
\script[src=packages/grid]
\makecolumns[columns=2]
\grid[spacing=15pt]
\grid:debug
\noindent
\float[]{\font[size=36pt]{A}}
\lorem[words=40]
\end{document}
```

image

Omikhleia:
This old PR now has conflicts, and it also seems some of it ("pageend" hooks etc.) went in the software eventually.
Aside node: We'd need to discuss true grid typesetting somewhere, the current solution is indeed unsatisfying.

PR #844, opened 2020-04-11
Revert "refactor(core): Use descriptive function names in pagebuilder"

simoncozens:
This reverts commit 657098ab1409bae2cb9d3fd051f73fcb08cc7d92, which was causing problems with overlapping footnotes and miscalculations of the footnote frame size.

### DIFF
--- a/core/typesetter.lua
+++ b/core/typesetter.lua
@@ -389,7 +389,7 @@ SILE.defaultTypesetter = std.object {
     if SILE.scratch.insertions then SILE.scratch.insertions.thisPage = {} end
     pageNodeList, res = SILE.pagebuilder:findBestBreak({
       vboxlist = self.state.outputQueue,
-      target   = self:getTargetLength(),
+      target   = self:pageTarget(),
       restart  = self.frame.state.pageRestart
     })
     if not pageNodeList then -- No break yet


packages/frametricks.lua
```diff
@@ -174,7 +174,7 @@ SILE.registerCommand("float", function (options, content)
   local hbox = SILE.call("hbox", {}, content)
   table.remove(SILE.typesetter.state.nodes) -- steal it back
   local heightOfPageSoFar = SILE.pagebuilder:collateVboxes(SILE.typesetter.state.outputQueue).height
-  if SILE.length(heightOfPageSoFar + hbox.height - SILE.typesetter:getTargetLength()) > 0 then
+  if SILE.length.make(heightOfPageSoFar + hbox.height - SILE.typesetter:pageTarget()).length > 0 then
```
alerque, line 177: This is a bad revert, you the measurements system was refactored _after_ this, you can't just start using the `.length` property again like this.

simoncozens:
Here's a M(ish)WE showing the problem:

```
\begin[class=cabook,binding=print]{document}
\begin{script}CASILE.metadata = {}\end{script}
\script[src=casile/casile]
\script[src=casile/layout-halfletter]
\define[command=meta:surum]
\define[command=meta:title]
\script[src=packages/lorem]
\imprint

lorem ipsum\footnote{Lausanne Congress on World Evangelism, 1974, Lausanne
  Covenant, section 9.} \lorem[words=30]

\lorem[words=100]X\footnote{\lorem[words=50]} \lorem[words=20]

\lorem[words=130]\footnote{\lorem[words=20]} being a case in point.
\end{document}
```

alerque:
I've stripped this revert down to only revert the ONE line of code that is actually needed to fix the regression. I have no confounded idea why this works. I'm still messing with it looking for a proper fix.

simoncozens:
CaSILE is still using the old names, which may be relevant.

alerque:
<strike>I'm stumped. There is some sort of crazy pants Lua scope issue here that I'm not seeing. Side my side the new function and old shimmed one (that just calls the new one) return the same lengths 99% of the time, but every once in a while the new ones returns 0 and the "old" one (which _should_ just be a pass through to the new one) returns some number that is not a length at all. I `grep`ed the code but can't find where it is being messed with...</strike>

Doh, per chat this is a CaSILE problem. It was copying the SILE interface, but since it changed the shim was not enough to catch this because `self` is different!

PR #861, opened 2020-05-01
Document some facts about \include

OlivierNicole:
I realized today, after spending some time, that a file included with `\include` should have `<sile>` as its root tag. (Or any tag in the case of TeX-like flavour, provided there is one root.)

I propose to document this in the manual.

This commit also adds `.sil` and `.xml` extensions to the included files in the example. I am aware that SILE has mechanisms to detect the format independently of the extension, but as-is, the example suggests that you can write `\include[src=chap1]` when the actual file is `chap1.sil` or `chap1.xml`. I believe that is not the case.

### DIFF
--- a/documentation/sile.sil
+++ b/documentation/sile.sil
@@ -952,18 +952,36 @@ for instance, you may wish to write a thesis like this:
 \begin{verbatim}
 \line
 \\begin[class=thesis]\{document\}
-\\include[src=macros]
-\\include[src=chap1]
-\\include[src=chap2]
-\\include[src=chap3]
+\\include[src=macros.xml]
+\\include[src=chap1.sil]
+\\include[src=chap2.sil]
+\\include[src=chap3.sil]
 \dots\par
-\\include[src=endmatter]
+\\include[src=endmatter.sil]
 \\end\{document\}
 \line
 \end{verbatim}
 
 \code{\\include}s may be nested, in that file A can include file B which includes
 file C.
+
+The contents of an included file should be put in a \code{\\sile} tag (or
+\code{<sile>} if the file is in XML flavour), like so:
+
+\begin{verbatim}
+\line
+\\begin\{sile\}
+
+\\chapter\{A Scandal In Bohemia\}
+
+To Sherlock Holmes she is always \em\{the woman\}.
+
+\\end\{sile\}
+\line
+\end{verbatim}
+
+This is because every file is required to contain a valid XML tree, which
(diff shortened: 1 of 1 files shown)


alerque:
Thanks for this documentation! Proper docs are always welcome.

The magic extension guessing mechanism between `<file>.sil` and `<file>.xml` is being removed, so adding the extensions explicitly is definitely the right move. Both files might exist and we cannot sanely pick or use both. I already removed the command line stuff that was doing this, I don't know if it's gone from `\include` yet or not, but it should be removed.

Note that the same does not apply to `\script[src=file]` because the only valid option there is `file.lua` and the Lua `require()` system takes care of adding that extension. Since there isn't ambiguity it doesn't (and shouldn't) be added explicitly.

The commit message format for this needs to start with `docs(manual): ...`. I can rewrite that for you so this can be merged, but for future reference we follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format to make the release process easier.

alerque:
For some reason Github isn't actually letting me push to your branch even though it shows you enabled that feature. Can you rewrite the commit message as something like this:

> docs(manual): Explain how included XML files must have a &lt;sile&gt; root tag

That way the prefix is right and it fits in the character limit.

documentation/sile.sil
```diff
@@ -952,19 +952,35 @@ for instance, you may wish to write a thesis like this:
...
 
+The contents of an included file should be put in a \code{\\sile} tag (or
+\code{<sile>} if the file is in XML flavour), like so:
+
+\begin{verbatim}
+\line
+\\begin\{sile\}
+  \\chapter\{A Scandal In Bohemia\}
+
+  To Sherlock Holmes she is always \em\{the woman\}.
```
alerque, line 976:
```suggestion
\\chapter\{A Scandal In Bohemia\}

To Sherlock Holmes she is always \em\{the woman\}.
```

Unlike in most programming languages where white-space is discarded and can be used to visually indent something in a block, leading white-space in the TeX-Like markup here can actually have an effect on the output, usually undesired. I would recommend not indenting lines like this.

OlivierNicole:
> For some reason Github isn't actually letting me push to your branch even though it shows you enabled that feature. Can you rewrite the commit message as something like this:
> 
> > docs(manual): Explain how included XML files must have a <sile> root tag
> 
> That way the prefix is right and it fits in the character limit.

I happily would, but I want to point out that included TeX files also need some root tag.

PR #864, opened 2020-05-02
fix(nodes): Fix calling non-existent nodefactory function

OlivierNicole:
Attempting to call a `SILE.nodefactory` function that doesn't exist
would cause a crash with an irrelevant error message:

```
\begin[class=plain]{document}

\script{
  SILE.settings.set("document.parskip", SILE.nodefactory.foobar())
}

\end{document}
```

```
This is SILE v0.10.4.r10-ge9dcf23-dirty
<depr.sil>

! SILE.nodefactory.foobar() was deprecated in SILE v0.10.0. Please use SILE.nodefactory.foobar() instead. nil at depr.sil:3:1: in \script

Error detected:
	...ile/master.local_install/share/sile/core/nodefactory.lua:667: attempt to call a nil value (local 'old_constructor')
```

This makes it crash with the actually helpful message:

```
This is SILE v0.10.4.r10-ge9dcf23-dirty
<depr.sil>

Error detected:
	[string "..."]:2: attempt to call a nil value (field 'foobar')
```

### DIFF
--- a/core/nodefactory.lua
+++ b/core/nodefactory.lua
@@ -662,9 +662,13 @@ end
 
 setmetatable(nodefactory, {
     __index = function (_, prop)
-      SU.deprecated("SILE.nodefactory." .. prop, "SILE.nodefactory." .. prop:match("n?e?w?(.*)"):lower(), "0.10.0")
-      local old_constructor = _deprecated_nodefactory[prop]
-      return string.find(prop, "^new") and old_constructor or old_constructor()
+      if _deprecated_nodefactory[prop] then
+        SU.deprecated("SILE.nodefactory." .. prop, "SILE.nodefactory." .. prop:match("n?e?w?(.*)"):lower(), "0.10.0")
+        local old_constructor = _deprecated_nodefactory[prop]
+        return string.find(prop, "^new") and old_constructor or old_constructor()
+      else
+        SU.error("Attempt to access non-existent SILE.nodefactory." .. prop)
+      end
     end
   })
 


core/nodefactory.lua
```diff
@@ -662,9 +662,11 @@ end
...
 setmetatable(nodefactory, {
     __index = function (_, prop)
-      SU.deprecated("SILE.nodefactory." .. prop, "SILE.nodefactory." .. prop:match("n?e?w?(.*)"):lower(), "0.10.0")
-      local old_constructor = _deprecated_nodefactory[prop]
-      return string.find(prop, "^new") and old_constructor or old_constructor()
+      if _deprecated_nodefactory[prop] then
+        SU.deprecated("SILE.nodefactory." .. prop, "SILE.nodefactory." .. prop:match("n?e?w?(.*)"):lower(), "0.10.0")
+        local old_constructor = _deprecated_nodefactory[prop]
+        return string.find(prop, "^new") and old_constructor or old_constructor()
+      else return nil end
```
alerque, line 669:
```suggestion
      else
        SU.error("Attempt to access non-existent SILE.nodefactory." .. prop)
      end
```

OlivierNicole: Sure. Do you want me to integrate it into the commit?

alerque:
Buggers. Something is really strange about Github right now. It might have to do with your fork being forked form another fork ... but I should be able to push commits to your branch and I can't.

I did just save a commit using the Github UI, but I messed up the commit message (`:)` instead of `):`) and now I can't fix it because I can't amend commits on this PR branch.

Sorry but I may need you to clean that up, either fixing the commit message or just squashing it into your commit. I don't really care which.

OlivierNicole:
Done.

alerque:
Thanks for the contribution and for helping to cleanup my mess ;-)

OlivierNicole:
Happy to help!

Issue #243, opened 2015-12-05
\dotfill spacing inconsistent when only two dots are required


alerque:
If the space requires many dots it works great. It also works if it can't fit any, and the case of one ends up nicely centered. However if space for two dots is allocated, the dots end up off center.

![selection_135](https://cloud.githubusercontent.com/assets/173595/11609459/b1f27678-9b8f-11e5-941c-0d5508c25a46.png)

simoncozens:
Does this look better? I think it's not just a "repetitions==2" bug but an odd/even bug:
![screen shot 2015-12-07 at 13 24 43](https://cloud.githubusercontent.com/assets/106728/11619014/f0ec7e9e-9ce5-11e5-852a-566b70aa86eb.png)

alerque:
Yes that looks better; but it doesn't look great either. Any one of those by themselves looks fine, but in the common use-case of a TOC there will be a bunch of them together as above and the inconsistent start/end points depending on the amount the space stretches looks bad. I'm wondering the prettiest way wouldn't be to calculated the width of the leading and trailing spaces as fixed width (i.e. no stretching allowed). In that case the single dot becomes a special case but everything after that becomes increments of `<fixed space><dot>(<flexible space><dot>)*<fixed space>`.

simoncozens:
Current attempt: 
![screen shot 2016-03-21 at 23 24 15](https://cloud.githubusercontent.com/assets/106728/13918520/25791da6-efbc-11e5-82e1-b1778d600b08.png)

alerque:
I like it and can't think of how I'd improve it. On a case by case basis there might be more optimal settings  but this seems to strike a good balance and looks good out of the box.

PR #874, opened 2020-05-26
Cleanup width calculations in leaders, especially dotfill

### LINKED ISSUE #859: Page numbers not aligned in table of contents in manual v0.10.2
The title says it all, I think. They used to be aligned in previous versions.

alerque:
Fixes #859.

Changes outcome of #243.

This includes both a fix for the line width problem and separately a different way of handling the spacing of dotfills. The previously discussion of dotfills attempted to keep one side of the dots (right) flush. This does slightly improve the looks of TOC where the page numbers happen to also be the same with (e.g. fonts with lining numbers) but it significantly negatively impacts edge cases such as when only 1 or 2 dots fit. Given the stretch/shrink scenario of spaces this had the potential to be very ugly.

This rehashes the solution to that. By not having any shrink and only stretch a lot of the worst cases look a lot better. At the cost of not lining up the end column any more the variation between odd/even (and the overall variance in general) is reduced.

If this new rendering is not agreeable it is possible to fix the buggy case and still use the fixed dot alignment, I just think it looks uglier.


alerque:
Before:

image

After:

image

alerque:
@dajare Any opinion on which leader alignment method is "better"?

dajare:
For me, **definitely** the second. You can't have wavy dot-leaders! (in vertical alignment).

See [this, for example](https://archive.org/details/earlypoemsofwill00morr/page/n13/mode/2up) more or less at random. Your "after" is in that family. (imo, ymmv, etc.) & FWIW!

alerque:
For reference here is what the manual's TOC will look like with this fix and change:

image

There are still some "bad" visual artifacts here, but they are the result of the different font sizes and the font's side-bearing increasing with size. That issue of visual alignment is a can of worms not related to the dotfill at all!

alerque:
@dajare Actually _both_ before and after situations here will result in wavy leaders. The situation is just sligtly less pronounced. The example you linked to is interesting in that it's a horizontal grid. No amount of fiddling with the current leader function will get anywhere close to that result, but it's an interesting option. If that's something desirable we should write a new package for `gridleaders` or something like that. Can you open an issue for that as a feature request? The way that will need to be implemented is just entirely different than what we have now, but the math for a horizontal grid might be very useful in other situations as well and probably a tool we should be providing.

alerque:
Also, *mad respect* for craftsmen that did typesetting like that my hand!

dajare:
> If that's something desirable we should write a new package for `gridleaders` or something like that. Can you open an issue for that as a feature request?

Can do. I'll try to supply some documentation, and get to this soon (before I forget!). (Btw - I thought a series [on this blog](https://dave.autonoma.ca/blog/) might interest you for obvious reasons!)

dajare:
@alerque P.s., also FWIW, Bringhurst [eschews dot leaders in ToCs](https://twitter.com/teropa/status/1028669612618141696), claiming they "force the eye to walk the width of the page like a prisoner being escorted back to its cell." Not sure I agree (the audacity!) — rather, they guide you gently to your destination, ensuring you don't get lost on the way!

dajare:
> ...Can you open an issue for that as a feature request?...

Done! Issue #875

PR #885, opened 2020-06-09
Add tracking setting to shaper that doesn't add nodes

### LINKED ISSUE #883: Adjusting tracking inhibits hyphenation
The current mechanism for adjusting tracking ruins the output for other purposes such as hyphenation.

```sile
\font[family=Brill,size=14pt]
\set[parameter=document.letterspaceglue,value=0.1pt]
Stereotypical characterisation Alkibiades
Stereotypical characterisation Alkibiades
Stereotypical characterisation Alkibiades
Stereotypical characterisation Alkibiades
\end{document}
```

Note hyphenation works fine until we enable the letter spacing glue. I assume the trouble here is that the order of shape->track->hyphenate means that by the time we do look for hyphenation points we have kerning glues between every letter node.

I'm guessing the best way to handle this is to actually mess with the metrics on the shaped letter nodes rather than faking it with glues that we have to sometimes use and sometimes ignore.

alerque:
Closes #883, not by fixing the "bug" *per se* but by providing a method of adjusting the tracking by fiddling with the shaped values rather than stuffing kern nodes between every glyph.

After implementing this, it occurred to be that Harfbuzz might itself have a function for this. It looks like it might ... in which case this fix is itself a hack. Oh well, it's a step.


alerque:
Nope, nope, not quite there.

alerque:
Okay maybe now.

simoncozens:
LGTM!

Issue #854, opened 2020-04-21
Manual requires fonts not part of build system


alerque:
Currently cutting releases requires Noto Sans CJK, Gentium Book Basic, and Scheherazade which are all readily available with open licenses (also Adobe Caslon Pro, which is not, see #853); but our build system doesn't actually fetch them so manuals are not consistent unless the builder knows to get these fonts first.

The build system should know that building docs requires these.

OlivierNicole:
It even seems to cause the build to fail completely for me, presumably because the expected fallback for a Japanese character is Noto Sans CJK JP and I don’t have it. When doing `make docs`:

```
This is SILE v0.10.4.r9-g2c13d78-dirty
<documentation/developers.sil>
<documentation/macros.xml>

! Font family 'Helvetica Neue' not available, falling back to 'Libertinus Serif' at documentation/macros.xml:9:67: in \par
[1] [2] 
! Font family 'Gentium Book Basic' not available, falling back to 'Cantarell' at documentation/macros.xml: in \hbox near <./packages/tableofcontents.lua:78>:6:90: in \process
[0] [1] [2] [3] [4] 
Harfbuzz version: 2.6.4
Shapers enabled: graphite2, ot, fallback
ICU support enabled

Fonts used:
/home/olivier/.local/share/fonts/Garamond-Math.otf:0	Version 2019-02-05 
documentation/Silson-Condensed.otf:0	Version 1.001;PS 001.001;hotconv 1.0.88;makeotf.lib2.5.64775
/usr/share/fonts/TTF/GentiumPlus-R.ttf:0	Version 5.000
/home/olivier/.local/share/fonts/Garamond-Math.otf:0	Version 2019-02-05 
/home/olivier/.local/share/fonts/Garamond-Math.otf:0	Version 2019-02-05 
/usr/share/fonts/cantarell/Cantarell-ExtraBold.otf:0	Version 0.201
/usr/share/fonts/OTF/LibertinusSerif-Semibold.otf:0	Version 6.10 
/usr/share/fonts/OTF/LibertinusSans-Italic.otf:0	Version 6.10 
/usr/share/fonts/TTF/DejaVuSansMono.ttf:0	Version 2.37
/home/olivier/.local/share/fonts/Garamond-Math.otf:0	Version 2019-02-05 
documentation/Silson-Condensed.otf:0	Version 1.001;PS 001.001;hotconv 1.0.88;makeotf.lib2.5.64775
[...]

PR #903, opened 2020-07-02
Build & publish releases in GH Actions

### LINKED ISSUE #759: Use CI to post release artifacts when pushing tagged releases
See #588.

After v0.10.0 the automatically generated change log should be much saner, and with the current state of make rules there is really no reason we can't use CI (maybe Actions instead of Travis) to with a release action for tagged commits that opens a Github release, adds the news, and attaches the matching source tarball and manual.

Later the Windows CI and/or disro packagers could also add attachments for their respective platforms.

Note I'm not even trying to to this for v0.10.0 because the changlog is a mish-mash of styles, the actual build system is going to need tweaking to get clean artifacts, etc.

alerque:
Closes #759.

This work is now at the point where GH Actions can compile SILE and for tagged commits post source builds back to a Github release. The main reason *not* to merge this yet is the Manual currently uses a bunch of fonts that aren't available in the CI environment and wont be until #854 is addressed. Basically we can't enable automatic releases until we are only using freely available fonts for the manual, examples, etc.


alerque:
Something not quite finished: `make dist` requires headers for all supported versions of Lua so that Luarocks can bundle all the dependencies for each Lua version. Compiling only uses one set, but making the source distro actually requires every supported version of Lua.

alerque:
No reason to block on this actually, the manual will be broken in the build (lacking fonts) but it's worth having access to the artifacts sooner rather than later, and even if we have to delete and swap the release builds when we tag it's worth running this to see if it even works.

Note this is testing the _build_ process and making distribution sources, **not** running the test suite. That will be a different setup entirely due to having to have all the different Lua versions and other dependencies in easy reach.

PR #908, opened 2020-07-03
chore(build): Enable parallel build/test runs by default for Make 4.3+

alerque:


alerque:
Old versions of autotools and Travis both hate this --- for different reasons. I'm going to have to rethink this, for now we just don't get to have nice things.

Issue #668, opened 2019-10-04
Test with Lua 5.4 so we're ready when it's released


alerque:
A beta release of Lua 5.4.0 has [been announced](http://lua-users.org/lists/lua-l/2019-10/msg00003.html).

* [x] Add it to our CI workflow so we know what's going on with it (as allow-fail)
* [ ] Fix anything that doesn't run on it properly so there are no hiccups when it's released and people start having it show up as the Lua interpreter on their systems.
* [ ] When we pass out tests on it, start making them required for new development work.

See also #606 for LuaJIT.

alerque:
http://www.lua.org/versions.html#5.4

It's out.

alerque:
This is currently blocking on [lua-luaepnf/issues/4](https://github.com/siffiejoe/lua-luaepnf/issues/4) at the least, possibly more...

alerque:
Luaepnf solved with upstream release.

We're now hung up on [bit32](https://luarocks.org/modules/siffiejoe/bit32).

alerque:
Looks like [bit32 support is in the works](https://github.com/keplerproject/lua-compat-5.3/issues/48).

alerque:
Next up we're going to be stuck on [lua-zlib](https://github.com/brimworks/lua-zlib/issues/49).

alerque:
That looks like the lot of them. Unfortunately lua-zlib isn't well maintained, we make have to fork that to get it up to speed. Also not up to speed with Lua 5.4 but if updating lua-zlib proves too painful we could look into [lzlib](https://luarocks.org/modules/hisham/lzlib).

alerque:
That last PR took care of bit32, the final hold-out is lua-zlib. I've tested it with a fork and gotten it to work, but the upstream needs to release a new rockspec before we can reasonably start requiring it!

PR #909, opened 2020-07-03
Test & build against latest versions of stuff

alerque:
Was going to deal with #668, but I'm putting off the Lua 5.4 stuff that started out here because there are enough _other_ fixes to the CI flows that I want to get it in master now and work on Lua 5.4 as a separate thing.


alerque:
Luarocks revert is because of [this upstream issue](https://github.com/luarocks/luarocks/issues/1204) and can be re-introduced when that is fixed. We'll need it for Lua 5.4 anyway!

PR #920, opened 2020-07-08
Fix up frame naming for jbook & language=main

### LINKED ISSUE #915: \language[main=ja] breaks \showframe
(This is a contrived example. By the time I discovered the problem, I realized frames are not a good solution to my problem. But, this is still a bug.)

```tex
\begin[class=book]{document}
\language[main=ja] % uncomment me to break everything
\begin[first-content-frame=tate1]{pagetemplate}
\frame[id=tate1,
left=left(content),right=right(content),bottom=top(content),next=content,height=40pt]
\end{pagetemplate}
\showframe
よくある質問（英語：FAQ）

よくある質問（英語：FAQ）

よくある質問（英語：FAQ）

楽しい答え（英語：FAQ）

楽しい答え（英語：FAQ）

楽しい答え（英語：FAQ）
\showframe
\end{document}
```

```
This is SILE v0.10.5
<showframe.tex>

Error detected:
	/usr/share/sile/core/libtexpdf-output.lua:151: attempt to index field 'nodes' (a nil value)
stack traceback:
	/usr/share/sile/core/libtexpdf-output.lua:151: in function 'debugFrame'
	/usr/share/sile/packages/frametricks.lua:149: in function '?'
	/usr/share/sile/core/sile.lua:284: in function 'call'
	/usr/share/sile/core/inputs-common.lua:65: in function 'process'
	/usr/share/sile/core/inputs-texlike.lua:164: in function 'process'
	/usr/share/sile/core/sile.lua:237: in function 'readFile'
	/usr/bin/sile:62: in function </usr/bin/sile:62>
	[C]: in function 'xpcall'
	/usr/bin/sile:62: in main chunk
	[C]: in ?
```

alerque:
Closes #915.

I blame @simoncozens who *knew* better when he went ahead with 2565771c.

https://github.com/sile-typesetter/sile/blob/2565771cee1f502cc4a7919c9fc520d3484981a7/core/libtexpdf-output.lua#L71

Someday [his code might actually eat him](https://twitter.com/CalebMaclennan/status/1280812395577905152). It will serve him right if his gravestone has terrible kerning. I'll contribute to a memorial fund to have a better one carved ... but not in the first week or two.

### DIFF
--- a/core/libtexpdf-output.lua
+++ b/core/libtexpdf-output.lua
@@ -148,14 +148,17 @@ SILE.outputters.libtexpdf = {
     self.rule(frame:left(), frame:bottom(), frame:width(), 0.5)
     --self.rule(frame:left() + frame:width()/2 - 5, (frame:top() + frame:bottom())/2+5, 10, 10)
     local stuff = SILE.shaper:createNnodes(frame.id, SILE.font.loadDefaults({}))
-    stuff = stuff[1].nodes[1].value.glyphString -- Horrible hack
-    local buf = {}
-    for i = 1, #stuff do
-      local glyph = stuff[i]
-      buf[#buf+1] = string.char(math.floor(glyph % 2^32 / 2^8))
-      buf[#buf+1] = string.char(glyph % 0x100)
+    local buf = "nil"
+    if stuff[1].nodes then -- Hide horrible hack from certain death
+      stuff = stuff[1].nodes[1].value.glyphString -- Horrible hack
+      buf = {}
+      for i = 1, #stuff do
+        local glyph = stuff[i]
+        buf[#buf+1] = string.char(math.floor(glyph % 2^32 / 2^8))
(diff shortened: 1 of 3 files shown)


alerque:
The patch here is just a band-aid. This bit that guards what glyphs can be sent directly to the PDF string probably should be a utility function, but I don't have a test case for what it's guarding against and why, so I'm leaving that for you @simoncozens. Also I'm not sure why we end up in this situation in the first place, so this is probably evidence of a deeper problem with the Japanese related `class=jbook` / `language=ja` stuff.

simoncozens:
Yeah, I don't think we should get to a situation where we have frames with no IDs. That's weird already. Taking a deeper look now.

simoncozens:
I'd like to fix this a different way. What's happening is that the frame name can't be shaped if the current font doesn't contain Latin characters. I have an alternative fix that resets the font temporarily to Gentium Plus just to do the frame debugging.

PR #922, opened 2020-07-08
Fix debug frame ids

### LINKED ISSUE #915: \language[main=ja] breaks \showframe

simoncozens:
My version of #920.

Closes #915.

### DIFF
--- a/core/libtexpdf-output.lua
+++ b/core/libtexpdf-output.lua
@@ -147,7 +147,8 @@ SILE.outputters.libtexpdf = {
     self.rule(frame:right(), frame:top(), 0.5, frame:height())
     self.rule(frame:left(), frame:bottom(), frame:width(), 0.5)
     --self.rule(frame:left() + frame:width()/2 - 5, (frame:top() + frame:bottom())/2+5, 10, 10)
-    local stuff = SILE.shaper:createNnodes(frame.id, SILE.font.loadDefaults({}))
+    local gentium = SILE.font.loadDefaults({family="Gentium Plus", language="en"})
+    local stuff = SILE.shaper:createNnodes(frame.id, gentium)
     stuff = stuff[1].nodes[1].value.glyphString -- Horrible hack
     local buf = {}
     for i = 1, #stuff do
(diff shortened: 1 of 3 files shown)


simoncozens:
Of course it strikes me that the original failure will still happen in the unhappy case that the frame ID contains no Latin characters.

alerque:
I was just going to say, how is Gentium going to handle the test case — for which it doesn't have glyphs. Maybe we shouldn't be mucking around with the glyph ids there at all. Otherwise this is more to the point than my PR.

simoncozens:
This handles the test case just fine - in test case, the frame ID was "content" all along. That was being mangled when we tried to shape it using Noto CJK JP. It displays fine in Gentium.

It turns out that having non-ASCII frame IDs causes cassowary to have conniptions, so maybe my concern isn't a big deal.

alerque:
> It turns out that having non-ASCII frame IDs causes cassowary to have conniptions

Really? I definitely saw some non-ASCII frame IDs in debugging and they didn't seem to be an issue. If fact the test case for this issue has one:

image


Is this a function of Cassowary not doing any real work on it? We should probably fix wherever the non-ASCII hang up is...

simoncozens:
Those *are* ASCII frame-IDs. They're being mangled by the shaping, which is what this PR fixes.

alerque:
Gah. I missed the fact that you (probably accidentally) *downgraded* the libtexpdf module as part of this PR. Maybe a little too aggressive with the `git add` foo?

simoncozens:
Oops, sorry about that. Serves me right for using the GUI not the command line.

PR #928, opened 2020-07-08
fix(packages): Ruby class should not affect document language

### LINKED ISSUE #926: ruby.lua should not set lang=ja
It seems strange that ruby.lua does this:

https://github.com/sile-typesetter/sile/blob/e5ee4679d2d6830f806183e3c11c34f83fbd972c/packages/ruby.lua#L2

I can't think of a legitimate reason to. Even if we assume ruby is only used for <span title="furigana">振り仮名<sup>🐣</sup></span> (it isn't!), they already should be using a Japanese class, right?

But ruby is often also used for pinyin, bopomofo, etc:

image

simoncozens:
Fixes #926

### DIFF
--- a/packages/ruby.lua
+++ b/packages/ruby.lua
@@ -1,5 +1,5 @@
 -- Japaneese language support defines units which are useful here
-SILE.call("language", { main = "ja" }, {})
+SILE.languageSupport.loadLanguage("ja")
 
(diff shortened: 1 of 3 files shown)


alerque:
The Travis failure is intermittent, it should get better based on #909 as I'm lightening the load on them. I'll rerun the failed job in a few minutes and it should be good.

PR #929, opened 2020-07-09
Support Lua 5.4

### LINKED ISSUE #668: Test with Lua 5.4 so we're ready when it's released

alerque:
This is Lua related stuff I was messing with in #909 but had to put off...

Closes #668.

### DIFF
--- a/sile-dev-1.rockspec
+++ b/sile-dev-1.rockspec
@@ -15,7 +15,7 @@ dependencies = {
   "cosmo == 16.06.04-1",
   "linenoise == 0.9-1",
   "lpeg == 1.0.2-1",
-  "lua-zlib == 1.2-0",
+  "lua-zlib == 1.2-1",
   "lua_cliargs == 2.3-3",
   "luaepnf == 0.3-2",
   "luaexpat == 1.3.3-1",
--- a/sile-dev-1.rockspec
+++ b/sile-dev-1.rockspec
@@ -21,7 +21,7 @@ dependencies = {
   "luaexpat == 1.3.3-1",
   "luafilesystem == 1.8.0-1",
   "luarepl == 0.9-1",
-  "luasec == 0.8.1-1",
+  "luasec == 0.9-1",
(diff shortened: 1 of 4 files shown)


sile-dev-1.rockspec
```diff
@@ -9,11 +9,10 @@ source = {
 }
 dependencies = {
   "lua >= 5.1",
-  "bit32",
```
siffiejoe, line 12: compat53 does not bundle bit32, because bit32 is deprecated in Lua 5.3. The code lives in the same repository, but there are still separate rockspecs for both.

alerque: Thanks. I figured that out after I tried this, but didn't push the change because I didn't actually figure out how to get it loaded.

Lua 5.3 only included `bit32` if/when compiled with 5.2 compatibility (which was the default). 5.4 doesn't seem to have it at all by default ... which leads me to wonder ... it seems this _IS_ still needed in order to write code that works across Lua versions.

alerque:
The BSD build here failed, but that's not related to the PR (it doesn't even touch Lua 5.4). I'm working on that in another PR, hence I'm going to merge this with an admin override for that skipped test.

Please welcome Lua 5.4 support.

P.S. Note that vstruct is deliberately being installed at an old version that does not have an upper limit on the Lua version. Later releases block 5.4 even though they *would* work just fine. This is [reported upstream here](https://github.com/ToxicFrog/vstruct/issues/41). Until that is merged we just need to hold back vstruct. All other Lua modules are the latest and greatest releases.

Issue #865, opened 2020-05-02
Pullquote with several paragraphs shows wrong indentation


OlivierNicole:
Trying to put several paragraphs in a pullquote, the first paragraph came out unindented (which is normal), but so did the second.

![image](https://user-images.githubusercontent.com/14031333/80868272-c6b72e00-8c99-11ea-92cd-72866720da16.png)

(The last line of every paragraph is stretched, but that’s an easy fix.)

I couldn’t find why this is happening. I only noticed that when commenting out [the mark typesetting](https://github.com/sile-typesetter/sile/blob/master/packages/pullquote.lua#L42), the second paragraph *is* indented.

alerque:
My eyes start twitching and I get erratic muscle cramps in my neck whenever I see something that looks like it might turn out to be a `pushBack()` bug...

OlivierNicole:
Would you happen to have any pointers to what a `pushBack()` bug is?

alerque:
@OlivierNicole The `typesetter:pushBack()` function (found in `core/typesetter.lua`) is the logic that tries to handle material that was originally shaped for one frame context but didn't end up fitting and gets pushed into the start of the next frame. The trouble is it has to _unshape_ the contents, drop some things that are relative to the frame (glues that might be margins, etc.) then hope it all gets shaped right in the next frame context. It mostly works, but is notorious for having edge case bugs. We've rewritten it several times and it still isn't very robust.

OlivierNicole:
I see.

PapistPenguin:
Has there been any progress on this issue? How should I fix the last line stretching issue in the meantime?

OlivierNicole:
@PapistPenguin One simply needs to remove [the temporary setting of `parfillskip` to a zero-width glue](https://github.com/sile-typesetter/sile/blob/193fded67edb7d33f65e3ebeb7b571285da5a8cd/packages/pullquote.lua#L39), which is the cause for that stretched line and is unnecessary given the way `pullquote` works.

I'll file in a merge request presently.

alerque:
@OlivierNicole Do you have a working example (stripped down to as small an example as possible) that shows the second paragraph not indenting?

OlivierNicole:
@alerque Sure.
```latex
\begin[class=plain]{document}
\script[src=packages/pullquote]

\begin{pullquote}
You wonder that there are so few followers of the Pythagorean opinion [that the
earth moves] while I am astonished that there have been any up to this day who
[...]

PR #939, opened 2020-07-13
fix(packages): Improve multi-paragraph pullquotes

OlivierNicole:
This removes the stretched lines at the end of every paragraph but the
last, when a `pullquote` (from the `pullquote` package) contains several
paragraphs. It does not fix the bug mentioned in #865.


OlivierNicole:
I can add a test if needed, but I'm not sure how to proceed.

alerque:
Thanks, this is a better default. I'm wondering it it should be explicitly reset to an `hfillglue`, but that probably shouldn't happen unless we process arguments for making it optional. Given there are several other things wonky with the pullquote package and it's short on options, I think it's okay to push this through without new tests and test the package properly when it actually is expected to behave properly.

PR #940, opened 2020-07-13
Escape texlike strings

### LINKED ISSUE #937: How should I escape characters in command parameters ( [] )?
MWE:

```tex
\fbox[width=1zw,_=\]\,sup]{hi}
```

If I set fbox to...
```
function(o,c)
  SU.dump(o)
  print()
  SU.dump(c)
end
```

```lua
This is SILE v0.10.5
<tkana.sil>
{
  _ = "\\",
  width = "1zw"
}
{
  col = 1,
  command = "fbox",
  id = "texlike_command",
  lno = 59,
  options = {
    _ = "\\",
    width = "1zw"
  },
  pos = 3382
}
```

alerque:
Fixes #937 ... and a number of other fragilities with the TeXlike parsing grammar.


alerque:
@ctrlcctrlv I believe this will fix the issue you raised and your (almost pathological) example `\fbox[width=1zw,_="],sup\""]{hi}` code should work. Any chance you can test this before I merge? It's passing our regular test suite plus I added some new tests specially for parsing commands, but it would be nice to have a real-life trial.

ctrlcctrlv:
Thanks for the patch, I'll build SILE later tonight and give it a try! 🙌

Seems like you were already confident enough to merge it, but I actually do need this feature, it's not just academic for me hehe.

alerque:
I've needed it too. An obvious use case is setting up the pullquote package with options for what quote marks to use. I side stepped this in CaSILE at one point because I happened to want the typographic curly quotes, and passing `\pullquote[start=“,end=”]{foo}` is pretty easy. But what if you wanted to pass a regular ASCII  double quote mark. Ya not so fun. Now there is a way.

Not covered yet is escaping a quote mark _not_ in quotes. I wasn't sure about possible unintended side effects of that so we started with the escape just inside quoted strings.

ctrlcctrlv:
Yes. Just wait until you see how I'm using it. Hackiness is something I love, as you know. ;-)

ctrlcctrlv:
@alerque See #947 and [LaTeXmode for SILE](https://github.com/ctrlcctrlv/sile-texmode).

I wish you would have used Rusty strings instead to escape. Perhaps it's not too late?

I explain why at [&sect; Known issues](https://github.com/ctrlcctrlv/sile-texmode#known-issues).

alerque:
@ctrlcctrlv I used a backslash escape because that's how we've done all the other escapes in TeXlike syntax. It did cross my mind that a Lua-like `[[...]]` syntax would be an interesting alternative. I didn't implement it though because I thought further thought & discussion would be necessary.

Ultimately though I don't think the answer is to change the escaping so much as change where you're trying to pass input in the first place. Rather than trying to pass CONTENT in `\cmd[_="CONTENT"]{}` you should just be able to pass it as content `\cmd{CONTENT}`. This is MASSIVELY more flexible, as commands also have an environment format `\begin{cmd}CONTENT\end{cmd}`. This is already used by `\script` for Lua code and you can see another example in my #499 package PR. With this syntax the end delimiter becomes "\end{your_cmd}", which is pretty hefty and unlikely to be a problem for most embedded languages. I think your `\texmode` should be re-implemented to operate on pass-through content as content instead of a big option string.

PR #946, opened 2020-07-14
Install manual in distro-configurable location

### LINKED ISSUE #873: Document Snell Roundhand and all other free fonts in examples
The font Snell Roundhand used in examples/test.sil is readily available on macOS and many Linux distros, but apparently not on Windows. It looks like there are freely available downloads, but we should probably document both that you need to get it and where to get it (and any other fonts used in examples) to avoid future confusions such as came up in #869.

### LINKED ISSUE #919: Documentation isn't installed where automake --docdir asks
We're installing README.md correctly, but the manual should also be installed using `dist_doc_DATA` so that it gets correctly placed per system layouts.

alerque:
Closes #919. Finally.
Closes #873.
Partially addresses #623 (title spacing is fixed, in-column line stretches are not).


alerque:
Mostly this is good, but it has the unwanted side effect of adding the manual to the list of things that gets build when running plain `make`. This is a speed killer because we really don't want to have to re-render the manual all the time. It should be updated for `make dist` and and installed with `make install` for sure, but we want a lighter default build target.

alerque:
Difinitely a massive catch 22. For now I'm putting the manual (and examples) installation behind a configuration flag. This both keeps installs light (frequently desired) and avoids rebuilding the manual on every `make` invocation (which is rarely desired).

```
$ ./configure --with-install-manual
$ ./configure --with-install-examples
```

(Or both of course.)

The actual installation location in wherever automake puts docs for the distro and can be configured using the usual method (changing `$(docsdir)`).

PR #948, opened 2020-07-15
Editing pass on chapters 1-5

simoncozens:


simoncozens:
Hmm, that example is not built in a straightforward way from a .sil file, so I've added a Makefile section. If needed we can generalise it. Please check my Make-fu and see if there's a better way.

alerque:
This build failing has a lot to do with CI stuff I fixed last night. I'm going to rebase this on master to get that fixed up and check the Make-fu while I'm at it. Just a few minutes heads up of a force push .... if you do anything between now and then no problem just push away. I have `git rerere` foo!

alerque:
> Hmm, that example is not built in a straightforward way from a .sil file, so I've added a Makefile section. If needed we can generalize it. Please check my Make-fu and see if there's a better way.

The existing PDF build target is already generalized to cover cases like this. The pattern rule tells make how to run SILE to build from either %.sil or %.xml files, but it passes arguments to SILE in a traditional make-ish way via flag variables. Your can override these variables on a per-target basis without writing a new rule like this (to append a flag):

```make
examples/docbook/article-template.pdf: SILEFLAGS += -I docbook.sil
```

With that in place to add the right flag, the only thing that needed to happen is adding the XML file to the EXAMPLESRCS list.

packages/leaders.lua
```diff
@@ -65,12 +65,12 @@ allow you to define your own leaders. For example:
 
 \begin{verbatim}
 \line
-A \\leaders[width=30pt]\{\\font[features="+ornm"]{iI}\} B
+A \\leaders[width=40pt]\{/\\\} B
```
alerque, line 68: This needs another escaped backslash I believe:

```suggestion
A \\leaders[width=40pt]\{/\\\\\\} B
```

Dizzying, but to show in the output that you backslash is escaped you need two pairs, plus one more to escape the closing brace to get an actual brace.

PR #953, opened 2020-07-16
Remove (most) of the wildcard hacks from automake

### LINKED ISSUE #951: CI‌ generated package artifacts aren't usable like regular dist builds
Not sure what's up, but the Github Actions artifact zip, which should be an archive of `make distdir` is not quite that.

1. It has a ready to run `configure` script but it isn't marked as executable.
1. The file `aclocal-1.15` and some friends are missing, requiring an autoreconf run which shouldn't be necessary.
1. It doesn't have a `lua_modules_dist` directory. This might be desirable actually, but it was unexpected because the dist package should have it.

alerque:
Closes #951.

This is a bit awkward, but I've been thinking about this now and then for years and nothing better has occurred to me. Fighting automake is not fun, using it as intended is actually not too bad. This punts most of the ways we were trying to be too smart for it to a shell script.

The biggest downside is currently you can't skip straight from `./configure` to `make dist`, you have to `make` in between. Working on that bit, maybe...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,6 +17,12 @@ jobs:
             .fonts
             .sources
           key: fonts-${{ hashFiles('Makefile-fonts') }}
+      - name: Cache Lua Rocks
+        uses: actions/cache@v2
+        with:
+          path: |
+            lua_modules_dist
+          key: luarocks-${{ hashFiles('Makefile-luarocks', 'sile-dev-1.rockspec') }}
       - name: Fetch Tags
         run: git fetch --prune --tags
       - name: Install System Dependencies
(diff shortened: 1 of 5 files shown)


alerque:
Okay back to one step from `./configure` to `make dist` (which wasn't that important because only people cutting releases and testing the process are going to hit that issue, but it does make testing less error-prone).

The whole lua_modules_dist thing is frankly a lot of pain for little gain. Realistically all the distros that care about self contained packages that don't need internet connections to build (i.e. don't want to call out to `luarocks install ...` are also the ones that use `--with-system-luarocks` and so don't care about the bundled sources anyway. It might be okay to drop that whole mess someday.

alerque:
For the record the only remaining use of `$(wildcard ...)` is okay I think. It can be allowed to fail with an empty list, it isn't used by automake at all (no packaging/dist related data), and is only an optimization to speed up rebuilds.

PR #963, opened 2020-07-20
ci(actions): Add a linter that covers C-code

alerque:


simoncozens:
I'm not sure that actually checked any C code. :-)

alerque:
It found a couple thousand warnings which I saw go by while it was running, then it all disappeared and I don't know what it did with the output. I don't know if it actually hit the C code or not. Since superlinter doesn't have a Lua ruleset at all and just scans for stuff it knows about I figured it would find it but maybe not. In any case this clearly isn't ready to merge yet ;-)

Issue #964, opened 2020-07-20
Stop distributing LuaRocks stuff


alerque:
I think our Lua Rocks distribution system has outlived its usefulness. New it's just causing trouble.

Current status:

When running `make dist` we fetch and build all the Lua Rocks we use for every supported Lua version (Now 4 versions) and stuff them in a directory (lua_modules_dist). When people build on a system `--without-system-luarocks` the rocks for their version of Lua are extracted to `lua_modules`, a directory which is then installed wholesale with `make install`.

Rational:

This arrangement helped ease the transition from the original method of including the full source copied from  these Lua modules as vendored code in our own repository, but now that other build methods are more refined it's becoming a millstone.
 
First of all, zero distributions use this stuff anyway: Arch, Ubuntu, Homebrew, Nix, Docker — so far every distro *bar none* with SILE packages has opted to use the system method, not the bundling method. 

That leaves the use-case of end users doing a source install on their own. For their purposes I can't think of a good reason why they can't be bothered to fetch the rocks and build them using `luarocks`. Note this is **not** a convenience issue because the build script is still going to do it automatically for them either way, it will just use LuaRocks to fetch and build to the local tree at compile time instead of coping files out of the dist dir. On the flip side I can't think of a single scenario where a system might have a working Lua installation but not a working LuaRocks that we are actually *helping*. In fact the only situations I know of where this has happened are broken anyway because our provided luasocket/luasec doesn't match their distro's openssl version anyway, so they _still_ have to work around that.

Proposed status:

Stop bundling 4 versions of Lua modules source code and prebuilt rocks in our dist packages. Start assuming the make system can run luarocks to fetch and build those things better anyway in the event people want a "bundled" install that doesn't require and Lua rocks on their system (only `luarocks` the executable at build time).

alerque:
@simoncozens If you can think of any objections to this, better speak up quick ;-)

rjmunro:
Do you mean "Stop" rather than "Step" in the title? I though it might be an extra "step" in the build process or something.

alerque:
@rjmunro Yes, I meant stop. :stop_sign:

Issue #961, opened 2020-07-18
Github built tarballs have automake issues


alerque:
For some reason the Github Actions generated release tarballs are missing an automake file. I've posted locally generated ones instead because I've already maee a muck of the last release cycle, but this needs to be reviewed.

alerque:
For future forensic comparison, here is a Github built artifact (to be compared against the one as posted on the release now, which I built locally): [sile-0.10.8.zip](https://github.com/sile-typesetter/sile/files/4942404/sile-0.10.8.zip).

alerque:
This might not be GH Actions specific, just different systems have different issues. The GH built one worked fine on Homebrew and Arch Linux last release cycle (v0.10.7), but failed on Nix. The current (v0.10.8) one built my me locally worked fine on Arch and Ubuntu, but failed on Homebrew (see [upstream pr](https://github.com/Homebrew/homebrew-core/pull/58199)).

PR #969, opened 2020-07-20
fix(build): Ship blank lua_modules install list in source packges

### LINKED ISSUE #951: CI‌ generated package artifacts aren't usable like regular dist builds

### LINKED ISSUE #968: CI build artifacts have lua module file list that may not match target
The makefile shipped from CI‌ artifacts lists files that would be installed if `make install` ran from the CI platform, but with `lua_modules` not being distributed and platforms being different, that file list doesn't match all platforms. The value should be blanked for distributions.

This seems to be a root cause of [homebrew](https://github.com/Homebrew/homebrew-core/pull/58199) and [nix](https://github.com/NixOS/nixpkgs/pull/92152) issues that have been playing hard to get. The automake version warning is a complete red herring. It only runs into that trying to rebuild something it shouldn't need to build because something earlier failed but reported success.

alerque:
Closes #968.

...Is the *actual* fix for #961.

...Has been underlying various issues we've had for a long time, but was made a lot more obvious by the refactor done in #965 for #964 because the data that needed to be cleared was previously mixed in with another list of files that was otherwise valid.

...Started turning up in more scenarios because of the work in #953 to fix #951. This was the right thing to do, but our previous hackery before was keeping the problem from showing up by (accidentally) make reruns where the second pass was fixing what the first one broke. Cleaning that up just make the underlying problem surface in more places.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,9 @@ jobs:
           ./configure --with-manual --with-examples
           echo "::set-env name=DISTDIR::sile-$(./build-aux/git-version-gen .tarball-version)"
           echo "::set-env name=MAKEFLAGS::-j$(nproc) -Otarget"
+      - name: Make
+        run: |
+          make
       - name: Make package
         run: |
           make dist
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ WORKDIR /src
 
 RUN mkdir /pkgdir
 
-RUN git clean -dxf ||:
+RUN git clean -dxf -e .fonts -e .sources ||:
 RUN git fetch --unshallow ||:
 RUN git fetch --tags ||:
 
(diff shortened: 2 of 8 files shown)


alerque:
Right problem, not quite fixed yet...

alerque:
So this PR‌ *passing* tests is a problem. It shows we aren't testing everything we need to be testing. We're building and testing from source, but we're not testing the source package we build for buildability. Automake has `make distcheck` just for this purpose, but ironically it won't actually catch the problem(s) in recent releases. The issues don't crop op building from the same platform that built the source, they only crop up *cross* platform. Ugg.

alerque:
Never mind the remote test runs, my local directories are starting to look like this (this is after blowing them away many times):

image

Automake behaves so differently between versions and platforms its making my head spin. I can hack it to work, but playing by its rulebook is a lot harder and I'm not convinced it can even work the way I want it too. Bah.

Issue #972, opened 2020-07-22
SILE's output is non-deterministic (in at least one case)


ctrlcctrlv:
Finally I found a somewhat short MWE that doesn't require any of my modules.

Consider:

(The text is a nonsensical pseudo-injoke devoid of context.)

## `nondet.sil`

```tex
\begin{document}
\script[src=packages/rebox]
\script[src=packages/raiselower]
\script{
local inputfilter = SILE.require("packages/inputfilter").exports
-- not done elsewhere so it doesn't mess up arguments to other commands
local function dblquotes(input, content) 
    return input:gsub('%"', '”')
end -- "Handle double quotes"
SILE.registerCommand("texmode", function(options, content)
local output = [[\rebox[height=0,width=0,depth=0]{\lower[height=72pt]{{Tinatawag nilá akón͠g ``magandá'' dahil ang akin͠g }{\em{ĉapelo}}{ ay nasa taman͠g lugár.}}}
\rebox[height=0,width=0,depth=0]{\lower[height=72pt - 1.25em]{Sa totoó lang, pan͠g{}unahin͠g kagandahan͠g{-}ásal sa m͠ga titik magsuót n͠g \em{ĉapelaĵoj} ganitó.}}
\rebox[height=0,width=0,depth=0]{\lower[height=72pt - 2.50em]{An͠g m͠ga titik na hindi, sila'y hindi edukado lan͠g pô kasí.}}]]
local tree = SILE.inputs.TeXlike.docToTree('\\begin{document}'..output..'\\end{document}')
SILE.process(inputfilter.transformContent(tree, dblquotes))
end)
local newline = function(_, content)
  SILE.settings.temporarily(function()
    SILE.settings.set("typesetter.parseppattern", "\n")
    SILE.doTexlike(" \n\\skip[height=0em]")
  end)
end

SILE.registerCommand("newline", newline)
SILE.registerCommand("\\", newline)
}
\font[family=Gentium Plus]
ĥ:
\texmode
\end{document}
```

Sometimes the output is:
![image](https://user-images.githubusercontent.com/838783/88129026-1294b780-cba5-11ea-9277-0511518c96ca.png)

Other times:
![image](https://user-images.githubusercontent.com/838783/88129210-8931b500-cba5-11ea-9df5-f284dadee631.png)

Notice how the word ''ĉapelaĵoj" moves about. Nothing changes in the document between invocations.

I do not know how to debug this. Help would be appreciated.

alerque:
[...]

Issue #355, opened 2016-08-08
Drop characters when hyphenating


alerque:
So the feature added at my request in #265 to setup custom hyphenation patterns is great for words that are exceptions to the normal rules, but I'm finding it leaves me hanging when it comes to the actual rules.

I've recently discovered that there is more than one style guide for how to hyphenate words with internal apostrophes in Turkish. My personal preference is to place the hyphenation points after the apostrophe, but this is not what all publishing houses want to see. Many of them require the apostrophe to be _dropped_ in the event that the word gets hyphenated at that point. For example

> müjdesinin

might normally hyphenate as

```
müj-de-si-nin
```

However in the event it is used as a proper noun the suffixes (after the third person possessive) are set apart with an apostrophe:

> Müjdesi’nin

```
Müj-de-si’-nin
```

However in the event that a hyphenation point is actually used, the apostrophe should go away.

> Mesih'in Müjdesi-
> nin tek gerçek

The problem here is two fold.
1. There are a huge number of words where this could be an issue, so adding case by case rules is cumbersome. My current solution for getting hyphens after apostrophes rather than before is to search the full text of my books for _all_ the apostrophized words and manually add them to an exception list with the pre-apostrophe hyphen points removed.
2. The current exception handler does not provide a mechanism for the input and output strings to be different. The input of the function is the hyphenated form and the function assumes the word is the text string with hyphens removed. This won't work to setup a publishing workflow that conforms to the expectation of apostrophes being dropped if used as hyphenation points.

simoncozens:
I'll need to come back to this, but I think the answer will be to replace the apostrophe with a `\discretionary` command like so: `\discretionary{-}{}{'}`. i.e. produce an apostrophe if not hyphenated, and produce a hyphen with no apostrophe if hyphenated. Of course this means I will have to check that the nobreak part of the discretionary works. I can't remember if I implemented that yet.

alerque:
At what stage in the processing would you envision this being implemented? An inputfilter?

khaledhosny:
The [Hyphen](https://github.com/hunspell/hyphen) hyphenation library, supports what it calls “[non-standard hyphenation](https://github.com/hunspell/hyphen/blob/master/doc/tb87nemeth.pdf)” where there are changes applied to the text when it is hyphenated. It might be what you are looking for.

This was requested in #277, but SILE will either need to use Hyphen instead of its own code, or support the enhanced hyphenation patterns.

alerque:
[...]

PR #974, opened 2020-07-22
Fix #532 by sorting table keys when processing content

### LINKED ISSUE #532: Content re-ordered
I've had a freaky problem come up in multiple projects now where _some_ rendering passes will miss-order content. Anything that is wrapped in some markup (font style or weight change, etc.) can get pushed to the end of the line it's supposed to be in. Usually running another rendering pass clears this up. It's not consistent on whether the first pass or the second or the third or whatever will fail like this.

Compare the order of italicized bits:

selection_046

With the source:

```TeX
...
\footnote{Smarto, \em{Pursued}, 122.}
...
\footnote{Anthony A. Hoekema, \em{Saved by Grace} (Grand Rapids, MI: Eerdmans, 1989), 178.}
...
```
Note the `\em{}` tagged content is typeset out of sequence.

simoncozens:
* This could do with a test.
* I tried to find other examples where we are using pairs() on AST content, but didn't see any. A review would be helpful.

### DIFF
--- a/core/utilities.lua
+++ b/core/utilities.lua
@@ -125,6 +125,22 @@ utilities.map = function (func, array)
   return new_array
 end
 
+utilities.sortedpairs = function (t)
+  local keyset={}
+  for k,v in pairs(t) do
+    keyset[#keyset+1]=k
+  end
+  table.sort(keyset, function(a,b)
+    if type(a) ~= type(b) then return false end
+    return a < b
+  end)
+  return coroutine.wrap(function()
+    for i =1,#keyset do
+      coroutine.yield(keyset[i], t[keyset[i]])
+    end
+  end)
+end
+
 utilities.splice = function (array, start, stop, replacement)
   local ptr = start
   local room = stop - start + 1
--- a/core/utilities.lua
+++ b/core/utilities.lua
@@ -274,7 +274,7 @@ end
 -- items — assuming that the current command is taking care of itself
 utilities.subContent = function (content)
   local out = { id="stuff" }
-  for key, val in pairs(content) do
+  for key, val in utilities.sortedpairs(content) do
     if type(key) == "number" then
       out[#out+1] = val
     end
--- a/core/utilities.lua
+++ b/core/utilities.lua
@@ -125,18 +125,19 @@ utilities.map = function (func, array)
   return new_array
 end
 
-utilities.sortedpairs = function (t)
-  local keyset={}
-  for k,v in pairs(t) do
-    keyset[#keyset+1]=k
+-- TODO: Replace with pl.tablex.sort()?
+utilities.sortedpairs = function (input)
+  local keys = {}
+  for k, _ in pairs(input) do
+    keys[#keys+1] = k
   end
-  table.sort(keyset, function(a,b)
+  table.sort(keys, function(a, b)
     if type(a) ~= type(b) then return false end
     return a < b
   end)
   return coroutine.wrap(function()
-    for i =1,#keyset do
-      coroutine.yield(keyset[i], t[keyset[i]])
+    for i = 1, #keys do
+      coroutine.yield(keys[i], input[keys[i]])
     end
   end)
 end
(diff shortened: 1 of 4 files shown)


alerque:
This fixes #972, but not #532. Its likely the same problem but since it doesn't uses inputfilter there is another case somewhere. I'm trying to track it down.

alerque:
I take back what I said, this does seem to fix #532. The project in question there has an elaborate work around for #355 that basically pre-processes everything with input filter fixing potential hyphenation issues at apostrophes. That does explain why it turned up regularly in a few projects and never in others.

As for tests ... I'm not to confident we have a way to reliably test non-determinism. The most likely candidate would be a Busted test unit that iterated over something a bunch of times and make sure it was the same every time. Hmmmm, yes let me try that....

core/utilities.lua
```diff
@@ -125,6 +125,22 @@ utilities.map = function (func, array)
...
+utilities.sortedpairs = function (input)
+  local keys = {}
+  for k, _ in pairs(input) do
+    keys[#keys+1] = k
+  end
+  table.sort(keys, function(a, b)
+    if type(a) ~= type(b) then return false end
+    return a < b
+  end)
+  return coroutine.wrap(function()
```
alerque, line 137: Clever and works, but I'm curious why you reached for coroutines instead of the classic Lua idiom of keeping an iteration counter outside the function.

simoncozens: Well, pairs() is an iterator, and we need our own version - I just thought coroutines were[ the way you do iterators](https://www.lua.org/pil/9.3.html) in Lua. :-) (Turns out you can do it by returning a generator function as well, but truth to tell I just grabbed the structure from `utilities.gtoke`)

alerque: Fair enough.

alerque:
I haven't come up a test that fails using `pairs()` yet.

Here'se what tried (plain in `spec/utilities_spec.lua` and run with `busted -t utilities`):

```lua
SILE = require("core/sile")

describe("The #utilities library ", function()
  it("should have a short alias", function() assert.is.truthy(SU == SILE.utilities) end)

  describe("sortedpairs() function", function()
    local input = { "a", _ = true, "c", "b", foo = "bar", "E" }
    it("avoid Lua’s non-determinism", function ()
      local testit = function ()
        for _ = 1, 100 do
          local i = 1
          -- for k, v in SU.sortedpairs(input) do
          for k, v in pairs(input) do
            if k == 1 and v ~= "a" then return false
            elseif k == 2 and v ~= "c" then return false
            elseif k == 3 and v ~= "b" then return false
            elseif k == 4 and v ~= "E" then return false
            elseif k == "_" and v ~= true then return false
            elseif k == "foo" and v ~= "bar" then return false
            end
            i = i + 1
          end
        end
        return true
      end
      assert.is.truthy(testit())
    end)
  end)

end)
```

alerque:
So I've still completely failed to create a test that reproduces the problem. It's hard to test that you're doing it right when you can't get a test you know in wrong to reliably fail as a starting point.

However I did some digging and came up with a few more instances of `pairs()` that either are or could be problematic. Some of them likely don't matter but technically were out of whack (defining hyphenation definition patterns in order, applying them randomly) others could clearly have tripped us up without warning (`outputYourself()` iterating nodes out of order).

For reference if this comes up again, I found these by first adding a wrapper to `pairs()`:

```lua
debugpairs = function (var)
  if var[1] then
    utilities.error("called pairs() on what looks like a sequential table")
  end
  return pairs(var)
end
```

Then aggressively replaced our use of `pairs()` with `debugpairs()`:

    $ git ls-files '*.lua' | xargs -n1 sed -i -e 's/in pairs/in debugpairs/g'

Then I‌ worked through running tests & building the manual replacing the bad lines with `ipairs()` or `SU.sortedpairs()` as appropriate until it stopped dying.

alerque:
In retrospect we could/should probably have just used `pl.tablex.sort()` for this. :man_shrugging:

alerque:
One I wasn't sure about and you may recognize @simoncozens:

https://github.com/sile-typesetter/sile/blob/175baad18d6f2b1591d2d941c346e7f7ba8b293c/lua-libraries/lomwithpos.lua#L42

Unfortunately this option doesn't seem to be getting tested at all because replacing `pairs()` here with `foo()` doesn't fail our tests at all.

PR #981, opened 2020-07-23
fix(packages): Combine unichar output with existing unshaped node

### LINKED ISSUE #979: unichar package doesn't work with combining characters
Compare output of
```
â
```
And

```tex
a\unichar{0x302}
```

simoncozens:
Fixes #979.

### DIFF
--- a/tests/bug-979.sil
+++ b/tests/bug-979.sil
@@ -1,0 +1,10 @@
+\begin[papersize=a6]{document}
+\nofolios
+\noindent
+\script[src=packages/unichar]
+\font[family=Symbola]
+\raggedright{
+â\break
+a\unichar{0x302}
+}
+\end{document}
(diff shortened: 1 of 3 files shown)


simoncozens:
Fixes #979.

tests/bug-979.sil
```diff
@@ -0,0 +1,8 @@
+\begin{document}
```
alerque, line 1:
```suggestion
\begin[papersize=a6]{document}
\nofolios
\noindent
```

Killing off the folios when not specifically testing frames saves quite a few false positives. Using a small paper size by default makes visually reviewing them easier when mass updates change something and they have to be inspected. A6 is what I've standardized on when papersize isn't important for any other reason. Killing off indentation just make inspection a tad easier.

Issue #986, opened 2020-07-27
What is the right way to set the footnotes font?


ctrlcctrlv:

alerque:
Redefine the command for `\footnote:font`.

```lua
SILE.registerCommand("footnote:font", function (_, content)
  SILE.call("font", { size = "9pt" }, content)
end)
```

ctrlcctrlv:
Hmm, it feels like there should be an option :-)

alerque:
That command (and many similar ones) exist explicitly so there is a place to hook in and set your own styling.

In CaSILE's book class `cabook` I overhauled the namespacing a bit, but the fundamental idea is the same. For example the default footnote font is specified as an inheritance from the "alternate" serif font (which by default just inherits from the default `cabook:font:serif`).

```lua
SILE.registerCommand("cabook:font:footnote", function (options, content)
  options.size = options.size or "0.74em"
  SILE.call("cabook:font:alt", options, content)
end)
```

Because of the way the options are set to be passed up the chain through these functions, you don't have to completely redefine the function to change something (although you *can* still do so). Instead you can just change the default contents of the options table:

```lua
SILE.setCommandDefaults("cabook:font:footnote", { size = "6.5pt", family = "Crimson Pro" })
```

I've been meaning to propose SILE's default classes be overhauled with namespaces and option chaining like this. How does it strike you?

ctrlcctrlv:
Quite well, my use case I think is pretty typical, using EB Garamond for the book; some of my other documents use URW's Palatino. I don't really know of why you'd want footnotes in a different font, so it seems a strange default to be honest.

alerque:
By default they aren't a different font, they are just a different size. I think the problem you are experiencing is that you are changing the activly used font of the current typesetter, but the footnote frame starts a new typesetter and hence reverts back to the default font. Perhaps you should change the default font for the scope of your documents instead, that way headers, footers, folios, footnotes, etc. (anything that starts a new frame or typesetter instance) keeps your font.

From TeXlike input:

```sile
\set[parameter=font.family,value=EB Garamond,makedefault=true]
```

For in Lua:

```lua
-- This is what it *should* be but I just reaziled we never implemented this!
SILE.settings.set("font.family", "EB Garamond", true)

-- This works for now
SILE.settings.defaults["font.family"] = "EB Garamond"
```

ctrlcctrlv:
[...]

PR #987, opened 2020-07-27
Add makedefault and set args to SILE.settings.set() and reset option to \set

alerque:
See comments in #986.

----

We previously added a way to set a setting and make it stick as the
default via the TeXlike command \set:

    \set[parameter=foo,value=bar,makedefault=true]

This is the complement of that for Lua so you don't have to directly
hack the settings array:

   SILE.settings.set("foo", "bar", true)


simoncozens:
Code looks good, but it's going to need docs.

alerque:
Roger that.

alerque:
Regression tests passed on Travis but failed on Cirrus because archive.org blocked the download of Symbola. This is unusual, normally it lets it through fine. I'm going to admin-override and merge this for now, if we keep having trouble with Cirrus I'll look into an alternative (...again).

PR #988, opened 2020-07-27
fix(utilities): Use deterministic sort for sorted pairs

### LINKED ISSUE #985: [inputfilter] SILE's output is non-deterministic (in at least one more case)
#972 and #532 have not been fully solved.

MWE on SILE v0.10.9.r36-gf16af44:

```tex
\begin[class=book]{document}
\script[src=packages/rebox]
\begin{script}
  local inputfilter = require("packages/inputfilter").exports
  local noop = function (input, _) return input end
  SILE.registerCommand("nondet", function(_, content)
    SILE.process(inputfilter.transformContent(content, noop))
  end)
\end{script}
\nondet{
\font[family=EB Garamond,size=12pt]
\define[command=SILE]{S\lower[height=0.5ex]{I}L\glue[width=-.2em]\raise[height=0.6ex]{\font[size=0.8em]{E}}}
\typeset-into[frame=folio]{\begin[size=12pt]{font}\hfill This document was typeset with \SILE.\end{font}}
}
\end{document}
```

Without `inputfilter` noop:
image

With:
image

alerque:
Closes #985.

This is a follow up to #974 which itself used a non-deterministic sort! This makes sure array values come before labeled table values in the sort.

### DIFF
--- a/core/utilities.lua
+++ b/core/utilities.lua
@@ -125,15 +125,16 @@ utilities.map = function (func, array)
   return new_array
 end
 
--- TODO: Replace with pl.tablex.sort()?
 utilities.sortedpairs = function (input)
   local keys = {}
   for k, _ in pairs(input) do
     keys[#keys+1] = k
   end
   table.sort(keys, function(a, b)
-    if type(a) ~= type(b) then return false end
-    return a < b
+    if type(a) == type(b) then return a < b
+    elseif type(a) == "number" then return true
+    else return false
+    end
   end)
   return coroutine.wrap(function()
     for i = 1, #keys do


alerque:
I still don't have a good idea how to test this. Lua's non-deterministic behavior seems to be locked in at run time (some sort of seed?) and no amount of iterating over the same input once the interpreter is spun up seems to change the outcome, we need something that actually starts the Lua interpreter from scratch between invocations. That doesn't fit either Busted or our regression testing model very well.

alerque:
Thanks. It's not quite as simple as mistakenly using `pairs()` when `ipairs()` would do the job. Otherwise we'd just use the right function and move on. The issue is we have data that combines array-like data (integer keys) with named key-value data (string keys) in a single table and need to iterate over both sets of data. Lua allows this data structure but does not have an function to iterate over it, so we had to write one. That's the function we need to test.

It doesn't sound like anybody in that chat understands the interpreter runtime issue at stake. That's okay we'll keep looking.

ctrlcctrlv:
FYI, I did:

```bash
git pull
wget https://github.com/sile-typesetter/sile/pull/988.patch
git apply 988.patch
make install
```

And this has been working well for me all night long!

image

Solved a bunch of issues in tkana.sil, and made the above possible.

In re: `ipairs()`/`pairs()`, you may find this unethical @alerque, but sometimes I simplify for IRC. Sometimes I fabricate alternate simpler scenarios where the details don't matter to the question I'm answering. Sorry if this makes me a bad person.

alerque:
No I don't think simplifying a problem for the sake of looking for somebody that could answer it is unethical at all. It fact it's a very good part of problem solving — sort of like working on an MWE. The more concise you can make the problem the easier it is to come up with a solution. My only objection was in this case the essence of the problem kind of got lost. I think they were right to point out "then they are using the language wrong", because if we were hoping to get ordered iteration out of `pairs()` then were **are** using the language wrong.

I'm *still* failing to come up with a reliable way to test this, and the fix seems to work pretty well so I'm going to merge for not so that at least `master` is usable. Whether we find a way to test or not before the next release this change should clearly be in there.

PR #998, opened 2020-07-30
Orient rules relative to writing & page advance directions

### LINKED ISSUE #723: rules package doesn't respect text advance direction
```TeX
\script[src=packages/rules]
\thisframeRTL

A \hrule[width=200pt,height=5pt] B
\end{document}
```

alerque:
Fixes #723.


alerque:
Hmm, interesting failure here. There are some problems with my test here. The main thing being tested seems okay (rules go the right way in all 4 tested directions) but there are also issues with them in TTB/BTT modes (that are also issues with other non-rule bits, so I didn't think trying to track them down as part of this PR was a good idea.

That being said I did introduce Mongolian with a rule, and that's where Harfbuzz 1.4.x is giving us different metrics than any of the newer versions in the matrix.

See the [diff](https://travis-ci.com/github/sile-typesetter/sile/jobs/366789351#L2009):

```patch
-Draw line	34.8819 238.2135 -10.0000 89.2913
+Draw line	33.0460 238.2135 -9.0820 89.2913
```

Should we ditch the relic of a Harfbuzz and bump our minimum supported for tests? Should I rework the test using a Latin script drawn TTB or something like that?

alerque:
Using Latin scripts in directions they were not meant to go makes the test output really confusing, and Mongolian etc. are problematic for other reasons. I rewrote the test to cover all 8 possible direction combinations using only rules, which made it quite a bit more obvious which ones are behaving and which are not:

* ✔ LTR-TTB
* ✗ LTR-BTT
* ✔ RTL-TTB
* ✗ RTL-BTT
* ✗ TTB-LTR
* ✗ TTB-RTL
* ✔ BTT-LTR
* ✗ BTT-RTL

image

I *think* … it's actually slightly confusing to reason about directions for which there are not actually orthographies to use as a reference and how they would be different than a rotated frame.

alerque:
I think this now does the right thing in all 8 possible directions, at least where `\hrule` is concerned. It's not clear to me that SILE is handling any of the BTT orientations correctly is regard to which direction ascender/decenders are pointing, but `\hrule` is currently doing the same thing as far as calculating line height as regular text would. Whether that's correct or not might depend on the actual language and script in use, which is deliberately not part of this test.

* ✔ LTR-TTB
* ✔ LTR-BTT
* ✔ RTL-TTB
* ✔ RTL-BTT
* ✔ TTB-LTR
* ✔ TTB-RTL
* ✔ BTT-LTR
* ✔ BTT-RTL

image

alerque:
This PR includes a major overhaul to the outputters including changing all the arguments to include `self` as the first argument (i.e. always call as `SILE.outputter:method()` not `SILE.outputter.method()`). This gives more flexibility for when some implementations want access to their own other functions and others don't. It also renames some of the functions for consistency and generally brings them a step closer to being able to class them.

@simoncozens You should probably sanity check this if you have a second.

PR #999, opened 2020-07-30
Use cluacov instead of just luacov for coverage tests

alerque:
In my local testing this is 4× to 5× faster generating coverage stats over our busted and regression tests. That should help out Travis just a bit.


alerque:
Somewhere north of 2×, close to 3× faster on Travis. I'll take it.

Issue #994, opened 2020-07-29
Character variants... cvXX . What am I doing wrong?


ctrlcctrlv:
I can't seem to figure out how to enable Character Variants, which are like Stylistic Sets, but different.

Right now, they aren't mentioned in `packages/features.lua`.

I tried really hard, but am sadly more confused than when I started.

HarfBuzz has a format for creating an `hb_feature_t` from a string: [[click]](https://fossies.org/linux/harfbuzz/docs/html/harfbuzz-hb-common.html#hb-feature-from-string)

I tried a bunch of things...

```tex
\font[features=+cv75]{Hoa hồng đẹp}
\font[features=cv75]{Hoa hồng đẹp}
\font[features="cv75[:]"]{Hoa hồng đẹp}
\font[features="cv75=1"]{Hoa hồng đẹp}
...etc.
```

I put a `SU.dump(options.features)` in `/opt/share/sile/core/harfbuzz-shaper.lua`.

It is getting passed to `hb._shape`.

Nothing I do results in the output SIL says I should get: [[click]](https://software.sil.org/wp-content/uploads/sites/20/2015/12/GentiumPlus-features5.000.pdf)

Fired up FontForge, verified the glyphs _are_ there.

![image](https://user-images.githubusercontent.com/838783/88846863-39aa4500-d1b4-11ea-9793-e09367921809.png)
![image](https://user-images.githubusercontent.com/838783/88846887-42028000-d1b4-11ea-821f-b72fa06cfe04.png)

I saw some reference to a "glyph number", but `cv75[7891]=1` doesn't work for me either. :disappointed: 

Sorry, I don't like using trackers for questions, it means I'm at the end of my rope :-)

simoncozens:
Can you confirm you have a graphite-enabled build of harfbuzz? (run `sile -d versions` on your file and look at "Shapers enabled")

ctrlcctrlv:
@simoncozens:

```
Harfbuzz version: 2.6.4
Shapers enabled: graphite2, ot, fallback
ICU support enabled

Fonts used:
/usr/share/fonts/truetype/gentiumplus/GentiumPlus-R.ttf:0	Version 5.000
/usr/share/fonts/truetype/gentiumplus/GentiumPlus-R.ttf:0	Version 5.000
/usr/share/fonts/truetype/gentiumplus/GentiumPlus-R.ttf:0	Version 5.000
[...]

PR #1002, opened 2020-07-31
\font-feature

### LINKED ISSUE #992: \add-font-feature; \remove-font-feature. Why not \font-feature?
Seems strange. Am I missing something?

ctrlcctrlv:
Closes #992.

0bbb1a3 adds SU.dumpString ; 
d809267 fixes an unreported bug I noticed related to #994 ;
541404b is the bulk of the work and adds `\font-feature` itself;
24dc59a adds tests for `\font-feature`.

This is my first PR. Hopefully first of many. Let's see how it goes. :game_die:


ctrlcctrlv:
OK, I don't know how to fix the failing test. SILE does not output OpenType features in a defined order. This doesn't matter normally: `+dlig;+hlig` and `+hlig;+dlig` are equivalent.

```diff
osboxes@osboxes:~/Workspace/sile/tests$ sile -b debug font-features.sil 
This is SILE v0.10.9.r36-gf16af44
<font-features.sil>
[1] 
osboxes@osboxes:~/Workspace/sile/tests$ diff font-features.debug font-features.expected 
39c39
< Set font 	Gentium Plus;24;400;;normal;+hlig;+dlig;LTR
---
> Set font 	Gentium Plus;24;400;;normal;+dlig;+hlig;LTR
61c61
< Set font 	Gentium Plus;24;400;;normal;+hlig;+dlig;;LTR
---
> Set font 	Gentium Plus;24;400;;normal;+dlig;+hlig;;LTR
osboxes@osboxes:~/Workspace/sile/tests$ sile -b debug font-features.sil 
This is SILE v0.10.9.r36-gf16af44
<font-features.sil>
[1] 
osboxes@osboxes:~/Workspace/sile/tests$ diff font-features.debug font-features.expected 
39c39
< Set font 	Gentium Plus;24;400;;normal;+hlig;+dlig;LTR
---
> Set font 	Gentium Plus;24;400;;normal;+dlig;+hlig;LTR
61c61
< Set font 	Gentium Plus;24;400;;normal;+hlig;+dlig;;LTR
---
> Set font 	Gentium Plus;24;400;;normal;+dlig;+hlig;;LTR
79c79
< Set font 	Gentium Plus;10;400;;normal;+cv81;+cv75;LTR
---
> Set font 	Gentium Plus;10;400;;normal;+cv75;+cv81;LTR
107c107
< Set font 	Gentium Plus;18;400;;normal;+cv81;+cv75;;LTR
---
> Set font 	Gentium Plus;18;400;;normal;+cv75;+cv81;;LTR
110c110
< Set font 	Gentium Plus;10;400;;normal;+cv81;+cv75;-cv75;+cv81;LTR
---
> Set font 	Gentium Plus;10;400;;normal;+cv75;+cv81;-cv75;+cv81;LTR
117c117
< Set font 	Gentium Plus;10;400;;normal;+cv81;+cv75;LTR
---
> Set font 	Gentium Plus;10;400;;normal;+cv75;+cv81;LTR
120c120
< Set font 	Gentium Plus;10;400;;normal;+cv81;+cv75;+cv75;-cv81;LTR
---
> Set font 	Gentium Plus;10;400;;normal;+cv75;+cv81;+cv75;-cv81;LTR
140c140
< Set font 	Gentium Plus;18;400;;normal;+cv81;+cv75;;LTR
---
> Set font 	Gentium Plus;18;400;;normal;+cv75;+cv81;;LTR
143c143
< Set font 	Gentium Plus;10;400;;normal;+cv81;+cv75;-cv75;-cv81;LTR
---
> Set font 	Gentium Plus;10;400;;normal;+cv75;+cv81;-cv75;-cv81;LTR
```

Fixing the failing linter stuff is easy and I'm pushing that now.

alerque:
The debug output backend should sort the OT features so that doesn't happen.

It might take a little bit to give this all a shakedown. I see some things that probably can be simplified a bit, but I want to actually put it through some paces before commenting on those and I'm knee-deep in JIT issues. In the mean time a couple low hanging fruit...

core/utilities.lua
```diff
@@ -103,9 +103,23 @@ utilities.debug = function (category, ...)
   end
 end
 
-utilities.dump = function (...)
+local _dumper = function(ret, space, ...)
```
alerque, line 106: I don't think we want to complicated the `dump()` function much more. Also some versions of Lua have issues with this mixed variable argument format.

The way we've been doing this before is actually adding `__tostring()` methods to tables that need this kind of dumping, then using `SU.debug()`. Looking at your use case I think that's probably appropriate here too.

packages/features.lua
```diff
@@ -1,7 +1,11 @@
+-- OpenType features for SILE.
```
alerque, line 1: No inline copyright headers please, the project level notice covers this.

tests/font-features.sil
```diff
@@ -0,0 +1,27 @@
+\begin[papersize=a5]{document}
+\script[src=packages/features]
```
alerque, line 2: For unit tests, less is more. I've spend days and days reviewing them when things break and the fewer things being rolled into a test the better. Some tips:

* Use `\nofolios` and `\neverindent` to simplify both input and output.
* You aren't testing rules, don't load that package.
* You aren't testing rebox, don't use it.
* Don't include fancy Unicode like ✓ just because you can, use the absolute minimum characters / words to demonstrate the thing being tested.

ctrlcctrlv: Would you perhaps like me to split it into multiple files? Since I'm testing multiple things.

First I'm testing if `Letters=SmallCaps` works, and if it's equivalent to the old way (`\add...`; `\remove...`) of using it.

Then I'm testing the needed multiple feature syntax I added, which I based off `fontspec.pdf`: `Ligatures=" { Historic, Discretionary }`, and making sure it's also equivalent to the old way.

Then I do a more complicated test, where I test nesting. I set up a parent `\font` with `CharacterVariant="{75,81}"`, and then in it put children `\font`s which do `CharacterVariant=No75`, `CharacterVariant=No81`. I then make sure that `\font[CharacterVariant="{No75, No81}"]` is exactly equivalent to that.

Finally, an unreleased version had a bug where `\font[size=3em]Test...` stopped working, because I had it wrapped in a `SILE.settings.temporarily`, which made the `\font` function stop working as expected when not fed content.

core/utilities.lua
ctrlcctrlv: Can @simoncozens please comment? I found this quite useful while debugging, and had it around in my files for a while. I can fix the mixed variable arguments. Having to add a metatable each time is annoying for what should be quick debugs, IMO. But, if @simoncozens hates it too, I'll remove it.

tests/font-features.sil
alerque: Yes you can split it up, but the other points still apply. Also some things like that last point would be caught be other tests anyway (I hope) and so don't need a new test. The fewer unique tests that die when a specific feature breaks the easier it is to figure out why.

packages/features.lua
ctrlcctrlv: Happy to remove it, but I'd like this to change:

https://github.com/sile-typesetter/sile/blob/3d0c0a3a22a758029e252e2e2da8f26af7f7c064/LICENSE#L3

On FontForge we solve this by having an `AUTHORS` file, and in the `LICENSE`, having:

## Contributors

See [`AUTHORS`](https://github.com/fontforge/fontforge/blob/master/AUTHORS).

And then in `AUTHORS`, we recognize our project's founder as being most significant by putting him on top, and then we list all committers alphabetically. Perhaps by number of commits would be more fair, I hadn't considered it deeply before this. Remembering, of course, that our project is much older, and went through several CVS, so in our case, AUTHORS doesn't equal one to one with commits, as sometimes George just thanked someone in a commit message and put them in authors, while still marking the code as his in Author field.

tests/font-features.sil
ctrlcctrlv: Well no it won't be caught by another test unless that test loads `features.lua`, which none of them do. `features.lua` overwrites `\font` per recommendation of @simoncozens.

ctrlcctrlv: And yes, what I'm saying is, if it's split up, suddenly we no longer need `rules.lua` to create visual separation. And we no longer need ✓ either because it's its own test so makes sense on its own without the prior semantics of "lines are equivalent".

packages/features.lua
```diff
@@ -104,53 +111,146 @@ local value = C(SILE.parserBits.integer)
...
-end)
+    SILE.settings.set("font.features", table2featurestring(t))
+    SU.debug("features", string.format("Features were: %s\nRemoved features: %s\nFeatures are now: %s",
+                                        SU.dumpString(false, t_cur), SU.dumpString(false, t_rm),
+                                        SU.dumpString(false, t)))
+end
+SILE.registerCommand("remove-font-feature", removefontfeatures)
+
+local fontfn = SILE.Commands.font
+SILE.registerCommand("font", function(options, content)
```
alerque, line 220: I don't think this should be replacing `\font` across all of SILE when loaded. There are some cases where changing a core function are okay (and certainly in your own projects you can do whatever) but I don't think replacing this is the right way to do it. If nothing else any changes we make to `\font` would need porting to more than one place. Instead use a different command and call out to the original where needed. Also easier to document this way!

ctrlcctrlv: Did I [misunderstand @simoncozens?](https://github.com/sile-typesetter/sile/issues/992#issuecomment-665575353)

alerque: I think so. He said "wraps" not "replaces".

ctrlcctrlv: Oh, OK. Sorry, I'm dumb. :man_facepalming: I understood "wrap" to mean "function that wraps `\font`", not "command that wraps `\font`.

What should I call it? `\font-and-features`? `\font-features`? `\Font`?

packages/features.lua
alerque: Bumping the date in the license and adding an MIT appropriate authors file would be welcome (the authors file if any should include a make target to update it before releases by scanning Git history), but that's not part of this PR. PRs need to stick to discrete topics.

core/utilities.lua
alerque: Any table being used as a data type that's being contributed to SILE core should have meta methods that facilitate this sort of thing. Anything you're hacking on your end prior to contribution can use whatever methods you like (I have my own collection too) but contributing a debug function to SILE core that isn't the most appropriate way to debug stuff that's in SILE core doesn't feel right.

What data do you have that `SU.dump(data)` doesn't spit out for "quick debugs" anyway? or `SU.dump{data1, data2, data3}` if passing multiple data points for debugging is your issue. What's to gain by stripping line break info and dumping only string elements?

tests/font-features.sil
alerque: The test suite isn't primarily meant for visual inspection. Don't write tests with visual inspection in mind, write them primarily so the `-b debug` output can be diffed. Secondarily as a fallback when things go wrong there should be a visual clue about what failed.

Overwriting `\font` won't be an issue since we aren't going to do that, but still do test all the possible modes of your function (with/without content, and so forth). Just do so in as terse a way as possible so as to not trigger diffs for unrelated changes.

packages/features.lua
simoncozens: No, @ctrlcctrlv didn't misunderstand, and this is the approach I would have taken - save the old `\font` command and extend it. I'm happy for a package to *extend* a built-in, and using the approach of saving and then calling the old command wouldn't mean we need to port changes to `\font` here too. I seem to remember someone wrote some code to do this kind of command wrapping, which is what I had in mind, but I  can't find it now.

alerque: My concern was less with whether in re-used the function name (although I'd suggest it would be a lot less confusing if it introduced a new function than replacing the original) and more with the fact that the entire body of `\font` was being re-implemented. Whatever you do with the command name I do think this needs to wrap the old font command.

`SILE.setCommandDefaults()` does something like this but in only designed to mess with default values for *options*, it won't work for this case. The technique is there though, just stash the old command, register a new one, parse the feature names down to a format the old `\font` will understand, then pass them along to the old command.

simoncozens: > The technique needed is not that different though, just stash the old command, register a new one, use the new one to parse the feature names down to a format the old \font will understand, then pass them along to the old command.

Isn't that what this change does?

```
local fontfn = SILE.Commands.font
 SILE.registerCommand("font", function(options, content)
    -- Do stuff with options
    fontfn(options, content)
end)

Comment (alerque) [reply packages/features.lua:226]: I apologize, I managed to mess up reading that bit in the code. Sorry to keep things so confusing, yes that is a wrap!

My remaining concern is that it's not possible to load the *features* package to use the other functions (e.g. `\add-feature`) without getting the re-mastered `\font` as part of the deal. I'm not sure how important that is or isn't, just something to think about.

**Comment by alerque** [comment]
The Lua 5.1 builds are failing by chance or because of implementation differences that affect the non-deterministic table output. We need to sort the feature list output by the debugger so this doesn't happen and tests don't semi-randomly break.

File: tests/font-features.expected [hunk 5]
```diff
@@ -0,0 +1,84 @@
...
+T	36 86	(As)
+Mx 	33.4683
+T	86 75 82 88 79 71	(should)
+Mx 	63.7630
+T	87 75 76 86	(this)
+Mx 	79.2952
+T	29	(:)
+Mx 	20.9764
+My 	116.1954
+Set font 	Gentium Plus;24;400;;normal;+hlig;+dlig;LTR
```
alerque, line 39: The list of enabled features needs to be sorted deterministically so this expected output is reliable.

ctrlcctrlv:
@alerque I don't know enough about SILE internals to fix this myself. `core/debug-output.lua` calls out to `SILE.font._key` (`core/font.lua`) which just uses `options.features`; meaning, I think, the very parser needs to change to always give options in the same order. Is this right?

ctrlcctrlv:
Another solution might be, in `core/debug-output.lua`'s `setFont`, break the array each time `+` becomes `-`, and then sort the sub-arrays alphabetically and glue them back together.

So:
```
1 +cv80;+cv75;-cv80;+dlig;-cv80;-cv75

2 {+cv75, +cv80}, {-cv80}, {+dlig}, {-cv75, -cv80}

3 +cv75;+cv80;-cv80;+dlig;-cv75;-cv80
```

Or, if we want to be really clever, we can read left-to-right and "cancel out" a plus and a minus, as:

```
1 +cv80;+cv75;-cv80;+dlig;-cv80;-cv75

2 +dlig;-cv80

3 -cv80;+dlig # alphabetized by second byte
```

Which do you prefer, my first solution (change all `options` orders), or one of these?

alerque:
I squashed together some of the commits that did something and then undid the same thing, then simplified and cleaned up a bit. The Lua ordering problem isn't fixed yet and I'm honestly not sure what the best approach is, but this should be a little bit closer to being mergable when we get that bit ready.

alerque:
I had a look at the trouble sorting the debug output and I'm pretty sure that is just a symptom of a deeper problem. It's not just the debug output that is wrong, it's potentially the actual output too. The debug backend is just dumping the same unsorted array of features that the real font selector is using. The debug output shouldn't have all the duplicated keys or an ordering problem — it shouldn't have that data in the first place. What needs fixing is the actual data ending up in the feature option array, not just it's debug representation.

alerque:
> Another solution might be, in core/debug-output.lua's setFont, break the array each time + becomes -, and then sort the sub-arrays alphabetically and glue them back together.

The SILE setting is a string, it shouldn't be troubled with sorting a table as part of the debug output, the problem is where the setting is getting set as a table, not where it's being debugged.

> Or, if we want to be really clever, we can read left-to-right and "cancel out" a plus and a minus, as

I don't think we can do that, these *might* not actually cancel. It's possible for `+dlig-dlig` to be different than an empty value (for a default-on feature an empty string leaves it on, turning it on explicitly and then off has the effect of turning it off).

alerque:
Hey @ctrlcctrlv would you be available to give this branch a quick spin again? I overhauled the Lua code a bit (okay a lot). Behind the scenes the way the functions iterate and save things is almost completely different. The output is close to the same, but not quite. Notably it now de-duplicates new options being set with those already in the font features setting, not just those in the current option list. Nesting commands and duplicating some features now does not result in a cluttered list in the settings. Also it goes ahead and saves negatives in case they are pertinent, not totally canceling out positives by not mentioning the feature at all.

If this still does what you expect I think we can merge since the tests are now deterministic.

alerque (approved):

alerque:
In the interest of this being in the next release, I'm going to assume the answer is yes. Since there aren't any breaking changes either way only a new package the worst case scenario is that the function isn't satisfactory and needs to be patched. I expect the next release cycle to be relatively short because I'm pushing to get math stuff included, but I really want to shake out some changes to the release cycle first before math lands so that important release has a better chance of being a smooth rollout.

PR #1017, opened 2020-08-06
Document better stair logo and \kern command

### LINKED ISSUE #1016: Better stair ("bumpy road") logo possible
I was going to PR this directly but was blocked by #1015, however I don't want to forget.

The stair logo in the manual (as of v0.10.9, &sect;6.1, p. 55), which I've copied into more than a few of my documents, could be improved.

Out with the old:
```tex
\define[command=SILEold]{S\lower[height=0.5ex]{I}L\glue[width=-.2em]\raise[height=0.6ex]{\font[size=0.8em]{E}}}
```

Right now, it uses a text size to make the smaller ⟨E⟩.

This looks, IMO, ugly, because it's essentially equivalent to the [fake small caps problem](https://practicaltypography.com/small-caps.html): the widest stroke of ⟨E⟩ is not equal to that of ⟨L⟩.

And in with the new:
```tex
\define[command=SILEnew]{S\lower[height=0.5ex]{I}L\glue[width=-.2em]\raise[height=0.6ex]{\font[features=+smcp]{e}}}
```

The problem is more acute with serif types, but keen eyes can notice it in sans-serif:

image

As luck would have it, my favorite font for use in SILE, EB Garamond, is worst affected of my sample.

I thought of this when rewriting the famous `\LaTeX` macro into SILE-speak:

```tex
\define[command=LaTeX]{\begin[family=Latin Modern Roman]{font}L\glue[width=-.3em]\raise[height=0.4ex]{\begin[family=Latin Modern Roman Caps]{font}A\end{font}}\glue[width=-.13em]T\glue[width=-.2em]\lower[height=0.5ex]{E}\glue[width=-.13em]X\end{font}}
```

Note the use of <kbd>Latin Modern Roman Caps</kbd>.

_Is this the most important issue facing SILE right now?_ **No**... It's still an issue, though!! 🤣

ctrlcctrlv:
Closes #1016.

# Stair ("bumpy road") logo
image
image

# `\kern`
image
image

### DIFF
--- a/documentation/c09-concepts.sil
+++ b/documentation/c09-concepts.sil
@@ -1,4 +1,5 @@
-\begin{document}
+\begin[class=book]{document}
+\include[src=documentation/macros.sil]
 \chapter{The Nitty Gritty}
 
 \noindent{}We are finally at the bottom of our delve into SILE’s commands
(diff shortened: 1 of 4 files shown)


ctrlcctrlv:
By the way, I must say, I very much disagree with the value judgment &laquo;It’s not a great logo&raquo;, especially now that it uses proper small caps. :wink: 

I love it, and will keep putting it in all my SILE documents~

simoncozens:
Looks good to me. @alerque, what's going on with docker? I'm getting 404s trying to see what it's doing wrong.

alerque:
@simoncozens Docker builds broke yesterday because the Arch Linux ecosystem got turned upside down by a half-baked transition to Lua 5.4 by default. I spent half my day working on patching and catching up packages that were not considered before somebody threw the switch. It's almost back to life.

Meanwhile the Docker CI job runner that posts status updates back to Github **never** reports the right links for the "details" link, their bot is just totally broken (and it has been reported many times). You have to go find the build logs yourself. The latest one for the PR is [here](https://hub.docker.com/repository/registry-1.docker.io/siletypesetter/sile/builds/5148f87c-ed06-465f-a13f-7d4669f63159).

alerque:
As of some Arch package fixes a few minutes ago `make docker` builds are working for me locally again so the next CI runs should too.

Meanwhile in spite of the green check on Travis that happened because I restarted some of the jobs, this PR is [actually failing](https://travis-ci.com/github/sile-typesetter/sile/builds/178827198) all of the Lua 5.1 & LuaJIT tests. Something introduced here isn't 5.1 compatible.

documentation/c09-concepts.sil
```diff
@@ -122,6 +123,62 @@ Adding horizontal and vertical penalties to the typesetter’s queues is similar
...
+\SILEglue\SILEglue\SILEglue\SILEglue\SILEglue\SILEglue\SILEglue%
+\line
+
+The 27\sup{th} of the 60 \SILEglue’s, in \color[color=#dd0000]{red}, is broken
+between its ‘L’ and ‘\raise[height=0.6ex]{\font[family=Gentium Plus,features=+smcp]{e}}’.
+Meanwhile, if we typeset the same line, except using \code{\\kern}’s as we had
+originally instead…
+
+\line
+\SILEkern\SILEkern\SILEkern\SILEkern\SILEkern\SILEkern\SILEkern\SILEkern%
```
alerque, line 156: This is totally fine as a demonstration, but one thing I'd like to fix is getting rid of all the warnings when the manual builds. Right now we're ignoring the fact that even our own manual has known-bad behavior going on such as overflows we *don't* intend. Which brings up the idea what we may need a way to suppress warnings for things we *do* intend, such as this overflow. There is nothing in place now, but someday this will need some kind of suppress mechanism.

documentation/macros.sil
```diff
@@ -36,6 +36,7 @@
 \define[command=changed]{
 \note{\bf{The material in this section has changed significantly since the previous release of SILE.}}
 }
+\define[command=sup]{\raise[height=0.3em]{\font[size=0.8em]{\process}}}
```
alerque, line 39: Not a big deal, but can we switch this to us `ex` units for the raise? Not a big deal here since we know the font, but as a demonstration if somebody copies this using a unit that is relative to font height rather than width is probably a better example to set.

alerque:
It wasn't obvious to me on a read through of the PR code just now why Lua 5.1 is affected...

alerque:
Disregard my Lua comments, the 5.1 issues are on the *other* PR, not this one.

@ctrlcctrlv If you can tweak the `em` → `ex` unit thing I mentioned and push again the Docker build *should* work again now, at which point I'd be okay to merge this.

documentation/macros.sil
ctrlcctrlv: Oh, so _that's_ what `ex` is for :eyes: 

Sure, no problem.

alerque:
Thanks for taking the time to contribute!

Issue #1018, opened 2020-08-08
Reseting the footnote counter with each page


isaz:
Resetting the counter in `newPage` of a book class kind of works, but in some cases, the footnote counter is reset before the next new page, even though the footnote itself is in the right page; something like this:
```
1. First footnote.
2. Second footnote.
1. Third footnote.
```
Is it possible to do some kind of rerun to fix the numbers? similar to what is done in `tableofcontents`.

alerque:
Hmmm. This is probably an issue with `pushBack()`. It's possible that `newPage()` isn't the best place to do the reset. I'll look into this, but do you have an example as minimalist as possible that replicates the incorrect behavior?

isaz:
Unfortunately, I don't have a minimalist example, this was part of a book which I can't share excerpts from. I will try to find an example though.
Thanks for the help.

simoncozens:
If you can't narrow it down, load the debug package and disable pushback at the start of your document:

```
\script[src=packages/debug]
\disable-pushback
```

If that works, we know where the problem is.

isaz:
> If you can't narrow it down, load the debug package and disable pushback at the start of your document:
> 
> ```
> \script[src=packages/debug]
> \disable-pushback
> ```
> 
> If that works, we know where the problem is.

@simoncozens tried with your suggestion, sorry to report that it didn't work.

alerque:
@isaz Don't be too sorry, in a round-about way that is actually good news. If the bug is *not* in `pushBack()` then it's much more likely to be an easy fix when we do find it.

alerque:
I haven't replicated the issue you showed yet, but there is a related issue in the opposite direction that will take some fixing. When a line gets shaped but doesn't fit onto the page and gets postponed to the next page, it might already have the footnote counters in it. This makes sense and we have no way to backtrack the counter and re-fill the node content with new values before shaping on the new page. The problem looks something like this:

```sile
\begin[papersize=a6,class=book]{document}
\nofolios
\script[src=packages/lorem]
% \script[src=packages/debug]
% \disable-pushback
\define[command=foo]{\bar \bar \bar \bar \bar \bar \bar\par}
\define[command=bar]{\lorem[words=3]\footnote{\lorem[words=5]}}
\begin{script}
local _newPage = SILE.documentState.documentClass.newPage
SILE.documentState.documentClass.newPage = function (self)
  local ret = _newPage(self)
[...]

PR #1032, opened 2020-08-14
Overhaul counters in insertions

alerque:
WIP branch for #1018, mostly so I can switch computers but help/ideas welcome.


Omikhleia:
Looking at the content, is this near-to-empty PR worth keeping?

Omikhleia:

PR #1040, opened 2020-08-19
Fix hole drawing in SVGs

### LINKED ISSUE #1005: [svg] "Holes" not appearing as holes
image

Should appear as:

MWE:
```
\begin{document}
\script[src=packages/svg]
hi % avoids #1004

\center{\include-svg-file[src=mojikyo_tkana.svg,height="10%ph"]}
\end{document}
```

SVG: 
[MWESVG.zip](https://github.com/sile-typesetter/sile/files/5010304/MWESVG.zip)

ctrlcctrlv:
Close #1005.

Essentially, the problem is, you were calling `f` after every path instead of after every shape.

Also, I made it so that `f*` will be used when called for.


ctrlcctrlv:
No idea why test fails; works fine on my machine and I have no way to replicate it. Looks like memory corruption; a lot of the numbers in `svg.c` (especially `256`) seem randomly chosen and the reasons for them aren't documented.

src/svg.c
```diff
@@ -80,13 +88,14 @@ int svg_to_ps(lua_State *L) {
...
-      if (output_l + 3 > max_output) {
-        output = realloc(output, max_output + 3);
-      }
-      output[output_l++] = strokeFillOper;
-      output[output_l++] = ' ';
-      output[output_l] = '\0';
     }
+    for (char* c = strokeFillOper; *c != '\0'; c++)
+        output[output_l++] = *c;
+    if (output_l + 4 > max_output) {
```
skef, line 94: It's fine to just increase the margin by 1 here but if there's insufficient space you need to realloc before copying strokeFillOper into the output, not after.

Rather than adding custom code to copy the string you might just call `safe_append()` instead?

src/svg.c
```diff
@@ -69,24 +70,34 @@ int svg_to_ps(lua_State *L) {
...
+            default:
+                strokeFillOper = "f*"; break;
+        }
+
         if (shape->stroke.type == NSVG_PAINT_COLOR) {
-          strokeFillOper = 'B';
+          strokeFillOper = "B";
         } else {
           if (output_l + 2 > max_output) {
+            max_output += 2;
```
skef, line 85: You're incrementing max_output and keeping the increment in the realloc() both here and with the +4 below. However, I don't (yet?) see how this could be causing the error.

src/svg.c
```diff
@@ -26,11 +26,12 @@ int svg_to_ps(lua_State *L) {
...
   struct NSVGimage* image;
   image = nsvgParse((char*)input, "pt", em);
-  int max_output = 256;
+  int max_output = 512;
   int output_l = 0;
   char *output = malloc(max_output);
   output[0] = '\0';
   for (NSVGshape *shape = image->shapes; shape != NULL; shape = shape->next) {
+    char* strokeFillOper = "s"; // Just stroke
     for (NSVGpath *path = shape->paths; path != NULL; path = path->next) {
```
skef, line 35: I'm suspicious of `safe_append()` above.

1. `strlen()` is the length of the string *without* the null terminator, so it seems like line 12 should have a `>=` to accommodate it (because `strncat()` adds it). That could result in a missing character, though, rather than the added garbage in the test failure.
2. More importantly, the limit in `strncat()` is supposed to be the number of characters to append, not the total size of the buffer. So the function is effectively functioning as `strcat()` here. 

I think together these problems mean that `safe_append()` can write a null character into the byte stored just past whatever `output` points to, which could explain the selective test failure. But it's hard to analyze these problems accurately. 

I would store the result of `strlen(s2)` before the first call, use it on line 12 but with the `>=`, pass the value to `strncat()` and then increment output_l by the value instead of calling `strlen()` again. Alternatively you could just call `strcat()` given that the code is already ensuring the buffer is large enough.

alerque:
Thanks for finding the root issue and taking a stab at this @ctrlcctrlv. It's looking promising. Thanks also @skef for the reviews on the C side of things, that's not a language I have a handle on so the input is very much appreciated.

As a matter of housekeeping Fredrick can you please open PRs from feature branches? This one is currently from the `master` branch of your fork which brings some limitations, the most notable of which is you can't give maintainers permission to push and hence help out with the PR. I believe Github actually added a way to switch base branches in open PRs recently so you should be able to push a branch and edit the info at the head of the PR to use the branch as the source. That would be helpful so I can help cleanup the test fixtures before merger. If you can't get it changed no worries I'll post patches here, but it would save rounds of communication if I could just work on the branch.

tests/svg.sil
```diff
@@ -0,0 +1,4 @@
+\begin{document}
```
alerque, line 1: I don't think we need this test. We do of course need to test SVG, but if we're testing specific aspects of SVG support having a separate test for SVG itself is not only duplicate work, it adds to the maintenance burden and makes it more complicated than it needs to be to review and update test fixtures when they change. Lets drop everything except the very specific minimized tests.

tests/bug-1005.sil
```diff
@@ -0,0 +1,4 @@
+\begin{document}
```
alerque, line 1: All tests (unless specifically testing paper size settings or having other good reasons to change it) should use the small paper sizes. This makes graphic review when needed much easier.

Also always include `\nofolios` when not testing the folios themselves because it simplifies the test output and the number of things unrelated to the test that might change. Likewise `\noindent` or `\neverindent` is quite useful to normalize tests where those settings aren't relevant.

```suggestion
\begin[papersize=a6]{document}
\nofolios
\neverindent
```

tests/bug-1005.sil
```diff
@@ -0,0 +1,4 @@
...
+\script[src=packages/svg]
+\svg[width=20%pw, src=./tests/bug-1005.svg]
```
alerque, line 3: In the interest of making visual checks easy, I would use most of the test canvas now A6):

```suggestion
\svg[width=100%lw, src=./tests/bug-1005.svg]
```

Alernatively specifying an absolute size in `pt` units (just less than the expected line width) might be better

```suggestion
\svg[width=265pt, src=./tests/bug-1005.svg]
```

tests/openpaths.sil
```diff
@@ -0,0 +1,4 @@
+\begin{document}
```
alerque, line 1: Request same changes re canvas and test image size as noted for `bug-1005.sil`.

alerque (changes requested):

tests/svg.sil
ctrlcctrlv: Since @skef looked at the C code and helped me fix it, probably worrying overly about memory management isn't such a big concern; he's really good at it and I trust his review. If it's OK with you I'll just scrap this test.

alerque: Sure. We can always add a test when we have a better idea what exactly needs testing. With a `busted` test we could assert the specific things you mention (not crashing, not having nulls, being ASCII, being at least X long. I'm happy to work up the test for that if there was a clear need to, but it also seems like most or all of those will get tested as side effects already anyway.

Dropping it for now is fine.

alerque:
@skef Feel free to mash the buttons man, your C is better than mine. I'm keeping an eye on the Lua / SILE side but your tips on this PR have been useful and official review or no I'm already deferring to you to review that end of things.

skef:
@alerque Consider the C code approved, then.

packages/svg.lua
```diff
@@ -3,6 +3,7 @@ local otparser = require("core/opentype-parser")
 
 local _drawSVG = function (svgdata, width, height, density, drop)
   local svgfigure, svgwidth, svgheight = svg.svg_to_ps(svgdata, density)
+  SU.debug("svg", string.format("PS: %s\n", svgfigure))
```
alerque, line 6:
```suggestion
  SU.debug("svg", function() return string.format("PS: %s\n", svgfigure) end)
```

This small hack is *almost* as silly as it looks. I ran some benchmarks a while back and discovered that one thing Lua does is evaluate all the expressions in these debug lines even if it doesn't use the result (i.e. the debug flag isn't active). This can cause significant overhead in some cases: ⓐ when the place being debugged is iterated over a lot or ⓑ the data being processed is large. I believe this falls into the latter category.

By passing a lambda function reference here the Lua interpreter is actually convinced not to bother evaluating the expression inside it unless the output is actually going to be used.

tests/openpaths.sil
```diff
@@ -0,0 +1,5 @@
+\begin[papersize=a6]{document}
+\nofolios
```
alerque, line 2:
```suggestion
\nofolios
\neverindent
```

tests/bug-1005.sil
```diff
@@ -0,0 +1,5 @@
+\begin[papersize=a6]{document}
+\nofolios
```
alerque, line 2:
```suggestion
\nofolios
\neverindent
```

alerque (changes requested):
Only minor tweaks left, stand-by and I'll actually fixup these details so this can be merged.

alerque (approved):

PR #1044, opened 2020-08-20
Cleanup libtexpdf outputter

### LINKED ISSUE #1012: Frame debug names are jibberish (again?)
I thought we found and fixed the frame name debugging sometimes being on the fritz, but apparently not:

image

It probably doesn't matter, but just in case ... this example is from PR #998 run with:

```console
$ make tests/bug-723.pdf SILEFLAGS='-d frames'
```

alerque:
Closes #1012 (cf. #915)

Also gets us one step closer to being able to run multiple backends at once...

### DIFF
--- a/core/libtexpdf-output.lua
+++ b/core/libtexpdf-output.lua
@@ -22,6 +22,10 @@ local _deprecationCheck = function (caller)
   end
 end
 
+local glyph2string = function (glyph)
+  return string.char(math.floor(glyph % 2^32 / 2^8)) .. string.char(glyph % 0x100)
+end
+
 local _dl = 0.5
 
 SILE.outputters.libtexpdf = {
@@ -117,8 +121,7 @@ SILE.outputters.libtexpdf = {
     -- is actually the shaped x_advance).
     if value.complex then
       for i = 1, #(value.items) do
-        local glyph = value.items[i].gid
-        local buf = string.char(math.floor(glyph % 2^32 / 2^8)) .. string.char(glyph % 0x100)
+        local buf = glyph2string(value.items[i].gid)
         pdf.setstring(cursorX + (value.items[i].x_offset or 0), cursorY + (value.items[i].y_offset or 0), buf, string.len(buf), self._font, value.items[i].glyphAdvance)
         cursorX = cursorX + value.items[i].width
       end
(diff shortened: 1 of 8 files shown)


core/libtexpdf-output.lua
```diff
@@ -117,30 +128,43 @@ SILE.outputters.libtexpdf = {
...
-      for i = 1, #(value.items) do
-        local glyph = value.items[i].gid
-        local buf = string.char(math.floor(glyph % 2^32 / 2^8)) .. string.char(glyph % 0x100)
-        pdf.setstring(cursorX + (value.items[i].x_offset or 0), cursorY + (value.items[i].y_offset or 0), buf, string.len(buf), font, value.items[i].glyphAdvance)
-        cursorX = cursorX + value.items[i].width
+      local x, y = self:getCursor()
+      y = SILE.documentState.paperSize[2] - y
+      for i = 1, #value.items do
+        local buf = glyph2string(value.items[i].gid)
+        self:setCursor(x + (value.items[i].x_offset or 0), y + (value.items[i].y_offset or 0))
```
alerque, line 135: @simoncozens Off the top of your head can you tell me anything about what this line was supposed to do? I ported it across with the math as-is, but I'm concerned because I don't get what it does and deleting the line entirely doesn't break any of our test fixtures.

core/libtexpdf-output.lua
```diff
@@ -117,30 +128,42 @@ SILE.outputters.libtexpdf = {
...
-      for i = 1, #(value.items) do
-        local glyph = value.items[i].gid
-        local buf = string.char(math.floor(glyph % 2^32 / 2^8)) .. string.char(glyph % 0x100)
-        pdf.setstring(cursorX + (value.items[i].x_offset or 0), cursorY + (value.items[i].y_offset or 0), buf, string.len(buf), font, value.items[i].glyphAdvance)
-        cursorX = cursorX + value.items[i].width
+      for i = 1, #value.items do
+        local buf = glyph2string(value.items[i].gid)
+        self:setCursor(value.items[i].x_offset or 0, value.items[i].y_offset or 0, true)
+        self:_drawString(buf, value.items[i].glyphAdvance)
+        self:setCursor(value.items[i].width, 0, true)
```
alerque, line 135: @simoncozens Off the top of your head can you tell me anything about what these 3 lines were supposed to do? I ported them across with the math as-is, but I'm concerned because I don't get what it does and substituting this doesn't break any of our tests:

```suggestion
        self:_drawString(buf, 0)
```

alerque: Nevermind, I finally found a case where this matters and will add a test for it so this code path actually gets covered.

PR #1046, opened 2020-08-28
Setup use *as* a GitHub Action

### LINKED ISSUE #907: Adapt for use *as* a Github Action
Given that we already have a Dockerfile and related tooling for creating an image out of any version, it would actually be really easy to adapt this repository for use *as* a Github Action. The only thing required is a config file and a couple tidbits to pass input / output information between the Docker image and GHA's runner environment.

alerque:
Closes #907.

Needs some experimentation and documentation.


alerque:
Here is the first PDF generated by SILE run as a GitHub Action, just for posterity: 
[mwe.pdf](https://github.com/sile-typesetter/sile/files/5144345/mwe.pdf)


image

alerque:
The grand total CI run time of 32 seconds from push to artifact post isn't bad either. It would be less if GitHub's own Docker container registry wasn't borked beyond usability (even they don't dog-food it and their popular actions pull from docker.io just like this did), but that is not out problem.

simoncozens:
Very, very nice!

ctrlcctrlv:
Cool :-) this is useful for my Unicode papers I think

alerque:
@ctrlcctrlv Seeing your work on Unicode papers is what prodded me to get back on this feature ;-) I've had SILE running on a CI server for years to serve my own projects but mine has been built using a custom CI runner on my own hosts. I've never cleaned up the idea into a building block that is easy for others to take and integrate.

This is a pretty simplistic start really. It will need some tweaking to be really useful on projects like yours where you want to utilize 3<sup>rd</sup> party SILE packages and fonts, but I figured starting with basic building blocks and working up from there into what full projects need was the best way forward.

PR #1057, opened 2020-09-14
Remove extra space in \code from packages/url

### LINKED ISSUE #1056: Extra space added by `\code` as defined in `packages/url`
Using SILE 0.10.10, and typesetting the following example:

```sile
\begin{document}
\script[src=packages/url]
The \code{\\code} command defined in \code{packages/url} typesets an extra space at the end, see:
\url{https://github.com/sile-typesetter/sile/blob/bd7e52e75663b861ee0bb57876204e740f518fbc/packages/url.lua#L47}.
Most likely I simply don’t know the use cases this extra space enables, but to the casual user,
this looks like a clear bug.
\end{document}
```

produces this output:

Bildschirmfoto von 2020-09-12 17-45-04

https://github.com/sile-typesetter/sile/blob/bd7e52e75663b861ee0bb57876204e740f518fbc/packages/url.lua#L47

Commenting out the line locally fixes the example above:

Bildschirmfoto von 2020-09-12 17-42-08

Does not typesetting this extra space break important stuff elsewhere?

I stumbled upon this while trying out @alerque's [WIP SILE writer for pandoc PR](https://github.com/jgm/pandoc/pull/6088) where using the standard markdown backticks syntax ends up using the `\code` command from `packages/url`.

jmaibaum:
Closes #1056

### DIFF
--- a/packages/url.lua
+++ b/packages/url.lua
@@ -44,7 +44,6 @@ SILE.registerCommand("code", function(_, content)
   SILE.settings.temporarily(function()
     SILE.call("verbatim:font")
     SILE.process(content)
-    SILE.typesetter:typeset(" ")
   end)
 end)
 
--- a/packages/url.lua
+++ b/packages/url.lua
@@ -40,11 +40,8 @@ SILE.registerCommand("url", function (options, content)
   SILE.call("code", {}, result)
 end)
 
-SILE.registerCommand("code", function(_, content)
-  SILE.settings.temporarily(function()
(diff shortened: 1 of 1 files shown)


alerque:
Thanks for taking the time to contribute this fix. I've tacked on a bit of a refactor that takes a convoluted set of steps and boils it down to about as simple as you can make it. Technically `\code` is not documented to take any arguments so it probably could get away without passing that through, but given it *could* be extended I guess this is okay too.

PR #1061, opened 2020-09-19
docs: Remove superfluous } in c10-classdesign.sil

jmaibaum:
The brace was already closed the line before. Copying this from the PDF into a text editor will result in unbalanced braces.


alerque:
Good catch, thanks for taking the time to contribute this fix!

PR #1063, opened 2020-09-19
docs(manual): Fix two-column layout example

jmaibaum:
Replacing `r` with `content` makes the example code work. This aligns the code with the prose above:

"We will use the boundaries of [the content] frame to declare our columns"

### DIFF
--- a/documentation/c09-concepts.sil
+++ b/documentation/c09-concepts.sil
@@ -267,11 +267,11 @@ column.
 
 \begin{verbatim}
 \line
-\\frame[id=leftCol, left=left(r), right=left(gutter),
-       top=top(r), bottom=bottom(r),
+\\frame[id=leftCol, left=left(content), right=left(gutter),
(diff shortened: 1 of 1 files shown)


alerque (changes requested):
Your commit comment suggest `content` as the correct frame name to use as a reference, and my memory from experience says the same. All the code in this pr says `column` instead, which I think is wrong.

jmaibaum:
Thanks for spotting the error @alerque ! I have fixed it now.

alerque (approved):

alerque:
Thanks again for taking the time to contribute. The manual having *working* examples is a very good thing.

I've considered these should sorts of things should probably actually be stand along files that we test for real, then `include` the source code for into the manual. In the past I've tried to update example code snippets manually for things like API deprecations, but the lack of automatic testing means any edits to examples are basically stabbing in the dark.

Issue #1062, opened 2020-09-19
Bringhurst class design example from the book is broken


jmaibaum:
The class design example in the SILE book does not work.

Following along the text, and copy-pasting all example code into `bringhurst.lua` will result in the following error trying to use `bringhurst.lua` with this minimal document:

```sile
\begin[class=bringhurst]{document}
\end{document}
```

This is the error:

```
$ sile test-bringhurst.sil 
This is SILE v0.10.10
<test-bringhurst.sil>

! A measurement tried to measure the paper size before the paper was defined at:
	test-bringhurst.sil: in <file> (\)
stack traceback:
	/usr/share/sile/core/utilities.lua:29: in function 'core/utilities.error'
	/usr/share/sile/core/units.lua:66: in upvalue 'checkPaperDefined'
	/usr/share/sile/core/units.lua:79: in field 'converter'
	/usr/share/sile/core/measurement.lua:82: in function 'core/measurement.tonumber'
	(...tail calls...)
	[C]: in function 'lpeg.match'
	/usr/share/sile/core/frame.lua:206: in method 'isAbsoluteConstraint'
	/usr/share/sile/packages/twoside.lua:12: in function 'classes/book.mirrorMaster'
	./bringhurst.lua:31: in main chunk
	[C]: in function 'require'
	/usr/share/sile/core/sile.lua:97: in function 'core/sile.require'
	/usr/share/sile/core/inputs-common.lua:5: in field 'init'
	/usr/share/sile/core/inputs-texlike.lua:182: in field 'process'
	/usr/share/sile/core/sile.lua:246: in function 'core/sile.readFile'
	/usr/bin/sile:62: in function </usr/bin/sile:62>
	[C]: in function 'xpcall'
	/usr/bin/sile:62: in main chunk
	[C]: in ?
```

For reference, this is the contents of my `bringhurst.lua`:

```lua
local book = SILE.require("book", "classes")
local bringhurst = book { id = "bringhurst" }

bringhurst:defineMaster({
  id = "right", firstContentFrame = "content",
  frames = {
    content = {
      left = "width(page)/13",
      top = "left(content) * 8 / 5",
      right = "100%pw - 2*top(content)",
      bottom = "top(footnotes)"
    },
    folio = {
        left = "left(content)",
        right = "right(content)",
        top = "bottom(footnotes)+3%ph",
        bottom = "bottom(footnotes)+5%ph"
    },
    footnotes = {
        left = "left(content)",
[...]

PR #1065, opened 2020-09-22
Refactor document classes using Penlight model

alerque:
This is long overdue work towards removing `stdlib`, in this case replacing the class model (the object oriented programming concept) used in SILE classes with Penlight classes.

In some ways this is a follow up to #1064 that started some much-needed code organization so that there would even be something regular to replace. See also #1062 and [wiki documentation of this project](https://github.com/sile-typesetter/sile/wiki/STD-to-Penlight-Migration).


alerque:
Wow, talk about timing. [Penlight 1.9.0](https://github.com/lunarmodules/Penlight/releases/tag/1.9.0) finally landed with [a fix](https://github.com/lunarmodules/Penlight/pull/289) more multiple inheritance that this work is going to need!

alerque:
This *so* needs to happen, the existing document "class" inheritance in so broken...

alerque:
This is much closer to landing than all the red CI marks suggest. I actually get a complete pass on tests locally last night, there are just some platform hickups and manual fixes to go.

alerque:
One thing not in here that I don't actually plan on doing is a regression test for the shims that make this work with extension classes still expecting stdlib. I've been reverting the changes to `jplain` and `jbook` to and running the tests that use those to make sure the shims work locally, but I'm going to release this with everything converted to the new model. Frankly I've spent **way too much time** making this backwards compatible in the first place. I could have just volunteered to fix EVERYBODY's code to the new model in less time. I need to move on.

PR #1066, opened 2020-09-25
Patch Penlight usage for forward support

alerque:
Work continues on #1065 and elsewhere to actually port our code to proper Penlight class usage, but this is an emergency fix that needs to get rolled into a release ASAP. The idiosyncratic way I worked around issues with the bug in old versions of Penlight regarding class inheritance turns out to be even weirder than I imagined. I inadvertently sidestepped the 'parent constructor' issue by making everything a grand-parent with no true parent classes. Hence now that in Penlight 1.9.0 constructor inheritance actually works properly we aren't actually getting anything inherited.

A bigger set of changes that fixes the idiosyncratic usage will be upcoming, but this should allow distros that updated to Penlight 1.9.0 already to keep using SILE.

### DIFF
--- a/core/tracestack.lua
+++ b/core/tracestack.lua
@@ -66,7 +66,7 @@ traceStack.commandFrame = pl.class({
 traceStack.contentFrame = pl.class({
     _base = traceStack.commandFrame,
     _init = function (self, command, content)
-      self:super(command, content, content.options)
+      self._base._init(self, command, content, content.options)
     end
   })
 


alerque:
Gah! Somebody yanked the Penlight 1.9.0 release and [replaced it with 1.8.1](https://github.com/lunarmodules/Penlight/issues/346). I presume this was either a bad typo and intended to be 1.9.1 or a very temporary loss of sanity. Either way half a dozen distros picked up 1.9.0 already, so supporting it is still an emergency fix for SILE (existing installs have stopped working on Arch Linux, Manjaro, Parabola, Raven, and others). However I'm dropping the upgrade to bundle a new Penlight and saving that for the next release. This enables SILE to work with ether old Penlight versions or the 1.9.0 that was (and is for some distros).

PR #1068, opened 2020-09-25
chore(actions): Rename GH action to not match somebody's username

alerque:
It is not urgent to cut a release for this because the syntax we documented for usage as a GitHub Action actually works now, but we won't show up in the Marketplace with push-button installation options until this is fixed.


alerque:
Oh spiffy, it let me publish the last release to [the marketplace](https://github.com/marketplace/actions/sile-typesetter) just by fixing this in master, we don't have to wait for the next release cycle for it to show up.

PR #1076, opened 2020-10-02
fix(packages): \verbatim:font can process text

jmaibaum:
This input:

```sile
\begin{document}
\script[src=packages/url]
Let’s try typesetting a URL: \url{https://sile-typesetter.com}.

And now a little bit of \code{inline code}.

And now a \verbatim{code block} inside normal text.
\end{document}
```

currently creates this output (using current master; zoomed in a bit):

Bildschirmfoto von 2020-10-02 11-48-51

With this PR applied, the output looks like this:

Bildschirmfoto von 2020-10-02 11-53-13

### DIFF
--- a/classes/markdown.lua
+++ b/classes/markdown.lua
@@ -54,10 +54,7 @@ SILE.registerCommand("bulletlist", function (_, content)
 end)
 
 SILE.registerCommand("link", function (_, content)
-  -- SILE.settings.temporarily(function ()
-    -- SILE.call("verbatim:font")
-    SILE.process(content)
-  -- end)
+    SILE.call("verbatim:font", {}, content)
 end)
 
 SILE.registerCommand("image", function (_, content)
--- a/packages/verbatim.lua
+++ b/packages/verbatim.lua
@@ -1,6 +1,7 @@
-SILE.registerCommand("verbatim:font", function(_, _)
-    SILE.settings.set("font.family", "Hack")
-    SILE.settings.set("font.size", SILE.settings.get("font.size") - 3)
+SILE.registerCommand("verbatim:font", function(options, content)
+  options.family = options.family or "Hack"
+  options.size = options.size or SILE.settings.get("font.size") - 3
+  SILE.call("font", options, content)
 end, "The font chosen for the verbatim environment")
 
 SILE.registerCommand("verbatim", function(_, content)


packages/url.lua
```diff
@@ -40,8 +40,11 @@ SILE.registerCommand("url", function (options, content)
   SILE.call("code", {}, result)
 end)
 
-SILE.registerCommand("code", function(options, content)
-  SILE.call("verbatim:font", options, content)
+SILE.registerCommand("code", function(_, content)
+  SILE.settings.temporarily(function()
```
alerque, line 44: Good catch.

This is ... indicative of a different problem. The `verbatium:font` command is not processing input. I think I'd rather fix this at the source. Of course this command isn't the only offender that doesn't process content even if sent, but lets weed them out rather than cater to them.

alerque (changes requested):
I think we can leave the `\code` command alone and fix this with something like this patch instead:

```diff
diff --git a/packages/verbatim.lua b/packages/verbatim.lua
index 523e71eb..b6fbcb4c 100644
--- a/packages/verbatim.lua
+++ b/packages/verbatim.lua
@@ -1,6 +1,7 @@
-SILE.registerCommand("verbatim:font", function(_, _)
-    SILE.settings.set("font.family", "Hack")
-    SILE.settings.set("font.size", SILE.settings.get("font.size") - 3)
+SILE.registerCommand("verbatim:font", function (options, content)
+  options.family = options.family or "Hack"
+  options.size = options.size or SILE.settings.get("font.size") - 3
+  SILE.call("font", options, content)
 end, "The font chosen for the verbatim environment")

 SILE.registerCommand("verbatim", function(_, content)
```

I haven't tested yet so there may be other usages of `\verbatim:font` that may need fixing.

packages/url.lua
jmaibaum: True, I was also thinking about modifying `verbatim:font` instead, so doing this now.

classes/markdown.lua
```diff
@@ -54,10 +54,7 @@ SILE.registerCommand("bulletlist", function (_, content)
 end)
 
 SILE.registerCommand("link", function (_, content)
-  -- SILE.settings.temporarily(function ()
```
alerque, line 57: Did this form stop working when we changed `\verbetim:font` or did you change it preemptively?

jmaibaum:
> I think we can leave the `\code` command alone and fix this with something like this patch instead:
> 
> ```diff
> diff --git a/packages/verbatim.lua b/packages/verbatim.lua
> index 523e71eb..b6fbcb4c 100644
> --- a/packages/verbatim.lua
> +++ b/packages/verbatim.lua
> @@ -1,6 +1,7 @@
> -SILE.registerCommand("verbatim:font", function(_, _)
> -    SILE.settings.set("font.family", "Hack")
> -    SILE.settings.set("font.size", SILE.settings.get("font.size") - 3)
> +SILE.registerCommand("verbatim:font", function (options, content)
> +  options.family = options.family or "Hack"
> +  options.size = options.size or SILE.settings.get("font.size") - 3
> +  SILE.call("font", options, content)
>  end, "The font chosen for the verbatim environment")
> 
>  SILE.registerCommand("verbatim", function(_, content)
> ```
> 
> I haven't tested yet so there may be other usages of `\verbatim:font` that may need fixing.

I'm still a learner of "idomatic SILE", as it seems. Your suggestion seems total valid, so I modified the PR accordingly.

Re other uses of `verbatim:font`, most are redefinitions:

```
$ grep -rn verbatim:font .
./classes/markdown.lua:58:    -- SILE.call("verbatim:font")
./documentation/macros.sil:32:\define[command=verbatim:font]{\font[family=Hack,size=9pt]}
./packages/verbatim.lua:1:SILE.registerCommand("verbatim:font", function(_, _)
./packages/verbatim.lua:16:    SILE.call("verbatim:font")
./packages/verbatim.lua:54:can redefine the \code{verbatim:font} command. For example you could change
./packages/verbatim.lua:58:<define command="verbatim:font">
./packages/url.lua:44:  SILE.call("verbatim:font", options, content)
./tests/bug-317.sil:4:\define[command=verbatim:font]{\set[parameter=font.family,value=Hack]}
./tests/bug-132.sil:4:\define[command=verbatim:font]{\font[family=Hack]}
```

The only "real" use besides `\code` from `packages/url` inside SILE’s tree is in `classes/markdown.lua` – which was commented out. I took the opportunity to re-enable it, and hope that this is fine? Did not yet find a way to test this though, as I don't really understand this classes use-case?

classes/markdown.lua
jmaibaum: See my comment from a few minutes ago, I basically re-enabled it as it looked like it should be working now. Didn't test yet, though.

alerque:
> Re other uses of `verbatim:font`, most are redefinitions:

I think we should probably check all the redefinitions too, otherwise the usage will be different. However the `\define`s from SIL macro land are a little different. They *should* automatically adjust to accepting content or not depending on how they are used. We need to confirm that though.

> The only "real" use besides `\code` from `packages/url` inside SILE’s tree is in `classes/markdown.lua` – which was commented out. I took the opportunity to re-enable it, and hope that this is fine? Did not yet find a way to test this though, as I don't really understand this classes use-case?

Don't worry about it too much, as far as I know that class is still at a proof of concept stage.

classes/markdown.lua
alerque: Oh right. I see now.

alerque (approved):

jmaibaum:
I checked the two redefinitions from the tests, they look OK to me (assuming I got the comments in the source code right):

`tests/bug-132.sil`:

Bildschirmfoto von 2020-10-02 16-38-16


`tests/bug-317.sil`:

Bildschirmfoto von 2020-10-02 16-39-06

I also rebuilt the book, and didn't notice anything wrong related verbatim passages, skimming through a few pages, so I guess the macros in `documentation/macros.sil` are also not affected.

Unless you want me to revert the change in the markdown class, I think we should be good here.

alerque:
Great, thanks!

PR #1079, opened 2020-10-10
fix(core): _drawString should take an offset

### LINKED ISSUE #1078: new version don`t work good in Arabic
image
this is test/sura-2.sil

simoncozens:
Fixes #1078

### DIFF
--- a/core/libtexpdf-output.lua
+++ b/core/libtexpdf-output.lua
@@ -109,9 +109,9 @@ SILE.outputters.libtexpdf = {
     return self:drawHbox(value, width)
   end,
 
-  _drawString = function(self, str, width)
+  _drawString = function(self, str, width, x_offset, y_offset)
     local x, y = self:getCursor()
-    pdf.setstring(x, y, str, string.len(str), self._font, width)
+    pdf.setstring(x+x_offset, y+y_offset, str, string.len(str), self._font, width)
   end,
 
   drawHbox = function (self, value, width)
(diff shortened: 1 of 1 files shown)


core/libtexpdf-output.lua
```diff
@@ -109,9 +109,9 @@ SILE.outputters.libtexpdf = {
     return self:drawHbox(value, width)
   end,
 
-  _drawString = function(self, str, width)
+  _drawString = function(self, str, width, x_offset, y_offset)
     local x, y = self:getCursor()
-    pdf.setstring(x, y, str, string.len(str), self._font, width)
+    pdf.setstring(x+x_offset, y+y_offset, str, string.len(str), self._font, width)
```
alerque, line 114: This will take care of the libtexpdf's internal comprehension of where the cursor is, but I think we need a call to self:setCursor() here to match our own internal comprehension of the same, otherwise they are going to be out of sync.

alerque (changes requested):

core/libtexpdf-output.lua
alerque: Or maybe I'm still misunderstanding this. I'm surprised it passed the BBT-RTL test for rules.

simoncozens: No! Offsets should not affect the cursor; that is the whole point of offsets.

simoncozens: ( https://simoncozens.github.io/fonts-and-layout//concepts.html#advance-and-positioning )

alerque: Oh crikey. The rules test no longer mixes text strings & rules....

simoncozens: The offsets should only non-zero when Harfbuzz gives us positioning information to draw a glyph at a certain displacement (e.g. mark positioning, cursive positioning). Any other time it should behave as before.

alerque (approved):

PR #1081, opened 2020-10-10
Test arabic shaping

### LINKED ISSUE #1080: Tests needed for cursive positioning and mark placement
As a follow up to #1078...

Part of the problem is that our test method doesn't even *register* the difference between broken and working. I assume this is because the debug backend isn't adding the relevant offset information from the shaping process.

alerque:
Closes #1080, or would in an ideal world. The PDF output of this test is broken in the before and works in the after of #1079. The test suite doesn't yet even *see* the difference.


simoncozens:
For this to work, the text outputter will also have to do the `if value.complex` test and position each character independently if true.

alerque:
And that's just what it now does.

alerque:
Regression tests are failing only because Travis can't get it's act together and *not* cache things. I've tried clearing the caches completely and we are still not getting fresh font downloads. I updated Libertinus and Amiri in the test suite, the tests are still somehow being run with the old versions.

alerque:
Okay Travis is badly broken. Their UI says we have no caches (we had some until I cleared them, now it says none), yet there *ARE* cached files that get loaded when caching is enabled again. I can bust them but as soon as I take away the busting we get old caches again. I think it might keep the new cache material once all the tests pass....

But now LuaJIT is giving us different rounding for `string.format("%.4f", number)` than other Luas. I've been able to replicate this locally and am looking for a debug solution that isn't tripped up by it.

alerque:
Sure enough once we passed tests once, backing out the cache clear hack doesn't break anything because the cache is fresh. I see quite a few posts on the Travis forums of people having similar issues, but no responses. If they don't get this sorted our I'll have to invent some cache clear mechanism of our own eventually, but today is not that day.

PR #1084, opened 2020-10-12
Don't abuse libtool internals

jsonn:
The whole internals of .libs are an implementation detail of libtool. I'm leaving the whole `to_copy` mess for now as it doesn't actually break something for pkgsrc, but the dependency certainly does.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -11,7 +11,7 @@ endif
 if SYSTEM_LIBTEXPDF
 LIBTEXPDF_LIB=-ltexpdf
 else
-LIBTEXPDF_LIB=../libtexpdf/.libs/libtexpdf.la
+LIBTEXPDF_LIB=../libtexpdf/libtexpdf.la
 endif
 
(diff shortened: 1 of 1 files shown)


alerque:
Thanks for these contributions @jsonn.

simoncozens:
Thanks @jsonn!

@alerque, for future reference, what's the best way to edit contributed commits to conform to .commitlint rules? Check out manually and rebase -i?

alerque:
@simoncozens  For PRs like this one with a single commit using the "Squash and Merge" option allows you to write your own commit message. The downside of that is that you skip CI checks for the re-written commit message, but that's what I did here. For longer ones where you want to preserve the commit the history, yes a chekcout and rebase is necessary. To speed things up I use [`gh`](https://github.com/cli/cli) to make it easy to fetch and checkout a PR like this (`gh pr checkout 1084`), and also use `git revise -i` instead of rebase for faster iterating (it doesn't require stepping through each commit as the working copy, it basically just revises the commits directly).

PR #1090, opened 2020-10-15
Update license holder (and revert overly aggressive Javascript→Lua port)

alerque:
This is *probably* moot either way given my access to the GitHub org, but see the commit message for my reasoning. I'm dealing with a repository transfer situation right now using Github's [Deceased User Policy](https://docs.github.com/en/free-pro-team@latest/github/site-policy/github-deceased-user-policy). Even though the repository owner was quite accepting of contributions and a number of people were involved during his lifetime, the lack on anything explicit naming other parties as being authorized to act on behalf of the project has made the whole process difficult.

@simoncozens if this sounds sane to you I think I'll leave you to merge this one so your name is on the merge commit.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2014 Simon Cozens
+Copyright (c) 2012-2020 Simon Cozens, Caleb Maclennan.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/core/typesetter.lua
+++ b/core/typesetter.lua
@@ -448,7 +448,7 @@ SILE.defaultTypesetter = std.object {
         end
       end
     end
-    SU.debug("pagebuilder", "Glues for self page adjusted by", adjustment, "drawn from", gTotal)
+    SU.debug("pagebuilder", "Glues for this page adjusted by", adjustment, "drawn from", gTotal)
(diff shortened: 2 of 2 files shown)


simoncozens:
Looks good, thank you!

rjmunro:
What does being copyright 2 people actually mean? I'm worried that it just means Simon owns his bits and Caleb owns his bits. You probably want it to mean something like a joint tenancy where you both own all of it.

Despite being interested in open source licenses and stuff for 20 years, I've never thought about this before. It's a huge can of worms. Sorry.

https://innovaccess.eu/wp-content/uploads/2019/08/jointownershipinintellectualpropertyrights_0000000649_00.xml_.pdf

alerque:
@rjmunro Yes this is a huge can of worms. Honestly this is just a shim not a full bullet proof fix. Yes one interpretation of licenses like this is that the parts remain copyrighted by the authors, but since they also *release* usage rights under the terms of the license it doesn't matter very much who holds the copyright. As I understand it this shim helps a little bit in most normal OSS project hand-offs as long as an actual legal action doesn't get involved, in which case there are loopholes that this doesn't close.

As far as I know the only way to really close the loop is to establish a legal entity to hold the copyright. That would make a lot more sense if we were using a more restrictive license and wanted to be able to enforce the terms. Given that the MIT license lets third parties take and use the code in proprietary projects anyway I don't think it makes any sense for SILE.

PR #1102, opened 2020-11-09
Turn off complex flag for items in \latin-in-tate

### LINKED ISSUE #1101: [tate, ja] Whether or not \latin-in-tate works is totally unpredictable
```tex
\begin[class=jplain,layout=tate]{document}
\script[src=packages/tate]
\script[src=packages/font-fallback]
\neverindent
暗号者・暗号学者・暗号実装者協議会

\parタイプセッター\latin-in-tate{SILE}はバグのある\latin-in-tate{Program}です
\bigskip\bigskip
\parタイプセッター\latin-in-tate{8888}はバグのある\latin-in-tate{9090909}です
\bigskip\bigskip
\parタイプセッター\latin-in-tate{S}\latin-in-tate{I}\latin-in-tate{L}\latin-in-tate{E}はバグのある\latin-in-tate{program}です
\bigskip\bigskip
\par* マスターキーとは、異なる管轄区域の複数の政府機関に与えられる単一の暗号化キーです。特定の暗号化されたチャットアプリケーション（\latin-in-tate{Signal}や\latin-in-tate{WhatsApp}等）を通過するすべてのメッセージやその他のデータを復号化することができます。
\bigskip\bigskip
\par* マスターキーとは、異なる管轄区域の複数の政府機関に与えられる単一の暗号化キーです。特定の暗号化されたチャットアプリケーション（\latin-in-tate{SignalW}等）を通過するすべてのメッセージやその他のデータを復号化することができます。
\bigskip\bigskip
\par* マスターキーとは、異なる管轄区域の複数の政府機関に与えられる単一の暗号化キーです。特定の暗号化されたチャットアプリケーション（\latin-in-tate{Signal}等）を通過するすべてのメッセージやその他のデータを復号化することができます。
\end{document}
```

image

Numbers always seem to work, words are unpredictable.

simoncozens:
Fixes #1101. Arguably breaks vertical Arabic in Japanese, but to be fair it was broken already.

### DIFF
--- a/tests/bug-1101.sil
+++ b/tests/bug-1101.sil
@@ -1,0 +1,7 @@
+\begin[class=jplain,layout=tate,papersize=a4]{document}
+\script[src=packages/tate]
+\neverindent\nofolios
+\font[family=Noto Sans CJK JP,language=ja]
+
+\latin-in-tate{SILE}と\latin-in-tate{Latin}。
+\end{document}
--- a/packages/tate.lua
+++ b/packages/tate.lua
@@ -91,6 +91,12 @@ SILE.registerCommand("latin-in-tate", function (_, content)
         SILE.typesetter.state.nodes[#SILE.typesetter.state.nodes+1] = nodes[i]
       end)
       local n = SILE.typesetter.state.nodes[#SILE.typesetter.state.nodes]
+      -- Turn off all complex flags.
+      for j = 1,#(n.value) do
+        for k = 1,#(n.value[j].nodes) do
+          n.value[j].nodes[k].value.complex = false
+        end
+      end
       n.oldOutputYourself = n.outputYourself
       n.outputYourself = outputLatinInTate
     end
--- a/tests/bug-1101.expected
+++ b/tests/bug-1101.expected
@@ -1,0 +1,20 @@
+Set paper size 	595.275597	841.8897729
+Begin page
+Mx 	539.9079
+My 	97.7650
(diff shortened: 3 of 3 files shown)


alerque:
Travis caching is going to be a problem here, I'll poke at that end of things.

alerque:
I split your commit into two (one for the fix and one for the test) then reverted the fix commit to test the test. Tests still pass — so either the test case isn't complex enough to even evidence this problem or it's not fixed.

tests/bug-1101.sil
```diff
@@ -0,0 +1,7 @@
+\begin[class=jplain,layout=tate,papersize=a4]{document}
+\script[src=packages/tate]
+\neverindent\nofolios
+\font[family=Noto Sans CJK JP,language=ja]
+
+\latin-in-tate{SILE}と\latin-in-tate{Signal}。
```
alerque, line 6: Running this test for me with the `tate.lua` fix reverted looks fine. I don't think this test is adequate to trigger the bug that #1102 represents, but I'm not sure what is either.

alerque (changes requested):

tests/bug-1101.sil
alerque: Per Fred's comments on the issue and my testing, the word "Signal" does not trigger this bug whereas "foobar" or "Latin" in the same context do. I've updated the test with one of those so that running the test without the fix in place actually fails as expected.

alerque (approved):

Issue #1092, opened 2020-10-19
Provide packages with a way to handle raw content


alerque:
The way the parser works, we currently have to tell the parser ahead of time what tags contain raw content that needs to be passed through to an external handler rather than recursively parsed for more SILE commands. This works great for our builtin `\script{}` command. It can also be amended quite easily as I've been assuming we would de for math (#220), Lilypond (#435), TikZ (#437), SVG (#1091), etc. if and when any of those things were bundled with SILE.

This does **not** work out very well for things like @ctrlcctrlv's `texmode` package (see discussion in #947) that we do not plan to bundle. In fact it means 3rd party packages are at a distinct disadvantage to bundled ones in what they can and can't do.

There is a workaround, but in requires invoking SILE with special arguments:

```console
$ sile -e 'SILE.inputs.TeXlike.passthroughCommands.foo = true' -d foo test.sil
```

We should either think of a way for packages to inject their requested handling early (I can't think of one other than using two parsing passes) or make a more convenient interface for this and document it for use by 3rd party packages.

Omikhleia:
(See https://github.com/sile-typesetter/sile/pull/675#r825074471 for the original context).

Instead of cramming more and more special commands into `SILE.inputs.TeXlike.passthroughCommands`, maybe a dumb idear, but why don't we modify the parser to accept (besides math and script, admittedly) only one _single_ 'inline-escaping' command, say `\inline[format=name, other options]{pass-through stuff}`. We then have the pass-through content and the extra options ready to be provided to some package.

Note that since we are at parser level, we could even decide for something that doesn't look as a real SILE command, so it could as well be `@inline(name)[other options]` or whatever (= think to the 3-backslash escape in Markdown, or the `<![CDATA...` esacpe in XML) and not necessarily use braces, but perhaps custom begin-end markers (= think to shell script escaping e.g. `cat <<  MEH ... MEH`)...

But for the sake of simplicity here, let's admit we may prefer using a SILE-familiar syntax...

Eg.
```
\svg[src=somefile.svg, width=dimension}
```
vs.
```
\inline[format=image/svg+xml, width=dimension]{<svg>....</svg>}
```
Or even:
```
\inline[format=base64:image/svg+xml, width=dimension]{PHN2Zz4uLi48L3N2Zz4=}
```
Another possible use:
```
\inline[format=text/x-lua]{some Lua code}
```
(And let's thus have a real verbatim thing, with possibly a package that even does syntax highlighting for the said language)

[...]

PR #1103, opened 2020-11-09
Native math typesetting support

OlivierNicole:
**Note:** This is a respawn of #578. I proposed several increments on top of @harrysummer's work, up to a point where the feature could be considered for merging. Since @harrysummer has not been considering my pull requests to their repo for a while, I open this new PR to unblock things.

image

This PR proposes a math typesetting engine as a non-invasive library. It could also be included as a package (if not for the necessity to tell the parser not to parse the contents of a `math` command). Currently supported and unsupported features:

- [x]  Parse MATH table from OpenType.
- [x] Two input languages: MathML (currently, only a subset of it) and a TeX-like language, nicer for humans (including multi-argument macros).
- [x] Subscript, superscript and mixed sub-superscript.
- [x] Fractions
- [x] ”Big operators”, i.e. sum, product, integrals…
- [x] Matrices *(very basic at this point)*
- [x] Brackets and parentheses that grow vertically to match their content, and other vertically growing glyphs
- [ ] Letter accents, radical, “under” symbols and generally all horizontally stretchable glyphs
- [ ] Kerning support for superscripts and subscripts
- [x] Full support of Unicode (akin to LuaLaTeX with `unicode-math`)
- [x] For convenience, include all “symbol macros” from`unicode-math` (`\alpha`, `\beta`, `\infty`…) as an alternative to Unicode
- [ ] Better integration with SILE typesetter 
  - [ ] Convert top-level horizontally layout math nodes into SILE's `nnode` to enable line breaking within math.
  - [ ] Insert penalty nodes into the `nnode` sequence to have better line breaking.
- [x] Test cases.
- [x] Documentation.

Whether this can be merged as-is is a matter of debate. I personnally think it can, even though it is still experimental and not yet complete.


core/opentype-parser.lua
```diff
@@ -185,15 +185,236 @@ local function parseSvg(str)
...
+  if s:len() <= 0 then return end
+  local fd = vstruct.cursor(s)
+  return vstruct.read(">majorVersion:u2 minorVersion:u2 fontRevision:u4 checkSumAdjustment:u4 magicNumber:u4 flags:u2 unitsPerEm:u2 created:u8 modified:u8 xMin:i2 yMin:i2 xMax:i2 yMax:i2 macStyle:u2 lowestRecPPEM:u2 fontDirectionHint:i2 indexToLocFormat:i2 glyphDataFormat:i2", fd)
+end
+
+local function parseMath(s)
+  if s:len() <= 0 then return end
+  local fd = vstruct.cursor(s)
+
+  local parseDeviceTable = function(offset, fd)
```
simoncozens, line 198: These (device table, coverage table) are general OpenType subtables and should be hoisted up to top level so they can be used by other routines.

src/justenoughharfbuzz.c
```diff
@@ -286,6 +286,61 @@ int shape (lua_State *L) {
...
+
+  hb_blob_t* blob = hb_blob_create(font_s, font_l, HB_MEMORY_MODE_WRITABLE,
+      (void*)font_s, NULL);
+  hb_face_t* hbFace = hb_face_create(blob, font_index);
+  hb_font_t* hbFont = hb_font_create(hbFace);
+  unsigned int upem = hb_face_get_upem(hbFace);
+  hb_font_set_scale(hbFont, upem, upem);
+
+  /* Harfbuzz's support for OT fonts is great, but
+     there's currently no support for CFF fonts, so
```
simoncozens, line 304: I'm not sure this is true any more (probably was when this code was originally written!). Instead use `can_use_ot_funcs`.

alerque:
Wow, I hadn't seen the PR's to Rui's fork. Thanks for all this work. I can't review it tonight but will certainly check it out.

Just a heads up that we're being bitten by an upstream Travis bug feeding us old fonts from our test suite even though we cleared the cache. When this gets closer to being ready to merge if they haven't gotten it fixed but `make test` passes locally I have a way to twist Travis's arm.

OlivierNicole:
Fixed MATH table parsing, now all math fonts should work.

OlivierNicole:
I think this set of features is a good starting point for basic math typesetting. Will you consider merging it (if I write suitable tests and documentation) ?

There is still the problem of https://github.com/sile-typesetter/sile/issues/1121, but I do not feel I will be able to fix it alone anytime soon, and I already know a temporary *ad hoc* fix for it.

I understand that mathematics typesetting is not a priority for the current maintainers. If you would rather not merge this, I can make it a third-party package (I also think having math typesetting as a package is an impressive proof of the modularity of SILE). However, to do that, it would help if you could merge the MATH table parser (240 loc) in the core, and make the `math` command passthrough (or provide a suitable interface as suggested in https://github.com/sile-typesetter/sile/issues/1092). Let me know if you would rather go for the 3rd-party option.

simoncozens:
I’m *desperate* to merge this (with documentation... 😉) and I think it should definitely go in core. I’m very excited about it. Thank you so much for all your work on this.

alerque:
Even though I don't have a need for this in my own work I too am excited to see it nearing readiness for use and will help the process of getting it merged.

However I adamantly disagree with @simoncozens about where it belongs. I strongly believe it should be a package, not in core! That being said I suspect my objection is a pedantic one in the way we are using words. I think this should be contributed to this repository and distributed with the base set of packages that make up a SILE‌ installation as opposed to being in a 3rd party package needing a separate installation. That's probably what Simon means too, but since "core" also refers to a set of Lua libraries that get loaded on every SILE run no matter what classes or packages are loaded I want to clarify that this doesn't belong in there. I've spent a great deal of time recently refactoring things *out* of core into proper modular namespaces. Especially given the potential for completely different math implementations this is something that should be modular and replaceable.

One small caveat to this in that we currently don't have a mechanism for packages to change the passthrough nature of commands, and we can certainly just hard code this one until that gets fixed.

OlivierNicole:
Sure, moving it to `packages/` sounds reasonable. It will probably consist in several files, however.

alerque:
If there are generic bits that make sense to have *always loaded up and on tap for everybody* for use in all documents and packages, then sure making some library space in `core` might make sense, but otherwise lets keep it as self contained as a package as possible. That doesn't mean it has to be one file (although our current package layout is pretty flat). I'm more worried about the Lua variable namespace and keeping each package as insular as possible using only public APIs and not making too many assumptions about the internals of other packages or even of core libraries. That way every other part of the system doesn't need to constantly adapt if this implementation changes and it would be easy to swap our for an alternative implementation.

I can help some with refactoring, organization, and test mechanics.

simoncozens:
Sorry, yes - I agree with @alerque, it should be in the `packages/` directory, but in the core SILE distribution. Not in the `core/` directory.

OlivierNicole:
Done. Also documented the package if you want to check. Next step: adding tests and fixing commit messages.

alerque:
I can probably handle commit messages. I'm pretty comfortable with rebasing across merges.

And no problem on requiring Libertinus Math.

packages/math/base-elements.lua
```diff
@@ -0,0 +1,1209 @@
...
+      compensatedY = SILE.length(y.length + self.value.items[1].depth
+        - self.value.items[1].fontDepth)
+    else
+      compensatedY = y
+    end
+    SILE.outputter:setCursor(scaleWidth(x, line), compensatedY.length)
+    SILE.outputter:setFont(self.options)
+    pdf.colorpush_rgb(0,0,0)
+    pdf.colorpop()
+    SILE.outputter:drawHbox(self.value, scaleWidth(self.width, line))
```
alerque, line 917: Related to #1121, why does the second argument to `drawHbox()` here have all this fancy math? In my testing setting this to and integer at all shows no difference in rendering.

packages/math/base-elements.lua
```diff
@@ -0,0 +1,1209 @@
...
+    return "space{w = "..self.width..", h = "..self.height..", d = "..self.depth.."}"
+  end,
+  _init = function(self, width, height, depth)
+    elements.terminal._init(self)
+    self.width = type(width) == "string" and SILE.length(width)
+      or width
+    self.height = type(height) == "string" and SILE.length(height)
+      or height
+    self.depth = type(depth) == "string" and SILE.length(depth)
+      or depth
```
alerque, line 728: This doesn't seem quite right to me. First, shouldn't height and depth at least (if not width as well) be `SILE.measurement`s not `SILE.length`s? Would they ever have stretch/shrink values?

Secondarily, if you are going to accept multiple types and do type casting, you should use `SU.cast()` to do it not re-invent the wheel. These routines would let a mix-and-match bag of types though (lengths, measurements, numbers) which is less than ideal.

Lastly, any type of casting at all is going to slow things down just a little. Since math is a new thing here and you don't have technical debt to deal with, it might be best to be very strick with the types accepted and not do any casting at all. Either that or take advantage of the casting so you don't need to verbosely specify things as for example in `newStandardHspace()` on L1029-1052.

core/opentype-parser.lua
```diff
@@ -188,15 +188,262 @@ local function parseSvg(str)
...
+  -- assembly.partRecords = vstruct.read("> "..assembly.partCount.."*{ &GlyphPartRecord }", fd)
+  -- assembly.partCount = nil
+  return nil
+end
+
+local parseMathGlyphConstruction = function(offset, fd)
+  local construction = vstruct.read(">@"..offset.." glyphAssemblyOffset:u2 variantCount:u2", fd)
+  local mathGlyphVariantRecord = vstruct.read("> "..construction.variantCount.."*{ &MathGlyphVariantRecord }", fd)
+  return {
+    glyphAssembly = construction.glyphAssemblyOffset ~= 0 and parseGlyphAssembly(offset + construction.glyphAssemblyOffset, fd) or nil,
```
alerque, line 355: The `parseGlyphAssembly()` function never returns anything but `nil` no matter what you pass it, and it doesn't even create any side effects. I replaced it with a literal nil to demonstrate (and get rid of the lint errors) but I don't understand what it was intended to do so I can't fix it properly.

alerque: I reverted some of my lint fixes because they broke the code. They broke the code because some of the functions are relying on side effects. Those will need to get refactored before this can be merged.

core/opentype-parser.lua
```diff
@@ -188,15 +188,254 @@ local function parseSvg(str)
...
+local function parseMath(s)
+  if s:len() <= 0 then return end
+  local fd = vstruct.cursor(s)
+
+  local fetchMathValueRecord = function(record, parent_offset, fd)
+    if record.deviceTableOffset ~= 0 then
+      record.deviceTable = parseDeviceTable(parent_offset + record.deviceTableOffset, fd)
+    end
+    record.deviceTableOffset = nil
+  end
```
alerque, line 251: These nested functions that use side effects need to be refactored to not rely of shadowed variables *not* having their declared scope (see others where first argument is declared `local` but isn't) and not to rely on side effects.

alerque (changes requested):

packages/math/base-elements.lua
OlivierNicole: You’re right, in the case of a text element, the width has no reason to be stretchable or shrinkable.

OlivierNicole:
Is “math” an acceptable scope for commit messages now? It surprises me, as math typesetting is merely a package now.

packages/math/base-elements.lua
OlivierNicole: I agree on your second and third points.

> This doesn't seem quite right to me. First, shouldn't height and depth at least (if not width as well) be SILE.measurements not SILE.lengths? Would they ever have stretch/shrink values?

This was already in Rui’s code, but I kept it, thinking that in the far future, it is not unthinkable to have vertical space adjustment in math. It also seemed simpler to have those three lengths of the same type.

alerque: Fair enough. Lets just use `SU.cast()` here and be flexible about what it accepts. I think some of Rui's code predates the current type system which explains why this got a little messy.

packages/math/base-elements.lua
alerque: It's not just that you don't need to do scaling here, it's that the value doesn't seem to be used for anything at all:

```lua
SILE.outputter:drawHbox(self.value, 0)
SILE.outputter:drawHbox(self.value, 999)
```

alerque:
> Is “math” an acceptable scope for commit messages now? It surprises me, as math typesetting is merely a package now.

I was thinking about making it an acceptable scope, yes. It's just a package, but it's also a world of its own with a lot of moving parts and people tend to think of it that way.

Also I was punting on the decision. I figured since I was going to have to rebase this entire tree to fix commit messages anyway it would be easier to rewrite 'math'→'packages' than the other way around if that's where this goes.

packages/math/base-elements.lua
OlivierNicole: Well, I’m not the one who wrote `SILE.outputter:drawHbox`, I must admit I did try to understand why it takes the width as an argument. I my latest commits I removed the unnecessary `scaleWidth`, but I would need some more time to understand why the width is here in the first place.

packages/math/base-elements.lua
OlivierNicole: No, I agree with you about avoiding superfluous casts. I removed them in my latest commits.

OlivierNicole (commented):
Got it.

core/opentype-parser.lua
OlivierNicole: Could you give an example of a shadowed variable not having its declared scope? I don’t really see what you mean.

Otherwise, as a functional programming aficionado, I’m all in for avoiding side effects as much as possible, so I changed `fetchMathValueRecord` to avoid relying on them.

OlivierNicole:
The only linter complaints remaining in `core/opentype_parser.lua` are about function arguments that shadow existing values. I am personally not shocked by that and think it is *sometimes* reasonable, namely, when the shadowed value is clearly not needed in the function and its role is “taken over” by the shadowing value, so to say. The advantage is that you don’t have to think about a new, uninformative name, like `fd2`.

That being said, this is a light-hearted opinion and I’m happy to remove shadowing if that makes the linter happy.

OlivierNicole:
Before writing tests, I ran `make test` on master to check if everything passes there. I get a weird error before make finishes printing the results:
```
• tests/rotate.sil
• tests/simplesvg.sil

Known bad tests that fail:
⚠ tests/balanced.sil
⚠ tests/bug-214.sil
⚠ tests/bug-307.sil
⚠ tests/color-fonts.sil
⚠ tests/font-fallback-empty.sil
⚠ tests/footnotes-twocols.sil
⚠ tests/frametricks.sil
⚠ tests/process-commands.sil
⚠ tests/space-after-command.sil
⚠ tests/sura-2.sil
set -f; IFS=';'
packagecpath=(./{,core/}?.so)
packagepath=(./{,lua-libraries/}?{,/init}.lua)
FONTCONFIG_FILE=/home/olivier/prog/sile/sile.master/fontconfig.conf busted --cpath="${packagecpath[*]};;" --lpath="${packagepath[*]};;" spec
/usr/bin/lua5.3: /usr/share/lua/5.3/busted/runner.lua:3: module 'pl.path' not found:
	no field package.preload['pl.path']
	no file '/usr/share/lua/5.3/pl/path.lua'
	no file '/usr/share/lua/5.3/pl/path/init.lua'
	no file '/usr/lib/lua/5.3/pl/path.lua'
	no file '/usr/lib/lua/5.3/pl/path/init.lua'
	no file './pl/path.lua'
	no file './pl/path/init.lua'
	no file '/usr/lib/lua/5.3/pl/path.so'
	no file '/usr/lib/lua/5.3/loadall.so'
	no file './pl/path.so'
	no file '/usr/lib/lua/5.3/pl.so'
	no file '/usr/lib/lua/5.3/loadall.so'
	no file './pl.so'
stack traceback:
	[C]: in function 'require'
	/usr/share/lua/5.3/busted/runner.lua:3: in main chunk
	[C]: in function 'require'
	/usr/lib/luarocks/rocks-5.3/busted/2.0.0-0/bin/busted:3: in main chunk
	[C]: in ?
make: *** [Makefile:1442 : busted] Erreur 1
```

Any clues that come to mind? I have `penlight-1.8.0-1` installed, as required.

alerque:
> I have `penlight-1.8.0-1` installed, as required.

I'll look into this, but I recall some Penlight release had some dud code regarding paths. The minimum version we require for SILE is actually in the 1.6 range I think, and 1.9 series should be fine. The rockspec has `==` operators so that we're testing a consistent set of packages. That doesn't mean we require OLD versions of packages, only that those versions were likely current the last time I bumped them.

I'm a bit confused why the tests on `master` fail for you though, those are exactly what gets tested on CI.

OlivierNicole:
It doesn't help. But in fact, I realized that `make regressions` works; it's only `make busted` that immediately fails with the message above. I'll open a separate issue.

Fortunately `make regressions` seems to be what I wanted to do.

OlivierNicole:
I did the cleaning gymnastics again and testing works today, strange but never mind me I guess.

So the only test that fails is `bidi.sil`, both on math and on master.

OlivierNicole:
Added tests. Only `bidi.sil` doesn’t pass on this branch, but neither does it on master.

Also, I don’t know why the math tests don’t pass on the CI, I suspect a font-related matter but I am incompetent on the topic.

I am waiting for your input on the remaining unresolved reviews.

alerque:
@OlivierNicole The `bidi.sil` test is broken on *master* too. It works for me locally but not in‌ CI. Whatever is going on with it isn't the fault of this PR, it probably has to do with UTF-8 changes and my recent update to the expected output are probably wrong. Don't worry about that particular test for now, I'll work on straightening out CI.

OlivierNicole:
@alerque This test fails on my local master as well. In case it gives you more clues.

alerque:
Just a heads up, it's hard to see this since CI is still down, but there are a number of regressions in the Lua support department. The math stuff works in Lua 5.3 and newer but fails in 5.2 and lower. The main culprit seems to be using our length objects direly in Lua math functions mixed with raw numbers. The length types support that with meta methods in newer Lua VMs, but older ones refuse to look for methods if the types don't match.

```lua
-- Will fail in old Lua
local foo = SILE.length(2)
if foo > 1 then ... end

-- Should always compare similar types
local foo = SILE.length(2)
if foo > SILE.length(1) then ... end
```

Here is an example of an error of that class:

```console
$ make test TESTSRCS=tests/math-subsup.xml CLEAN=true
rm -f tests/math-subsup.actual
FONTCONFIG_FILE=/home/caleb/projects/sile-typesetter/sile/fontconfig.conf ./sile -m .deps/tests/math-subsup.d -d versions -f fontconfig -b debug tests/math-subsup.xml -o tests/math-subsup.actual
SILE v0.10.14.r124-ga752a40 (LuaJIT 2.1.0-beta3)
<tests/math-subsup.xml>

Error detected:
        ./packages/math/base-elements.lua:207: bad argument #1 to 'max' (number expected, got table)
```

In this case `math.max()` won't work. I found a work around for these before but I forgot what it was. I'll poke around to find the cure for those usages.

OlivierNicole:
These errors are now fixed in all tested versions of Lua (5.1 to 5.4).

OlivierNicole:
Just in case this wasn't clear, I don't intend to implement the missing features for the time being. I will only make the changes required by the review process. Also as far as I can tell, I've responded to all review queries, but if I missed something let me know.

alerque:
Yes I believe most things have been dealt with. I was trying to get some housekeeping stuff out of the why so we had proper testing going on. That's back in order I think, now I just need to clean up the history a bit and get this ready for a release. I'll probably cut a patch release first to take care of a couple little things, then make this a bigger point release. Ball in my court.

alerque:
Postscript to my last comment on versioning...

I keep forgetting we're in the zero-dot phase, which in semver terms means everything is moved up a level—API breaking changes go in minor and everything else in patch level. In other words this will be a "patch" level release even though it adds new features not just fixing bugs. I can't wait for the slightly more versatile numbering of 1-dot something ;-)

alerque:
My apologies for not getting this merged yet. Life has gotten the better of me for a while. I haven't forgotten entirely...

OlivierNicole:
Life is rarely easy on us. As someone who hasn't devoted any time to free software in two months, I can sympathize.  I do hope that soon you will be able to spend time doing what you really want —whatever it is!

alerque:
I finally sat down and got a few more hours into this. It is shaping up but it is also poking holes in my Git ninja foo. :open_mouth:

alerque:
Wow guys just going through the commits fixing commit messages for the sake of upcoming release notes I'm blown away, you did a lot of work to get it to this point. @harrysummer and @OlivierNicole great work.

I'm about to blow away this branch, but I'm going to post this one on my fork at [olivernicole/math-1103](https://github.com/alerque/sile/tree/olivernicole/math-1103) as a backup. The incoming push is going to be a byte-for-byte match with this one, but the history will look a bit different.

OlivierNicole:
Oh, nice! Thanks for reworking the commit history.

alerque:
Sorry it took so long.

It would be great if this could get a little testing in the *master* branch. There are a few other things that may or may not land in the next release depending on how long they takes, but there is also really no reason not to cut a release just for this. If it gets put through some paces I can cut a release pretty much at any time.

I suspect this will get some publicity, so it would be nice to know it's actually ready for some use.

simoncozens:
This is absolutely amazing. Thank you both so much (and also to @harrysummer) for all your work on this.

OlivierNicole:
@alerque You mean, in addition to the already existing tests?

I cannot vouch for the perfect usability of the package. In fact, I can safely say that it is not quite there yet. For instance, syntax errors in the `math` command will not report the relevant location where the error happened.

alerque:
> @alerque You mean, in addition to the already existing tests

I guess if no major bloopers show up preparing a showcase example for the release announcement I'll be comfortable releasing. After all this is **initial** support and the  debut can hardly be expected to also be the gold standard.

I think we can move to the [website issue](https://github.com/sile-typesetter/sile-typesetter.github.io/issues/19) for any remaining discussion.

PR #1116, opened 2020-12-16
fix: small error

0xflotus:


simoncozens:
Thanks very much! I've merged this manually in 6024bde.

PR #1117, opened 2020-12-27
Update Homebrew install commands

nico-bachner:
2 changes have been made to Homebrew since this was originally written:

1. The location of caskroom/fonts has changed: 
    ```
    Error: caskroom/fonts was moved. Tap homebrew/cask-fonts instead.
    ```
2. Home-brew casks are installed using the flag `--cask` now
    ```
    Error: Calling brew cask install is disabled! Use brew install [--cask] instead.
    ```


simoncozens:
Thanks (merged manually!)

PR #1122, opened 2021-01-26
Decode UTF16-BE in name tables

### LINKED ISSUE #1120: Trace output can have invalid UTF-8
The new (upcoming 0.3.0) CLI for CaSILE uses a wrapper to execute SILE and captures the output (both STDOUT and STDERR) on a line by line basis. The text processing library in Rust is complaining (very loudly in the form of a panic) that the input isn't valid UTF-8. Otherwise the SILE run seems to execute properly.

It's possible this is an issue with either the project (because source content snippets get echoed in the trace) or even a CaSILE issue with something leaking a sequence into the shell wrapper output, but for now I'm inclined to blame SILE.

simoncozens:
See #1120.


alerque:
I edited description because this won't close out the related issue (see prior comments there), but it is a good fix to an issue in general, thanks for tracking it down!

tests/bug-1120.sil
```diff
@@ -0,0 +1,15 @@
+\begin[papersize=a6]{document}
```
alerque, line 1: Let's not use SILE document rendering for tests that aren't actually testing the output of typesetting functions; they are a huge maintenance burden to keep fresh compared to more targeted tests. In this case I think a Busted basted spec will suffice since all we want to know is if a string got decoded correctly.

I'll rig it up.

alerque (changes requested):

alerque (approved):

PR #1127, opened 2021-02-01
Redo CLI arg parsing without stdlib

alerque:
- refactor(cli): Swap stdlib opts parser for cliargs
- chore(cli): Work around parsing limitation in cliargs


simoncozens (approved):

alerque:
Merging with admin privs for now because Travis is down. I did run the regression suite locally, although we only have a handful of tests that check whether the CLI is working and nothing that actually checks that arguments are processed properly so this is mostly down to my hand checking it.

PR #1134, opened 2021-02-09
Overhaul Docker build

alerque:
* Update base image and switch to tagged releases
* Move dependency list to place it doesn't have to be repeated
* Use hook to build so Docker Hub and locally built containers match
* Work around new GitHub Actions limitations
* Reuse utility scripts from Fontship & CaSILE
* Normalize intermediate container names for easier cross-project diffs
* Rewrite container tags based on autoconf binary renaming
* Package everything necessary to build Docker container in source dist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,31 @@
-FROM docker.io/library/archlinux:base-devel AS sile-builder
+FROM docker.io/library/archlinux:base-devel-20210131.0.14634 AS builder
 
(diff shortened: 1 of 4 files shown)


alerque:
This (and all builds of up to date distros) on Docker Hub is failing right now because of a new syscall in glibc vs. outdated hosts that don't have it whitelisted. The fix has landed upstream but I have no idea how long it will take to propagate to the servers that host CI builds.

PR #1144, opened 2021-02-15
Add font hooks, auto-load color-fonts

### LINKED ISSUE #1142: Make color font support more obvious
It isn't obvious that SILE supports color fonts. If you load a font with COLR/CPAL table, that could be output as color, SILE should try to help you more than [making you ask Simon on Twitter](https://twitter.com/simoncozens/status/1360937214608293894) and being directed to `packages/color-fonts.lua` :-)

simoncozens:
This adds a post-load hook which is called on a font the first time it is loaded. So far the hook parses the OpenType tables, and loads the `color-fonts` package if a COLR table is found. This PR will fix #1142.

### DIFF
--- a/core/font.lua
+++ b/core/font.lua
@@ -42,6 +42,11 @@ SILE.registerCommand("font", function (options, content)
     SILE.settings.set("font.hyphenchar", SU.utf8charfromcodepoint(options.hyphenchar))
   end
 
+  -- We must *actually* load the font here, because by the time we're inside
+  -- SILE.shaper.shapeToken, it's too late to respond appropriately to things
+  -- that the post-load hook might want to do.
+  SILE.font.cache(SILE.font.loadDefaults({}), SILE.shaper.getFace)
+
   if type(content) == "function" or content[1] then
     SILE.process(content)
     SILE.settings.popState()
@@ -95,9 +100,19 @@ SILE.font = {
     local key = _key(options)
     if not SILE.fontCache[key] then
       SU.debug("fonts", "Looking for "..key)
-      SILE.fontCache[key] = callback(options)
+      local face = callback(options)
+      SILE.font.postLoadHook(face)
+      SILE.fontCache[key] = face
     end
     return SILE.fontCache[key]
+  end,
+
+  postLoadHook = function(face)
+    local ot = SILE.require("core/opentype-parser")
+    local font = ot.parseFont(face)
+    if font.cpal then
+      SILE.require("packages/color-fonts")
+    end
   end,
 
   _key = _key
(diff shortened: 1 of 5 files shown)


core/font.lua
```diff
@@ -42,6 +42,11 @@ SILE.registerCommand("font", function (options, content)
     SILE.settings.set("font.hyphenchar", SU.utf8charfromcodepoint(options.hyphenchar))
   end
 
+  -- We must *actually* load the font here, because by the time we're inside
+  -- SILE.shaper.shapeToken, it's too late to respond appropriately to things
+  -- that the post-load hook might want to do.
+  SILE.font.cache(SILE.font.loadDefaults({}), SILE.shaper.getFace)
```
alerque, line 48: Is this going to mess up the `-d versions` output by outputting information about font files that are not ever actually used? Also won't this error or systems without the default font (Gentium Plus) even in the case of documents or classes which explicitly set a different font early?

tests/bug-1142.sil
```diff
@@ -0,0 +1,4 @@
+\begin[class=book,papersize=129mm x 198mm]{document}
```
alerque, line 1: Unless testing the class, papersize, or other spacing things directly (which I don't think is happening in this case) lets stick to `plain` instead of `book` and use `papersize=a6`. Suppressing folios and indentation also helps keep the tests from updating for unrelated changes.

```suggestion
\begin[papersize=a6]{document}
\noindent
\nofolios
```

core/font.lua
```diff
@@ -95,10 +100,20 @@ SILE.font = {
     local key = _key(options)
     if not SILE.fontCache[key] then
       SU.debug("fonts", "Looking for "..key)
-      SILE.fontCache[key] = callback(options)
+      local font = callback(options)
+      SILE.font.postLoadHook(font)
```
alerque, line 104: If it's a font here but a face in the next function that then does something extra to make it a font, maybe keeping the naming consistent between functions would make it easier to parse for humans.

```suggestion
      local face = callback(options)
      SILE.font.postLoadHook(face)
```

alerque (changes requested):
We can ignore the Docker and FreeBSD build tests for this PR. Docker is fixed, this was just branched before the fix landed. The Ciruss issue is still unresovled but unrelated to the PR.

core/font.lua
simoncozens: 1) Possibly, and 2) I don't think so, no. Note that this is inside the implementation of the `\font` command, so will only be called if the user (or class) calls that. It won't happen just from setting the default font - in fact, the other way around, it will only happen when a font other than the default is chosen!

alerque (approved):

PR #1148, opened 2021-02-16
feat(packages): Add complex-spaces package

### LINKED ISSUE #1146: SILE wrongly assumes a space will never contain any shapes, resulting in bizarre output
image

Those spaces shouldn't be there.

https://github.com/ctrlcctrlv/FRBAmericanCursive/blob/main/specimens/combos.sil

simoncozens:
Fixes #1146

### DIFF
--- a/packages/complex-spaces.lua
+++ b/packages/complex-spaces.lua
@@ -1,0 +1,41 @@
+SILE.nodeMakers["x-spaces-are-nodes"] = pl.class({
+    _base = SILE.nodeMakers.unicode,
+    makeGlue = function (self, item)
+      self:addToken(" ", item)
+      self:makeToken()
+    end
+})
+
+SILE.languageSupport.languages["x-spaces-are-nodes"] = true
+
+SILE.settings.declare({
+  parameter = "shaper.complexspaces",
+  default = true,
+  type = "boolean",
+  help = "Whether the font's space glyph should be emitted, rather than a glue"
+})
+
+if SILE.shaper then
+  local origSpaceNode = SILE.shaper.makeSpaceNode
+
+  SILE.shaper.makeSpaceNode = function (_, options, item)
+    if SILE.settings.get("shaper.complexspaces") then
+      local myoptions = pl.tablex.deepcopy(options)
+      myoptions.language = "x-spaces-are-nodes"
+      local nnodes = SILE.shaper:createNnodes( " ", myoptions)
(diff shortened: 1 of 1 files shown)


alerque:
Do we have any fonts in our already-downloaded test set that have a space glyph?

packages/complex-spaces.lua
```diff
@@ -0,0 +1,41 @@
...
+SILE.languageSupport.languages["x-spaces-are-nodes"] = true
+
+SILE.settings.declare({
+  parameter = "shaper.complexspaces",
+  default = true,
+  type = "boolean",
+  help = "Whether the font's space glyph should be emitted, rather than a glue"
+})
+
+if SILE.shaper then
```
alerque, line 18: The current module load tree is seriously borked –this test should be superfluous— but I'll refactor that when I finish fixing the class inheritance for shapers.

alerque (changes requested):
Just needs a test. I can address that...

packages/complex-spaces.lua
simoncozens: I, um, only added that to keep the busted "can we load this package" thing happy.

packages/complex-spaces.lua
```diff
@@ -0,0 +1,41 @@
...
+  local origSpaceNode = SILE.shaper.makeSpaceNode
+
+  SILE.shaper.makeSpaceNode = function (_, options, item)
```
alerque, line 21: Am I understanding this right that it will permanently affect *all* fonts in all languages after load unless the shaper is reloaded? It seems like this should be tied to the current font more than the current language/shaper, no?

alerque:
Rebased on master and force-pushed because the CI failures here are likely fixed since the original branch point.

packages/complex-spaces.lua
simoncozens: Yes, this probably needs rethinking to make it per-font. The question is how to automatically turn the feature back off again when a different font is chosen. (And back on again afterwards if we go back to the original font.)

alerque: Sounds like we need some sort of stack…

alerque:
Following the the same logic of merging #1002, I think I'm going to merge this now and expect refactors to make it better later. Since it's a package and not a core function we don't need to worry about breakage so much, and allowing this to be used sooner seems better than waiting for a perfect implementation.

Note that #994 is also in need of a setting that tracks the current font similar to what this needs.

PR #1168, opened 2021-04-20
refactor(utilities)!: Change breadcrumbs:contains() to return <bool, index>

alerque:
BREAKING CHANGE: Previous return value for breadcrumbs:contains() was
just an depth index with -1 indicating no match. This made sense when
I wrote it, but coming back to it for a new project I expected a boolean
return value. Returning two values seems like the best option, but given
the function naming it seemed to make sense to return the boolean first,
hence the API breakage.

### DIFF
--- a/core/utilities.lua
+++ b/core/utilities.lua
@@ -573,9 +573,9 @@ utilities.breadcrumbs = function ()
 
   function breadcrumbs:contains (needle)
     for i, command in ipairs(self) do
-      if command == needle then return #self - i end
+      if command == needle then return true, #self - i end
     end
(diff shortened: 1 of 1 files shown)


simoncozens:
Could we instead return `false, -1` so that we have the same number of return values each time and only need to check for one thing?

alerque:
We could. I initially thought of that, but I wasn't sure `-1` was ever going more useful than `nil`. On reconsideration I guess it does make it strictly better from a typing perspective since `type(-1) =~ type(nil)`. I'll do that.

Issue #749, opened 2020-01-03
0 parskip in grid mode results in full blank grid line


alerque:
See also #748, tracking this separately because it could be fixed different ways.

Note one side effect of the current method of hacking the leading every time a glue is pushed is that a `document.parskip` of 0 currently results in a full grid skip between paragraphs. Avoiding this requires a negative value of some small amount. Whether the current leading system is fixed or where the calculations are even done is changed, this behavior is counter-intuitive and should be fixed.

alerque:
Note fixing this will be a breaking change. Issuing a warning about what happened on 0pt parskips might be a good plan.

alerque:
I've confirmed this is fixed by #1175. I'm not quite sure what or how to warn, but I think release notes will be the place to do it.

alerque:
I added 6424369a just so the next release gets the appropriate warning.

PR #1175, opened 2021-06-04
grid.lua: Don't add a full gridSpacing where none would suffice

zavislak:
Previously when calculating the space toadd in makeUp, a cursor positioned exactly at a multiple of the gridSpacing would insert a full gridSpacing vglue, even when a zero vglue would also maintain proper spacing.  In other words, the range of possible values of toadd was (0, gridSpacing].

This change modifies the range of toadd to instead be [0, gridSpacing), which prevents the addition of unnecessary space.

### DIFF
--- a/packages/grid.lua
+++ b/packages/grid.lua
@@ -2,7 +2,7 @@
 local gridSpacing = SILE.measurement()
 
 local function makeUp (totals)
-  local toadd = gridSpacing - SILE.measurement(totals.gridCursor) % gridSpacing
+  local toadd = (gridSpacing - SILE.measurement(totals.gridCursor)) % gridSpacing
   totals.gridCursor = totals.gridCursor + toadd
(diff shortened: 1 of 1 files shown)


simoncozens:
This is brilliant, thank you!

simoncozens:
(Looks like we will also need a fix to the test expectations, which I will do elsewhere.)

alerque:
Thanks a lot for this fix @zavislak. I'm testing now but I'm pretty sure this will fix my bug report in #749. As such I might add on a warning to the changelog of the next release about this being a breaking layout change, even though the "breakage" will mostly be in the direction of expected behavior.

Also @simoncozens not your fault, but `commitlint` gave this a pass when it shouldn't have because its use of regular expressions is just that bad. For single commit PR's that need fixing in the future, the "squash" option from GH allows you to rewrite the commit message.

Issue #1174, opened 2021-06-03
Typesetting issues in sile-0.10.15.pdf


zavislak:
I'm new to SILE so I can't debug, but I wanted to note the following oddities:

On page 4 the spacing between two paragraphs is insufficient, probably due to the placement of the figure:

![image](https://user-images.githubusercontent.com/371987/120570526-3338bb00-c3cd-11eb-8f9a-9bbb13ad0890.png)

On page 74 in the two column demo, the left column has a vertical offset that takes the text out of the frame:

![image](https://user-images.githubusercontent.com/371987/120570642-7004b200-c3cd-11eb-9fae-be0ba40e495e.png)

alerque:
Thanks for taking the time to report these @zavislak! (And welcome to SILE. Don't be afraid to post about any trouble you run into here even if it's a question about how to accomplish X and not strictly a bug report).

The first issue here is a duplicate of #394. Drop-caps (floats) and everything to do with them are a long-standing disaster situation. Lets leave that out of this issue and comment on the old tracking issue if needed.

The second issue I don't remember seeing reported before. Lets make this issue about tracking down that problem.

Omikhleia:
_"A one-year old unaddressed issues...", thinks Omikhleia. "Hey, it's time to **pinpoint** it!_" :rofl: 

Let's recap:
![image](https://user-images.githubusercontent.com/18075640/177049280-0f912354-1d61-454f-ba06-47c6a2481c15.png)

Problem (2), i.e. some small frame overflow at the bottom, is easy to understand - it's a mix of two thing (fixed baseline skip from very early in `sile.sil`, aggravated by #1404). "_I am not going to investigate it further._"

But problem (1), i.e. some extra vertical space in the columns, is much more interesting. 
If one looks closely, the issue is not only on the first column (where it is indeed very visible), but also on the second one. Erm. Wait! Looking even more closely, it's actually all the frames top which are affected, from a certain point in the document. Let's use our eyes:
![image](https://user-images.githubusercontent.com/18075640/177049921-5c1bcafe-2000-4a57-8156-6cec7d4cdd0f.png)

We just need to track where this weird thing started occurring...  _Omikhleia puts on his spectacles... "**Gotcha**!_". 

**And the culprit is**.... That good old **grid** package, via it's package documentation.

[...]

PR #1177, opened 2021-06-16
Hunt down vertical space mishandling in twocol example

alerque:
See #1174. So far this doesn't find the problem, just tweaks incidentals found while looking for it


alerque:
I'm shutting down this PR in favor of #1184 because I accidentally opened this against the master branch on my personal fork. This blocks my being able to fiddle which that branch which is how I do some testing of CI jobs such as the release system that would otherwise be disruptive if done in this repository.

PR #1184, opened 2021-07-21
Hunt down vertical space mishandling in twocol example

alerque:
See #1174. So far this doesn't find the problem, just tweaks incidentals found while looking for it.

### DIFF
--- a/documentation/c09-concepts.sil
+++ b/documentation/c09-concepts.sil
@@ -285,14 +285,15 @@ And now finally we can end our \code{pagetemplate}.
 \end{verbatim}
 
 Let’s do it.
-\eject\par%
+\eject
+
 \begin[first-content-frame=leftCol]{pagetemplate}
 \frame[id=gutter,width=3%pw]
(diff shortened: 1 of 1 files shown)


Omikhleia:
I think this could be closed without merge, further to #1501 (original issue #1174 may remain open if we want to address the remaining points mentioned there)

ctrlcctrlv:
just wanted to recognize all the amazing work being done on this project lately, no idea if my sponsoring of @alerque has significantly helped with that but i have noticed a notably huge increase in how much SILE stuff i see in my inbox afterwards so am going to just assume i helped lol

alerque:
It did (and is!) helping a lot. It didn't get the ball rolling because I missed GitHub notifications and first noticed the change much after the fact after I'd already started a big push for v0.13, but it sure has helped keep the ball rolling. In particular it has really helped motivate me to persevere with cleaning up technical debt, hopefully improving the ergonomics for SILE users that aren't also core developers, and facilitating contributions that are not required for my own current use cases.

Issue #1185, opened 2021-07-28
Paragraph overflow from floating frames not shapped properly


alerque:
On the same dreaded topic as #394, I have another iteration that might not be the same bug.

Everything in a paragraph after a dropcap is not getting wrapped properly, and the *next* paragraph is not getting indented properly. Everything in the 2 lines of dropcap space is wrapped fine and everything in the next paragraph is fine, it's just the paragraph that comes over from the overflow that does not get hyphenated and wrapped properly.

![image](https://user-images.githubusercontent.com/173595/127354677-50501e59-6a90-4348-bdc6-ceb8fdead618.png)

Omikhleia:
Greetings,
This issue has not been discussed further for 1+ year and the requested MWE (Minimal Working Example) has not been provided to demonstrate the problem. Moreover in this case:
 - It seems to show frame overflow issues which might be the usual ones (partly tunable, partly not fixable without tracking etc.)
 - It uses the old "dropcaps" approach based on "floating" frames

Therefore, I am adding the "**pending closure**" label to this issue, as part of a backlog cleaning pass and triage attempt (= global pass on issues with the "**needs MWE**" label.)
Interested users are invited to say their word, and to provide a minimal example to reproduce or illustrate the point.
If nothing occurs, it will be rejected after some delay, without further notice.

PR #1191, opened 2021-08-26
Add MWE and mess around with pushback bugs

alerque:
Included so far is an MWE for #1185 and a bizzare hack that gives better results (probably by covering up the actual bug).


Omikhleia:
This 3+ year old PR never came to fruition as-is and now has conflicts.

Issue #1054, opened 2020-09-08
numbers in headings and TOC/PDF bookmarks


ousia:
I don’t know whether this is intended or not, but the manual (https://sile-typesetter.org/manual/sile-0.10.10.pdf) displays numbers in headings and not in the TOC and PDF bookmarks.

Sorry, but I’m afraid dissimilar references for the same entries don’t help the reader to find them.

Just in case it helps.

alerque:
I'm guessing this was by intention, but it's certainly something we can discuss.

Personally the current arrangement makes sense to me. I'm I'm hanging out in the TOC I'm probably looking for a topic, and the section numbers are not relevant. When I find what I want in the TOC the page number is going to be the fastest way to get me where I want to go. The same is true for the PDF bookmarks, which are basically a UI for the TOC.

On the other hand if I'm talking *about* the manual somewhere (say here on GitHub) I can mention a section number as an easy references that isn't locked to a specific version of the manual. Telling somebody the answer to their question is on "page 72" of the manual is not helpful if I'm looking at the latest development version of the manual and they happen to have a manual from the version packaged in their distro that is a version or two back. The pages might have changed, but the section numbers are much more stable (not 100%, just more so). Saying "check §8.3 on hyphenation" gives them an easy way to thumb through a manual and find the heading they want (given that the info is in the page header).

I'd be happy to make the manual more usable, but we're also going to need to strike a balance between usefulness and clutter. What is the use case for needing section numbers in the TOC when there are page numbers?

ousia:
> Personally the current arrangement makes sense to me. I'm hanging out in the TOC I'm probably looking for a topic, and the section numbers are not relevant. When I find what I want in the TOC the page number is going to be the fastest way to get me where I want to go. The same is true for the PDF bookmarks, which are basically a UI for the TOC.

Well, page numbers might contain multiple section headings. In the case of the TOC, linked entries would help a lot. 

> What is the use case for needing section numbers in the TOC when there are page numbers?

Not needing to check the page number. Any given section number is easier to check that a page number (section numbers are very similar to unique identifiers [when properly set]).

I agree that your proposal works when you are familiar with the document. It was not my case and I missed numbered section headings (wherever they happen to be).

Omikhleia:
Some documentation (e.g. technical) could benefit from having the sectioning details in the TOC. So I concur with @ousia that it could be an interesting feature.

There are actually three parts in it:
[...]

PR #1213, opened 2021-09-10
Tableofcontents leveraged

Omikhleia:
A proposal covering main aspects of #1054 

The distinct features are in separate commits:
1. Fix for the book class: unnumbered sections shouldn't increment the counters, so one can have, e.g. a "Prologue" and then a "Chapter 1")
2. The book class passes the section number to the `\tocentry`.
3. Table of contents support for:
    - depth control
    - configuration for showing the section numbers in TOC entries 
4. Table of contents support for:
    - option for TOC entries to be turned into links if the pdf package is loaded
5. Finally, output a warning to console if SILE needs to be rerun to update the TOC.

### DIFF
--- a/packages/tableofcontents.lua
+++ b/packages/tableofcontents.lua
@@ -27,6 +27,7 @@ end
 
 SILE.registerCommand("tableofcontents", function (options, _)
   local depth = SU.cast("integer", options.depth or 3)
+  local linking = SU.boolean(options.linking, true)
   local tocfile,_ = io.open(SILE.masterFilename .. '.toc')
   if not tocfile then
     SILE.call("tableofcontents:notocmessage")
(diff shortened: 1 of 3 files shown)


simoncozens:
Very nice! Thank you!

alerque (approved):

packages/tableofcontents.lua
```diff
@@ -23,9 +24,15 @@ local writeToc = function ()
   if not tocfile then return SU.error(err) end
   tocfile:write("return " .. tocdata)
   tocfile:close()
+
+  if not pl.tablex.deepcompare(SILE.scratch.tableofcontents, _tableofcontents) then
+    io.stderr:write("\n! Warning: table of contents has changed, please rerun SILE to update it.")
```
ctrlcctrlv, line 29: I think this was a missed opportunity. It could have been written such that SILE would return a non-zero return value.

Perhaps… :monocle_face: 

```
[fred@🍇葡萄🍇~]$ rg '#define\s+EAGAIN' /usr/include/ 
/usr/include/asm-generic/errno-base.h
15:#define	EAGAIN		11	/* Try again */
```

Omikhleia: @ctrlcctrlv See https://github.com/sile-typesetter/sile/issues/230#issuecomment-916789361 ;)

PR #1215, opened 2021-09-10
Add Symbola as fallback for math in code samples

### LINKED ISSUE #1214: Use different font for code blocks in manual, at least for math
Checking the manual on the new Math features, I see the verbatim code blocks don't have all the necessary.

image

We should switch to something that has all the symbols we need, or add a fallback for extra symbols, or something.

cc @OlivierNicole

alerque:
Fixes #1214

The only problem right now is this doesn't actually work and I don't understand why. I tried adding it to the whole document verbatim:font command and also adding it inside the specific verbatim block in question, nothing is actually working.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -173,15 +173,16 @@ examples: $(EXAMPLES)
 # garantee the TOC is up to date, simplify when #230 is fixed.
 hastoc = [ -f $(subst .pdf,.toc,$@) ] && echo true || echo false
 pages = $(PDFINFO) $@ | $(AWK) '$$1 == "Pages:" {print $$2}' || echo 0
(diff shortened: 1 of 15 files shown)


alerque:
Mystery solved. Rendering the documentation for the color-fonts package was replacing the shaper, hence killing off any ability to use font fallbacks. See #1244. I'm fixing it only enough so that **documenting it** doesn't trigger the shaper being replaced, not actually fixing the incompatibility.

alerque:
What a yak shave. Fixing the font used for one glyph in the math documentation has turned up bug after bug, first in the package loader in general, now in the Japanese / Tate / Ruby stuff. I seriously need to finish the document class inheritance stuff so we can enforce package scope where it belongs.

PR #1216, opened 2021-09-11
Fix footnotes and folios line and font issue

### LINKED ISSUE #986: What is the right way to set the footnotes font?

### LINKED ISSUE #1027: Why is there so much space between lines of my footnotes?
See e.g. footnote &numero;7 on page 6. 

https://raw.githubusercontent.com/ctrlcctrlv/PamudpodUnicodePaper/863487eca016bb938452c698705be329b5abe99e/pamudpod.pdf

Omikhleia:
Fixes #986
Fixes #1027

Test script:

```sile
\begin[papersize=6in x 9in, class=book]{document}
\script[src=packages/lorem]
\language[main=en]

\font[family=Ubuntu Mono, size=8pt]

\hbox{}

\vfill

\lorem[words=50].\footnote{\lorem[words=50]}

\break

\end{document}
```

Before:

image

After:

image


alerque:
This looks pretty good to me. The one concern I have is that (if I'm reading it right) it's going to be a pretty significant render change to existing documents that either relied on the old behavior or solved the same problem another way internally. I'm not sure whether we should consider this just an improvement or actually call it a breaking change.

Omikhleia:
> The one concern I have is that (if I'm reading it right) it's going to be a pretty significant render change to existing documents that either relied on the old behavior or solved the same problem another way internally. I'm not sure whether we should consider this just an improvement or actually call it a breaking change.

Another (simple) option, non-breaking, could have been to introduce a `\fontdefault` command (this was one of the alternatives I mentioned in #986. Despite what I said there - yes it looks a bit weird - I'm not against it either, formally.

alerque:
This really does seem to sit half way between being a big fix and being a breaking change. It is quite likely to be what people expect to have happened all along, but changing the way it works is breaking.

I have an idea. Lets flip a coin and (heads or tails) call it breaking just for the fun of it. That way we can merge earlier without holding up the release cycle for a major version bump, and as an added "benefit" the new math features will land in a bigger version number ;-) Sort of a cheat, but hey this *is* a breaking issue.

alerque:
Merged despite test failures, I've pushed so many tests lately some CI jobs are failing because font downloads are being rate limited. I'm testing the full suite locally.

alerque:
Dang it. Wrong PR. Yuck sorry, I was supposed to force merge #1220. This one was not ready.

alerque:
Fixed and merged properly with some manual fiddling.

Issue #1218, opened 2021-09-11
dropcaps (or floats) and footnotes do not behave very well


Omikhleia:
Greetings,

SILE 0.10.15 on Ubuntu 20.04

I tried to add a dropcaps to the initial (text) paragraph of a document. _Caveat emptor:_ My attempt at getting dropcaps might be wrong. SILE does a pretty good job without it.

Dropcaps and footnotes do not seem to behave very well. It might depend on how the dropcaps is made (so my definition follows, at the end of this ticket)

I tried to add a dropcaps to the initial (text) paragraph of a document.

- Pretty nice formatting without dropcaps :+1: - SILE does a good job that far: 
 
  ![image](https://user-images.githubusercontent.com/18075640/132963097-2c5aa166-af96-4cfb-a2bf-aa83493613b1.png)

  Well, nearly perfect - but to tell the truth, on p. 4 the notes get really close to the text (but that's not the main point here):

   ![image](https://user-images.githubusercontent.com/18075640/132963425-e7636722-229d-412a-b62d-84da3f771860.png)

- Any how, when the dropcaps is used on the first text paragraph :-1: 

   ![image](https://user-images.githubusercontent.com/18075640/132963139-6e45c151-fd3d-4c00-83c9-5dfad77fb7ec.png)

   - On page 1, lots of overlaps between the text and the footnotes. For a closer view:

     ![image](https://user-images.githubusercontent.com/18075640/132963233-1f183600-5aca-407b-82c1-0885dd26af9f.png)
 
   - The layoyut of subsequent pages  seems affected too (huge vertical spaces, etc.)

Of course, maybe I did somethings wrong. The definition I tried for obtaining the dropcaps is `lettrine{E}some more text` where:

```
\define[command=lettrine]{\noindent\float[bottomboundary=1bs]{\font[size=3bs, family=Linux Biolinum Keyboard O]{\process}}\noindent{}}
```

(lFWIW, "lettrine" = dropcaps, in French ;)

alerque:
I've already spent half my week trying to fix this. I have an alternate float implementation better suited to dropcaps in the works, but it still has some cases that don't work.

Everything I see here is iterations of the bugs noted in #394, #420, #474, #491 and #1185. Somewhat amazingly I think you've managed to trigger every one of those separate but related issues in one example.

I'm closing this as a duplicate, but I do feel your pain.

[...]

Issue #1118, opened 2021-01-04
Proposal: Change default leading algorithm


simoncozens:
SILE currently uses the TeX leading algorithm (baselineskip/lineskip). Literally the first thing I do when starting a SILE document is pull in the `linespacing` package and change it to fixed baseline distance. I'm sure that TeX users will prefer the TeX system rather than the system used by *literally every other typesetting package*, but I wonder if it's time to prioritise getting-things-done over TeX compatibility.

alerque:
Ya I think I'm on board with this. Let's think about how to do it gently though because it will be a *huge* reflow / breaking change for most documents.

I think starting with a warning whenever the TeX leading algorithm would be the right place to start, combined with a notice in the release notes. Once that's had some time to percolate and active users have had some lead time (pun intended) to adjust we can switch the default and include a warning the other direction for all cases where the default is used (i.e. don't warn when the document loads and sets the leading, but only warn when the bare SILE default is unchanged) so that people that missed it realize something changed behind their backs. A later release cycle could completely remove the warnings.

OlivierNicole:
Reading this thread, I was curious to learn what TeX’s behaviour was, but it’s the only one not documented in SILE’s manual.

alerque:
It's hard enough to keep up with documenting SILE, if documenting how TeX works were to be part of our goals we'd never come up for air. We can probably summarize...

Maybe [start here](https://tex.stackexchange.com/q/410250/5100) for a rough outline of factors. At the end of the day it's a bit of a fudge and a guess based on the current font with compensations for what happened in the two lines being computed. Descenders in the first line and ascenders in the second line cause compensations in the leading meaning the distance might change from line to line.

simoncozens:
It’s documented here: https://github.com/sile-typesetter/sile/blob/fa23fcd20aa05f51d5bbf96f0760c5985e0a74dd/documentation/c07-settings.sil#L166

OlivierNicole:
Sorry, I didn’t mean to sound like saying you should document it. Thanks for the pointer!

Omikhleia:
I'll be honest (seeing this is re-prioritized again, now to 0.13) - I don't think this would be a good move yet.

alerque:
Well it hasn't landed yet, so the timing is clearly up in the air. I'd be interested in hearing any rational for why to delay / when *would* be a good time. But also landing this will require extra careful thought to a deprecation warning and probably shimming first and changing the default later.

[...]

PR #1219, opened 2021-09-13
feat(core): Implement hangafter/hangindent + dropcaps package

Omikhleia:
Greetings,

This PR is an implementation for the idea I mentioned in #1218, as an alternative for implementing initial capitals (drop caps) without using float frames.
- It implements the hangafter/hangindent algorithm from TeX
- As such, it touches two "core" components, the typesetter (a small bit) and the line breaking routines (in a more tricky way, and the pseudo-code from Knuth's book isn't the simplest thing to read).

I haven't noticed anything broken in my tests (though limited), but this _certainly_ has to be reviewed and checked with both eyes wide open. Hence the PR, if anyone wants to play with it. - not necessarily to check drop caps, but also to ensure there's no hiccup on normal documents.

As a by-product of this hangafter/hangindent, I have included:
- The proposed `drocaps` package _per se_, with an example (in the "examples" directory). It works by:
  - Computing the dimensions of the "insert" based on the target font for the initial
  - Using that to set the appropriate hangs on the top left of the paragraph
  - Invoking some box magic to fit the initial in the space left blanks by the hangs
- A test for the hangafter logic, using all 4 cases (I have no idea what the use case for the 3 other corners, but it's here), in the tests directory.

N.B. As opposed to TeX, the hangafter is not reset at the start of a new paragraph (but being implemented via settings, it's easy to set only temporarily).


alerque:
I'm putting off reviewing and merging this until after the current release cycle because we have an issue with feature creep and can't get everything together for a good release if we don't draw some boundaries. I will not be opposed to an early follow up with whatever new features and fixes we have lined up after that, and this is very high of my personal list of wants/needs.

alerque:
Whew. Goodbye math, safe voyages.

Hello there my old ~~nemesis~~ friend dropcaps.

alerque:
This implementation has some great sides, but I'm going to combine it with one of mine before merging. The split between core shaping issues vs. dropcap package issues is a good one, and that is much better than any iteration I have cooked up. From my implementations though I have some measurement stuff that I think is a little more robust, a function to trigger the "next" paragraph to have a dropcap without needing to split out the character in markup yourself, and a few other tweaks.

I'm posting partial work on this because my brain is going fuzzy and I need to put it to bed.

alerque:
@Omikhleia Can you have another look at this? Other than splitting up the commits a bit I left your hangAfter/hangIndent implementation alone entirely. On the other hand I basically rewrote the dropcaps package in b76a175. It functions roughly the same way, but I changed to just two cycles creating hboxes, the second one being the final shape. What is being measured changed just a bit along the way (using the requested dropcap font for placement metrics, not the paragraph) but I think it makes sense. I also re-used some APIs that exist (rebox, lower) to achieve the same affect in perhaps a more understandable way for anybody that tries to copy and modify the code.

Besides that I moved the example to the website and cleaned it up a bit for that context, and I pared down the test a bit to be closer to a unit test.

alerque:
Force push right now took care of a Lua lint issue. The remaining test failures have to do with this being branched before some changes to the CI font caching. Even though I merged from master into this, the way the GH Actions are setup the fonts are cached for test runs on the PR. I *think* this will be okay, but if not I'll take care of that separately.

alerque:
I just took yet another pass at this in f31438d. This time I moved the content of the initial itself from an argument to the command to the content itself. The rest of the paragraph can just be typeset as usual afterwards. This makes it a bit easier to integrate into other systems and contexts I think.

alerque:
My last couple commits have added some additional features so this can be customized in more ways. Just for reference, here is where we are at (from the [website example PR](https://github.com/sile-typesetter/sile-typesetter.github.io/pull/26)):

```sile
\begin[papersize=a5,class=book]{document}
\nofolios
\script[src=packages/lorem]
\script[src=packages/dropcaps]
\set[parameter=document.parskip,value=4ex]

\dropcap[lines=2]{T}his paragraph has a pretty plain initial (drop cap),
using the default document font:
\lorem[words=26].\footnote{You didn’t expect more than that here, with such a font choice, right?}

\dropcap[lines=3,shift=-2pt,raise=-2pt,size=5em,family=Acorn Initials]{T}\font[features=+smcp]{his paragraph},
provided you have the Acorn Initials font, has a much better initial. The size of glyphs in Acorn tends to run a a little big, but we have exagerated this example by specifying a size ourselves. We've also tweaked the location left and down by 2pt.

\dropcap[lines=4,family=Zallman Caps]{H}\font[features=+smcp]{ere is an even nicer initial},
provided you have the Zallman Caps font:
\lorem[words=44].

\dropcap[lines=4,join=true,family=Zallman Caps,standoff=1.5spc]{T}his paragraph shows the effect of the joining option that you can use when the initial belongs to the first word in a sentence.
The first line of text will be flush against the drop cap.
Each additional line spanned with be indented with an extra space.
In old-style typography, this was often preferred to the bare rectangular box most modern word-processing software render.
With SILE, the choice is all yours!

\language[main=tr]
\dropcap[lines=2,style=Italic,join=true]{2.} gösterebileceğimiz nokta, dropcap fonksiyonun birden fazlı karakter içerebilmesi. İşte su paragrafın başında bulunduğu gibi bazı metinde nokta da dahil olması gerekebilir. Ayrıcı \em{Italic} stil kullanıldı.

\dropcap[lines=2,join=true]{\kern[width=-0.8en]“N}\font[features=+smcp]{ever say never”}, the saying goes.
But what does it mean?
Maybe someday your dropcap \em{will} need to include leading punctuation and a hanging indent.
No worries.
\lorem[words=20].
\end{document}
```

dropcaps

alerque:
> N.B. As opposed to TeX, the hangafter is not reset at the start of a new paragraph (but being implemented via settings, it's easy to set only temporarily).

In order to work as expected in my redesigned usage (passing the initial letter content as the function content and letting the rest of the paragraph follow naturally) this will need to change to reset itself much like paragraph indent does when you change the current paragraph setting.

Omikhleia:
@alerque Looks great, and the extra options are most welcome. The case for leading punctuation is impressively cool. I'll try to have a closer look and test as soon as I can (might be hard before the end of the week, some tough days to come ^^)

alerque:
@Omikhleia No worries. I'm going to use this is a book project soon and I try to make sure there is a stable SILE release to match, but this week is just ISBN application and I don't expect to go to press before next week. I'll try to hold off stuffing this in a release until you get a chance to review.

By the way you might check your email from a couple days back for special invitation notices from GitHub ;-)

alerque:
@Omikhleia and @simoncozens The function provided here is `\dropcap`, but the package name is currently plural, `packages/dropcaps`. Does that make sense or should we move it to singular `packages/dropcap`? There will only ever be one dropcap function provided, but it is quite likely there will be more functions *related to* dropcaps (such as setting up auto detection of the first letter of paragraphs following chapter functions, something I have working in CaSILE for Turkish and English but needs better language handling before adding here).

packages/dropcaps.lua
```diff
@@ -0,0 +1,90 @@
...
+  local raise = SU.cast("measurement", options.raise or 0)
+  local shift = SU.cast("measurement", options.shift or 0)
+  local size = SU.cast("measurement or nil", options.size or nil)
+  local color = options.color
+  options.size = nil -- we need to measure the "would have been" size before using this
+
+  if color then SILE.require("packages/color") end
+
+  -- We want the drop cap to spanning N lines is N-1 baselineskip plus the height of the first line.
+  -- We Define the height of the first line based on measuring the size the initial would have been.
```
Omikhleia, line 30: (Cosmetic) The comments there have some typos and might be clear. Suggested: 

> We want the drop cap to **span over** N lines, **that** is N-1 baselineskip plus the height of the first line.

> We **d**efine the height (...)

alerque: Thanks, fixed.

packages/dropcaps.lua
```diff
@@ -0,0 +1,92 @@
...
+
+  if color then SILE.require("packages/color") end
+
+  -- We want the drop cap to span over N lines, that is N - 1 baselineskip + the height of the first line.
+  -- Note this only works for the default linespace mechanism.
+  -- We determine the height of the first line by measuring the size the initial content *would have* been.
+  -- This gives the font some control over its relative size, sometimes desired sometimes undesired.
+  local tmpHbox = shapeHbox(options, content)
+  local extraHeight = SILE.measurement((lines - 1).."bs"):tonumber()
+  local targetHeight = tmpHbox.height:tonumber() + extraHeight
```
Omikhleia, line 35: An option here could be interesting... Let me try to explain what I mean, the target height currently is:

image

But as can been seen in this example, the height the initial would have taken is bigger than the line content (in the main text font, cf. especially the "V" later on). That can be checked _without_ dropped initial.

image

Of course, the first line content can be anything, use several fonts that don't match in height, even contain other things than raw text... So it's unlikely there is a general solution here. On the other hand, the main usage should be for text, and wouldn't it be nice if we could get:

image

Here above, the initial fits well with the line top (aligned on the V). What I did here is obviously not general:
```
  local targetHeight = SILE.shaper:measureChar("V").height + extraHeight
```
If you remember my initial commit, that's something I had tried (but measuring a "i" rather than a "V" - kind of _ad hoc_ I admit, and even less general)

Still, I think we could try to address it. The default would still be `tmpHbox.height:tonumber()`, but say for instance we have an option `aligntopon=V` (I can't think of a better name)? Or a better way to find out the height of the subsequent characters.

Perhaps even more general, a command to store some content such as:

```
\dropcap:alignement{any content I want}
``` 
And then
```
  local lineAdjustBox = dpContentForVerticalAlign and shapeHbox({}, dpContentForVerticalAlign) or tmpHbox
  local targetHeight = lineAdjustBox.height + extraHeight
```
So one would have more flexibility adjust the vertical top alignment. E.g.

image

Here I decided to have the initial top-aligned to the smallcaps (which in _this_ font, are of course smaller than a capital, but slightly higher then an "x").

I am not sure how to formalize it (good options or commands for it), and maybe there would be even a nicer way to achieve that result. But it sure would look good to have this kind if flexibility, no? 

EDIT: Fixed a few typos in my comment.

Omikhleia: Yes I understand you viewpoint..After a second thought, why not just a "first line factor" (default 1.0) applied to that `tmpHbox.height:tonumber()` i(and the responsibility of computing the ratio is left to user) ?

On the other point, I suppose you mean the (bottom, now) alignment is not perfect with the last indented line, do you? And that the baselineskip is only indicative as the TeX paragraph building can stretch/expand the actual lineskip, right? I'm bothered by it too (not only for dropcaps). Started to look into alternate line spacing from the "linespacing" package and into ##1118, trying to understand how it works there...

alerque: The optional `size` parameter can already be used if the user wants to calculate their own size, but I see what you mean about a relative scale factor being more ergonomic for some use cases. I just added `scale` (defaulting to 1.0). The [tweaked example shows usage](https://github.com/sile-typesetter/sile-typesetter.github.io/pull/26/commits/0847d4b0d8522b025113ffb4cf2929b90b5e9559), here is the relevant output of that use of scale:

image

Omikhleia:
> > N.B. As opposed to TeX, the hangafter is not reset at the start of a new paragraph (but being implemented via settings, it's easy to set only temporarily).
> 
> In order to work as expected in my redesigned usage (passing the initial letter content as the function content and letting the rest of the paragraph follow naturally) this will need to change to reset itself much like paragraph indent does when you change the current paragraph setting.

Sorry to notice it that late - but this change, if the dropcap is in the main document, breaks any inserts (e.g. footnote if present in the paragraph with a dropcap, folio if a pagebreak occurs in that paragraph) as they inherit the document settings since #3795a4e83823642a69e481cecd15a4966053fd71 

image

packages/dropcaps.lua
Omikhleia: As for the bottom alignment, I think I now get it: We use "bs" units converted an to absolute number, but
- "bs" is derived from "document.baselineskip"
- In the TeX linespacing, "document.baselineskip" as some stretch, and "document.lineskip" may also occur - resulting in an imperfect alignment
- The other linespacing mechanisms don't at all rely on "document.baselineskip", so using "bs" is broken by nature in these cases.

This raises two main questions:
- What is supposed to be a "bs" unit when the linespacing is not the default (TeX) one?
- Either the dropcap must be computed _after_ the paragraph is built (as you implied above), or it seems acceptable by definition that such a paragraph should have stretch between baselines.

alerque:
> if the dropcap is in the main document, breaks any inserts (e.g. footnote if present in the paragraph with a dropcap, folio if a pagebreak occurs in that paragraph) as they inherit the document settings since [3795a4e](https://github.com/sile-typesetter/sile/commit/3795a4e83823642a69e481cecd15a4966053fd71)

Hmmm, my current production project has both situations in the wild (footnotes in first paragraph with dropcap) and pagebreak in the first paragraph. I don't see this issue there, do you mean when *both* those situations are true?

<details>
<summary>One each sample condition (not combined)</summary>

image



</details>

Omikhleia:
> Hmmm, my current production project has both situations in the wild (footnotes in first paragraph with dropcap) and pagebreak in the first paragraph. I don't see this issue there, do you mean when _both_ those situations are true?

One or the other, not necessarily at the same time.

```
\begin[papersize=6in x 9in, class=book]{document}
\script[src=packages/dropcaps]
\script[src=packages/lorem]

\dropcap[lines=2, join=true]{E}ventually,
we will get it right... \lorem[words=50].\footnote{Houston, we got a problem}

One day, \lorem[words=50]\footnote{I am another footnote}

\end{document}
```
image

alerque:
Ah hah! My production project uses CaSILE's implementation of footnotes, not SILE's. This is a function of your recent contributions to SILE's footnotes package, the typesetter of which is now inheriting settings from the document scope. That worked pretty nicely for font family and relative size treatments, but it looks like it is a bit aggressive. We need to either blacklist some settings it should typically *not* inherit or whitelist the ones it should.

alerque:
For now I added a blacklist for core settings that users are pretty unlikely to expect to carry over to footnotes. This isn't an ideal implementation because it is *in* the footnote package and hence any new packages that come along that might have inappropriate settings won't be able to extend this easily.

I have to wonder if perhaps the defaults array might be a better source for footnote settings rather than the top level document settings. Only if people explicitly mark their settings as *default* would they get inherited. In any event that fix can go in another PR.

packages/dropcaps.lua
alerque: For my current project (which uses linespace method `fixed`), I set the baselineskip value to be a copy of what I used for fixed:

```lua

-- Set so bs unit works. Also note dropcaps would be based on this if we didn't manually set them.
SILE.settings.set("document.baselineskip", SILE.settings.get("linespacing.fixed.baselinedistance"))
```

Obviously that's a bit less than ideal, but having this working for the (current) default leading calculator seems like a good enough place to start. We can add detection and support for others later.

Honestly I think the solution to this is to refactor the entire linespacing system so that all of the methods are modular drop in packages that load their own settings and can redefine some generic unit for use in situations like this and/or provide a callback to get absolute values from actual lines rather than predictions. That can go in #1118.

alerque:
Hm, folios needs the same fix.

PR #1220, opened 2021-09-13
Drop examples from source repository

alerque:
Examples are moving to the website repository. See https://github.com/sile-typesetter/sile-typesetter.github.io/pull/23

Few of the examples are even current (many don't even build) so having them in the SILE source repo is counter productive. The raster artifacts are super problematic, and the bloat problem makes it prohibitive to keep the examples fresh.

Moving them to the website and building them in CI will mean we can update the example gallery as fast as contributions come it without worrying about SILE releases or repo bloat.


alerque:
Need to bring back the `simple.sil` example in a way that can be documented as a self-check for people building from sources. That makes a really weak showcase file for the website anyway. Otherwise this just needs some documentation touchups to reference the website.

alerque:
Need to bring back the `simple.sil` example in a way that can be documented as a self-check for people building from sources. That makes a really weak showcase file for the website anyway. Otherwise this just needs some documentation touchups to reference the website.

Omikhleia:
Not sure this was wanted but this commit https://github.com/sile-typesetter/sile/commit/f295974d0ce0cdb542135ce2f2b22185d858fc9e broke something in the 0.12 manual for the url package (some of my earlier text removed, duplicated text and a weird `|||||| 190683a8` sequence.

alerque:
`|||||| 190683a8` looks like the second part of a 3 way merge resolution. That would be my fault somewhere along the lines.

Issue #621, opened 2019-05-25
French conventions with punctuation not respected


OlivierNicole:
I don't know whether this is implemented yet, but I realized that SILE doesn't get the right space before double punctuation signs in French language settings:
```
\begin{document}
\font[family=Linux Libertine, language=fr, size=15pt]
\glue[width=1.1cm]La peine de mort est nécessaire, disent les partisans de l’antique et barbare routine ; sans elle il n’est point de frein assez puissant pour le crime. Qui vous l’a dit ? Avez-vous calculé tous les ressorts par lesquels les lois pénales peuvent agir sur la sensibilité humaine ? Hélas ! avant la mort, combien de douleurs physiques et morales l’homme ne peut-il pas endurer !
\end{document}
```
This code creates an orphan interrogation mark at the beginning of a line:
![Capture d’écran_2019-05-25_23-38-33](https://user-images.githubusercontent.com/14031333/58374861-42671180-7f46-11e9-8dca-be8223c49d78.png)

If you remove the spaces before punctuation in the source, there is no more orphans, but the space before double punctuation signs is too small (there is supposed to be a space before them, more or less thin depending on the typesetting school, but definitely larger than in English).

simoncozens:
Thanks for this report - a really good one to look at, and it is giving me the impetus to clean up and document language-specific tokenization. To get French working optimally, I will need to

- [*] Clean up the Unicode node maker.
- [ ] Test and refactor other languages which have their own node makers.
- [ ] Document how the node maker workers and how it can be extended on a language-specific basis.
- [ ] Write a French node maker.

OlivierNicole:
I'll be happy to help with French-specific advice if needed. I would also help with the code, but I probably need to get a bit more acquainted with it first.

simoncozens:
How does this look?

![Screenshot 2019-05-26 at 14 24 42](https://user-images.githubusercontent.com/106728/58382435-05406500-7fc2-11e9-91af-e8f277c84eb2.png)

See bug-621 branch if you want to play with it.

OlivierNicole:
It seems indeed than I no longer have the orphan interrogation mark. Great!

With this font, I find myself a bit happier with the setting `punctuationsspace` at `0.3en`, since with `0.5en` some of the spaces look larger before the punctuation mark than after, which definitely looks awkward. I don't know if that setting will be very portable across other font sizes and families, however.

[...]

PR #1221, opened 2021-09-14
Fix(languages): Correct punctuation rules for French

Omikhleia:
This PR is my take at #704, starting anew and fresh. 

I used it it on my "test" article and found it neat, with no hanging punctuation at the beginning of lines, and good spaces where they should be. Just an extract :

image

I also tested the "exception" cases:

image
'
I noticed that with some fonts, the shaper could have pre-processed some ligatures (such as !?) as one "item", so I took that too into account.

Any other Frenchman than I, for a double-check ? ;)


Omikhleia:
Hmm, there's something wrong later in my "test" article - a missing hyphenation dash.

image

Not sure how it could be related, though.

Omikhleia:
> Hmm, there's something wrong later in my "test" article - a missing hyphenation dash.

An another one
 
image

The root cause seems to be here:

image

If I remove the penalty, the issue disappears - but it somehow means not using much of that ICU logic... Actually I don't understand that ICU logic here, since we have hyphenation patterns and the Knuth line-breaking algorithm. Could it be that ICU inserts inappropriate breakpoints, or puts the line-breaking algorithm at pain?

Omikhleia:
Worked around the abovementioned issue, added a test case for #704, fixed the two other tests for #621 and #702 after I had checked their correct result with the new punctuation definitions. AFAIC, I am done on this topic for now - moving to the next one !

simoncozens:
Wonderful - thank you!

alerque:
For future reference the conventional-commit scope tags on these commits will make them show up a bit strange in the next release: 95c2398 is good, ed8734c should have been `chore(languages)` or maybe `feat(languages)`, and 8e9b056 should have been `test(languages)`.

Thanks for the expertise and time to contribute this @Omikhleia ! (← thin space)

simoncozens:
I’ve been thinking about this. The original purpose of conventional commit messages for us was to help generate release notes. I’m eager not to turn a maintainer aid into a contributor burden. For the purpose of generating release notes (which deal with changes less granular than individual commits, and more often at the PR level of granularity), we don’t actually need to enforce them on every single commit. And now that master is protected, we get to rewrite the commit message into a PR-level summary when we squash and merge. So why don’t we just require conventional commits, only by maintainers, at squash time?

alerque:
Because we don't require granular enough PR's that we can squash them. In fact they very frequently touch several different things, hence the commit level is usually where the release notes come from, not the PR level. We could have a PR/squash level workflow but we would need to enforce a lot more granular PRs. Personally I think that's more of a contributor burden than commit messages. I can and frequently do fixup commit messages in PRs so I'm not that concerned with contributors getting it right (see e.g. the math PR history vs the original branches!), you just merged this one before I got to it. Whether it is checking the messages in a PR or writing PR squash messages you and I as maintainers are still responsible for this.

PR #1223, opened 2021-09-15
feat(build): Add flake.nix

akavel:
Initial rough flake.nix, allowing to build SILE with Nix (with Flakes
feature enabled). Based on default.nix from nixpkgs:

https://github.com/NixOS/nixpkgs/blob/74b3e9ff87af90f9eae6c4a64ce920babfa3e59f/pkgs/tools/typesetting/sile/default.nix

## Draft:

*The "polishing" tasks I want to try completing before "upgrading" this to a non-draft PR:*

 - [ ] fill in the flake's `description =` attribute + research if I should move the meta information (`homepage =`, etc.) into flake's main block as well;
 - [ ] fix to use a deterministic commit of libtexpdf instead of `master`; I'm thinking to see if I can make the hash be one of the flake's `inputs`
 - [ ] delete the commented-out line with the old `version` attribute's value
 - [ ] maybe drop the `isDarwin` conditional tweaks; the original `default.nix` is marked as `broken = stdenv.isDarwin` anyway; also, current flake is only for `x86_64-linux` anyway; seems more honest to drop the false pretense of somehow trying to support Darwin for now, until someone with a Darwin machine tries to revive that variant.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,25 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1631596344,
+        "narHash": "sha256-Wn61o5G6egxhqyYyZxDw6z1EhpdwM6TCUrP+8roHM6E=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "6f5e9a6f7bc5b83bcfd09f4a2d32e3a35d62fb16",
+        "type": "github"
+      },
+      "original": {
+        "id": "nixpkgs",
+        "type": "indirect"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,131 @@
+{
+  #description = "...",
+
+  outputs = { self, nixpkgs }: {
+    defaultPackage.x86_64-linux =
+      with import nixpkgs { system = "x86_64-linux"; };
+      let
+        luaEnv = lua.withPackages(ps: with ps; [
+          cassowary
+          cosmo
+          compat53
+          linenoise
+          lpeg
+          lua-zlib
+          lua_cliargs
+          luaepnf
+          luaexpat
+          luafilesystem
+          luarepl
+          luasec
+          luasocket
+          luautf8
+          penlight
+          stdlib
+          vstruct
+        ]);
+        libtexpdf-src = fetchFromGitHub {
+          owner = "sile-typesetter";
+          repo = "libtexpdf";
+          # FIXME(akavel): use specific hash, not master which can move
+          rev = "master";
+          sha256 = "Y2NdojNSjSrTeVPO/Vih+ZT+bYsN50jllSVfwL3RDMk=";
+        };
+      in
+
+      stdenv.mkDerivation rec {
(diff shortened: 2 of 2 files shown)


suhr:
Overall, looks good.

ctrlcctrlv:
In case anyone else is wondering what this file is for, [this page explains it](https://nixos.wiki/wiki/Flakes). It's for an OS called NixOS and allows you to integrate software you install from Git repositories with Nix's package management.

alerque:
This should be perfect for being able to use HEAD builds to update the website examples!

Any input on this @doronbehar? I am completely unfamiliar with flakes. I'm guessing there are a few things that should be adjusted from the `default.nix` layout though. Also, once this is merged (and/or makes it to a tagged release) is there anything that should be adjusted in the nixpkgs repo?

akavel:
@alerque would you possibly fancy elaborating a bit more on what do you mean by "a few things that should be adjusted from the `default.nix` layout"? I tried to think what you could mean by this, but I'm afraid I have no clue as of yet :/

doronbehar:
My suggestion is a bit different: https://github.com/sile-typesetter/sile/pull/1231 . Mainly, it is relying on Nixpkgs' `default.nix` and not rewriting it's contents here as well. I also used `flake-compat` for legacy nix support.

> Also, once this is merged (and/or makes it to a tagged release) is there anything that should be adjusted in the nixpkgs repo?

@alerque not really. Ideally there was some support in Nixpkgs to import package's build instructions from the package's `flake.nix` files, but it's not possible yet. It's called IFD, see:

- https://nixos.wiki/wiki/Import_From_Derivation
- https://discourse.nixos.org/t/is-importing-a-derivation-from-another-repository-bad-practice/5016/4

That's why I chose to use Nixpkgs' build instructions here and not the other way around.

> - [ ] fix to use a deterministic commit of libtexpdf instead of master; I'm thinking to see if I can make the hash be one of the flake's inputs

I think it's better to keep the management of `libtexpdf`'s commit in the hands of non Nix tools, so other contributors will be able to deal with it if needed without using Nix. It still should be noted though, that currently, in my PR as well, git submodules, should be checked out when trying to build `sile` from source. I wonder how would that play with a user wishing to run `sile` with:

```
nix run github:sile-typesetter/sile#sile
```

As I see it, the tasks for you @alerque is to think how to handle the fonts, and the git submodule along with nix, so the command above could be written in the README :).

akavel:
I'm more than fine with closing this PR in favor of a different one, esp. seeing that @doronbehar seems more advanced in Nix than me. That said, I'm in something of a pickle with #1231's  "`nix build` not working": for me personally, being able to compile it easily with `nix build` locally, for development, was the very whole point of trying to add a Nix wrapper :/ so #1231 as is now would kinda make my situation harder than before, due to having to *delete* its `flake.nix` before pasting-in the one that works locally for me :/ As such, is it possible we could try and make it work for local building? E.g., could we add the fonts into `buildInputs` (I'll add this suggestion as a comment on #1231)? I'm happy to try and help with that, though as I said, I'm not super experienced with Nix :cry:

alerque:
> I tried to think what you could mean by this, but I'm afraid I have no clue as of yet :/

The re-used of the dependency list and other things in the other PR was a lot more along the lines of what I was expecting but didn't know how to articulate. I hope you don't mind I close this in favor of the other PR. I do appreciate getting the ball rolling on this!

I'll hold off merging the other PR though until you confirm it actually works for your use case.

PR #1230, opened 2021-09-17
fix(math): Cope with MathVariant coverage tables being NULL

### LINKED ISSUE #1222: Unsupported coverage table format 100
I have a simple text document with at its beginning, a global:

`\font[family=Linux Libertine O]`

SILE 0.11.1 = works fine, and the document _seems_ to be ok.

Current master (as of 0cb304f7eba46b81a1e46914ccc7603ea18e08cb) = aborts on error:

`! Unsupported coverage table format 100 at test.sil:4:1: in \fonttable: 0x560e3a4016e0`

Something might have been broken on master - maybe it should just be a warning and/or have some more clear message?

(FWIW, I'm switching my documents to Libertinus - which doesn't have this issue - but just reporting an issue in case some other fonts people might be tempted to use just stop working. I don't really care about "Linux Libertine O", but other fonts might fail too with this obscure message, I guess, and it's always good to know about breaking changes if it is not addressed.

simoncozens:
Fixes #1222


alerque:
The commit scope for this should be `chore(math): Fixed....`. You can use "Fixed blah" as the human readable message, but if you make that the scope it will show up in the release notes as a bug fix. That would be appropriate if you were fixing something that was released, but not when patching up a feature that isn't released yet.

alerque (changes requested):
Looks good to me other than the message. Lets fix that so the release notes are readable without a bunch more interventions. It's really nice to be able to release without a lot of fuss and work on release day.

alerque (approved):

Issue #1212, opened 2021-09-09
Docker image `siletypesetter/sile:latest` seems broken?


ctrlcctrlv:
```
[fred@🍇葡萄🍇SILE]$ docker run -it --volume "/usr/share/fonts:/fonts" --volume "$(pwd):/data" --user "$(id -u):$(id -g)" siletypesetter/sile:latest
Unable to find image 'siletypesetter/sile:latest' locally
latest: Pulling from siletypesetter/sile
3cebb2be42a5: Pull complete 
ea4b88892555: Pull complete 
207849b83b60: Pull complete 
58653039bd47: Pull complete 
87ffd86569c5: Pull complete 
8c895c379b4e: Pull complete 
7106344c4f72: Pull complete 
4f4fb700ef54: Pull complete 
fe860da897c2: Pull complete 
Digest: sha256:8641d6ee4bee962d24fd82a3cd1db433a0dd8c86ceadc6d36520677486aaede5
Status: Downloaded newer image for siletypesetter/sile:latest
SILE v0.11.1 (Lua 5.4)
/usr/sbin/lua: error loading module 'justenoughlibtexpdf' from file './core/justenoughlibtexpdf.so':
	libtexpdf.so.0: cannot open shared object file: No such file or directory
stack traceback:
	[C]: in ?
	[C]: in function 'require'
	./core/libtexpdf-output.lua:1: in main chunk
	[C]: in function 'require'
	./core/sile.lua:77: in function 'core/sile.init'
	/usr/local/bin/sile:44: in main chunk
	[C]: in ?
```

alerque:
I am unable to replicate this. I even get the same Digest checksum for the image I downloaded.

```console
$ docker pull siletypesetter/sile:latest
latest: Pulling from siletypesetter/sile
Digest: sha256:8641d6ee4bee962d24fd82a3cd1db433a0dd8c86ceadc6d36520677486aaede5
Status: Image is up to date for siletypesetter/sile:latest
docker.io/siletypesetter/sile:latest
$ docker run -it --volume "/usr/share/fonts:/fonts" --volume "$(pwd):/data" --user "$(id -u):$(id -g)" siletypesetter/sile:latest
SILE v0.11.1 (Lua 5.4)
> SU.dump{"seems good?"}
{
  "seems good?"
>
```

I also use the same image to build some actual SIL files and got valid PDFs.

[...]

PR #1231, opened 2021-09-17
Add flake.nix and other files

doronbehar:
Use `flake-compat` for `shell.nix` and `default.nix` compatibility.

~~Currently `nix build -L` doesn't work, due to the fonts not available in the source, and they can't be downloaded in the sandbox.~~


flake.nix
```diff
@@ -0,0 +1,55 @@
...
+  flake-utils.lib.eachDefaultSystem (system:
+    let
+      pkgs = import nixpkgs {
+        inherit system;
+      };
+      # Use the expression from Nixpkgs instead of rewriting it here.
+      sile = pkgs.sile.overrideAttrs(oldAttr: {
+        version = "${(pkgs.lib.importJSON ./package.json).version}-flake";
+        # TODO: Building from source doesn't work yet as the fonts are not
+        # available during build and can't be downloaded in the sandbox.
```
akavel, line 28: Could we add the fonts into `buildInputs` (maybe as in #1223's FONTCONFIG_FILE)? (And patch out `include Makefile-fonts` + `LOCALTESTFONTS` if necessary, again maybe as in #1223?)

Honestly, for me personally, the whole point of needing a flake.nix is to be able to build locally to make development on #220 easier...

akavel:
Hm; if `nix build` doesn't work with this flake, could you help me understand how can I actually use it? I'm not a super experienced Nix user, and not sure what to do now to take advantage of having this flake.nix after I pulled this PR locally :/

flake.nix
```diff
@@ -0,0 +1,55 @@
...
+        src = builtins.filterSource
+          (path: type: type != "directory" || baseNameOf path != ".git")
+          ./.;
+        nativeBuildInputs = oldAttr.nativeBuildInputs ++ [
+          pkgs.autoreconfHook
+        ];
+        # TODO: This change can be updated in nixpkgs on the next version bump
+        postPatch = oldAttr.preConfigure;
```
akavel, line 36: thought: could it make sense to try and modify .gitmodules file to have a specific revision of libtexpdf instead of master, then try to extract the revision from `builtins.readFile ./.gitmodules` and fetchGitHub it and copy into ./libtexpdf directory?

alerque:
I haven't had time to review this in detail, but thanks for the contribution!

> Currently `nix build -L` doesn't work, due to the fonts not available in the source, and they can't be downloaded in the sandbox.

A quick note on fonts, there is already a `gentium` nixpkg which should be the only font needed to build and run SILE out of the box. The other two sets of fonts are for building the manual (which is a time consuming process that does not need to be done for most SILE builds) and for test fixtures (also not needed if what you want is to run SILE). A third group for building examples is being removed from this repo shortly anyway (see #1220).

If anybody is doing development work on SILE itself I would expect them to have the source checked out anyway and be able to do the font downloads on the host system on in a nix shell (with curl). They should not get in the way of *running* a dev version.

doronbehar:
@akavel I'm not sure I'm more experienced in Nix then you :) That `FONTCONFIG_FILE` was a good piece missing (excuse me for not reading thoroughly your PR and noticing that the build in your version works). I haven't added that attribute though, because I disabled the build of the manual, so no fonts are needed at build at all.

The current usages of this flake are:

1. Test build that is similar to what would happen in Nixpkgs, with `nix build -L`.
2. Enter a development environment shell with `nix develop`.
3. Allow Nix users to test drive the software easily with `nix run github:sile-typesetter/sile`.

Feature 3 doesn't work, due to the Git Submodule issue, but we can test it once we decide how to tackle it with `nix run github:doronbehar/sile/nixFlake`.

@akavel `nix shell` failed for you righteously because the build was broken. `nix shell` is meant for entering a shell where the program `sile` is available. Perhaps now it will work for you, with the build working? I forced pushed.

doronbehar:
Some ideas for tackling `libtexpdf`:

1. Add it as a flake input (pros: easy handling inside this flake.nix) but it might go out of sync with the submodule.
2. Add it as a package to nixpkgs and add it as a `buildInput`. Might go out of sync with the submodule as well..
3. Add it with `fetchFromGitHub` inside this `flake.nix` like in #1223 but then the hash there would have to be updated every time it's version is bumped.

akavel:
@doronbehar I'm afraid `nix build` still doesn't work for me :cry:  :

```console
$ git fetch origin pull/1231/head:pr1231.02
$ git checkout pr1231.02
$  nix build
error: builder for '/nix/store/v96hg6vzn6pd3xnivkv7hr4ivacgpbqp-sile-0.11.1-flake.drv' failed with exit code 2;
       last 10 log lines:
       > config.status: creating sile
       > config.status: executing depfiles commands
       > config.status: executing libtool commands
       > Making all in libtexpdf
       > make[2]: Entering directory '/build/xblbbmm57jdrjb7va26va9ggdblzp19a-source/libtexpdf'
       > make[2]: *** No rule to make target 'all'.  Stop.
       > make[2]: Leaving directory '/build/xblbbmm57jdrjb7va26va9ggdblzp19a-source/libtexpdf'
       > make[1]: *** [Makefile:772: all-recursive] Error 1
       > make[1]: Leaving directory '/build/xblbbmm57jdrjb7va26va9ggdblzp19a-source'
       > make: *** [Makefile:501: all] Error 2
       For full logs, run 'nix log /nix/store/v96hg6vzn6pd3xnivkv7hr4ivacgpbqp-sile-0.11.1-flake.drv'.
$ git describe
v0.11.1-141-gd0174488
$ ls -1 libtexpdf/ | wc -l
133

**Comment by alerque** [comment]
Of the 3 options there, ② makes the most sense long term. We're planning on splitting it out with it's own release cycle anyway. I've already setup the Arch Linux packaging getting it ready for a split for that to be it's own library package. Right now we *always* have it pinned to the *master* branch HEAD, but as soon as we start tagging releases there it will just be pinned to whatever the latest tag is.

**Comment by doronbehar** [comment]
> Of the 3 options there, ② makes the most sense long term. We're planning on splitting it out with it's own release cycle anyway. I've already setup the Arch Linux packaging getting it ready for a split for that to be it's own library package. Right now we _always_ have it pinned to the _master_ branch HEAD, but as soon as we start tagging releases there it will just be pinned to whatever the latest tag is.
> 
> Option 3 is also viable. I have a make job to confirm the submodule is up to date, I can have it update the hash in the flake too.

Perhaps it will be convenient to have a libtexpdf as a submodule for now, or perhaps forever, so it'd be easier to test changes in that library while developing? I added a TODO in the code. Now, using `builtins.fetchGit`, this works for me:

```
nix run -L --refresh github:doronbehar/sile/nixFlake
```

And also the README is updated with a similar command

And maintaining this flake.nix only requires making sure that the file `./libtexpdf.git-rev` has the same git rev as the submodule. No need to update a sha256, as `builtins.fetchGit` is used, and not `fetchFromGitHub`.

@akavel I hope _now_ it will work for you :pray: .

File: flake.nix (hunk 1 above)
Comment (doronbehar) [reply flake.nix]: > Honestly, for me personally, the whole point of needing a flake.nix is to be able to build locally to make development on #220 easier...

Does sile need certain fonts available during sile's build in order to test math support? Or could they be available only when sile is building documents?

**Review by doronbehar — COMMENTED** [review]
One thing that still bothers me @alerque is that when I `nix run . -- --version` I get:

```
SILE vUNKNOWN (Lua 5.2)
```

In which part of the build would it be correct to write `.version` or `.tarball-version` so that it'd be picked up by the build?

File: flake.nix (hunk 1 above)
Comment (alerque) [reply flake.nix]: Not to build no. If you want to run the regression test suite then it needs not only a MATH font but a whole army of other fonts besides.

However the default math font is *Libertinus Math*, so if people try to test any math stuff they will either need to use a font they have or get that. I don't know if we should start baking that is as a requirement, but since it is used in a package not in core I think it's probably okay that we just make the user supply a font rather than making everybody install one as a dependency.

**Comment by alerque** [comment]
> In which part of the build would it be correct to write `.version` or `.tarball-version` so that it'd be picked up by the build?

Does the build environment *see* the rest of the repository (i.e. the `.git` dir)? If so then something else is wrong. If not then I think you'll need to pretend to be a source-dist by stuffing the version in `.tarball-version`. If you skip ahead to `.version` (which is where it eventually gets used from) the configure/make process is likely to blow it away.

Or if needed I can come up with another mechanism.

File: flake.nix (hunk 1 above)
Comment (doronbehar) [reply flake.nix]: We could suggest a `FONTCONFIG_FILE` in `.envrc` or in `devShell`, but that could be part of a future PR. The only issue is that Nix users don't like fonts to be downloaded during build, so hopefully we'll be able to rely upon fonts from Nixpkgs.

**Comment by doronbehar** [comment]
> > In which part of the build would it be correct to write `.version` or `.tarball-version` so that it'd be picked up by the build?
> 
> Does the build environment _see_ the rest of the repository (i.e. the `.git` dir)? 

No. Putting the version in `.tarball-version` worked.

File: flake.nix (hunk 1 above)
Comment (alerque) [reply flake.nix]: Understood, and yes I agree.

By default the build system shouldn't muck about with downloading any fonts at all provided the manual and tests are both disabled at configure time. The only other thing that is *optionally* online during the build is the vendored Luarocks option, but that's irrelevant to Nix since it uses system supplied dependencies for those already.

**Comment by alerque** [comment]
I'm having trouble testing this locally too. Interestingly I can use `nix run github:doronbehar/sile/nixFlake` and it seems to run great. What I can't seem to do is any incantation to run it from the Git working directory:

```console
nix run .
/nix/store/g786j9lkivd7fida2hbz6gncvl1mk4r7-lua-5.2.4/bin/lua: error loading module 'justenoughicu' from file './core/justenoughicu.so':
        libicuio.so.69: cannot open shared object file: No such file or directory
stack traceback:
        [C]: in ?
        [C]: in function 'require'
        ./core/utilities.lua:496: in main chunk
        [C]: in function 'require'
        ./core/sile.lua:40: in main chunk
        [C]: in function 'require'
        ...a272cdbvayq834axmlq1439zhkirl-sile-0.11.1-flake/bin/sile:36: in main chunk
        [C]: in ?
```

Any idea what's up with that? It seems like they should be the same thing since my local branch *is* what is in this PR branch right now.

flake.nix
```diff
@@ -0,0 +1,74 @@
...
+      # TODO: Should this be replaced with libtexpdf package from nixpkgs? or
+      # should we keep it that way, so that it'd be easy to test new versions
+      # of libtexpdf when developing?
+      libtexpdf-src = builtins.fetchGit {
+        url = "https://github.com/sile-typesetter/libtexpdf";
+        rev = "${(pkgs.lib.fileContents "${self}/libtexpdf.git-rev")}";
+      };
+      # Use the expression from Nixpkgs instead of rewriting it here.
+      sile = pkgs.sile.overrideAttrs(oldAttr: rec {
+        version = "${(pkgs.lib.importJSON ./package.json).version}-flake";
```
alerque, line 33: This is a nice touch and definitely better than not having it at all. However I'm wondering if we can do one better. Do you have access to the actual commit SHA at this point? Normally when building from Git, SILE will identify itself with a counter and short SHA of the sources it was built from, e.g.:

```console
$ SILE v0.11.1.r144-g8935c52 (Lua 5.4)
```

Obviously this is based on `git describe` output, but I think those are useful bits to know.

Not a big deal if there isn't an obvious way to do it, but since I'm planning on using this Flake to build the SILE website it would probably be useful to see from CI which version was used to generate examples.

doronbehar: That wasn't trivial, because `.git` is not available inside the source during build. I did found a way though. I pushed a fixup commit for this.

doronbehar:
> Any idea what's up with that? It seems like they should be the same thing since my local branch _is_ what is in this PR branch right now

No idea :/ Did you try perhaps using `nix run --refresh .`?

flake.nix
doronbehar: BTW would you like to add packages of sile built with different `lua` versions? Currently Nixpkgs defaults to lua 5.2

akavel:
Woohoo @doronbehar @alerque , worked for me now!!! :tada: :heart: 

Both `nix build` and `nix run .` were successful on my machine; weird that it didn't work for @alerque :/ could some impurity find its way into the build? sounds ugly, given the whole point of Nix is to not have situations like this :/

flake.nix
akavel: Would be cool indeed; the current `SILE v0.11.1-flake (Lua 5.2)` is kinda less than ideal in that it doesn't show it's not *really* 0.11.1, there are quite some commits over it already. But for me, I'm happy this PR is making my life easier, even if imperfect, so I personally (FWIW) see it as something that could be perfected post-merge.

With that said, I seem to vaguely recall there were in theory some extra "magic" attributes that are maybe available in a flake expression with some extra info; though I wasn't able to google them up yet. One thing I found is ["Flake reference attributes"](https://github.com/NixOS/nix/blob/master/src/nix/flake.md#flake-reference-attributes), but I didn't manage to access them; maybe those are only for inputs? Dunno. I always have trouble finding stuff in Nix docs *\<sigh>*

akavel:
@doronbehar oh, sweet, the `self.rev` trick is so awesome!!!

```console
$ nix run .
warning: Git tree '/home/akavel/prog/sile' is dirty
SILE v0.11.1-dirty-flake (Lua 5.2)
> 

$ git stash
Saved working directory and index state WIP on pr1231.04: 789be0b4 fixup! feat(tooling): Enable use as a Nix flake

$ nix run .
SILE v0.11.1-789be0b4-flake (Lua 5.2)
> 
```

flake.nix
alerque: Re Lua versions: https://github.com/NixOS/nixpkgs/issues/138420

alerque:
> No idea :/ Did you try perhaps using `nix run --refresh .`?

Yes I tried that.

Playing with it some more now, it seems that the trouble is the working directory I was in has a *built* version of SILE — built on my host system, Arch Linux. If I blow that away with `git clean`¹ then a then `nix run .` works fine. For some reason if it finds the built stuff in the current repository it tries to use them, and of course they are a binary miss-match for the system. I presume this is actually SILE's fault for looking at the CWD first, system later.

¹ <sub>`git clean -dxff -e node_modules -e .fonts -e .sources -e .husky`</sub>

flake.nix
```diff
@@ -0,0 +1,76 @@
...
+        rev = "${(pkgs.lib.fileContents "${self}/libtexpdf.git-rev")}";
+      };
+      # https://discourse.nixos.org/t/passing-git-commit-hash-and-tag-to-build-with-flakes/11355/2
+      version_rev = if (self ? rev) then (builtins.substring 0 8 self.rev) else "dirty";
+      # Use the expression from Nixpkgs instead of rewriting it here.
+      sile = pkgs.sile.overrideAttrs(oldAttr: rec {
+        version = "${(pkgs.lib.importJSON ./package.json).version}-${version_rev}-flake";
+        src = builtins.filterSource
+          (path: type: type != "directory" || baseNameOf path != ".git")
+        ./.;
```
doronbehar, line 38: > `git clean -dxff -e node_modules -e .fonts -e .sources -e .husky`

Perhaps we should filter these files as well right here?

ctrlcctrlv:
> I presume this is actually SILE's fault for looking at the CWD first, system later.

That is also the root cause of #1212, but just affecting you this time and not me. :laughing:

flake.nix
alerque: I don't think that will help, if `$CWD` is still set to the project directory when it is actually run it won't matter what was or wasn't copied into the Nix build, SILE will still prefer local resources to it's distribution ones. This is intentional so that users can override virtually anything in the system, but it does make the assumption that any resources in `$CWD` should be usable.

In any case this is a developer only problem and I'm not too worried about it. I've scheduled this for merge so I can test it out on the SILE website before tagging a release. We can issue any adjustments for more developer convenience in a follow-up.

alerque: As @ctrlcctrlv noted, this is really the same as #1212. I'm not sure what the solution should even be (maybe nothing), but lets move discussion there.

doronbehar:
One thing that bothers me, is that the `src` is changed a lot, causing rebuilds to be frequent, even with changes such as adding a comment to `flake.nix`, even with the current `filterSource`. I asked a question at https://discourse.nixos.org/t/how-to-make-src-in-a-flake-nix-not-change-a-lot/15129

PR #1232, opened 2021-09-17
Correct Armenian support to use ISO 639 code ‘hy’

alerque:
Hey @simoncozens I'm not sure what bottle of good stuff you opened on August 16th, 2018 when you added 3c0dfe35 but if you have any more of it around feel free to invite me ;-)


simoncozens (approved):
Not a clue. But yeah, `hy` seems like a better option for that. :-)

Issue #1229, opened 2021-09-17
URL and line breaks


Omikhleia:
Nowadays, it's a common practice to have URLs in print article (whether it is a _good_ practice or not is another topic).

The **url** package says it will 
> automatically insert breakpoints into unwieldy URLs, 

.... but it actually breaks only after `/` by default.

I think it could be made more powerful and I would suggest:
- Removing the `breakpat` option on `\href` and `\url`: 
  - As currently implemented, it could be any regular expression, which makes the urlFilter some some sort of badly named 
    generic filter
  - It's hard to use in SIL language (not even tried to check we could escape `[` and `]` in command options!)
  - Some breaks should be before, some after (see below)
- Implementing breakpoints at approriate places:
  - After:
    -  `:`, `/`, `.`, `;` (URL path elements)
    -  `?`, `&`, `=` (URL query parts)
    -  `_`, `-` (goodies that feel safe breaking at, e.g. https://en.wikipedia.org/wiki/Alan_Turing, https://github.com/sile-typesetter/sile#what-is-sile)
    - maybe `!` (IIRC the `#!` is/was used by Twitter, Facebook etc. I am not on these social networks, but heh why not)
  - Before:
    -  `%` (URL escape sequence)
    -  `#` (URL fragment)
    - I am not an expert at penalties, but maybe some of the above "After" elements could be breaks before, with a lower (higher?) penalty. I guess i'll need to read more about penalties.
      I mean, _in the worst case_, a line break before them could be acceptable.
- Temporarilly switching language to 'und'
  - To avoid hyphens being inserted.
    (While bits of an URL can be actual words, in most international URLs several languages can be mixed, and
    for latin-script languages, accents are not used, so it can become quite messy anyway)
  - To avoid language-specific formatting: currently that might affect French only, but one doesn't want the typography rules
    (space before `:`, `?` etc.)
  - If this change is felt too drastic,
    - Either do it for French only...
    - Or allow an option `language` to be set (and one author could take the decision to override 'und' 
      with  `SILE.setCommandDefaults` if he really wants hyphenation rules to apply...)

I will be doing this (at least some parts) for my own needs, so I you feel this could become part of the default **url** package,
assign the issue to me and I'll propose a PR at some point :D

simoncozens:
I like this a lot! I don't think it's just a French language issue, so please do it for everything. :-)

alerque:
Sounds good to me too.

alerque:
Resolved by #1233.

PR #1233, opened 2021-09-18
feat(packages): URLs are allowed many more breakpoints

Omikhleia:
Here is my proposal for #1229 - Compared to the ticket, I have ended up adding a bit more flexibility. Sounds good to me in all my tests so far (but I can't have a test submitted, as the url package needs the pdf package which needs the libtexpdf backend, not the debug backend - If I am not mistaken).

Besides a brute force test (see below), looks decent too in my "test" article:

image

(I would have preferred the `_` tto have a break after it, not before, but toying with the penalty settings, it was not much better with a very dense line, so I think they are quite OK finally... But of course the default penalties are at best a wild guess)

The brute force test, if one wants to play with it - 12 pages of URLs :smile_cat: - the decent scenario being of course the last.

```
\begin[papersize=6in x 9in]{document}
\script[src=packages/url]

\language[main=en]

1. Check the undocumented href is not broken

\href{https://my.site.org/proposition;2021/sile/test_url?name=Jack&age=50#hash-my-anchor}

2. Test url without language (default)

\script{
    local url = "https://my.site.org/proposition;2021/sile/test_url?name=John%20Doe&age=50#hash-my-anchor"
    for i = 7, 74 do
      SILE.typesetter:leaveHmode()
      for j = 1, i do
         SILE.typesetter:typeset("x")
      end 
      SILE.typesetter:typeset(" ")
      SILE.call("url", {}, { url }) 
    end
}

3a. test url with a language (English)

\script{
    local url = "https://my.site.org/proposition;2021/sile/test_url?name=John%20Doe&age=50#hash-my-anchor"
    for i = 7, 74 do
      SILE.typesetter:leaveHmode()
      for j = 1, i do
         SILE.typesetter:typeset("x")
      end 
      SILE.typesetter:typeset(" ")
      SILE.call("url", { language = "en" }, { url }) 
    end
}

\language[main=fr]

3b. test url with French: punctuation spacing rules should be disabled; except in this line!

\script{
    local url = "https://my.site.org/proposition;2021/sile/test_url?name=John%20Doe&age=50#hash-my-anchor"
    for i = 7, 74 do
      SILE.typesetter:leaveHmode()
      for j = 1, i do
         SILE.typesetter:typeset("x")
      end 
      SILE.typesetter:typeset(" ")
      SILE.call("url", { language = "fr" }, { url }) 
    end
}

\language[main=en]

4. test url with some flexible text content

\script{
    local url = "https://my.site.org/proposition;2021/sile/test_url?name=John%20Doe&age=50#hash-my-anchor"
    for i = 4, 44 do
      SILE.typesetter:leaveHmode()
      for j = 1, i do
         SILE.typesetter:typeset("x ")
      end 
      SILE.typesetter:typeset(" ")
      SILE.call("url", {}, { url }) 
    end
}

\end{document}
```

### DIFF
--- a/packages/url.lua
+++ b/packages/url.lua
@@ -5,22 +5,76 @@ if pdf then SILE.require("packages/pdf") end
 
 local inputfilter = SILE.require("packages/inputfilter").exports
 
-local urlFilter = function (node, content, breakpat)
+-- URL escape sequence, URL fragment:
+local preferBreakBefore = "%#"
+-- URL path elements, URL query arguments, acceptable extras:
+local preferBreakAfter = ":/.;?&=!_-"
+-- URL scheme:
+local alwaysBreakAfter = ":" -- Must have only one character here!
+
+local escapeRegExpMinimal = function (str)
+  -- Minimalist = just what's needed for the above strings
+  return string.gsub(str, '([%.%?%-%%])', '%%%1')
+end
+
+local breakPattern = "["..escapeRegExpMinimal(preferBreakBefore..preferBreakAfter..alwaysBreakAfter).."]"
+
+local urlFilter = function (node, content, options)
   if type(node) == "table" then return node end
+
   local result = {}
-  for token in SU.gtoke(node, breakpat) do
+  for token in SU.gtoke(node, breakPattern) do
     if token.string then
       result[#result+1] = token.string
     else
-      result[#result+1] = token.separator
-      result[#result+1] = inputfilter.createCommand(
-        content.pos, content.col, content.line,
-        "penalty", { penalty = 100 }, nil
-      )
+      if string.find(preferBreakBefore, escapeRegExpMinimal(token.separator)) then
+        -- Accepts breaking before, and at the extreme worst after.
+        result[#result+1] = inputfilter.createCommand(
+          content.pos, content.col, content.line,
+          "penalty", { penalty = options.primaryPenalty }, nil
+        )
+        result[#result+1] = token.separator
+        result[#result+1] = inputfilter.createCommand(
+          content.pos, content.col, content.line,
+          "penalty", { penalty = options.worsePenalty }, nil
+        )
+      elseif token.separator == alwaysBreakAfter then
+        -- Accept breaking after (only).
+        result[#result+1] = token.separator
+        result[#result+1] = inputfilter.createCommand(
+          content.pos, content.col, content.line,
+          "penalty", { penalty = options.primaryPenalty }, nil
+        )
+      else
+        -- Accept breaking after, but tolerate breaking before.
+        result[#result+1] = inputfilter.createCommand(
+          content.pos, content.col, content.line,
+          "penalty", { penalty = options.secondaryPenalty }, nil
+        )
+        result[#result+1] = token.separator
+        result[#result+1] = inputfilter.createCommand(
+          content.pos, content.col, content.line,
+          "penalty", { penalty = options.primaryPenalty }, nil
+        )
+      end
     end
   end
   return result
 end
+
+SILE.settings.declare({
+  parameter = "url.linebreak.primaryPenalty",
+  type = "integer",
+  default = 100,
+  help = "Penalty for breaking lines in URLs at preferred breakpoints"
+})
+
+SILE.settings.declare({
+  parameter = "url.linebreak.secondaryPenalty",
+  type = "integer",
+  default = 200,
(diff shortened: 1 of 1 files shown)


packages/url.lua
```diff
@@ -32,25 +125,55 @@ SILE.registerCommand("href", function (options, content)
       borderoffset = options.borderoffset },
       content)
   else
+    -- FIXME
```
alerque, line 128: I'm not a big fan of "fixme" comments it the code as they are hard to discuss and track. Lets add any user facing documentation to the manual, any developer-only explanation as regular comments, and then open issues for any todo/fixme type comments or anything else that might generate developer discussion.

packages/url.lua
```diff
@@ -5,23 +5,116 @@ if pdf then SILE.require("packages/pdf") end
 
 local inputfilter = SILE.require("packages/inputfilter").exports
 
-local urlFilter = function (node, content, breakpat)
+--[[
+  FIXME: For reference, but likely too long for the SILE Manual.
```
alerque, line 9: See other comment, but "fixme" stuff that either needs developer work or perhaps discussion should be an issue.

Lets split the rest of this up into a) code comments for technical explanation or b) documentation in the manual for anything users would benefit from knowing. I'm not at all afraid of the manual being too long. People can skip the documentation for packages that aren't interesting to them.

packages/url.lua
```diff
@@ -32,25 +125,55 @@ SILE.registerCommand("href", function (options, content)
...
+    else
+      SILE.settings.set("document.language", 'und')
+    end
+
+    local result = inputfilter.transformContent(content, urlFilter, {
+      primaryPenalty = primaryPenalty,
+      secondaryPenalty = secondaryPenalty,
+      worsePenalty = worsePenalty
+    })
+    SILE.call("code", {}, result) -- important: do not pass the options through
```
alerque, line 172: Would there be any practical use for passing the options through? We can always clear whatever font/language property is an issue and pass everything else through — but only if there is any conceivable use case for that being a net benefit.

alerque (changes requested):
The substance of what this does looks great, thanks for the contribution. Lets just shuffle the documentation around a little bit.

packages/url.lua
Omikhleia: Agreed, fixed.

packages/url.lua
Omikhleia: Comment removed, ok.

packages/url.lua
Omikhleia: Right. So #1239 opened, and FIXME comment removed here, and missing documentation added.

packages/url.lua
```diff
@@ -33,43 +87,90 @@ SILE.registerCommand("href", function (options, content)
...
+      if options.language == "fr" then
+        -- Trick the engine by declaring a "fake"" language that doesn't apply
+        -- the typographic rules for punctuations
+        SILE.hyphenator.languages["_fr_noSpacingRules"] = SILE.hyphenator.languages.fr
+        -- Not needed (the engine already defaults to SILE.nodeMakers.unicode if
+        -- the language is not found):
+        -- SILE.nodeMakers._fr_noSpacingRules = SILE.nodeMakers.unicode
+        SILE.settings.set("document.language", "_fr_noSpacingRules")
+      else
+        SILE.settings.set("document.language", options.language)
```
alerque, line 118: Isn't this effectively a noop?

packages/url.lua
```diff
@@ -33,43 +87,90 @@ SILE.registerCommand("href", function (options, content)
...
-  SILE.call("code", {}, result)
-end)
+  SILE.settings.temporarily(function ()
+    local primaryPenalty = SILE.settings.get("url.linebreak.primaryPenalty")
+    local secondaryPenalty = SILE.settings.get("url.linebreak.secondaryPenalty")
+    local worsePenalty = primaryPenalty + secondaryPenalty
+
+    if options.language then
+      SILE.languageSupport.loadLanguage(options.language)
+      if options.language == "fr" then
```
alerque, line 109: I understand French needs special handling here, but this doesn't feel like the right place to set this up. We're almost certain to run into other languages with similar needs, and having language specific decision trees in packages doesn't feel right.

I am okay with merging this as is to get the ball rolling, but I think we should consider refactoring this so that the language support modules have a generic flag of some kind for applying (or not applying) more invasive typographical details.

Right now we use `und` to disable hyphenation, but that also disables other language features. We should probably offer better control over this.

alerque (approved):

packages/url.lua
alerque: I opened an issue #1242 for this topic.

PR #1238, opened 2021-09-18
Test Nix functionality

alerque:
As a follow up to #1231, setup a CI workflow to test that the Nix Flake and other features stay in working order.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,3 +57,29 @@ jobs:
             sile-${{ env.VERSION }}.pdf
             sile-${{ env.VERSION }}.zip
             sile-${{ env.VERSION }}.tar.xz
+
+  nix-flake:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@v14
+        # https://github.com/cachix/install-nix-action/#usage-with-flakes
+        with:
+          install_url: https://nixos-nix-install-tests.cachix.org/serve/i6laym9jw3wg9mw6ncyrk6gjx4l34vvx/install
+          install_options: '--tarball-url-prefix https://nixos-nix-install-tests.cachix.org/serve'
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+      - name: Check nix build
+        run: |
+          nix build
+      - name: Check nix flake
+        run: |
+          nix flake check
+      - name: Check nix shell
+        run: |
+          nix shell -c true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,8 +72,22 @@ jobs:
         with:
           install_url: https://nixos-nix-install-tests.cachix.org/serve/i6laym9jw3wg9mw6ncyrk6gjx4l34vvx/install
           install_options: '--tarball-url-prefix https://nixos-nix-install-tests.cachix.org/serve'
+          nix_path: nixpkgs=channel:nixos-unstable
           extra_nix_config: |
             experimental-features = nix-command flakes
+
+      - name: Print nixpkgs version
+        run: nix-instantiate --eval -E '(import <nixpkgs> {}).lib.version'
+
+      - name: Print whether nixpkgs' stdenv is darwin
+        run: nix-instantiate --eval -E '(import <nixpkgs> {}).stdenv.isDarwin'
+
+      - name: Evaluate nixpkgs' sile.meta
+        run: nix eval --json nixpkgs#sile.meta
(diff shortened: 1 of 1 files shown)


alerque:
@doronbehar [This error](https://github.com/sile-typesetter/sile/pull/1238/checks?check_run_id=3641246078#step:5:6) has be baffled. The `default.nx` in nixpkgs is only marked as broken for Darwin. I get the same error locally by the way. Any ideas?

> error: Package ‘sile-0.11.1-dirty-flake’ in /nix/store/2c1s0cbii0v7gqxyklknp32ya3xa3nr6-source/pkgs/tools/typesetting/sile/default.nix:100 is marked as broken, refusing to evaluate.

doronbehar:
> @doronbehar [This error](https://github.com/sile-typesetter/sile/pull/1238/checks?check_run_id=3641246078#step:5:6) has be baffled. The `default.nx` in nixpkgs is only marked as broken for Darwin. I get the same error locally by the way. Any ideas?
> 
> > error: Package ‘sile-0.11.1-dirty-flake’ in /nix/store/2c1s0cbii0v7gqxyklknp32ya3xa3nr6-source/pkgs/tools/typesetting/sile/default.nix:100 is marked as broken, refusing to evaluate.

No idea at all.. Perhaps it would help to print the Nixpkgs revision used, as explained in here:

https://github.com/cachix/install-nix-action/#how-do-i-print-nixpkgs-version-i-have-configured

And perhaps we can print similar stuff to debug why it evaluates as broken:

```yaml
- name: Print whether nixpkgs' stdenv is darwin
  run: nix-instantiate --eval -E '(import <nixpkgs> {}).stdenv.isDarwin'
```

This might work and help:

```yaml
- name: Evaluate nixpkgs' sile.meta
  run: nix eval --json nixpkgs#sile.meta
```

I also noticed you didn't specify a nix channel with `nix_path`, do you think it doesn't matter because you use the Nixpkgs flake?

doronbehar:
I'd try also:

```yaml
- name: Evaluate nixpkgs' sile.meta
  run: nix eval --json .#sile.meta
```

when Inside the repository.

alerque:
Thanks for the ideas. You can see the output of the debug info requested in [this test run](https://github.com/sile-typesetter/sile/runs/3654839729).

> I also noticed you didn't specify a nix channel

I tried it with and without, it just happened to end on "without". It's back, but that isn't the (only?) problem.

Everything else looks as expected, including isDarwin being falsey.

doronbehar:
I found it, we just had to RTFM:

> This command verifies that the flake specified by flake reference flake-url can be evaluated successfully (as detailed below), and that the derivations specified by the flake's checks output can be built successfully.
>
> If the keep-going option is set to true, Nix will keep evaluating as much as it can and report the errors as it encounters them. Otherwise it will stop at the first error.

It failed because it evaluated the derivation for all systems. You have a few options:

1. Remove the `nix flake check`
2. Not use `flake-utils.lib.eachDefaultSystem` and use: `flake-utils.lib.eachSystem [ "supported systems here as a list" ]`
3. Use `nix flake check --keep-going` but it will still exit with a non-zero code.

alerque:
Thanks for the research! Is there a definitive list of supported systems we can use and just subtract Darwin?

alerque:
I found the exact opposite of what I was looking for:

```nix
flake-utils.lib.eachSystem (
	flake-utils.lib.defaultSystems ++ ["x86_64-darwin"]
) (system:
	...
```

What I need is a list subtraction instead of a list concatenation operator.

alerque:
`lib.lists.subtractLists` might do the job, or there may be a way to iterate `defaultSystems` and use `stdenv.isDarwin` to filter out the ones we don't want (similar to [this](https://gitlab.tryp.io/nix/nixpkgs/-/blob/4cb53831410ab0212b99e4324488df67615b20dc/pkgs/servers/mpd/default.nix#L93) but for systems not features), but I can't figure out an exact usage that works.

doronbehar:
> `lib.lists.subtractLists` might do the job, or there may be a way to iterate `defaultSystems` and use `stdenv.isDarwin` to filter out the ones we don't want (similar to [this](https://gitlab.tryp.io/nix/nixpkgs/-/blob/4cb53831410ab0212b99e4324488df67615b20dc/pkgs/servers/mpd/default.nix#L93) but for systems not features), but I can't figure out an exact usage that works.

I guess we can do that, but in general, what value do you see in `nix flake check`? It only checks that the derivations evaluate.. If I were you, I'd run only `nix build -L .#sile` and perhaps test that the compiled binary can compile a sample document? Perhaps even that is not needed, since the derivation runs the tests of the build system during build.

alerque:
I believe this attempted work was obsoleted by 0ae6771f and we've been test building the flake in CI since before tagging v0.13.0.

PR #1240, opened 2021-09-19
Implement and use munder and mover

OlivierNicole:
Implement the `<munder>`, `<mover>` and `<munderover>` MathML tags.

Also fixes translation of TeX-like syntax into MathML: e.g. `\sum_i` should be translated (using the `<munder>` tag, not the `<msub>` tag) as `<munder><mo>&sum;</mo> <mi>i</mi></munder>`.

**Note:** I was not able to run the regression tests because of a font problem on my machine.


alerque:
Thanks for the contribution! Looks pretty good, I'll get this merged shortly. I'll probably split the original commit to be a fix + a feat plus squash the lint fixups. No big deal though.

OlivierNicole:
Thanks! It might be difficult to separate the fix from the new feature, however, as the fix makes use of the new feature.

PR #1241, opened 2021-09-20
Use italic mathvariant for \mi only if the content is a single character

cmfwyp:
This changes the default script style for `\mi` to upright when the content is more than one character, so `\mi{sin}(\mi{x})` is sin(*x*) instead of *sin*(*x*).

MathML considers function names to be identifiers and uses U+2061 to add a space between the function name and argument in cases like sin *x*.


OlivierNicole:
> This changes the default script style for `\mi` to upright when the content is more than one character, so `\mi{sin}(\mi{x})` is sin(_x_) instead of _sin_(_x_).

I had missed this point of the MathML spec. Thanks!

alerque:
Thanks for taking the time to contribute @cmfwyp!

alerque (approved):

PR #1249, opened 2021-09-22
fix(math): Fix underover error with sub wider than base but no sup

cmfwyp:
```
\begin[mode=display]{math}
	\munder{base}{subsubsubsub}
\end{math}
```

Using `\munder{base}{sub}` with sub wider than base (as in the above example) produces this error:

```
/usr/local/share/sile/packages/math/base-elements.lua:695: attempt to index a nil value (local 'widest')
```

This fixes that.


simoncozens:
Failing checks seem irrelevant, so I'm going to force it.

PR #1255, opened 2021-09-29
feat(nix): Ignore many files when creating src

doronbehar:


doronbehar:
In response to https://github.com/sile-typesetter/sile/pull/1231#issuecomment-923795968

flake.nix
```diff
@@ -33,9 +35,31 @@
...
+            # CI files
+            ".cirrus.yml"
+            "action.yml"
+            "azure-pipelines.yml"
+            "Dockerfile"
+            # Git files
+            ".github"
+            ".gitattributes"
+            ".gitignore"
+            ".git"
```
alerque, line 60: `.git` is not itself tracked by Git. That makes me wonder, do we need to list other things here that might exist as artifacts in a working copy for one reason or another, or are we starting with just a list of *tracked* files here?

flake.nix
```diff
@@ -33,9 +35,31 @@
...
       sile = pkgs.sile.overrideAttrs(oldAttr: rec {
         version = "${(pkgs.lib.importJSON ./package.json).version}-${version_rev}-flake";
-        src = builtins.filterSource
-          (path: type: type != "directory" || baseNameOf path != ".git")
-        ./.;
+        src = nix-filter.lib {
+          root = ./.;
+          exclude = [
+            # Nix files
+            "flake.nix"
```
alerque, line 42: Shouldn't changes to the `*.nix` files be *included* in the sources considered? Changing them should cause rebuilds right?

alerque (commented):
Thanks for looking into this.

Just a few question marks in my mind. See comments, plus more generally does this filter have any kind of wildcard/globbing/regex support? Could it accept a file list from another source / format? We now have include/exclude lists for quite a few purposes: git ignores, linting include / exclude, vendored libraries, autotools dist_files, docker builds, etc. Someday we're going to want to unify them a bit.

flake.nix
doronbehar: That's a good question. When Nix "builds" the derivation of the source directory, it already _evaluated_ all the Nix files it had to evaluate, so it doesn't need any Nix file in the source in order to build sile.

flake.nix
doronbehar: The exclude list I wrote here is a list of files not needed for compilation. I think it will be appropriate to also exclude generated files or dist_files to make the source even cleaner in case a real developer (not me) is using `nix build`.

doronbehar:
> more generally does this filter have any kind of wildcard/globbing/regex support?

Sort of, but it's not that well developed, see: https://github.com/numtide/nix-filter#design-notes

> Could it accept a file list from another source / format?

Yes, we can read a list of files from another file, we can even read that list of files and ignore the file itself along.

> We now have include/exclude lists for quite a few purposes: git ignores, linting include / exclude, vendored libraries, autotools dist_files, docker builds, etc. Someday we're going to want to unify them a bit.

I'm not against that, but to my view many of these exclude lists perhaps intersect but not match a lot since they serve different purposes. The sole purpose of this exclude list is to make the source directory of the build cleaner so that running `nix build` two times in a row won't require a real rebuild if the source is exactly the same - exactly the same regarding source files that are actually used in the build.

I do see a reason to ignore both what's in the list here, and what's written in `.gitignore`. I'll see how can I use https://github.com/hercules-ci/gitignore.nix as well.

doronbehar:
`.gitignore` is used now to ignore files, and the other list of files are also excluded.

alerque (approved):

PR #1256, opened 2021-10-01
Suppress incorrect overfull warning

alerque:
A frame isn't overfull if everything already fit and we have nothing more to add.


simoncozens:
Oh, good catch. I wonder if a better fix would be to turn ll.459-460 into an `elseif`:

https://github.com/sile-typesetter/sile/blob/84507a546ac63f7f40e65abbfe6831ff4c19f046/core/typesetter.lua#L457-L463

alerque:
> I wonder if a better fix would be to turn ll.459-460 into an `elseif`:

No, because those two conditions are not exclusive. Both may or may not need running independently.

There is a slightly neater to handle the logic I'll post, but unsettling the previousVbox is something that may be needed for `initFrame()` too.

simoncozens:
Ah, yeah. OK, then!

simoncozens (approved):

PR #1257, opened 2021-10-01
Fixup early merge of PR #1256

alerque:
Oops. I merged #1256 before I pushed the tweak I mentioned there.


alerque:
This is wrong. Some cases *do* need a `chuck()` here. I'm going to keep it around because this is an interesting edge case to my last attempt to refactor `pushBack()`, but as it stands #1256 as merged is correct, this is not.

Omikhleia:
This 3+ year old PR never came to fruition as-is and now has conflicts.

PR #1264, opened 2021-10-23
feat(core): Implement paragraph shaping (parshape)

Omikhleia:
These are the changes I made to the linebreak algorithm as mentioned in https://github.com/sile-typesetter/sile/discussions/1263

I am suggesting them for inclusion in the standard SILE. They are pretty limited, @simoncozens had already prepared all the way to them. They are a good basis to start playing with, I hope.

- I renounced to use a fixed pre-computed array as TeX does, a function (ideally "memoizing" already computed lines) is much easier and we have all the power of Lua to make fancy things.
- On the way, I cleaned up a bit the hangafter/hangindent stuff I had earlier done, it should be easier to maintain and understand this way.
- The small change to the typesetter also makes things easier IMHO, as the right/left skips are handled differently in TeX and SILE.
- A test is included, using a fixed pre-computed circle (that's better for a non-regression test!)

As for an actual package using this feature and computing things dynamically... Honestly, I am not sure my current **colophon** package should make it into the standard SILE now... I mean, it works quite nicely for my needs, but it is quite fragile with several wild assumptions. I don't plan something better for now, but I committed it to my own package [repo](https://github.com/Omikhleia/omikhleia-sile-packages) - So you may look at the PDF in the "examples" folder (for the Woah! effect, I hope :smile: ) and to the actual documentation in the "docs" package (to see why and where it could be so fragile that it might be a mess to maintain in the official distribution), and make your own opinion. And beyond the inherent flaws of the approach, it's also fairly specialized (with "ornaments" I selected, etc.), and I might also add more shapes eventually (triangles/trapezes obviously come to mind), if I ever need them... at some uncertain point... But this just express my concerns and I am open for discussion.


Omikhleia:
Following https://github.com/sile-typesetter/sile/issues/1268 I realize I should check what happens to headers/folios/footnotes when this shaping is active when the page break occurs (and the current solution for hangindent/hangafter fix in footnotes and folios should also probably be addressed, it's kind of _ad hoc_). Kind of minor (I don't think users would normally use the parshaping accross pages, but heh).

core/break.lua
```diff
@@ -75,30 +74,42 @@ function lineBreak:trimGlue() -- 842
...
+  -- numbers.
+  --
+  -- Using a Lua function rather than a fixed array allows one to perform
+  -- computations, but remember the linebreak methods will be called multiple
+  -- times to try good breaks, so it is strongly recommended to "memoize"
+  -- values for lines already computed.
+  --
+  -- TeX wizards shall also note that this is slighty different from
+  -- Knuth's definition "nline l1 i1 l2 i2 ... lN iN".
+  self.parShape = nil
```
ctrlcctrlv, line 98: I must say that this way of doing things has a very strong code smell. This `parShape` should not be here at all as it's never used by the function and you show that the only way to use it is to rewrite the whole function…it would be better for there to be a `lineBreak:parShape` method, which by default is a noop (rather than being tested for `nil`), but which can be rewritten by user scripts. This would also mean you don't need to save and restore `lineBreak:setupLineLengths`, only `parShape`; and indeed if the noop function is simple to rewrite not even that.

tests/parshaping-simple.expected
```diff
@@ -0,0 +1,297 @@
...
+Mx 	188.0200
+T	87 75 72 w=12.2212	(the)
+Mx 	202.6196
+T	79 68 85 74 72 85 w=22.2451	(larger)
+Mx 	224.8647
+T	17 w=2.0610	(.)
+Mx 	201.3418
+My 	565.3031
+Set font 	Gentium Plus;10;400;;normal;;LTR
+T	20 w=4.6924	(1)
```
ctrlcctrlv, line 295: As I was taught by @alerque before you, tests should set `\nofolios` :-)

tests/parshaping-simple.sil
```diff
@@ -0,0 +1,89 @@
...
+  { width = SILE.measurement("167.28269918420pt"), indent=SILE.measurement("84.190650407902pt") },
+  { width = SILE.measurement("164.66117660805pt"), indent=SILE.measurement("85.501411695975pt") },
+  { width = SILE.measurement("159.09112080046pt"), indent=SILE.measurement("88.286439599772pt") },
+  { width = SILE.measurement("150.24495450145pt"), indent=SILE.measurement("92.709522749275pt") },
+  { width = SILE.measurement("137.49177425846pt"), indent=SILE.measurement("99.086112870769pt") },
+  { width = SILE.measurement("119.58808312009pt"), indent=SILE.measurement("108.03795843996pt") },
+  { width = SILE.measurement("93.624309129279pt"), indent=SILE.measurement("121.01984543536pt") },
+  { width = SILE.measurement("47.922780548840pt"), indent=SILE.measurement("143.87060972558pt") },
+  { width = SILE.measurement("167.54343340503pt"), indent=SILE.measurement("84.060283297483pt") }
+}
```
ctrlcctrlv, line 26: This is scary to put in a script, surely with Lua at your disposal you can calculate something as simple as points along a circle?

tests/parshaping-simple.sil
```diff
@@ -0,0 +1,89 @@
...
+            -- Paragraphs that extend beyond the circle...
+            return {
+                width = self.hsize,
+                left = 0,
+                right = 0
+            }
+        end 
+        -- Assume we should never reach 500 lines. If it does, stop
+        -- the mess.
+        self.lastSpecialLine = 500
```
ctrlcctrlv, line 60: I don't like this magic number. Bibles may have many small lines…please try to figure out a termination condition that isn't just "no one will ever want more than 500 lines", it's too Bill Gates and his 640KB of RAM.

tests/parshaping-simple.sil
```diff
@@ -0,0 +1,89 @@
...
+
+    -- Restore the original linebreaks.
+    SILE.linebreak.setupLineLengths = oldSetup
+    -- Also do not forget to deactivate the parShape
+    SILE.linebreak.parShape = nil
+  end)
+end, "Format paragraphs in a fixed centered circle")
+}
+
+Galileo, in 1638, wrote:
```
ctrlcctrlv, line 75: @alerque also taught me not to put unnecessary paragraphs in tests. :-)

ctrlcctrlv (changes requested):
Neat feature, but I think the code needs work before it enters SILE core.

Also, was any thought given as to how this works in TTB mode?

tests/parshaping-simple.sil
Omikhleia: Simple here: no termination condition and trust parShape to always spit out values. Problem is that the line breaking algorithm expects a value here, and I frown upon messing too much with it.

tests/parshaping-simple.sil
Omikhleia: My assumption was that in a test script, one wants repeatable values (known in advance). It doesn't even needs to be a circle here.

core/break.lua
```diff
@@ -598,24 +606,41 @@ function lineBreak:postLineBreak() -- 903
...
 
   repeat
-    -- local width
-    local indent
-    -- ParShape case should be handled too here.
-    if self.hangAfter > 0 then
-      -- width = line > nbLines - self.hangAfter and self.firstWidth or self.secondWidth
-      indent = line > nbLines - self.hangAfter and self.firstIndent or self.secondIndent
-    elseif self.hangAfter == 0 then
-      -- width = self.hsize
```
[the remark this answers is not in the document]
Omikhleia: > developers can refer to the commit before yours.

Developers do that? I very seldom did it in my whole career so far, or only not too long after the commit. Personally I find this kind of comments useful (well, maybe the "SILE handles the actual width differently" afterwards comes a bit too late and could be reformulated, e.g. "we don't need the actual line widths for now as SILE handles them differently than TeX later in the typesetter instance"), but I don't care if it's removed.

core/break.lua
Omikhleia: You are right (well rather than a noop, just spit out the h.size) + a setting to enable/disable it would allow to fix the case of folios and headers (as we do for the hangs). Working on it.

Omikhleia:
> Also, was any thought given as to how this works in TTB mode?

Nope. I don't know how to check RTL, TTB, BTB with scripts I can't even read. The line-breaking algorithm is agnostic to the writing direction, so I would expect it to somehow work, but I have no real idea what it implies.

tests/parshaping-simple.sil
Omikhleia: Should avoid relying on the book class page masters though, an use only plain.

tests/parshaping-simple.sil
Omikhleia: Done.

tests/parshaping-simple.sil
Omikhleia: Done

tests/parshaping-simple.expected
Omikhleia: Indeed! Done.

core/break.lua
Omikhleia: Done, and indeed clearer IMHO

core/break.lua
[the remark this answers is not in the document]
Omikhleia: Reformulated the comments, keeping the values but with more explanations. Hope this is better.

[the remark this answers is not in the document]
alerque: Yes, I refer to previous code all the time. I frequently have my editor open with a split diff view that I can cycle one side back through previous iterations. This helps make out what current code is supposed to do and why in a lot of situations.

That being said comments that explain why something is done a certain way are even better. Code that no longer works is not. Any obsolete code should be removed, and if the current implementation needs explaining an explanation can be added. Obsolete code as comments rarely helps future work.

alerque:
Thanks for the contribution, this sounds great. Sorry I had been traveling and haven't been very on top of reviews, but I'll get this checked out eventually.

tests/parshaping-simple.sil
alerque: Hard coded values in tests are fine. Using *plain* instead of *book* for tests is almost always preferred except where the thing being tested stands any reasonably chance of being affected by the class. As with folios and other things, the less things that get stuffed into unrelated tests the better.

tests/parshaping-simple.sil
alerque: There are a pile of hard coded values in the break and shape mechanisms already, many of them inherited from TeX history. I've been meaning to abstract them to a known place in SILE memory that *could be* tuned, but that doesn't belong in this PR.

tests/parshaping-simple.sil
alerque: Yup, the less content the better. The more content you have the more change future versions of fonts or other unrelated changes will kick in. Testing the minimum possible surface area for a given feature or bug is the most maintainable way. Even now and with a bunch of custom tooling on my side (such as iterating through visual diffs of outputs) updating tests when things change is a significant pain point.

alerque:
The example of the circle was really cool, but that's what I think it should be: an example. I'm going to move it to the example gallery and drastically simplify the test case to something where the numbers can be debugged easier by hand. Basically just enough to prove that the shape table is being used but not much more.

core/break.lua
```diff
@@ -74,34 +75,60 @@ function lineBreak:trimGlue() -- 842
   nodes[#nodes+1] = SILE.nodefactory.penalty(inf_bad)
 end
 
+function lineBreak:parShape()
+  return { width = self.hsize, left = 0, right = 0 }
```
alerque, line 79: I would like to refactor this a little bit to make the end-user interface more flexible. Rather than expecting a table with mixed types and line data that is clearly redundant mathematically, I would like this function to return 3 values, any one of which can be nil to use defaults, and all of which may be measurements.

core/break.lua
```diff
@@ -74,34 +75,60 @@ function lineBreak:trimGlue() -- 842
...
+  -- numbers.
+  --
+  -- (Obviously, one should ensure left + right + width = self.hsize, so
+  -- returning the two indents could have been avoided, but it feels more
+  -- readable and natural this way, and easier afterwards back into the
+  -- typesetter).
+  --
+  -- Using a Lua function rather than a fixed array allows one to perform
+  -- computations, but remember the linebreak methods will be called multiple
+  -- times to try good breaks, so it is strongly recommended to "memoize"
```
alerque, line 101: Additionally I would like to build the memorization into the breaker so that the end user supplied function doesn't have to optimize this, we will just know not to keep asking the same question.

alerque (commented):
I'm happy to make these changes ... and probably will right now at least as a POC, but I'm open to feedback on the changes I'm proposing here.

core/break.lua
Omikhleia: Maybe then, for consistency, should we also make the hangIndent a measurement rather than an integer, what do you think about it?

core/break.lua
Omikhleia: I'm not strongly opinionated on the issue, but I'm not fully sure about it, for the following reasons at least:
- If using a pre-computed  shape (as done in the test), additional memoization (https://en.wikipedia.org/wiki/Memoization) is of course not needed (= would cost extra memory usage for no reason).
- If computations are kept simple (e.g. a mere triangle/trapezoid), the small hit on performances might still be acceptable vs. caching values.
- If one wants to implement more complex algorithms, e.g. shaping several paragraphs at once in a common shape, it may require additional things to be overloaded (in the typesetter instance, etc.) to keep tracks, etc. The breaker (as it currently stands) might be too low level then anyway. I've not explored this topic much, however (what I have been exploring a bit, though, without a satisfying result yet, is shapes with one hole... Until I abandoned, it was a bit frustrating and for needs that might not be that relevant - I kind of know that LaTeX has the `shapepar` package which fully reimplements Knuth's parshaping differently - with apparently edge cases too though - but I am not sure it is that widely used, beyond demonstration...)

core/break.lua
simoncozens: Yes, it should be.

core/break.lua
alerque: I have implemented the API changes I suggested. I'm adding a caching layer onto it now, will post for review soon.

core/break.lua
alerque: I'm not sure about this one. There is a bit of difference between lua function arguments and a declared setting that has a type that we actually typecheck. Given that the typesetter only actually *utilizes* an integer here, accepting the setting as `integer or nil` seems like the most logical and lowest friction thing to do. Anything else introduces a weird edge case where it isn't clear to the end user when the setting would be `:absolute()`ised. At time of setting? At paragraph init? At time of use? Once and done? An every paragraph?

I am struggling to see the utility in that.

I setup the `lineBreak:parShape()` function return values to accept anything and cast them to what we use internally. Since the entire mechanism is cached this is pretty low overhead. This doesn't seem equivalent of the hangIndent setting.

alerque:
I'm working on porting the circle example to something more suited to be adjustable and show off the caching, but my ideas for changing the parShape API are implemented above. Feedback welcome, I don't claim to have hit this nail on the head.

alerque:
Here is an example using the modified parshape API drawing a paragraph in a circle using a center point and radius:

```sile
\language[main=en]
\hyphenator:add-exceptions[lang=en]{iso-pe-ri-me-tric}% No idea where that weird word should really break.
\nofolios
\begin{script}

SILE.settings.set("linebreak.tolerance", 200)
SILE.settings.set("document.baselineskip", "1.6em")
SILE.settings.set("document.parindent", 0)

-- The math here finds intersection points for a line crossing a circle, making a bunch of assumptions about
-- the line always being horizontal. Only touching doesn't count, only a full intersection is a win for this.
-- http://csharphelper.com/blog/2014/09/determine-where-a-line-intersects-a-circle-in-c/
local sliceofcircle = function (c, r, h)
  local cx, cy, y = c, r, h
  local B = 2 * -cx
  local C = cx * cx + (y - cy) * (y - cy) - r * r
  local det = B * B - 4 * C
  if det <= 0 then
    return false
  else
    local enter = (-B - math.sqrt(det)) / 2
    local exit = (-B + math.sqrt(det)) / 2
    return enter, exit
  end
end

SILE.registerCommand("incircle", function (options, content)
  local center = SU.cast("measurement", options.center or "50%fw")
  local radius = SU.cast("measurement", options.radius or "50%fw")
  local oldParShape = SILE.linebreak.parShape
  SILE.typesetter:leaveHmode()
  SILE.settings.set("linebreak.parShape", true)
  SILE.linebreak.parShape = function (self, line)
    local c = SU.cast("number", center)
    local r = SU.cast("number", radius)
    local h = SILE.measurement((line-1).."bs"):tonumber()+3
    local enter, exit = sliceofcircle(c, r, h)
    if not enter then return 0, self.hsize, 0 end
    local width = SILE.measurement(exit - enter)
    return enter, width, nil
  end
  SILE.process(content)
  SILE.typesetter:leaveHmode();
  SILE.linebreak.parShape = oldParShape
  SILE.settings.set("linebreak.parShape", false)
end, "Format paragraphs in a fixed circle")

\end{script}
% Galileo, 1638:
\begin[radius=30%fw]{incircle}
The area of a circle is a mean proportional between any two regular and similar
polygons of which one circumscribes it and the other is isoperimetric with it.
In addition, the area of the circle is less than that of any circumscribed
polygon and greater than that of any isoperimetric polygon. And further, of
these circumscribed polygons, the one that has the greater number of sides has
a smaller area than the one that has a lesser number; but, on the other hand,
the isoperimetric polygon that has the greater number of sides is the larger.
\end{incircle}
\end{document}
```

Omikhleia:
> Here is an example using the modified parshape API drawing a paragraph in a circle using a center point and radius:
> ...
> SILE.linebreak.parShape = oldParShape
>  SILE.linebreak.parShapeCacheClear()

The explicit cache cleaning sounds real weird. Also, the example could set the settings temporarily so that the default parshaping stops being called afterwards.

core/break.lua
Omikhleia: > I am struggling to see the utility in that.

For API consistency, and the setting to accept eg. "2cm" rather than an unqualified integer value (i.e. absolute points).

Omikhleia:
Also this:
>  SILE.settings.set("linebreak.tolerance", 200)
SILE.settings.set("document.baselineskip", "1.6em")
SILE.settings.set("document.parindent", 0) 

Could be done temporarily in "incircle". The example is not very useful otherwise - one would not want to have this globally in a normal use case. (If one does so, we'll get question why plenty of other things starts working weird - a non-stretchable baselineskip gives pretty awful results on pages with footnotes etc.)

alerque:
> The explicit cache cleaning sounds real weird.

I grant the interface perhaps could use refining, but I couldn't think of any way to get the benefit of cached results but also allow the use of new shapes after the first one without some sort of cache flush mechanism. The `SILE.linebreak` object is almost completely stateless.

> Also, the example could set the settings temporarily so that the default parshaping stops being called afterwards.

True, before that example actually hits the gallery issues like that should be addressed and it could be commented etc. I just needed a POC that was a little more like somebody might use it rather than the test file which did the minimal work possible to hit all the possible interfaces.

> Could be done temporarily in "incircle". The example is not very useful otherwise - one would not want to have this globally in a normal use case.

Of course.

> (If one does so, we'll get question why plenty of other things starts working weird - a non-stretchable baselineskip gives pretty awful results on pages with footnotes etc.)

Sure. The `\incircle` function could take care of using the current baselineskip but temporarily dropping any strech/shrink. That sounds like a useful addition before the example goes in the gallery. But that is for a PR on the website repo...

Lets keep this issue more about the API and whether the right things are exposed vs. handled.

Omikhleia:
> > The explicit cache cleaning sounds real weird.
> 
> I grant the interface perhaps could use refining, but I couldn't think of any way to get the benefit of cached results but also allow the use of new shapes after the first one without some sort of cache flush mechanism. The `SILE.linebreak` object is almost completely stateless.

At the end of the loop in `lineBreak:doBreak()` , the paragraph has its final breaks all computed (which `lineBreak:postLineBreak()\ just builds into a result list, so that's perhaps where the cache could be cleared. Next time the algorithm is invoked (for a subsequent paragraph), the inputs (hsize and sideways) could be different so the cache would need to be invalidated. True, in many cases they'd be the same (e.g. in a sequence of identical paragraphs), but it cannot be guaranteed (the next paragraph submitted to `lineBreak:doBreak()` by the typesetter could have a different size).

core/break.lua
```diff
@@ -74,34 +75,73 @@ function lineBreak:trimGlue() -- 842
   nodes[#nodes+1] = SILE.nodefactory.penalty(inf_bad)
 end
 
+-- NOTE FOR DEVELOPERS: this method is called when the linebreak.parShape
+-- setting is true. The arguments passed are self (the current line) and a
```
Omikhleia, line 79: `self` is the linebreaker instance (rather than the current line)

alerque: Roger that. I'm also fixing LTR vs RTL handling of extra space when some shape values are left unspecified.

Omikhleia: by unspecified, ypu mean when some is "nil" as stated here:
> The expected return is three values, any of which may be nil to use default values or a measurement to override the defaults. The values are considered as left, width, and right respectively.

I am unsure what default values are and where it is handled (should it even have defaults?)

core/break.lua
alerque: Interestingly the way we have these parameters setup all SILE settings are automatically cast to their expected types at assignment time. This means you can already pass "2cm" just fine. The difference in changing the setting from a number to a measurement will change when relative values (i.e. not "2cm" but "2%fw") are absolutized: at assignment on shortly prior to use. The later sounds potentially more useful to me so I'll go ahead and make the handIndent setting measurement.

alerque (approved):

core/break.lua
```diff
@@ -74,34 +75,84 @@ function lineBreak:trimGlue() -- 842
...
+  local width = SILE.measurement(w or hsize)
+  if not w and a then width = width - SILE.measurement(a) end
+  if not w and b then width = width - SILE.measurement(b) end
+  local remaining = hsize:tonumber() - width:tonumber()
+  a = SU.cast("number", a or (b and (remaining - SU.cast("number", b))) or 0)
+  b = SU.cast("number", b or (a and (remaining - SU.cast("number", a))) or remaining)
+  return a, width, b
+end
+
+-- Wrap linebreak:parShape in a mnotized table for fast access
```
Omikhleia, line 112: typo : "mnotized"

core/break.lua
alerque: Defaults are `0, self.hsize, 0`, but you don't have to return those and they aren't configurable except by supplying a `parShape()` function that does something different. Anything left `nil` will default to those with the addition that changing any of them also makes the appropriate adjustment to the defaults. For example setting a non-zero left with reduces the width from `self.hsize` to `self.hsize - left`.

core/break.lua
```diff
@@ -74,34 +75,84 @@ function lineBreak:trimGlue() -- 842
...
+function lineBreak:parShapeCache(n)
+  local cache = parShapeCache[n]
+  if not cache then
+    local l, w, r = self:parShape(n)
+    local left, width, right
+    if SILE.typesetter.frame:writingDirection() == "LTR" then
```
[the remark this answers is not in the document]
alerque: Good catch. In fact RTL frames are reversing the values anyway. Better not to mess with it at this low level...

alerque:
> Also, was any thought given as to how this works in TTB mode?

@ctrlcctrlv Yes. We're just twiddling standard parameters already used used "linewise" by the typesetter. The typesetter already handles flipping them around based on the writing and advance directions.

Issue #1267, opened 2021-11-02
Clarifying weird units, please...


Omikhleia:
(a) In the "image" package documentation:
> If the size parameters are not given, then the image will be output at its ‘natural’ pixel size.

And the examples go on with several widths in **px** (pixels?)

But what does it mean? If I give a width of 50px, it doesn't seem much different than 50pt apparently. So I am lost here, what is this "px" unit on the images, exactly?

(b) In the "svg" package documentation:
> Optional width or height options will scale the SVG canvas to the given size calculated at a given density option (which defaults to 72 ppi).

So now we have a new **ppi** (points per inch, I assume), and I am lost again. Changing the density from 1 to 600 didn't seem to affect the output in my raw test. Huh. 72ppi (the so-said default) looks like a screen resolution of some sort, but why? All I want is to ensure that in print, I'd have at least, say, 300 **dpi** (yeah, let me introduce another unit...) if things have to be rasterized at some point. Do they, for SVG? What I am supposed to understand and do?

(c) _Where it goes even weirder_

So we have "SILE.**length**(...)" in several places (sometimes behind a `SU.cast(..)` - which I realize I have abused in my own use of it - a length can have a plus or minus (shrink/stretch).

But we also have SILE.**measurement**` in other commands. And the SILE manual doesn't describe it, but the behavior is different. Wow... Sounds messy, when these concepts aren't introduced properly...

But besides a documentation issue (and a consistency issue between packages?), it goes even further... Some length/measurement values are actually passed via yet another mean: `SILE.parseComplexFrameDimension` - The "raiselower" and "image" package do this.... Huh? Wait, this method understands **frame-related specifications**, and I am sort of amazed, but it actually works:

```
\img[src=documentation/gutenberg.png, width=width(content)]
```
Wow... real weird.... Was this intended? But if so, it fails on the  "svg" package, which will not accept such a width specification. Doh.

_Let's recap_ : length vs. measurement, pt vs. px or  ppi, ordinary units vs. frame-style width. ~~Isn't there some inconsistency to fix? Maybe not: Frankly, I got lost.~~ Update: 

- [x] Length vs. measurement = documentation issue --> See #1768
- [x] SVG ppi = likely hides a bug #1449 - marked as done here since it now has its own bug issue
- [x] Image size = documentation issue, fixed in #1277 
- [x] Use of frame dimensions in some unrelated commands = likely a bug. Notably to consider/investigate
[...]

PR #1277, opened 2021-11-09
Doc manual image fixes

Omikhleia:
Fixing, in 3 dedicated commits:
- The easy part of #1267 (image package), that is:
  - Changed the resolution of the gutenberg image, so it would get a fine "natural" size (exactly 2in).
  - Reformulated the package doc and removed the "px"
- A typo (`current.parindent`) in the manual, seen in passing, as well as a bunch of small typography details (apostrophes, etc.) both in the manual and in the packages' docs.
- Several wrong commands (`SILE.Glue`) or deprecated commands in the manual (I've not checked the code bits, though, beyond renaming as now expected.)


Omikhleia:
Fixed my bad english ^^

simoncozens:
Thank you!

PR #1281, opened 2021-11-12
fix(core): Fix core UTF-16 encode/decode functions

### LINKED ISSUE #1279: pdf:metadata val vs. value
The documentation for `pdf;metadata` in the **pdf** package gives `\pdf:metadata[key=Author, value=J. Smith]` but the actual implementation expect a `val` option.

Suggestion (sounds better that way that the other way round, unfortunately):
- deprecate `val` and support `value` as documented
- at some point remove support for `val`

### LINKED ISSUE #1280: pdf:metadata does not accept unicode
SILE 0.12

Unicode characters in metadata...
```
\pdf:metadata[key=Author, val=Erwin Schrödinger]
\pdf:metadata[key=Title, val=Erwin Schrödinger’s cat]
```
... do not work well:

image

The PDF standard says:
> The text string type shall be used for character strings that shall be encoded in either PDFDocEncoding or the 
UTF-16BE  Unicode  character  encoding  scheme.  PDFDocEncoding  can  encode  all  of  the  ISO  Latin  1  
character set and is documented in Annex D. UTF-16BE can encode all Unicode characters.

This (very) convoluted way indeed works, confirming that UTF16BE is the solution:

```
\script{
local pdf = require("justenoughlibtexpdf")
SILE.outputters.libtexpdf._init()

local info = pdf.get_dictionary("Info")
pdf.add_dict(info, pdf.parse("/Title"), 
     pdf.parse("<"..SU.utf8_to_utf16be_hexencoded("Erwin Schrödinger’s cat")..">"))
}
```
image

I tried to find an easier workaround, but `pdf_metadata` in the C code does some `strlen()` https://github.com/sile-typesetter/sile/blob/master/src/justenoughlibtexpdf.c#L328 so simply passing the UTF16LE-encoded string there cannot work. It should likely have been a `lua_rawlen(L, 2)` there - but I haven't found yet how to compile this library.

_In other terms_, solving this issue likely implies
- fixing the C code in justenoughlibtexpdf
- encoding in UTF16BE in the Lua "pdf" package

ctrlcctrlv:
Make sure they always use a BOM for UTF-16, (required per Adobe PDF
spec), and add a test for them.

Fix related issue found by @Omikhleia, #1280, by taking most of their
recommendations regarding a possible patch, but actually editing the C
code rather than using `pdf.parse`.

Closes #1280.
Closes #1279.

image
```postscript
13 0 obj 

<<
/CreationDate (D:20211112023651-05'00')
/Producer (SILE v0.12.0.r33-g3ad14d5-dirty \(Lua 5.4\))
/Author <feff0045007200770069006e0020005300630068007200f600640069006e006700650072>
/Title <feff0045007200770069006e0020005300630068007200f600640069006e006700650072201900730020732bd83ddc08>
>>
endobj
```


packages/pdf.lua
```diff
@@ -108,15 +108,24 @@ end)
...
+  local value
+  if options.val ~= nil then
+    SU.deprecated("\\pdf:metadata[…, val=…]", "\\pdf:metadata[…, value=…]", "0.12.0", "0.13.0")
+    value = SU.required(options, "val", "pdf:metadata")
+  else
+    value = SU.required(options, "value", "pdf:metadata")
+  end
+
+  -- see comment in pdf:bookmark
+  value = SU.utf8_to_utf16be(value)
```
Omikhleia, line 120: Should we do it always, even if the key is ModDate or CreationDate? The PDF specs do not say dates can be UTF16-BE encoded. This might work if viewers/readers are flexible enough, but might not be compliant.

ctrlcctrlv: Fixed in https://github.com/sile-typesetter/sile/pull/1281/commits/7df0242865c4c4ef44e94bb26d3e61440244d165.

ctrlcctrlv:
Unclear why this has sat so long.

alerque:
> Unclear why this has sat so long.

Because I was traveling for the 7 weeks period covering when it was reported until just this past week and I'm just now trying to dig out of GitHub issue notifications.

alerque (approved):
Great work, thanks for the contribution. I shuffled some commit stuff around to split the test changes from the other work. Also most of the scope was not `core` so much as a mix of `utilities`, `backends`, and `packages` so I picked the closest match for those. Should be zero code changes.

PR #1284, opened 2021-11-12
ci(azure): Update to `windows-2019` before deprecation

### LINKED ISSUE #1282: [ci] The windows-2016 environment will be deprecated on November 15, 2021, and removed on March 15, 2022.
CI states:

The windows-2016 environment will be deprecated on November 15, 2021, and removed on March 15, 2022. Migrate to windows-latest instead. For more details see https://github.com/actions/virtual-environments/issues/4312

E.g. on https://dev.azure.com/sile-typesetter/sile/_build/results?buildId=1016&view=results

Offending line seems to be—

https://github.com/sile-typesetter/sile/blob/3ad14d54b9176577fe5553c522bdafcea0cfbd56/azure-pipelines.yml#L8

ctrlcctrlv:
I'm far from an expert on this…just couldn't figure out how to stop telling `cmake` an explicit version (seems impossible), so couldn't use `windows-latest`, so went with `windows-2019` so it doesn't randomly break in future.

Cf. https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software

Closes #1282.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,7 +5,7 @@ trigger: none
 jobs:
 - job: Windows
   pool:
-    vmImage: 'vs2017-win2016'
+    vmImage: 'windows-2019'
   steps:
     - checkout: self
       submodules: true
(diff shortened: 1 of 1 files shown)


Omikhleia:
Is that related to #650 ?

alerque:
Yes this probably belongs in #650, but since merging to there is complicated and we want to get rid of the CI error before that is merged I'll just take it as is then handle conflicts on that branch.

PR #1289, opened 2021-11-23
Assumes that fractions have both a numerator and denominator

arichard4:
In math/base-elements.lua, elements.fraction is written to support not receiving a numerator or denominator. The support is incorrect/doesn't function, though; in shape, if the numerator or denominator is missing, shape will set other to nil, then crash on trying to access other.width and trying to set other.relx.

This doesn't matter *in practice*, because the only caller to elements.fraction is math/typesetter.lua, which fails earlier if there's a missing numerator or denominator:

```
    local children = convertChildren(content)
    if #children ~= 2 then SU.error('Wrong number of children in mfrac: '
      ..#children)
    end
    return b.fraction(children[1], children[2])
```

### DIFF
--- a/packages/math/base-elements.lua
+++ b/packages/math/base-elements.lua
@@ -965,35 +965,20 @@ elements.fraction = pl.class({
     elements.mbox._init(self)
     self.numerator = numerator
     self.denominator = denominator
-    if self.numerator then table.insert(self.children, self.numerator)
-    end
-    if self.denominator then table.insert(self.children, self.denominator)
-    end
+    table.insert(self.children, numerator)
+    table.insert(self.children, denominator)
   end,
   styleChildren = function(self)
-    if not (self.numerator or self.denominator) then
-      SU.error("Fraction cannot have both no numerator and no denominator")
-    end
-    if self.numerator then
-      self.numerator.mode = getNumeratorMode(self.mode)
-    end
-    if self.denominator then
-      self.denominator.mode = getDenominatorMode(self.mode)
-    end
+    self.numerator.mode = getNumeratorMode(self.mode)
+    self.denominator.mode = getDenominatorMode(self.mode)
   end,
   shape = function(self)
     -- Determine relative abscissas and width
     local widest, other
-    if self.numerator and self.denominator then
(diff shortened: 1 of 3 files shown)


OlivierNicole:
Thank you for this! The code changes look fine to me.

That being said, have you checked that everything works as expected when there is indeed an empty numerator or denominator, e.g. when typesetting `<mfrac> <mrow></mrow> <mrow><mn>667</mn></mrow> </mfrac>` or `\frac{1 + i}{}`?

It might be worth adding this kind of cases to `tests/math-fractions.xml`.

arichard4:
> That being said, have you checked that everything works as expected when there is indeed an empty numerator or denominator

The existing code calls SU.error in math/typesetter.lua:

```
  elseif content.command == 'mfrac' then
    local children = convertChildren(content)
    if #children ~= 2 then SU.error('Wrong number of children in mfrac: '
      ..#children)
    end
    return b.fraction(children[1], children[2])
```

SU.error calls os.exit, which halts execution of the program:

```
utilities.error = function(message, bug)
  utilities.warn(message, bug)
  io.stderr:flush()
  SILE.outputter:finish()
  os.exit(1)
end
```

So elements.fraction can't get reached without a numerator and denominator as-is, and the behavior of the program is unchanged by this commit: it will still fail in math.typesetter.lua and fail to produce any output with the message "Wrong number of children in mfrac: $num".

OlivierNicole:
Sorry I haven't been clear, I was referring to what happens, not when one of the children is `nil`, but when one of them is an empty `stackbox` (which should be the case when the parser encounters an empty `<mrow>`, or an empty group `{}` in the TeX-like syntax).

arichard4:
Sorry for misunderstanding!

Those two cases aren't affected by this commit, both produce identical looking fractions missing a top or bottom, both before or after this.

I added a testcase to tests/math-fractions.xml, the expected output didn't differ before/after my commit.

simoncozens (approved):
Thanks for this, looks good to me too.

OlivierNicole:
Great, thank you!

PR #1296, opened 2021-12-03
fix(languages): Repair broken French hyphenation patterns

### LINKED ISSUE #1295: Broken French hyphenation rules
SILE 0.12

**Expected**
lo-go-gramme
igno-rer
com-pa-gnon
di-gres-sion

**Observed**
logog-ramme
ig-no-rer
com-pag-non
dig-res-sion

The patterns in `languages/fr.lua` are actually severely broken
- Some comments made it into pattern strings
- A whole group of patterns is missing after  "résur3gent."
- Some other patterns missing compared to v2.13 (2016/05/12) at https://tug.org/tex-hyphen/

I have fixed it from the above TeX source and can make a PR with the changes. (But likely these patterns were converted automatically, so perhaps the issue exists for other languages.)

### LINKED ISSUE #1297: Hyphenation issue after apostrophe?
SILE 0.12

**Observed**
```
grou-pe d’-hommes (French, 0x2019)
grou-pe dʼ-hommes (French, 0x02BC)
grup d’-ho-mes (Catalan, 0x2019)
grup dʼ-ho-mes (Catalan, 0x02BC)
```
I am getting line breaks with an hyphen after the apostrophe (whether using the 0x02BC apostrophe or the 0x2019 quote mark: I thought this could have been my issue initially, with received texts from different software having one or the other, but they behave the same apparently).

image

**Expected** 

It does not seem to me that the hyphenation point is legit. Seems to occur in more than one language (e.g. French and Catalan above).

Omikhleia:
Closes #1295 
- Regenerate the full patterns

Closes #1297
- Found a French-specific issue due to the removed spacings
- Commented out the broken "linebreak type change" code that was inactive due to a typo but doesn't seem to work. It might need to be re-addressed at some stage, but trying to make it work broke other things (and for the "qu" type notably, which is ambiguous, it has side-effects on other languages such as Turkish, which uses apostrophes inside words - I have no idea what would be expected here...
- Therefore, I went for the least impacting change, adopting an ad hoc logic for French and Catalan. As for French, I could not find any issue now on a >150-page document which initially had all the above (wrong breaks due to the patterns, and a few breaks after the apostrophe).


languages/unicode.lua
```diff
@@ -75,7 +75,7 @@ SILE.nodeMakers.unicode = pl.class({
         self:addToken(char, item)
         self:makeToken()
         self:makePenalty(0)
-      elseif self.lasttype and (self.thistype and thistype ~= lasttype and not self.isWordType[thistype]) then
+      elseif self.lasttype and (thistype and thistype ~= self.lasttype and not self.isWordType[thistype]) then
```
Omikhleia, line 78: But line 85 below seems wrong too, assigning a lasttype...

Omikhleia: And then I notice that comment above
```
-- luacheck: globals lasttype
-- XXX - this is wrong and broken, but is also confusing. See bug #687
```
So the `self.thistype` was likely wrong, but the rest is too weird to understand ^^

alerque (approved):

Omikhleia:
Commit deb88c476c3c6556d08994fa2dff12764540468d has "Update Fresh hyphenation expectations" as headline. Read "French", I assume.

alerque:
Er ahem, yes that should read "French". Unfortunately we're kind of stuck with that, and with 100% automatic changelog entries and releases that typo of mine is going to propagate. Sorry.

Issue #579, opened 2019-04-19
negative space widths results in odd line lengths


mhosken:
Currently, when typesetting nastaliq with its negative spaces between clusters, the shrink and stretch on associated glue becomes .. interesting. Using the absolute spaceWidth ensures that shrink is shrink and stretch is stretch:

```lua
  makeSpaceNode = function(self, options, item)
    if SILE.settings.get("shaper.variablespaces") == 1 then
      spacewidth = item.width
      w = SILE.length.new({
        length = spacewidth * SILE.settings.get("shaper.spaceenlargementfactor"),
        shrink = math.abs(spacewidth) * SILE.settings.get("shaper.spaceshrinkfactor"),
        stretch = math.abs(spacewidth) * SILE.settings.get("shaper.spacestretchfactor")
      })
      SU.debug("typesetter", "newGlue "..w)
      return (SILE.nodefactory.newGlue({ width = w }))
    else
      return SILE.nodefactory.newGlue({ width = SILE.shaper:measureSpace(options) })
    end
  end
}
```

This seems to work in the docs we need.

PR #1308, opened 2022-01-27
Fix regression of #579

mimbrown:
I noticed that #579 had regressed, I'm not sure when. This pull request should take care of it I think.

Before:

[test.pdf](https://github.com/sile-typesetter/sile/files/7952848/test.pdf)

After:

[test.pdf](https://github.com/sile-typesetter/sile/files/7952854/test.pdf)


alerque:
Thanks for catching this regression. The error was mine in f4a64e7Df which did a huge amount of refactoring on lengths — mostly for the better but obviously missed some things too.

Can we come up with a test case that would show this error early if it ever goes unnoticed again? Your test PDF shows the problem, but lets make it as simple as possible. Perhaps a6 paper, and just a couple words with a font we already have and whatever the minimum scenario we can arrange to make it show this error without the right code.

alerque (changes requested):
The fix itself looks great. We just need a regression test to warn us in the future if this math gets overlooked.

alerque:
I tweaked the commit message to match our release system. I can help work out the test too, but if you could help me with some sample nastaliq text and settings that exhibit the problem? The PDF isn't very useful to copy from but I should be able to work up a test from the source of that PDF if you can share!

mimbrown:
If it helps, here's a sample doc that can generate the problematic PDF (if you have Awami Nastaliq installed):

```
\begin[direction=RTL-TTB]{document}

\set[parameter=document.baselineskip, value=2.5em plus 2pt minus 2pt]
\font[family=Awami Nastaliq, size=10]

\script{
local function typesetRepeatedly (str, num)
  local concatenated = ""
  for i=1, num do
    concatenated = concatenated..str
  end
  SILE.typesetter:typeset(concatenated)
end

typesetRepeatedly(
  "کی خواہش ",
  100
)
}

\end{document}
```

And here's a unit test (unvalidated because I've never done unit testing in Lua before) for testing to make sure this gets caught (I'm not sure where the best place for this is):

```lua
it("should always have positive stretch and shrink", function()
  SILE.settings.set("shaper.variablespaces", true)
  SILE.settings.set("shaper.spacestretchfactor")
  SILE.settings.set("shaper.spaceshrinkfactor")
  local space = SILE.shaper:makeSpaceNode({}, SILE.nodefactory.glue("-4pt plus 1pt minus 1pt"))
  assert.is.truthy(space.width.stretch > 0)
  assert.is.truthy(space.width.shrink > 0)
end)
```

alerque:
Fantastic! The best approach might actually be both, I'll check it out.

alerque (approved):

alerque:
Thanks again for the contributions. I massaged both tests around a bit to simplify the end-to-end test and make the unit test valid, but thanks for the general content needed that was super helpful.

This is a relatively simple fix to a relatively bad regression so I'll be looking to cut a release soon just to get this out there. Just so you know though we have one little hold up on that — Lua 5.4.4 was just released and some of our font loader code is failing in it. Our CI currently only tests Lua 5.4.3 so I'm looking for a fix on my machine locally. I'll try to get a release out as soon as I can track that down.

PR #1322, opened 2022-02-25
Fix typesetNaturally() by enabling BiDi in default typesetter

### LINKED ISSUE #1317: SILE.typesetNaturally is broken
```
\begin[direction=RTL]{document}
\font[family=Amiri,size=100pt]
\nofolios

\begin{script}
SILE.typesetNaturally(SILE.getFrame("content"), function()
 SILE.typesetter:typeset("عَرَبي")
end)
\end{script}

\par

\begin{script}
SILE.typesetter:typeset("عَرَبي")
\end{script}
\end{document}
```

The first line here is broke, the second line is not.
image

The line resetting the typsetter to the `SILE.defaultTypesetter {}` seems to be the culprit:
https://github.com/sile-typesetter/sile/blob/58b54ec78df6290e7d5860578bbb61654e5d3e9a/core/typesetter.lua#L721-L729

No idea of this intentional or not, but it breaks things like `\typeset-into` from `frametricks`.

alerque:
Closes #1317


alerque:
@simoncozens How does this solution sit with you? By stashing the "bidi off" version of the `boxUpNodes()` in the typesetter instead of relying on the template one we can both enable BiDi on the default typesetter while still having a way to turn it off on a per-typesetter basis.

Issue #1196, opened 2021-08-29
Standard way for extracting a node from the AST tree


Omikhleia:
Greetings,

SILE 0.10.15 on Ubuntu 20.04

In my package attempt (*), I tried to do something similar to the dockbook example in the SILE manual, where it uses `SILE.findInTree` to locate a command within the immediate children of the current tree, and then a `docbook.wipe` helper to nullify it before processing the content. 

In the manual's example, it works well, as it extracts a "title", processes it, and then _afterward_ wipes it and processes the content.

The problem in my case is that the extracted command is intended to be processed _after_ the content (basically, displayed below it), so I could not use the same trick. I ended up explicitly removing it from the tree, as follows (boilerplate removed)

```
omikhleia.extractFromTree = function (tree, command)
  for i=1, #tree do
    if type(tree[i]) == "table" and tree[i].command == command then
      return table.remove(tree, i)
    end
  end
end
```
It works, but:
* It doesn't feel right to do it in a package. It requires too much knowledge of the internal AST for what could be a fairly common use case (e.g. extracting a caption from a figure environment, etc.) - I have the same remark for the `docbook.wipe` helper, actually.
* This could (should, IMHO) be a standard command, in the same vein as `SILE.findInTree`
* If so, it could be made more robust and efficient, e.g.
  * It could check there is no extraneous command by that name remaining and issue a warning (the command may or may not exist in the broader scope of the document)
  * I am not sure `table.remove` is efficient if the processing tree gets big. If so implementation could differ (e.g. just flag the node as already processed / to be ignored later, but it then requires going deeper into the AST processing)  

EDIT (*) Just for the curious-minded, my [package attempt](https://github.com/Omikhleia/omikhleia-sile-packages/)

PR #1323, opened 2022-02-25
fix(packages): Make underline respect shrink/strech (rules package)

### LINKED ISSUE #1320: \underline and \boxaround do not honor stretcheability
SILE 0.12.1

The `\underline` command and the (undocumented) `\boxaround` command from the **rules** package do not respect shrinking/stretching (caused by an explicit glue or even line justification)

```
\begin{document}
\script[src=packages/rules]
\language[main=und]% Do disable hyphenation and have stretched line easily
\nofolios

Some \underline{underlined text}

Some \boxaround{boxed text}

Some \underline{underlined\hfill text}\hfill

Some \boxaround{boxed\hfill text}\hfill

Some \underline{test lorem ipsum} dor sit amet. Some test lorem ipsum dor sit amet. Some test lorem ipsum dor sit amet. 
Some ametloremipsum dor sit amet. Some test lorem ipsum dor sit amet. Some test lorem ipsum dor sit amet. 

Some \boxaround{test lorem ipsum} dor sit amet. Some test lorem ipsum dor sit amet. Some test lorem ipsum dor sit amet. 
Some ametloremipsum dor sit amet. Some test lorem ipsum dor sit amet. Some test lorem ipsum dor sit amet. 

\end{document}
```
image

I got the problem with my own framebox package and I am attempting a fix for it - if it works there, I could possibly try to check the rules package.

By the way, the `boxaround` command is not documented, should we even keep it (it lacks options - perhaps I could propose my framebox package for replacement, let me know...)

Omikhleia:
Closes #1320 

- Changed implementation to support shrinking/stretching in underlined text (`\underline` command).
- A fixed the undocumented `\boxaround` command for the same reason.
- Marked the latter as deprecated:
  - It was undocumented
  - It's pretty rough in what it does (no padding, no option, etc.) and unsatisfying
- Regarding underlining, the OpenType specs state that underlinePosition is the distance of _the top of the underline from the baseline_. I interpreted it strictly, i.e. the thickness is drawn below. The previous implementation lowered the line, but drew it upwards, which seems wrong to me.

Caveat: I did try RTL (arabic with Amiri font), and the commands do not do what I would have possibly expected. The older implementation was broken too, however, so this fix is not causing more harm...


packages/rules.lua
```diff
@@ -43,25 +43,72 @@ SILE.registerCommand("underline", function (_, content)
...
+      -- should expand downwards
+      SILE.outputter:drawRule(oldX, Y + underlinePosition, newX - oldX, underlineThickness)
+    end
+  })
 end, "Underlines some content")
 
 SILE.registerCommand("boxaround", function (_, content)
+  -- This command was not documented and lacks feature.
+  -- Plan replacement with a better suited package.
+  SU.deprecated("\\boxaround (undocumented)", "\\framebox (package)", "0.12.0", "0.13.0")
```
alerque, line 77: This bit needs splitting out into a different commit for the sake of the release notes. Deprecations stuffed inside `fix:` commits won't get the airtime they require.

packages/rules.lua
```diff
@@ -43,25 +43,72 @@ SILE.registerCommand("underline", function (_, content)
   local underlineThickness = font.post.underlineThickness / upem * fontoptions.size
 
   local hbox = SILE.call("hbox", {}, content)
-  local gl = SILE.length() - hbox.width
-  SILE.call("lower", {height = SU.cast("measurement", underlinePosition)}, function()
-    SILE.call("hrule", {width = gl.length, height = underlineThickness})
-  end)
-  SILE.typesetter:pushGlue({width = hbox.width})
+  table.remove(SILE.typesetter.state.nodes) -- steal it back...
```
alerque, line 46: Neither here no there for this PR, but I've thought several times this little "trick" needs to be upgraded to be a first class API of some kind, phantom mode or a reset command with mark points or a push/pop stack or something...

packages/rules.lua
alerque: P.S. I don't mind doing the split, I have `git revise -c` + interactive diff edits and am not afraid to use them. While we're in here I'd fix the underline and boxaround commands in two commits, again for the automatic release note benefits.

alerque (changes requested):

alerque (approved):

packages/rules.lua
Omikhleia: You were quicker than I :)
Great, thanks !

Omikhleia: Oops the commit messages say "borderbox" for actually "boxaround"

packages/rules.lua
Omikhleia: Thinking alike, as this is something we do in a lot of places... I'd think of something such as `SILE.helper.box.hbox(content)` that would do the same things as the "hbox" command except the final `table.insert` so one doesn't need the nearly immediate `table.remove` in these cases ;

Might look like a long name, but `SILE.findInTree` could be e.g. relocated to `SILE.helper.ast.findInTree`, where other goodies (e.g. my old #1196 from my initial try at SILE, lol) could also find a place, without polluting the global SILE namespace.

Omikhleia: Altenately, we could have an `hbox:unpushed` command (build the box in the current context, but doesn't pushes it to the node queue). Less general but maybe clearer. (But that's beyond this PR indeed, just for the discussion)

alerque: Opened an issue to track this someplace it won't get lost when I close this PR:‌https://github.com/sile-typesetter/sile/issues/1326

PR #1325, opened 2022-02-26
Use class direction option (if set) to inform default frame direction (if not set)

### LINKED ISSUE #1321: Mark all frames to inherit class' direction option by default
Noticed while solving #1317...

Setting `direction=RTL` at the document level (as a class option) sets that property on the default content frame for the class. It does **not** get set on any additional frames (headers, foootnotes, folios, etc.) and it does **not** get set on new frames created by the user.

My expectation is that this class level setting would change the default for all frames created in the document, obviously with the possibility to override that with specific frame properties.

I'm not sure whether this change should be considered a breaking change or not (i.e. do we warn about it, provide a migration path, do a major revision bump, etc.) or whether this is just so expected that we should just change it already and call it good. Perhaps @khaledhosny you have an opinion on this?

alerque:
Closes #1321

* Sets default on all default frames in a class when setting direction option.
* Inherits the class direction value on new frames that do not otherwise specify their own direction.

This *may* be a breaking change for some use cases, but as discussed the new behavior is so much more logical that people were probably working around this as a bug to get this behavior anyway.

Tests are not expected to pass until #1322 is merged since the BiDi inheritance issue affects the test output and I already filled in the correct expectations for having that issue fixed.


classes/plain.lua
```diff
@@ -34,7 +34,14 @@ plain.defaultFrameset = {
 plain.firstContentFrame = "content"
 
 plain.options.direction = function (value)
-  if value then plain.defaultFrameset.content.direction = value end
+  SILE.documentState.direction = value
+  if value then
+    for id, frame in pairs(plain.defaultFrameset) do
```
Omikhleia, line 39: Shouldn't rather the frame initialization logic look for the preferred direction (in a setting ?), rather than forcing it from here? I am concerned it may likely break for all dynamically generated frames created _after_ class initialization (e.g. new frames created programmatically by commands - I used them in my "parbox" implementation, but as of the core SILE, there's also the "frametricks" package that can be used to do such things.)

PR #1332, opened 2022-03-03
fix(packages): Avoid infinite loop when re-enabling BiDi

### LINKED ISSUE #1330: \bidi-on causes infinite loop
If you call `\bidi-on` when bidi is already on, `nobidi_boxUpNodes` now becomes equal to `bidiBoxupNodes`, and when `bidiBoxupNodes` calls `nobidi_boxUpNodes` it ends up calling itself recursively and endlessly.

MWE:

```
\bidi-on

aaaaa
\end{document}
```

alerque:
Closes #1330

### DIFF
--- a/packages/bidi.lua
+++ b/packages/bidi.lua
@@ -246,14 +246,18 @@ local bidiBoxupNodes = function (self)
 end
 
 local bidiEnableTypesetter = function (typesetter)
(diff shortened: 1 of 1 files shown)


simoncozens (approved):
Looks good, thank you!

PR #1334, opened 2022-03-04
Multi-line links and underlined contents conditional support

### LINKED ISSUE #424: Wrapping of pdf:link elements not implemented
The function to [tune the wrapping of URL's](https://github.com/simoncozens/sile/blob/master/packages/url.lua#L32) that are pdf:links is pointless because as it stands now the linker [forces everything into an hbox](https://github.com/simoncozens/sile/blob/master/packages/pdf.lua#L66). This allows for truly unsightly output.

### LINKED ISSUE #1239: pdf:link does not allow line breaks
SILE 0.11.1

Links made with pdf:link cannot span over multiple lines.

Omikhleia:
This PR is a proposal for supporting multi-line PDF links #1239 and multi-line text underlining. It is split in three commits:
- Actual support in the `typesetter` (core)
- Use in the `rules` package (= `\underline`)
- Use in the `pdf` package (= `\pdf:link`, therefore the `\href` command from the `url` package automatically inherits the feature too).
  - Closes #1239,
  - Closes #424

The idea follows my early experiments mentioned in discussion #1292 but on other grounds and a different approach: start and end of contents in lines are "marked" with specific boxes, and unterminated links/underlines are repeated on each line until the termination is met. So technically, each line is a different link annotation (etc.)

While a cool feature, I consider it (a) still experimental and (b) somehow as a breaking change, so the feature is disabled by default (= behind a `typesetter.support.multiline` setting wihich is false by default). As it is, it SHOULD be transparent for users, and only enabled on-demand. My idea is that this setting could be removed or reversed to true in a subsequent feature-breaking release, while still being able to be checked and used cautiously.

Why would it be a breaking change? 
1. A moot point, but maybe some people were used to hbox'es overflowing the line.. :roll_eyes: 
2. For links with (rectangle) borders enabled, it will use the line height rather than the former hbox'd content height so it might be looking _slightly_ different... (though in a way more consistent)
3. It would likely need to be checked for the RTL and BTT cases... :shark: I am not able to do it properly.

Other than that, the result is pretty cool IMHO.

image

(blue underlines are `\href`'s annotations, regular black line are `\underline`'s, showing various cases and even nested at the end.)

### DIFF
--- a/packages/pdf/init.lua
+++ b/packages/pdf/init.lua
@@ -82,38 +82,27 @@ function package:registerCommands ()
     })
   end)
 
-  self:registerCommand("pdf:link", function (options, content)
+  SILE.multiliner:register("pdf:link", function(x0, y0, x1, y1, height, _, options)
     local dest = SU.required(options, "dest", "pdf:link")
     local target = options.external and "/Type/Action/S/URI/URI" or "/S/GoTo/D"
     local borderwidth = options.borderwidth and SU.cast("measurement", options.borderwidth):tonumber() or 0
     local bordercolor = borderColor(SILE.color(options.bordercolor or "blue"))
     local borderoffset = SU.cast("measurement", options.borderoffset or "1pt"):tonumber()
     local borderstyle = borderStyle(options.borderstyle, borderwidth)
-    local llx, lly
-    SILE.typesetter:pushHbox({
-      value = nil,
-      height = SILE.measurement(0),
-      width = SILE.measurement(0),
-      depth = SILE.measurement(0),
-      outputYourself = function (_, typesetter, _)
-        llx = typesetter.frame.state.cursorX:tonumber()
-        lly = (SILE.documentState.paperSize[2] - typesetter.frame.state.cursorY):tonumber()
-        pdf.begin_annotation()
-      end
-    })
-    local hbox = SILE.call("hbox", {}, content) -- hack
-    SILE.typesetter:pushHbox({
-      value = nil,
-      height = SILE.measurement(0),
-      width = SILE.measurement(0),
-      depth = SILE.measurement(0),
-      outputYourself = function (_, typesetter, _)
-        local d = "<</Type/Annot/Subtype/Link" .. borderstyle .. bordercolor .. "/A<<" .. target .. "(" .. dest .. ")>>>>"
-        local x = typesetter.frame.state.cursorX:tonumber()
-        local y = (SILE.documentState.paperSize[2] - typesetter.frame.state.cursorY + hbox.height):tonumber()
-        pdf.end_annotation(d, llx , lly - borderoffset, x, y + borderoffset)
-      end
-    })
+
+    local llx = x0:tonumber()
+    local lly = (SILE.documentState.paperSize[2] - y0):tonumber()
+    pdf.begin_annotation()
+    local d = "<</Type/Annot/Subtype/Link" .. borderstyle .. bordercolor .. "/A<<" .. target .. "(" .. dest .. ")>>>>"
+    local x = x1:tonumber()
+    local y = (SILE.documentState.paperSize[2] - y1 + height):tonumber()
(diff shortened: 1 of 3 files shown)


Omikhleia:
As for RTL, since the contents are not longer boxed, it seems neat.

image

(top = current `\underline`, bottom = with setting `typesetter.support.multiline` set to true)

alerque:
This looks great. I'm going to stew on the issue of whether this is "breaking" or not. For the most part we are *not* considering document reflows to be breaking if the changes are clearly better/more correct/expected. We do consider anything requiring programming changes to even render to be breaking or changes that might be different and unexpected (even if both before and after are correct).

Given what I see here I'm included to just call this a combination of fixing and improving, it's not clear to me that the old behavior would ever be more expected or preferable. It might obsolete some workarounds, but so do other things such as the frame direction fixes I just pushed out in v0.12.3 :wink:.

Omikhleia:
N.B. Rebased onto 0.12.5 to fix (documentation) conflicts with rules and pdf packages.

alerque:
Is this ready to take out of draft mode and work on finalizing?

alerque (approved):
It looks to me like the class model here is a bit sideways. This creates a Penlight class but doesn't define a constructor and then just uses one instance to stuff all the possible commands using the model. Since I'm working on Penlight stuff anyway I'll take a closer look and see if that can be refactored, but the overall effect here is good.

Omikhleia:
Of course the "typesetter.support.multiline" was done in the case that this could go in a non-breaking release without taking any risk. If this makes it in an otherwise breaking release, we could remove it and simplify the code (= removing the previous implementations).

Omikhleia:
Update:
- Rebased on 0.13.x
- Extended to `\strikethrough` in addition to `\underline` (rules package)
- Removed the compatibility option (`typesetter.support.multiline`)
As per @alerque comments on the global multiliner class, I am not that convinced of it either. The proposed API might not be perfect. I am not that satisfied with the `addmarks` logic either. The Typesetter is already cluttered with plenty of weird things, not to say probably a bit broken (the pushback logic, the rlskips management, etc.) that likely will need some re-consideration at some point. On the other hand, it does solve an issue and provide a cool feature, IMHO. So I am in two moods:
- I like what it does
- I am not fond of my proposed implementation, though.
Yet, it might be better than nothing?

Omikhleia:
Update: rebased on current master (7925d45257d2fb228d1c569eabeb5f7ade35d9ef)

@alerque BTW, testing this, I noticed an (unrelated) line which is likely wrong, in the url package's `_init`, likely breaking links on master:

```
   pdf = SILE.outputter == SILE.outputters.libtexpdf
```
This is always false, I think that further to your recent changes, one is an instance, the other a class.
`SILE.outputter._name == "libtexpdf"` does the correct thing, but I don't know if this is the way you want to go?

alerque:
Good catch, yes I think that is wrong and solution is probably the best. There is a `:is_a()` method on instances, but the usage is easy to get wrong and checking the `_name` attribute directly is something I've done quite a few other places because it's simple and stable even if a bit ugly.

Omikhleia:
One year has passed, and SILE still doesn't have multi-line PDF links and multi-line text underlining, &c.

Omikhleia:
I am closing this PR for several reasons:
- Time has passed, which much changes in those files, so none of this stuff is applicable "as is"
- I was never really happy with the proposed API
- The implementation was not really nice inside hboxes either (= in horizontal-restricted mode).
- As stated above, this implementation only supported effects above content, so failed for effects below content -- such as a background color, etc. I wanted the latter to work for Djot/Markdown "hightlights" (resp. `{=mark=}` and `=mark=`)

While some ideas were sound, for the record I am working on a new experimental approach in **sile·x** (https://github.com/Omikhleia/silex.sile/pull/10). If this ever gets satisfying enough, I"ll consider a backport PR here (if you are interested).

Issue #1194, opened 2021-08-29
Package documentation


Omikhleia:
Greetings

SILE 0.10.15 on Ubuntu 20.04

In order to properly document my own package, I tried understanding the **autodoc** package, and have several concerns.

First, I couldn't find any documentation about it, some made test and trial attempt looking at the SILE manual code. I ended up with the following kind of script (for illustration):

```
\begin[papersize=6in x 9in, class=book]{document}
\script[src=packages/autodoc]
\script[src=packages/color]
\script[src=packages/rules]

\define[command=code]{\process}
\define[command=note]{\process}
\define[command=dots]{\process}

\bigskip
\package-documentation[src=packages/autodoc] % A standard package

\bigskip
\package-documentation[src=packages/color] % A standard package
\end{document}
```
Main issues:
* For compilation to succeed, I had to provide (here just as passthru) some commands (= code, note, dots...), whereas one would have expected either the packages to provide the necessary commands, or those to be provided by some supporting package (either **autodoc** iself or a dependency)... These commands are nowhere to be found in the packages or classes directory.
* The PDF output _only_ shows the documentation for the **color** package. The **autodoc** package does have a documentation, though. I suspect it has to do with it including blank lines / white spaces ? If so, shouldn't the code trim the text before parsing it? Several other "standard" packages also fail to have their documentation generated.

All in all, it summarizes to "how to get working documentation for existing packages", and "how should package authors make working documentation"

Secondary issues:

I do feel that enforcing (to some extent) packages to be properly documented improves maintainability and usability. The current **autodoc** packages looks (no offense intended) as a quick "hack" for the SILE manual. Are there plans or ideas how to correctly handle package documentation? That could include:
* Managing metadata (e.g. author, date, package version)
* Extracting the documented settings and commands (`SILE.settings.declare` and `SILE.registerCommand` both have support for help details, so these should be put to use) 
* and perhaps other similar things as well.

Regards. 

EDIT: 
Beyond the (small) initial step proposed in PR #1335 ("stage 0"), here is how I would consider subsequent stages... At a slow pace... from simplest ideas to the more complex or challenging:
- [x] An `\autodoc:command` to typeset commands in a clever and consistent way. First, there are a lot of inconsistencies in the documentation (in the way parameters are presented, etc.). Secondly, it's a bit annoying to have all these escape sequences where we could _just_ rely on the AST parsing to do the job for us... (+ indirect benefits, syntax check)
[...]

PR #1335, opened 2022-03-05
Leverage package documentation and manual (stage 0)

Omikhleia:
This is a subset of the concerns addressed in #1194 
I have been sitting far too long on it - and my previous attempts ended up being very messy, so it's preferable (I think) to proceed by small steps. Hence the "stage 0" in the PR title.

1. Removal of the `\dots` command in the SILE Manual and package documentations.
   - _SILE doesn't do that_... There are plenty of other characters directly in Unicode in the Manual and package documentations (e.g. em-dashes, quotes, etc.), so there's no good reason introducing this only shortcut in the Manual and have package documentations depends on it...
   - One may use some LaTeX-[compatibility layer](https://github.com/ctrlcctrlv/sile-texmode) for such things, if one really wants them.
2. Addition of the `autodoc` package documentation (fixed) to the SILE manual (in the "experimental" section)
   - It has to be there eventually, for package designers to know about it!
   - We want packages to be properly documented, don't we?
3. Introduction of `\autodoc:setting{<setting name>}` in the autodoc package
   - Automatically inserts good breakpoints, avoiding the need for designers to manually add some `\goodbreak` here and there.
   - (Conditionally) ensures the setting exists - It actually helped finding a few typos in the documentation (= e.g. in `linebreak.parShape` and `chordmode.offset`), so I am considering more commands of that type in another stage...

(On the way, I also fixed the `value` argument in the Manual's call to `pdf:metadata`, to avoid the deprecation warning.)

### DIFF
--- a/packages/autodoc.lua
+++ b/packages/autodoc.lua
@@ -63,23 +63,23 @@ end, "Outputs a settings name in code, ensuring good line breaks and possibly ch
 
 return {
   documentation = [[\begin{document}
-This package extracts documentation from other packages. It’s used to
-construct the SILE documentation. Doing this allows us to keep the
-documentation near the implementation, which (in theory) makes it easy
+This package extracts documentation information from other packages. It’s used to
+construct the SILE manual. Keeping package documentation in the package itself
+keeps the documentation near the implementation, which (in theory) makes it easy
 for documentation and implementation to be in sync.
 
-For that purpose, it provides the code{\\package-documentation[src=\em{package}]}
+For that purpose, it provides the \code{\\package-documentation[src=\em{package}]}
 command.
 
 Properly documented packages should export a \code{documentation} string
 containing their documentation, as a SILE document.
 
-For documenters and package designers, it also provides commands that can be used in their package
-documentations to present various pieces of information in a consistent way.
+For documenters and package authors, it also provides commands that can be used in their package
+documentation to present various pieces of information in a consistent way.
 
-Setting names can be fairly long (\em{namespace.area.some-stuff}).
+Setting names can be fairly long (e.g. \em{namespace.area.some-stuff}).
 The \code{\\autodoc:setting} command helps line-breaking them automatically at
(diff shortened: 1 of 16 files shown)


alerque (approved):
As far as I'm concerned these changes all look great! Thanks for the work put in, improving docs is not the most glamorous work.

The only thing I did is `git revise -e` on the commit messages to add the manual scope.

packages/autodoc.lua
```diff
@@ -12,12 +16,71 @@ SILE.registerCommand("package-documentation", function (options, _)
...
-  documentation = [[
-  \begin{document}
+  documentation = [[\begin{document}
 This package extracts documentation from other packages. It’s used to
 construct the SILE documentation. Doing this allows us to keep the
 documentation near the implementation, which (in theory) makes it easy
 for documentation and implementation to be in sync.
-  \end{document}]]
+
+For that purpose, it provides the code{\\package-documentation[src=\em{package}]}
```
alerque, line 71:
```suggestion
For that purpose, it provides the \code{\\package-documentation[src=\em{package}]}
```

packages/autodoc.lua
```diff
@@ -12,12 +16,71 @@ SILE.registerCommand("package-documentation", function (options, _)
...
+command.
+
+Properly documented packages should export a \code{documentation} string
+containing their documentation, as a SILE document.
+
+For documenters and package designers, it also provides commands that can be used in their package
+documentations to present various pieces of information in a consistent way.
+
+Setting names can be fairly long (\em{namespace.area.some-stuff}).
```
alerque, line 80:
```suggestion
Setting names can be fairly long (e.g. \em{namespace.area.some-stuff}).
```

alerque (changes requested):
A couple minor copy-edits.

alerque (approved):

simoncozens (approved):
Looks good to me too, thank you!

Issue #1106, opened 2020-11-23
Provide nested enumerations and bullets


the42:
Apparently one can get far without enumerations and bullets.

Bullets are frequently used in text, maybe overused. They might lead to bad type setting style. Nonetheless the lack of proper support of bullets and enumerations leads to imperfections as can be seen in the Sile manual, section 4.1 the possible arguments to `\font`.

I hope this ticket gets accepted as tracking ticket to make Sile eventually feature-rich to fulfill casual type setting needs - for your definition of casual.

## Enumerations

Support enumerations. In first iteration arabic consecutive numerals only with the option to set the starter manually and to specify an enumeration stop symbol (usually a dot as in `1.` - 1 is the enumeration, the dot is a visual helper ) Provide a mechanism to input longer batches of text without breaking out of the enumeration environment.

In a later iteration support non-arabic enumerations like letters, roman numbers, chinese, and
to style the enumeration  eg. by defining the color, font weight, etc.

## Bullets / Itemization

Support bullets and itemization, Provide an option to specify, besides defaults, the unicode rune to use as the item symbol.

In a later iteration provide the means to style bullets by changing color, font weight, distance to text, ...

## Nesting of bullets and enumerations
Provide a means to nest Enumerations and bullets.

In the case of enumerations, a sub-enumeration is usually indented and uses a new enumeration domain, eg. letters

1. Text here
  a. Sub-item level

A nested bullet list often reduces the weight of the used bullet eg. switching from a solid circle an nesting level 0 to an unfilled circled on nesting level 1. 

Also provide the means to nest enumerations and bullets.

Omikhleia:
Nearly a one-year old issue ^^

Hacked the quick'n dirty way during lunch pause

![image](https://user-images.githubusercontent.com/18075640/140319080-ce0f67d3-b134-4a2a-a73c-aa74c191323b.png)

Some of the above requirements could be questioned
- Bullet:
  -  "Provide an option to specify, besides defaults, the unicode rune to use as the item symbol." - Er. Why. Or rather why not defining your _styles_ for all levels, and use them consistently without micro-management?
- Enumerations:
  -  "the option to set the starter manually": Use case?
  - "input longer batches of text without breaking out of the enumeration environment": Quite vague.

the42:
Any eta to try that out? Looks good for a quick shot and a welcomed feature

Omikhleia:
> Any eta to try that out? Looks good for a quick shot and a welcomed feature

[...]

Issue #428, opened 2016-12-29
Correct for advance direction changes when switching to/from italics 


alerque:
Per discussion on my recent [Libertinus font issue](https://github.com/khaledhosny/libertinus/issues/50), it sounds like SILE could (optionally?) implement some sort of correction when switching to/from italic text to avoid this kind of display issue.

Omikhleia:
After those 6+ years... You meant having proper "italic correction", right?

Would I be close here?

![image](https://user-images.githubusercontent.com/18075640/235096961-8d7e5a06-1a3d-4b8c-8ecc-6eb3b2ef9dae.png)

alerque:
Yes.

I think the hard part of implementing this is we're necessarily in different chunks, and should probably only apply this when the only difference between font properties is the style or weight but not size or script.

Omikhleia:
Oh it's tricky... and (relative) size matters:

![image](https://user-images.githubusercontent.com/18075640/235107510-b3ca8499-9eba-44ec-b5c2-9672d378a879.png)

I'll clean up my messy code and propose a PR, eventually.

PR #1342, opened 2022-03-08
Leverage package documentation and manual (stage 1)

Omikhleia:
This PR is a proposal for the first two items in the task list from #1194 - or "low hanging fruits"
- `\autodoc:command` typesets a command (or more) without needing to escape characters, relying on the AST parsing. Due to this (i.e. because of the things skipped at parsing and thus not "reconstructible", such as spacing, option ordering, grouping), it may have a few shortcomings, but on the other hand, besides the sudden ease of typesetting, it comes with the benefits of syntax check and even syntax highlighting.
- `\autodoc:package` just takes a package name and typesets in a different style instead of "code".

Additionally,
- `\autodoc:environment` just takes a command name (string), but checks its validity - needed to be on par with `\autodoc:command` regarding styling.
- `\autodoc:parameter` typesets a parameter name or a parameter=value, again to be in line with the other commands.

And a few extra goodies, such as recognizing angle brackets around parameter or content value, and presenting them in a nicer way. See the package's documentation for more details.

The syntax highlighting setting is disabled by default. Would it be enabled, I tried to propose very discrete colors (i.e. quite dark, so as to remain readable if printed on B&W paper, and also not too intrusive for readers, yet possibly giving the impression that there's a little something to this documentation.)

Hope you like the principle.

### DIFF
--- a/packages/autodoc.lua
+++ b/packages/autodoc.lua
@@ -18,6 +18,13 @@ end)
 
 -- Styling hook
 
+SILE.settings.declare({
+  parameter = "autodoc.highlighting",
+  default = false,
+  type = "boolean",
+  help = "Whether audodoc enables syntax highlighting"
+})
+
 SILE.scratch.autodoc = {
   theme = {
     command = "#1d4851", -- oil blue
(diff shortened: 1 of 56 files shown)


Omikhleia:
No urge obviously - anyhow I will be offline this week-end and very possibly with limited access next week - but just to mention here what could be the steps after this one, so you know with a bit of advance where I am heading to.

Thriving to improve the look of the manual, I plan to donate to SILE a trimmed-down version of my enumeration/itemization package (trimmed-down in the sense that it doesn't have my _still in-the-work_ style inheritance and customization features for bullets, colors, numbering schemes etc., in other terms kept to the minimal functionality.)
- It would address at least partially #1106 in the short term
- It would make the bullet list in the manual nicer and consistent, e.g.
  image

The package is fairly ready, with tests etc. - so it's almost ready for a PR, depending whether the above `\autodoc:*` proposal makes it. It's in the `enumitem-lite-package` branch in my fork, currently on top of this very branch so using the proposed autodoc stuff in its documentation, as well as for the related changes in the manual). You can start thinking whether this is something you would possibly wish (IMHO, obviously, it looks great ;)

Omikhleia:
N.B. Rebased onto master to fix conflicts with  `packages/leaders`.

packages/autodoc.lua
```diff
@@ -77,10 +282,37 @@ containing their documentation, as a SILE document.
 For documenters and package authors, it also provides commands that can be used in their package
 documentation to present various pieces of information in a consistent way.
 
-Setting names can be fairly long (e.g. \em{namespace.area.some-stuff}).
-The \code{\\autodoc:setting} command helps line-breaking them automatically at
+Setting names can be fairly long (e.g. \em{namespace.area.some-stuff\kern[width=0.1em]}).
```
alerque, line 285: c.f. #428

alerque (approved):
This all looks great, thanks for all the work and contributions!

Issue #1206, opened 2021-09-03
`simpletable.lua` doesn't insert linebreaks in a cell when the cell overflows its frame


ctrlcctrlv:

simoncozens:
Simpletable is deliberately simple. We actually need a proper table implementation. :-/

alerque:
cf. #45

Omikhleia:
Backlog cleaning attempt:
- Are we going to consider changing the *simpletable* - I don't think so (_per nature_ of it), so this would be a "won't fix".
- More complex tables are already the topic of #45, with an existing solution (and I'll comment there if need be regarding the possible future of the latter)

Perhaps we can close this ticket as "won't fix" (for simpletable) or "duplicate" (in the larger scope)?

Omikhleia:
Perhaps we could deprecate **simpletable**...

Issue #529, opened 2018-03-25
Draw a box around content (both inline and vertical box)


simoncozens:

alerque:
I think [ptable.sile](https://github.com/Omikhleia/ptable.sile) probably satisfies this need. I'd be interested in some of the lower lever tools being upstreamed, but I think this need is covered in the package ecosystem. If not feel free to re-open.

Issue #367, opened 2016-09-05
Support vbox[width=...]


simoncozens:
First \vbox should put a box into the typesetter's queue. Second, this should make two columns:

      \begin[papersize=a5]{document}
      \script[src=packages/lorem]
      \hbox{\vbox[width=60pt]{Text here will wrap \lorem}}
      \hbox{\vbox[width=60pt]{Text here will wrap \lorem}}
      \par
      \script{SILE.typesetter:debugState()}
      \end{document}

alerque:
Won't this need to introduce some alignment parameters too? Sort of like `display: inline-block` in CSS? Or `\begin{minipage]| in TeX. If I'm understanding this correctly the same layout logic would also be useful for image positioning.

simoncozens:
Not really. The idea is that you put a vbox onto the horizontal list (which is supported but the output routine doesn't quite work).

Consider the following piece of plain TeX:

```TeX
hello
\qquad \vbox to 20pt {\hbox to 40pt{Stuff in a}\hbox to 25pt{vbox}}
\qquad \vbox to 20pt {\hbox to 40pt{Stuff in a}\hbox to 25pt{vbox}}
\qquad world.
\end
```

![screen shot 2016-09-05 at 16 11 55](https://cloud.githubusercontent.com/assets/106728/18238756/813091f6-7383-11e6-9b4a-c238b9e7c4f9.png)

(This is, essentially, how minipage is implemented under the hood.)

PR #1344, opened 2022-03-17
Paragraph boxes, framed boxes, and tables

### LINKED ISSUE #45: Support ctable-like tables
Please support tables with the (as much as possible) syntax of cable, see also http://www.ctan.org/tex-archive/macros/latex/contrib/ctable

### LINKED ISSUE #544: help please? want to have a box with a rule on one side
Consider packages/linebox.lua:
```
SILE.registerCommand("linebox", function (options, content)
  local box = SILE.Commands["hbox"]({}, content)
  local backgroundColor = SILE.colorparser(options.color)
  SILE.outputter:pushColor(backgroundColor)
  SILE.outputter.rule(cursorX, cursorY, 0.5, box.height)
  SILE.outputter:popColor()

  table.insert(SILE.typesetter.state.nodes, box)
end, "Place the output within a box that has a left rule")
```

The `SILE.outputter.rule(cursorX, cursorY, 0.5, box.height)` bit was cribbed from libtexpdf-output.lua.  And when I try to use this, I get an error that boils down to cursorY is undefined, which upon inspection I see is a variable declared elsewhere in libtexpdf-output.lua. Fooey. 

I can't seem to find any examples of working out the size of box so you can draw near it. 

(The end goal is to have email style quoting, because I'm trying to include email in a document.)

### LINKED ISSUE #911: Braces that span lines
Ran across this and it made me wonder: could we even pull this off if we wanted to? What changes would we have to make to get there?

### LINKED ISSUE #1024: Keep a group of lines together on a page
Sorry if this is very stupid question, but I can't figure it out.

I have:

```tex
\define[command=line]{\novbreak\skip[height=5pt]\noindent\hrule[width=450pt,height=0.3pt]\par\novbreak\skip[height=5pt]}
\define[command=sup]{\raise[height=0.6ex]{\font[size=0.6em]{\process}}}
\hbox{
\line
\pullquote{Thank you so much Fr. Ferdinand Bajao for taking the time to explain so well Postma's greatest contribution in preserving and keeping alive the rich Mangyan cultural heritage---the extensive documentation of the Mangyan script, and not to mention the over 20,000 \em{ambahan} he documented and [which are] now preserved [in] the Library of Congress in D.C. In fact, the vowel killer which is called \em{pamudpod} in the \em{Surat Mangyan}  is not his invention. [It] was introduced in other writing systems as early as the 17\sup{th} or 18\sup{th} century for the Tagalog Baybayin. It was also used in the 900 \font[features=+c2sc]{AD} Laguna Copperplate Inscription, the oldest Philippine document, which was discovered after Postma introduced the \em{pamudpod}. Even in the Tagbanwa script, a vowel killer was also introduced. Postma believed that the use of the \em{pamudpod} is necessary to keep the script alive and the Mangyans approved this for instruction purposes. Postma did not alter the original script.} 
\line
}
```

The `\hbox` doesn't do what I expect though, it still ends up broken across two pages:

image

Is my only choice to put an <kbd>\eject</kbd> before it? But I want SILE to intelligently place it, not necessarily at top of next page.

What am I missing?

Omikhleia:
In this PR, I am proposing to donate some of my packages to SILE, now that they have achieved some sort of maturity.

The PR contains a set of 3 related inter-dependent packages:
- struts package
- parbox package (uses struts)
- ptable package (uses parbox)

It includes as well a package that doesn't necessarily depend on the above, but becomes most interesting in combination with them (esp. struts and/or parbox):
- framebox package (also comes with a 'small' path rendering library)

_Rationale_
This PR aims at proposing a solution for at least:
- Closes #911 = `\bracebox` in combination with `\parbox`.
- Closes #1024 = doable with a mere `\parbox`.
- Closes #544 = `\parbox` has the necessary options to do it, for the simple case requested here.
- Closes #45 = `ptable` environment.
- Complements #1320 which lead to the deprecation of `\boxaround` (from the rules package) = `\framebox` is an appropriate feature-rich replacement.

Other related issues, possibly addressed too:
- #367 = alternate approach with `\parbox`.
- #1206 = `ptable` is a possible solution for more advanced tables that support line-breaking in cells.
- #529 = no description, but I'd assume the combination of `framebox` and `parbox` covers it.
- #318 = not attempted, but I think it is feasible via Lua e.g. parbox the text, steal it back, use its height/depth to display the image, then push the parbox back. Exercise left to the reader?

All in all (beyond tackling with 5-9 open issues), "struts" and "parbox" are somewhat low-level building blocks that can benefit other packages. I really think they should make it in the core distribution.

While tables have been requested (and the default "simpletable" is often not sufficient for anything serious), the proposed "ptable" is only one possible (experimental) approach. I still think it's better than nothing and would fill a long-standing gap. Maybe it can also foster interest and new ideas...

The "framebox", admittedly, is a bit more specialized and far less critical. But IMHO, it's cool for the show and has its Woah! effect.

_Other considerations_
I am using the `ptable` and `parbox` extensively in my own repo documentation, and also to some extent in my current book project. It helped finding several issues and/or edge-cases, and refine the concepts. Each feature was gradually added based on actual needs. I also intend to use `framebox` at some later stage.

All these packages are documented and have several non-regression tests... Though it could be a never-ending tasks to cover all corners. But Simon once mentioned the "stone-soup" principle, so i'll abide by it and suggest this might be robust enough for regular LTR cases and scenarios, and a ground for further enhancements.

_Caveat / Known issues_
  - RTL and BTT writing directions are untested.
  - Linespacing algorithms other than the (TeX-based) default one (e.g. the linespacing package, but also very likely grid-typesetting and Co.) are untested and probably not that well supported.

I hope you'll be interested, comments are of course welcome.


Omikhleia:
So seeing the CI failures: number rounding (1.0 vs. 1) and type conversions (operations on length/measurement, likely) behave differently in various versions of Lua. That's annoying. Another symptom of #649?

alerque:
The test failures have to do with mixing and matching types. Different versions of Lua handle type casting differently and none have strict type checking (see #871). The solution is to make sure you always cast yourself to the best type at the right time (`SU.cast()` can be useful here) and only do math on like types (`SILE.measurements` with `SILE.measurements`, raw `number`s with raw `number`s, etc.).

Omikhleia:
> Different versions of Lua handle type casting differently and none have strict type checking

Oh... Lua is annoying sometimes. And the change in ouputting 1.0 as "1.0" vs. "1" in Lua 5.2 is annoying too. Seems I got them right now, though I hated doing it :roll_eyes: - But thanks for the hint on what was going wrong! :tophat:

packages/parbox.lua
```diff
@@ -0,0 +1,498 @@
...
+    if minimize then
+      -- We go through all nodes in the line to compute the maximal line width.
+      -- This is not so efficient, we could have modified the typesetter to
+      -- compute it when it was processing the lines...But is is buried deep
+      -- in its breakpointsToLines() method, which some packages may moreover
+      -- overload, so we went for the easiest solution, albeit CPU-intensive.
+      local w = SILE.length()
+        if vboxes[i].nodes then
+          for n in ipairs(vboxes[i].nodes) do
+            if vboxes[i].nodes[n].width > SILE.length(0) then
```
Omikhleia, line 201: This is too naive and doesn't minimize the parbox as much as possible: negative width can of course occur (e.g. via kerns, or lskip/rskip/indent) and shouldn't be skipped.

I was all too quick here. I just noticed that recently... FWIW, the fun part of the story is that I liked so much the book example shown by @alerque in #911 that I decided to use it for a "fancy" custom table of contents with 2 levels (parts and chapters)... and met that case. I'll propose a fix and for a better width computation. The fix sounds decent now, with the nice TOC I had in mind, but I'll of course want to check it more before submitting it, and I'll also extract a dedicated test:
image

packages/parbox.lua
```diff
@@ -0,0 +1,544 @@
...
+  end
+  if #spec == 1 then
+    return { spec[1], spec[1], spec[1], spec[1] }
+  end
+  if #spec ~= 4 then SU.error("Invalid "..opt.." specification: "..rawspec) end
+  return spec
+end
+
+local computeNaturalWidth = function (nodes)
+  -- Derived from typesetter:computeLineRatio()
```
Omikhleia, line 156: Which does plenty of wrong things (#1632), so this is gonna be wrong too.

alerque (changes requested):
I like where this is going and what it is accomplishing, but I have some ideas about refactoring it before it goes in. Not all of these are fully thought out, just some initial comments:

1. ~~The PR title is missing the requisite Oxford comma.~~ Aa yes that's better!
1. Something doesn't sit right with the `graphics` package name space. Yes some of those are utilities tangentially related to graphics, but it doesn't seem to jive with what I might expect a SILE 'graphics' package to include.
1. The PRNG and RoughGenerator utilities really look like something that should be stand along Lua rocks. Other than a couple error messages and such I don't see anything SILE specific in there, just a utility library that could just as well be used elsewhere. How would you feel about refactoring those to be stand alone? I have lots of tooling to make it really easy to publish LuaRocks and would be happy to set it up if you are interested. We could house them in this org (like we already have for cassowary.lua) or in your user or whatever.
1. I love lots of tests, but this seems like it might be a little bit overkill. How confident are you that these are granular and not duplicating eachother?
1. Another test detail is I really prefer tests to use small pages where possible. I sometimes generate PDFs of everything and make a montage of them for review. Big A4 pages waste a lot of space and make these hard to review. A6 for singe pages tests and A7 for multi-page tests are preferred if possible. This way the default font sizes make the tests easier to review *en masse*.

Omikhleia:
>  2. Something doesn't sit right with the `graphics` package name space. Yes some of those are utilities tangentially related to graphics, but it doesn't seem to jive with what I might expect a SILE 'graphics' package to include.

It contains basic PDF graphics routines / primitives... I didn't find a better name for that ^^
I really don't have expectations on a SILE "graphics" packages; I could have some expectations for a "charts" package (e.g. histograms, pie charts, etc.) - that would likely use some of these graphics primitives and extend them (e.g. with circles and arcs etc.), but that's another story ;)

>  3. The PRNG and RoughGenerator utilities really look like something that should be stand along Lua rocks. Other than a couple error messages and such I don't see anything SILE specific in there, just a utility library that could just as well be used elsewhere. How would you feel about refactoring those to be stand alone? I have lots of tooling to make it really easy to publish LuaRocks and would be happy to set it up if you are interested. We could house them in this org (like we already have for cassowary.lua) or in your user or whatever.

I'll be plain honest: I don't have luarocks and don't plan to install and use it. It doesn't mean, of course, that this code cannot be extracted and moved elsewhere indeed, just that I don't feel interested in learning how to properly make "rocks" and maintain them separately in a standalone place.

>  4. I love lots of tests, but this seems like it might be a little bit overkill. How confident are you that these are granular and not duplicating eachother?

- framebox: 3 tests, each testing a different feature.
- parbox: 10 tests.
  - They all have a purpose, testing documentation claims or implementation issues.
  - Admittedly, 001 (a global safeguard) and 072 (an edge-case of sorts) could _perhaps_ be considered as slightly extraneous.
  - However, I'd tend to keep them all as far as I am concerned... But at your convenience.
- ptable: 6 tests
  - It initially had 7 tests = I removed one that can be considered as tested indirectly via the parbox tests
  - 5 tests cover documentation claims. I don't think I would do less than that.
  - 010 and 011 check we can match LaTeX ctable/booktabs examples. We'd likely want to include these as examples in a showcase, and I'd prefer they remain tested then. But do at your convenience.

>  5. Another test detail is I really prefer tests to use small pages where possible. I sometimes generate PDFs of everything and make a montage of them for review. Big A4 pages waste a lot of space and make these hard to review. A6 for singe pages tests and A7 for multi-page tests are preferred if possible. This way the default font sizes make the tests easier to review _en masse_.

Tests changed from A4 to to A6 (except one for parbox, changed to A5 only = 040, a bit too hard to make smaller).

alerque (changes requested):
I still really don't like the name spacing here, but I also don't want to hold up getting it included in core just because it's not perfect yet. What I would like to see as a minimum is getting it to be self contained so any future refactoring is more isolated to its own namespace.

Lets at least change this:

```
packages/
├─ framebox.lua
graphics/
├─ prng.lua
```

To this:

```
packages/
├─ framebox/
│  ├─ init.lua
│  ├─ graphics/
│  │  ├─ prng.lua
```

Omikhleia:
Moving forward steps identified so far:
- [ ] Fix the two computation problems noted in my reviews (I am close to it, but testing more thorough-fully this time)
- [ ] Refactor the tests (= even less test files, I've worked on grouping them more efficiently and the above fixes helped)
- [ ] Refactor package stucture according to 0.13 (subdir, init.lua, register/declare stuff) = I'll only do this last, so as to be able to compare line by line with my own implementation for 0.12.5, and only break after both are in line)
- [ ] Decide where the graphics primitives should go (as an aside note, I've worked on another package using them, not depending on framebox, so I am not that happy for them to move _below_ it, but I don't know...)
- [ ] Drink coffee or tea.

My ETA was tentatively for this week-end, 'cause I'm pretty busy after that -- But I can't guarantee it. We'll see...

alerque:
Re 0.13 refactoring: you can probably do that earlier and then shim them to be 0.12 compliant by with a `if SILE.version < "0.13" then init(SILE.documentState.DocumentClass); registerCommands() end` or something to that effect. I haven't tried it because I don't have anything I have any interest in keeping compatible with old versions, but it might work to reverse the idempotency stuff.

About graphics primitives, I still have an idea this should get factored out into a separate library. Even if not though remember that with packages being idempotent it shouldn't be a problem to load them when not in use and/or load parts of them from other packages (for any packages included in the same source).

packages/parbox.lua
Omikhleia: Reimplemented based on `typesetter::computeLineRatio()` also returning the natural totals since  PR #1378

packages/parbox.lua
```diff
@@ -0,0 +1,544 @@
...
+  if bordercolor then SILE.outputter:popColor() end
+end
+
+local insertStruts = function (vboxlist, strut)
+  -- The core assumption here is that first/last vboxes are actual text
+  -- lines. Could be wrong...
+  local h, d = SU.cast("length", strut.height), SU.cast("length", strut.depth)
+  for i = 1, #vboxlist do
+    if vboxlist[i].is_vbox and vboxlist[i].height < h then
+      vboxlist[i].height = h -- Hack height of the first vbox
```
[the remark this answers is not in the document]
Omikhleia: All vertical adjustment logic reimplemented avoiding box hacks.

Omikhleia:
> Drink coffee or tea.

With moderation, but I actually think a glass of wine is well deserved :clown_face: 

I hope I did the 0.13 re-shuffling quite as expected.

One additional remaining task would be to add these 4 packages to the manual.

packages/framebox/graphics/renderer.lua
```diff
@@ -0,0 +1,301 @@
...
+--
+
+local RoughGenerator = require("packages.framebox.graphics.rough").RoughGenerator
+
+-- HELPERS
+
+local _r = function(number)
+  -- Lua 5.3+ formats floats as 1.0 and integers as 1
+  -- This is annoying.
+  return math.floor(number) == number and math.floor(number) or number
```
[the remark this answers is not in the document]
Omikhleia: Done (fixing the mentioned line and also some other places where rounding was forgotten).

Omikhleia:
Why is this tagged for some "0.14.x" future when
- It _decently_ worked with 0.12.x
- It now works with 0.13.x even including additional fixes

Omikhleia:
Update: rebased on current master at this date (7925d45257d2fb228d1c569eabeb5f7ade35d9ef) and updated to 0.14 API.

That's a long way since the first experiments on [Oct 28, 2021](https://github.com/sile-typesetter/sile/issues/45#issuecomment-953667479) to this version. In the meantime, I have been working on its successor (e.g. with automatic column sizing and a few other "vanilla" improvements), but I am not satisfied with it yet and it might not deliver anytime soon -- this version is as good as I could make it.

Omikhleia:

Omikhleia:
There is goes: https://github.com/Omikhleia/ptable.sile

PR #1345, opened 2022-03-18
Update and cleanup repo tooling

alerque:
Minor updates to Docker builds, CI tooling, etc. Some is routine maintenance, some is lessons learned from other recent projects.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,7 +50,7 @@ jobs:
           path: sile-${{ env.VERSION }}.zip
       - name: Release
         uses: softprops/action-gh-release@v1
-        if: contains(github.ref, 'refs/tags/v')
+        if: startsWith(github.ref, 'refs/tags/v')
         with:
           body_path: sile-${{ env.VERSION }}.md
           files: |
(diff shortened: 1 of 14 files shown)


simoncozens:
Code looks good, thanks, but haven’t looked at the failing tests yet. Perhaps we should also get something to automatically PR a bump the JS deps.

alerque:
@simoncozens We actually *had* something to automatically bump the JS deps, but I turned it off because it was too noisy. They update very frequently and frankly we don't care that much, they are only developer conveniences for the commit / changelog management stuff and only really affect me during release cycles. Bumping them constantly when I'm not actually somewhere towards the end of a release cycle isn't worth the PR noise (at least to me).

Failing tests are because the Lua installer action changed how it handles build flags and we were testing on customized Lua builds with compatibility stuff disabled just for broadest coverage. I'm working out the new syntax requirements now.

alerque:
... and now we have an issue with Lua 5.4.4, luasec won't download right because luasocket doesn't work on Lua >= 5.4.3. Unfortunately the release process for luasocket is at least partially up to me now, so I'm going to shelve this PR and go shave yacks.

PR #1347, opened 2022-03-20
Add inline-escaping in SIL-language

### LINKED ISSUE #1092: Provide packages with a way to handle raw content

Omikhleia:
Here is a WIP showing my proposed idea in action for #1092, with samples (tests) for "text/plain" (a.k.a. some sort of true verbatim) and "image/svg+xml" (a.k.a. inline SVG). - To show by example how it could possibly work and foster discussion about it.

Incidentally, I think I found an unrelated "bug" (unintended side-effect) - via test `/inline-svg-002.sil` - the C function `svg.svg_to_ps()` apparently modifies the Lua input string (which is not a problem currently, as the input string is always read from a file, but turned out to be an issue here when the inline input string could be reused in further calls).

Closes #1092


alerque:
Just for the record, I am somewhat in favor of having at least *one* general purpose raw pass-through command that could be used by non-core packages.

Perhaps the biggest thing that bugs me about this specific proposal is the naming. I don't like calling it `\inline` because that drags the semantics of block vs. other content into the equation. HTML/TeX/SILE/Markdown and others all have a concept of hmode vs. vmode (block vs. inline, etc.) and calling a raw passthrough "inline" messes with that semantics.

Secondarily I don't like using the mime type. This might not be as defensible on my part because I do see the logic, it just seems verbose and smacks of HTML rather than Lua. Is there a way we could just call it "handler" or something that was a name of a callback function that would be more idiomatic of Lua than of HTTP headers?

alerque (changes requested):
Yes, but...

1. Lets not call it `\inline`.
1. Lets consider if mime types are the best identifiers or if just declaring a handler might be more idiomatic of Lua (also the same mime type might have more than one handler!)
1. Lets bring the sample test data down to a bare minimum. We don't need to drown in pages of diffs when something breaks, a one or two line SVG would suffice without testing a full piece of art.

alerque:
Per [discussion here](https://github.com/sile-typesetter/sile/issues/437#issuecomment-1150703614) I agree that finding a term idiomatic of Lua and programming doesn't need to be the goal here. What are our options? And do we have to pick just one or might a general purpose raw/passthrough/input command benefit from more than one of these args?

* mime / mimetype
* type / filetype
* format
* syntax
* language
* ?

Omikhleia:
> do we have to pick just one

Personally, I don't see the use case for more than one specific argument (for selecting the appropriate routine handler), especially if all other options are passed through to to handler.

I agree my quick `\inline[format=<mime>]` is probably ill-named, `\raw[type=<name>]` is probably better. 

I am not that fond of the `\raw` name, though it not too bad. For for the record it is what reStructuredText uses (e.g. `.. raw:: name`) for "raw data pass-through" (although the semantic is different there).

Out of the top of my head, another proposal could be `\embed`, `\incorporate`... er... `\transclude` (lol).

alerque:
> Personally, I don't see the use case for more than one specific argument [...]

I can think of some use cases that might call for two or more: one declarative markup (e.g. "mimetype") and one programmatic (e.g. "handler"). This way we could have a default set of mime type handlers (and even allow access to override them of course) but also have a one-off way to pass the content to a different handler instead. This would be a shortcut for setting up phony mimetype handlers for variants of the same type that should be handled by a different function.

Another use case might be a pre-processing filter to use at the parser level before passing it to the handler.

Other options of course would be passed through to the handler function, so perhaps the latter (filter) use case could just be covered on a per-handler basis.

alerque:
I am currently leaning towards `\raw[type=<name>]` myself but not dead set on it yet. That leaves us some flexibility to decide if we want to use mime style names or something simpler; I lean towards simpler, more like the `filetype` list in VIM or other text editors.

The `\input` option has some attraction although it is at odds with what LaTeX means by that, and we may want to reserve it for something similar.

As far as I'm concerned `\inline` gets a veto on the grounds that it confuses the block vs. inline (horizontal vs. vertical mode) layout paradigm common to many markup languages.

Omikhleia:
Moving forward:
- Rebased on 0.13
- Changed to `raw[type=<name>]`
- Tests simplified = one single test with a much simpler SVG

Omikhleia:
As an aside note, as far I am concerned, I am theoretically done with this. (SIL-escaping of this kind is needed for one of my projects, but for now I'll workaround it by re-implementing an override for 0.12.5 in my local packages.)

ctrlcctrlv:
Really not too interested in the name, would just like to see it merged ;-)

alerque:
Definitely coming.

alerque:
Just a funny note, even without this PR or the other one we've had the ability to handle this for ages and ages right in front of our noses. The existing `\script` function could have been over-ridden by any class/package to add different handlers for any format desired.

This is still coming as I think it's more ergonomic and discoverable (will merge after #1373), but I'm amused there was so much grief (my own included) over something we already had the flexibility to do...

alerque:
The only difference brought in by my merge (besides resolving code conflicts) was to make the same scope change to this that I made to command registrations: `SILE.registerRawHandler()` → `class:registerRawHandler()`. This currently does not *function* relative to the class as they still just register a global list, but it lays the usage framework so that in the future when we fix that and allow more than one class to be loaded at a time (e.g. processing one document with two classes in parallel in one SILE instance) we will be able to fix it without it being a breaking change from the usage perspective.

alerque (approved):

Omikhleia:
Not very important, but my name is lost on the test I designed, as you split the commit. :clown_face:

alerque:
Yes, sorry about that. You didn't add the expected output, so when I was adding that I split out the test and moved it to that commit. The issues is I did the re-combination in reverse order. If I have made my addition of expected output a fixup for the split commit it would have kept your authorship but I did it the other way around (used the split with the test bits as a fixup for my commit) to keep my commit message for it. I didn't think about the author meta data until too late ;-)

PR #1348, opened 2022-03-20
dotfill stretch (#1343) and leaders vertical alignment (#875)

### LINKED ISSUE #875: Can we have dot leaders that are vertically aligned, please?
I see there has been [some kerfuffle with dot leaders][i] recently, and that [a solution][s] (of sorts, at any rate) is on the way.

However, where dot leaders are used (typically, but not exclusively in tables of contents), then  one would expect all the leaders to align vertically, e.g.:

- [__0-0a-sile-toc1](https://archive.org/details/earlypoemsofwill00morr/page/n13/mode/2up)
- [__0-0a-sile-toc2](https://archive.org/details/manualoflinotype00merg/page/7/mode/1up)

Of course, there are innumerable examples, but this is so deeply engrained when dot leaders are used (of course, sometimes [they're not][t]) that to have any other layout arrangement would be odd. 

I would think a "typesetting engine" would want to make such vertically aligned dot leaders possible, at any rate. (Apparently, [it can be a challenge!][p])

[p]: http://www.ftp.tug.org/TUGboat/tb33-3/tb105veytsman-leaders.pdf
[t]: https://twitter.com/teropa/status/1028669612618141696
[i]: https://github.com/sile-typesetter/sile/issues/859
[s]: https://github.com/sile-typesetter/sile/pull/874

### LINKED ISSUE #1343: \dotfill doesn't stretch to the end of a single paragraph line
SILE 0.12.4

I would have thought `\dotfill` to behave the same way as an `\hfill`, as far as streachability is concerned, but it doesn't:
- When it is on the last line of a paragraph (or below, on its own unique paragraph line), it doesn't stretch to the full width.
- It only does so if `typesetter.parfillskip` is cancelled.

Could be a documentation issue, but I am also concerned it might not be the expected behavior, and it's kind of delicate to always notice and fix.

Test script:
```
\begin[class=book, papersize=a5]{document}
\script[src=packages/leaders]
\script[src=packages/lorem]
\nofolios
\neverindent

|\hfill{}|

|\dotfill{}|

\lorem[words=13]

Kill parfillskip:

\set[parameter=typesetter.parfillskip,value=0pt]

|\hfill{}|

|\dotfill{}|

\lorem[words=15]

\end{document}
```

Results:
image

Omikhleia:
Closes #1343 

That dubious hard-coded arbitrary "0pt plus 100000pt" stretch was even mentioned in #859... Whatever it means for an hfill(glue) to have a pseudo-infinite (1e13) stretch, the dotfill shall have the same.

Closes #875 

- Contrary to what I initially thought, I renounced having a setting for the dot separation width in `\dotfill`, as it's so easy after all to redefine the command if need be (see point below).
- All leaders (including those with a specified width) are aligned, see test `feat-875-leaders-fixed-alignement` for the actual effect.
- In the former implementation, the width was mandatory on `\leaders`, I made it optional (defaulting to an hfill glue) - so redefining `\dotfill` or other line-filling patterns becomes obvious.

N.B. Running SILE with `-d leaders` will output some leader debug information and will draw some visual hints in the output file, showing repeated leaders elements and skipped elements.


Omikhleia:
Gorgeous, no?

image

Before anyone asks - It's quite normal for different levels in a TOC not to have the same dots alignment (as they are using a different font size), if one sticks with the default (font-relative 0.5em) spacing between dots. This is why TOCs in real books often do not use dot-fills at _all_ levels. Our `tableofcontents` packages doesn't offer this feature yet (well, unless redefining `\tableofcontents:item`, which no one would logically want to do). This could be a customization feature to add. Oh, wait, my `omitableofcontents` already has it via TOC-styling :smirk_cat:

simoncozens:
Wow. Very nice!

alerque (approved):

alerque:
At first glance this looks great!

I'm working on something that will fix that build test, we need to release luasocket upstream to work right with Lua 5.4.3+ ... we're almost there. I'll check back in on PRs here when that is working.

packages/leaders.lua
```diff
@@ -1,34 +1,87 @@
...
+  if frame:writingDirection() == "LTR" then
+    w = frame:right() - frame.state.cursorX
+  elseif frame:writingDirection() == "RTL" then
+    w = frame.state.cursorX - frame:left()
+  elseif frame:writingDirection() == "TTB" then
+    w = frame:bottom() - frame.state.cursorY
+  elseif frame:writingDirection() == "BTT" then
+    w = frame.state.cursorY - frame:top()
+  else
+    SU.error("Unknwown writing direction")
```
Omikhleia, line 16: Can't I spell "Unknown" right? ;)

packages/leaders.lua
```diff
@@ -1,34 +1,87 @@
...
+      local outputWidth = SU.rationWidth(self.width, self.width, line.ratio):tonumber()
+      local leaderWidth = self.value.width:tonumber()
+      local ox = typesetter.frame.state.cursorX
+      local oy = typesetter.frame.state.cursorY
+
+      -- Implementation note:
+      -- We want leaders on different lines to be aligned vertically in the frame,
+      -- from its end edge (e.g. the right edge in LTR writing direction).
+
+      -- Compute how many leader we can fit from the current initial position.
```
Omikhleia, line 34: (cosmetic) how many leader**s**

packages/leaders.lua
alerque: This is just an onomatopoeia.

PR #1350, opened 2022-03-23
fix(packages): Better TOC title extraction for PDF bookmark (#1029)

### LINKED ISSUE #1029: SU.contentToString and \tocentry
This issue arises out of end of discussion in #1026, where @alerque helped me find code causing my problem.

Consider e.g.

```tex
\chapter{Dynasties in \em{MacBeth}}
```

For PDF bookmark (table of content entry) purposes, SILE interprets it as `Dynasties in `.

This is ultimately because in `pdf.lua` `SILE.Commands.tocentry` calls `SU.contentToString` to get the title, which, per its doc-comment, "[f]latten[s] content trees into just the string components".

I want to PR a new function which would recursively pull strings out and concatenate them all together, so we rightly get `Dynasties in MacBeth`.

Should `SU.contentToString` change, or should a new function be added?

Omikhleia:
Closes #1029 

Test case: `\chapter{The \em{great} John Reuel \font[features=+smcp]{Tolkien}}`

Before:

image

Now:

image

N.B. The same kind of "technique" is currently used in the indexer package (though with issues in its implementation) and in the bidi package (in a specialized case as far as I can tell).

EDIT. Sorry JRR, I forgot "Ronald". Hasty example :rofl:


Omikhleia:
FWIW, I do see one big **caveat**, though - If there's things such as a footnote in title section, problems arise... But this is a more general case, there are other situations and other commands which need to be "cancelled" in various places. In my unaddressed or partially addressed issues on my own tests, for instance, footnotes and labels/refs cross-references are an issue in the TOC, in running headers, etc.

packages/tableofcontents.lua
```diff
@@ -87,13 +87,42 @@ SILE.registerCommand("tableofcontents:item", function (options, content)
   end)
 end)
 
+-- Flatten a node list into just its string representation.
+-- (Similar to SU.contentToString(), but allows passing typeset
```
alerque, line 91: Along with SU.contentToString() we also have the text output backend. This makes a third implementation of roughly the same problem, and I think I have a fourth in CaSILE. I think we should review the problem and see if we can't come up with a utility function that works for more cases. It's possible we need to annotate commands with whether their output is suitable for various contexts.

Omikhleia: Well `SU.contentToString()` acts on the AST. It is possibly interesting e.g. for mere logging (as does the typesetter, for instance, in boxUpNodes), but other that than I don't really see it's usefulness: it doesn't even recurse, and it's far too early to have any clever use: at AST parsing, one has no idea what to expect from it... It's, in a certain way, plain wrong and misleading.

The text output backend is another kind of beast: it does things at a _much_ later stage, where nodes are actually likely all shaped, line-broken and boxed (which we don't really need here, by the way), and expands all contents to text (including hboxes), but with cursor positioning information, etc. - So besides the fact that coming that late implies a lots of counter-performance things one doesn't need in our case, it's not really appropriate anyway, IMHO (it's far too late and too tied to the actual pseudo-page output).

The function proposed here (also already used, as noted, with some variations, in the indexer and the bidi logic), acts upon processed (hence typeset) content (which, here at the point where I used it, actually ends up at being, normally, for text, as still unshaped nnodes). It's more or less what we need: extracting output _at that point_. (N.B. The indexer logic, though, is fragile and can do it wrong, though by nature it probably doesn't expect complex structures in its content. That's however a problem/limitation, it just also means that the current indexer has no way to pass on typesetting decisions in an indexed entry...).

Annotating commands, AFAIK, isn't a solution either. A command can process it's internal content(s) in my ways (extracting things, etc.) and can output bits of it in many ways (typeset or not, in boxes or not, etc.). So one cannot just know what parts of the contents is a reusable "output"... except by processing the command, which is what the above proposal actually does.

Annotating boxes could be more in line with the expectations (so if a command outputs several boxes, some could be marked to allow their contents to be extracted and other not, knowingly). Still, that would be an extension of the proposed function (when it meets an hbox in the processed nodes). The limit of such an approach, obviously, is that review all hboxes possibly pushed in the current code base is a hard tasks, possibly an overkill for most reasonable cases.

Omikhleia: > Annotating commands, AFAIK, isn't a solution either.

Well, I was too quick here. Passing some sort of context to commands, so that they possibly know what one intends for them at the point they are called, is a possible way. It ends up being something such as `SILE.process(content, context)` (were the context is optional, but could be used to take informed decisions). That's what I've briefly experimented so far for some commands to act accordingly (footnotes, crossrefs, etc.). It's a heavy hammer for a small case, though.

alerque: My current take is that this is a fair improvement to start with and we can iterate later. There doesn't seem to be any world in which this is *worse* than what we have now, and refactoring it into something different use cases can apply isn't at the top of anybodies todo list before a patch release goes out.

alerque (approved):

PR #1360, opened 2022-04-13
French punctuation spacing must honor current font size

### LINKED ISSUE #1359: French special punctuation spaces do not honor current font size
"One train can hide another", as we say here... 

(SILE 0.12.4) After tackling with #1358 (either removing the kern feature on the font or using the suggested fix), I still had some minor but visible justification issues in French, especially in footnotes and content typeset in smaller font. So...

Test script:
```
\begin[papersize=6in x 9in]{document}
\nofolios
\neverindent
\language[main=fr]
\font[size=50pt]

\font[size=10pt]{Lorem ipsum : « dolor » ; sit ? amet !}
\end{document}
```

Expected:

image

Observed:
image

Analysis:

The french settings for special spaces are usually font-relative (in "spc" since my changes in #1221, but it was in in "en" before). This needs to be "absolutized" at this level... (I didn't notice it earlier as the effect is small when font size changes are small (contrary to the test here!), which is usually the case for footnotes, headers or block-quotes etc.)

I'll submit a PR.

(N.B. With that fix and #1358 workaround one way or another, I got almost all lines ok over 350 pages. I still have a few lines off, though, whether in French or English, under investigation...)

Omikhleia:
Closes #1359


simoncozens (approved):
Looks good to me, thank you! If this pattern crops up again we should turn it into a utility function.

alerque (approved):

alerque:
The test failures here look like Ubuntu is having trouble with their package hosting (Azure?) and the GitHub Action runners are having trouble downloading dependencies. I'll give it a little bit and try again.

PR #1364, opened 2022-04-15
Go back to testing latest lua

alerque:
Waiting on upstream luarocks release (should be this week) and action update to match


Omikhleia:
@alerque The CI tests on 5.4.2 currently. Is this old PR still applicable?

alerque:
Last I checked the CI env still had issues with 5.4.4, but it's been a while and notably GitHub Actions has a new Ubuntu image we're running in, so it may work now.

alerque:
Looks like we're good. Lua 5.1 has been giving us trouble lately—the VM has been getting stuck regularly. I don't think this has anything to do with SILE and put a timeout on it lately so it doesn't eat through our CI quota. I guess I'll merge this and if we have any trouble we can re-evaluate.

alerque:
Hm, also it looks like I cross wired a branch in here. Stand by.

PR #1366, opened 2022-04-15
Cross-referencing package

### LINKED ISSUE #1007: Referring to a section
In LaTeX you can easily reference a section by using \label{} next to a section and then \ref{} to create the reference.

If hyperlink is loaded, it also is used. 

How can I do this in SILE?

Omikhleia:
Rationale for this PR:
- Closes #1007 
- Moreover, the [SILE Wiki](https://github.com/sile-typesetter/sile/wiki/easy-jobs) mandated me to propose it...
  image
  (so challenge taken, and I claim the bonus points too ;) )

What this packages proposes:
- Generic cross-references, with a LaTeX-inspired syntax (`\label`, `\ref`, `\pageref`...).
- An API for other packages or classes to support it (that is, register their own numbering items): here applied to the book class and footnotes package.

If it makes it into the distribution, I intend it as a replacement for my own "omirefs" package (it's the exact same code, the only change here being the documentation, using our new autodoc stuffs).

As for the TOC, it requires one or more re-run(s) of SILE. It doesn't come with tests (until we have a clean way to automate those re-runs), nevertheless, FWIW:
- Its own package documentation uses it.
- I used it in my own package and class documentations
- I used it in my book project... Including a gamebook-like chapter (of the "Choose Your Own Adventure" type).


classes/book.lua
```diff
@@ -100,6 +93,7 @@ end
 function book:finish ()
   local ret = plain.finish(self)
   self:writeToc()
+  self:writeLabelRefs()
```
alerque, line 96: This brings up an interesting point ... not really the fault of this package contribution but a larger fault in our current core implementation: it really should not be necessary to add code to the base class to get stuff to happen in these locations. We need more hook/callback functions so that packages can get their work done without needing core or class adjustments.

Omikhleia: On the other hand, there could be cases where the class knows more regarding when and what to do (e.g. in a different order, at some other point, etc.). Actually we could need to distinguish real '"packages" (that provide extra commands without needing class support) and "modules" that are somewhat reusable in various classes but depend on them. I am not fully sure yet, but it could be opened for discussion. (But it's true that the current way is a bit weird and even the naming is inconsistent, e.g. writeTocNodes, writeLabelRefs, buildIndex)

alerque: Yes, order is an issue. In a hook system probably that would come down to order in which modules were first loaded. Given there are a couple stages involved that could be hooked, it may not even be pertinent to need an order, and it if was whatever package cared about the other one could just make sure it was loaded first.

Perhaps the hooks would need to be tagged by package so that, for example, a cross-ref package could request that (if present) any toc package hooks be called before it's own...

classes/book.lua
```diff
@@ -63,6 +65,7 @@ end
 
 function book:endPage ()
   self:moveTocNodes()
+  self:moveLabelRefs()
```
alerque, line 68: Will the `endpage` hook location I added in #1414 work for this?

packages/footnotes.lua
```diff
@@ -69,6 +69,11 @@ local function registerCommands (class)
       SILE.settings:set(v, SILE.settings.defaults[v])
     end
 
+    -- Cross-reference suppport.
+    local hasRefs = type(class.pushLabelRef) == "function"
+    if hasRefs then
+      class:pushLabelRef(SILE.formatCounter(SILE.scratch.counters.footnote))
```
alerque, line 75: Lets try to think of a way to let the labelrefs package set a hook that makes footnotes respond this way without having to edit the footnotes package.

alerque (changes requested):
I think we're 99% of the way towards making this package independent so it doesn't need to touch the class or other packages. Lets get it all the way there before merging. I've spent days recently trying to untangle other packages so they can be used in isolation and/or are idempotent to load. I think that's going to be a requirement for new core packages going forward and this one only needs minor tweaks to get there.

packages/footnotes.lua
Omikhleia: I don't get it. The laberefs package here provides hooks (push/pop) so that any 3rd party packages that wishes to have labeled items can register them for a duration. Do you mean it should be the other way round? I am not sure I understand how it would work. I am not even using the standard footnotes package (and my footnotes are not even necessarily numbered) - and I have other packages using such hooks (e.g. poetry). Surely the labelrefs package cannot know all of these packages, shouldn't they be the ones responsible for registering their labeled items?

Omikhleia: Forget what I wrote above. You could certainly introduce a new paradigm and have some class hooks that packages would register to.

Omikhleia: So, how do we move forward here?

This worked for 0.12.x. I have no idea how to deal with the "requested changes" status.

Omikhleia:
Update: Rebased on current master at this date (7925d45257d2fb228d1c569eabeb5f7ade35d9ef) and upgrade to 0.14 API.

Caveat: _In the manual, this is not working very well any longer_, but for two root causes which IMHO are unrelated:
- the same package can be (re)initialized multiple times (i.e. re-instantiated in `class:loadPackage`), e.g. infonodes gets initialized 2 or 3 times in the manual, leading to `Attempted to set the same function hook twice` warnings very early when processing the manual).
- Autodoc is quite broken and/or is messing up (so here, tableofcontents re-register its commands, and labelrefs' `tocentry`  override is disabled.

classes/book.lua
Omikhleia: Done.

classes/book.lua
Omikhleia: Now using the "finish" hook.

Omikhleia:
> The same package can be (re)initialized multiple times (i.e. re-instantiated in `class:loadPackage`), e.g. infonodes gets initialized 2 or 3 times in the manual, leading to `Attempted to set the same function hook twice` warnings very early when processing the manual).

[Here](https://github.com/sile-typesetter/sile/blob/7925d45257d2fb228d1c569eabeb5f7ade35d9ef/classes/base.lua#L143), each time a package loads another one on which it is dependent, with `self.class:loadPackage()`, then a new instance is created.

Omikhleia:

Omikhleia:
Here it goes https://github.com/Omikhleia/labelrefs.sile

PR #1373, opened 2022-04-18
Overhaul include/preamble semantics (and lots of plumbing for the instantiation / run cycle)

### LINKED ISSUE #653: Overhaul include flag / preamble system
While working on #648 I ran into a series of problems involving the `-I / -- include` flag which does more than one thing: it can process SILE input files inline _OR_ execute Lua scripts _OR_ load Lua classes or modules. It is not possible to specify which, and it doesn't always guess correctly. Part of the guess work involves looking in different paths and trying a sniff approach to file contents, but the sniff happens differently depending on where the file was found. This created some really unpredictable results.

Fixing this started to make the already overgrown PR morph into a monster, so I'm backing out that change and leaving it for another day.

A further look through the code turned up some issues with the "preamble" designation. It isn't clear why all SILE runs that have an include specified need an extra finishing stage run on their class and all ones without such an include do not.

I think we need a little more regularity and control over this from the command line, maybe even multiple flags for different kinds of loading/parsing. Maybe even a way to run things _after_ the input too. And maybe a way to string multiple inputs (documents? fragments?) together in sequence.

### LINKED ISSUE #798: 0.10.x breaks -I input for *.sil classes
SILE no longer recognizes *.sil classes specified on the command line using -I. For example the docbook examples given on page 9 of the docs fails with the errors below. This also breaks other packages that use the -I command, such as [ftml-sile](https://github.com/silnrsi/ftml-sile).

```
sile -I docbook examples/article-template.xml
This is SILE v0.10.1
Loading docbook

! Unable to find file docbook
	no file 'classes/docbook'
	no file 'docbook'
	no file 'examples//classes/docbook'
	no file 'examples//docbook' at 	<nowhere>


! Unable to find file examples/article-template.xml
	no file 'examples/article-template.xml'
	no file 'examples//examples/article-template.xml' at 	examples/article-template.xml


! Could not find file at 	examples/article-template.xml
```

### LINKED ISSUE #1092: Provide packages with a way to handle raw content

### LINKED ISSUE #1413: Ability to include documents without a \begin{sile} ... \end{sile} environment
## Ability to include documents without a \begin{sile} ... \end{sile} environment

Having to wrap every included sub document in a *sile* environment is document ceremony to fit the requirement of having a root node to be valid XML. However, we are not coding XML documents, we are coding SILE documents. This places burden onto the document writer to conform to XML requirements.

When sile compiles the document, when it finds no sile environment wrapping on an included text, it could wrap that itself thus freeing the document writer from the XML conformance burden yet adhering to XML rigour that it requires itself.

alerque:
Closes #653
Closes #798
Closes #1092
Closes #1413
Lays groundwork for #1438

----

BREAKING CHANGE: The `-I` / `--include` option whose use was overloaded for more than one purpose is deprecated in favor of more specific replacements: `-r` / `--require` for loading code into SILE before input processing, `-p` / `--preamble` for processing content prior to a document and `-P` / `--postamble` for processing content after a document.

c.f. especially my [latest romment](https://github.com/sile-typesetter/sile/issues/798#issuecomment-1101466854)

> Having reviewed this a bit further, I don't think the `-I` option is even fixable. It tries to be too smart and as a result can't easily be manipulated to do any specific thing.
> 
> * Does it load Lua code? or process SILE content?
> * Does it run before loading the parser or after?
> * Does it search for files relative to the PWD? Or the input document? Or the Lua search path?
> 
> Some of these combinations are mutually exclusive and guessing can lead to breakable. I propose deprecating it entirely in favor of more specific and predictable alternatives.


alerque:
I haven't actually implemented this change yet, just updated the CLI hints and man page for what I am proposing. This draft PR is a request for comment, especially from @simoncozens.

alerque:
A couple more notes:

* We could expand the `-r` option to be able to read SIL files using the Lua search path (and hence finding things in the SILE installation rather than the current project) but I don't really see the point. We can rewrite the docbook class in Lua, and I haven't come up with other use cases for this that really make sense. Yet.

* We currently have something internally called "preamble", but it isn't actually handled as a preamble. I would suggest the actual `--preamble` content be handled *after* processing the document declaration from the main input AST, but before diving any farther into the AST. This seems like it would be more utilitarian than doing anything with it before we even know the papersize or class.

Omikhleia:
> refactor(core)!: Move inputters/shapers/outputters to isolated classes

Moving files is always risky, so just a notice in passing not to forget taking care of #1469 when this overhaul is merged...

(By the way, this change of yours here doesn't seem _necessarily_ related to "include/preamble semantics" overhaul, perhaps it could be a separate PR?)

alerque:
> Moving files is always risky, so just a notice in passing not to forget taking care of #1469 when this overhaul is merged...

No sweat, that's what merges are for, especially with `git rerere`...

> (By the way, this change of yours here doesn't seem _necessarily_ related to "include/preamble semantics" overhaul, perhaps it could be a separate PR?)

It is inextricably linked. Handling the new arguments forced me to cleanup the scope of inputter code to handle the various content types, and doing that forced the outputters to have similar scopes. I've tried piecemeal but it is kind of an all or nothing deal.

alerque:
The scope creep on this is turning into a monster, but at the same time it's sorting out a lot of intertwined issues and I'm really excited about how many different things it will improve: including normalizing (programming) classes & types such that it will be feasible to swap in Rust interfaces.

classes/base.lua
```diff
@@ -182,24 +182,133 @@ function base:runHooks (category, args)
...
+    else
+      SILE.processString(content[1], "sil")
+    end
+  end, "Process sil content. The content may be supplied either inline or using src=...")
+
+  self:registerCommand("xml", function (options, content)
+    if options.src then
+      SILE.processFile(options.src, "xml")
+    else
+      SILE.processString(content[1], "xml")
```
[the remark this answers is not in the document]
alerque: Nope, raw XML.

You can now nest XML in SIL or SIL in XML or Lua in SIL or XML (ala what `\script` has been). This just generalizes that to be a 3 way interchangeability rather than separate hacks for each (`doTexlike` from Lua, `\include` from XML but only as an external file, etc.). You can also programmatically pass strings to be parsed as content from any of those three languages. Lua inputs can also return an AST directly.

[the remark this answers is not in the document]
alerque: Caveat: the embeded SIL parser is limited to proper closing tags. For example if you open with `\begin{xml}` the next instance of `\end{xml}` will terminate the XML embedding even if it happened to be content of the XML. For `\xml{` the next *matching* `}` will close. Both cases can be worked around by including balancing tags that are not desired as content as comments.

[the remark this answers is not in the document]
Omikhleia: Ah I see indeed, you made `lua`, `sil` and `xml` all reserved in the passthroughCommands. I am unsure how it relates to include/preamble semantics, but do as you deem fit. (I don't really get why one would want to mix xml with sil, beyond what would have been achievable with `\raw` + handler, but I'm just missing the picture. Question answered.)

inputters/sil.lua
```diff
@@ -1,26 +1,50 @@
...
-    __call = function(self, command)
-      return self[command]
-    end
-  })
+
+function sil:_init (tree)
+  -- Save time when parsing strings by only setting up the grammar once per
+  -- instantiation then re-using it on every use.
+  self._parser = self:rebuildParser()
+  base._init(self, tree)
```
Omikhleia, line 41: Base init doesn't seem to have two args (i.e. one besides "self")
image

alerque: It does have two args, `self` and `tree`. The call here is notably *not* using the Lua shortcut for passing self, it is calling the function from the parent instance (in this case `base`) but passing it the instantiation of the current class as `self`.

```lua
-- These two are equivalent in lua:
base:_init(tree)
base._init(base, tree)

-- This (which we are using) is different
base._init(self, tree)
```

The latter is different because at the time we pass it `base ~= self`.

Omikhleia: Hmm... Not sure why VSCode Lua checker hightlights this line then.... 
image

Ah, the base init is a noop, but it does expect one single argument indeed (= self).

alerque: Correct, the base class is a NOOP and hence does not need the second argument. However VSCode might actually have a good point there, we should use `(_, _)` there just so even the unused prototype has the same number of arguments.

alerque: Actually I'm full of it. That appears to be cruft and no longer need. The `_init()` function never uses the tree in any variant. It gets instantiated earlier then passed content to parse later.

PR #1374, opened 2022-04-18
Setup SILE interface to be localized

alerque:
With the tooling in place to localize, I got to thinking ... passing args to a string localizer instead of generating strings in place will actually fix the issue with unused debug messages actually slowing down our runtime. This is some initial work to swap out the internals and

a) quantify if this will enough of a gain to bother with and
b) see if it makes working with the code harder

**NOT FOR CONSUMPTION**, just a proof of concept.

### DIFF
--- a/.commitlintrc.yml
+++ b/.commitlintrc.yml
@@ -38,6 +38,7 @@ rules:
     - inputs
     - installation
     - languages
+    - l10n
     - manpage
     - manual
     - math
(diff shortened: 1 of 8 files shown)


Omikhleia:

alerque:
I'm still working on this. It's low priority but even in the last few weeks of radio silence on the issue tracker I've done work that furthers this along both inside SILE and it other library projects that will eventually make this easier. Some of the content here will be worth recovering when the rest of the pieces are in place.

(Having BCP-47 locale code handling and being able to apply that to the system locale ENV vars is a major missing piece not being developed *for* this feature but will suddenly put it within easy reach.)

Issue #1362, opened 2022-04-14
Ligatures and line justification issues


Omikhleia:
And after #1358, #1359 and #1361 there comes another case of improper justification.

```
\begin[class=book, papersize=6in x 9in]{document}
\neverindent
\nofolios
\script[src=packages/lorem]

\font[family=Libertinus Serif, features="+liga"]

Lorem difficult and effects affairs of finition \lorem[words=20]

\font[family=Libertinus Serif, features="-liga"]

Lorem difficult and effects affairs of finition \lorem[words=20]

\font[family=DejaVu Serif, features="+liga"]

Lorem difficult and effects affairs of infinite findings, difficult affairs and effects of infine findings \lorem[words=20]

\font[family=DejaVu Serif, features="-liga"]

Lorem difficult and effects affairs of infinite findings, difficult affairs and effects of infine findings \lorem[words=20]

\end{document}
```

![image](https://user-images.githubusercontent.com/18075640/163379735-3945bdbc-e0dc-482b-96ca-2fa04b3f9652.png)

Seems to occur with fonts having ligatures enabled, and the ligatured text is smaller than the original text. So the test uses Libertinus, but also DejaVu, to show it"s not font-specific...

By the way, I am not sure hyphenation plays a role here, I have lines with ligatures but no hyphenation that also show the same pattern:
![image](https://user-images.githubusercontent.com/18075640/163380954-501d753c-c49e-46ff-8a0e-51bcc25ec663.png)

On the bright side: if this could be fixed too, it seems to me that this is the last source of "justification issue" I saw so far in my book, besides the above-mentioned issues...

Omikhleia:
Digging with a shovel,
1. the hyphenation and discretionaries logic actually seems wrong...
2. the line ratio computation is clearly plain wrong, and inconsistent esp. with regards to 3.
3. the nnode ouputYourself routine is pretty wrong too in the general case.

Omikhleia:
A possible course of action here could be (as a partial workaround) to at least align (2) with (3) so that both behave accordingly and the justification issues don't occur. As of today they are inconsistent, (3) ouputs the parent nnode if not hyphenated (so ligatures are applied there) but (2) computes the width based on all split (children) nnodes and discretionaries (so a potential ligature with a break point in between is ignored).

[...]

PR #1378, opened 2022-04-21
Properly justify lines containing ligatures (workaround)

Omikhleia:
Partial workaround for #1362, with two concerns in mind:
- Re-shuffle things in computeLineRatio to at least meet the expectation of nnode output so that lines wish ligatures, with fonts where those can be shorter than the non-ligatured text, are correctly justified.
- Store the natural width of the line, so that other packages that may need it (e.g. see #1344) don't need to recompute it and meet the same issues again.

Before:

image

After:

image

It's kind of beating the wind (it does not address the core issue and as the second part of the test shows, doesn't fix therefore ligatures in hyphenated words...), but at least it provides correct justification.


Omikhleia:
Two existing tests are impacted:
- tests/bug-283.sil
- tests/bug-79.sil
Unless I have dust in my eyes (which is possible!), the effect on them is hardly noticeable and seem correct to me. My feeling is that some font kerning (similar but much smaller effect than the ligatures that triggered this ticket) is at play. I leave you judge...

ctrlcctrlv:
Do you mind retitling the PR to "Justify lines with ligatures more narrow than unligated" or something? If I was confused…

ctrlcctrlv:
Also I know there's a limit on commit name length, but not on PR title length, so no need to change your commit title

alerque:
> "Justify lines with ligatures more narrow than unligated"

That would be a good *issue* title, but this isn't the issue report it is a proposed action to fix it.

ctrlcctrlv:
@alerque Sure, it just made me think all my documents were broken until I carefully read the issue :-)

alerque:
> Sure, it just made me think all my documents were broken until I carefully read the issue :-)

Most of mine *are* :frowning:. The level of brokenness is relatively minor in my case because Turkish rarely runs into ligatures, but it I reviewed a bunch of old publications and clearly the kerning has been affecting my justification just a little bit on a regular basis.

Omikhleia:
@alerque If I am not mistaken (based on the front matter), you typeset a book named "Plan". It could be interesting to check if the suggested workaround at least improves what seems to be similar justification issues there, or if they are other causes?
test-plan-justify

alerque (approved):

alerque:
Definitely a bit of a hack I'd like to see refactored, but it isn't immediately apparent to me what a better interface would look like without other major refactorings first. In the mean time it is hard to deny the current output is bogus and this is clearly correct in more cases.

After squinting at the two affected tests I agree the output is more correct. In some cases the overflow is *bigger*, but that appears to be because the bug causing hyphenation overflows is not being partially compensated for visually by otherwise bogus under-calculations.

PR #1380, opened 2022-04-23
Drop std.string.monkey_patch()

alerque:
The Lua (scare quotes) "stdlib" has some *serious* funny business under the hood. I understand why some bits of it are *convenient* at times, but debugging anything it has touched is a nightmare.

This PR is actually bits extracted from #1065. It was getting really confusing trying to figure out what I broken by changing the class model and what was related to the monkey patch business conflicting with explicit meta methods I was trying to set using Penlight. Separating just the de-monkify bits let me work through all the tests.

By the way in case anything ever `git bisect`s back to here, one way to test if you problem is related to this change or not (especially if you have downsteam code that might have built on the assumption SILE's Lua strings were monkey patched) is to load it manually yourself:

```console
$ sile -e 'require("std.string").monkey_patch()' your_file.sil
```

If that works, then

a) add it as a require for *your* project, but leave SILE out of it
b) consider refactoring to use standard Lua or explicitly using `std.string` instead of monkey patching `string`.

### DIFF
--- a/core/debug-output.lua
+++ b/core/debug-output.lua
@@ -127,7 +127,7 @@ SILE.outputters.debug = {
     else
       buf = table.concat(value.glyphString, " ") .. " w=" .. _round(width)
     end
-    writeline("T", buf, "("..value.text..")")
+    writeline("T", buf, "(" .. tostring(value.text) .. ")")
   end,
 
   setFont = function (self, options)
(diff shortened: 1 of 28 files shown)


alerque:
Just a heads up @Omikhleia I suspect this breaking change will probably affect your downstream packages as it will some of mine. Almost always using `tostring()` on the failing bit to explicitly trigger `__tostring()` metamethods on `table`s and automatically cast `nil`s to `string`s will do the trick, but odder ones like a7dd754 are popping up too.

PR #1381, opened 2022-04-23
Refactor string handling with luautf8

alerque:
My claim that this is a performance boost needs confirming with an actual benchmark, but we don't even have a *test* that uses this file so in all honestly I don't even know if this chance works at all. It might not even do the right thing.


Omikhleia:
The contributor himself said "It might not even do the right thing."

Omikhleia:
BTW: If is is reopened, it would need an MWE with a benchmark.

PR #1383, opened 2022-04-29
An hrule with a depth shall not affect current baseline

### LINKED ISSUE #1382: hrule with depth option wrongly affects the current base line
SILE 0.12.5

Test script
```
\neverindent
\nofolios
\script[src=packages/rules]

Lorem \hrule[depth=2mm, width=1cm] Ipsum
\end{document}
```
Observed:

image

Expected:

The depth should not have moved the baseline, i.e.

image

Moreover:

- Package documentation states: "The first command is `\hrule` which draws a line of a given length and thickness, although it calls these `width` and `height`. (A box is just a square line.)" 
- The in-code command documentation states `Creates a line of width <width> and height <height>`
- EDIT: package documentation also says "Like images, rules are placed along the baseline of a line of text." This caused me a big LOL ;) (that a mere blob of ink is being compared to an image?)

This is misleading and probably ill-formulated: the command also has a `depth` option... BTW, as far as I can tell, only one existing test uses this depth option, but on a standalone line without surrounding context, so the issue could not be seen there. The box vs. line wording is also weird (the command rather makes a rectangular blob of ink -  It can be used to produce horizontal or vertical lines depending on the dimensions, and the `h` in the name just refers to the fact that is is an horizontal box, no ?)

Personally I don't think the depth option should necessarily be removed. Not only it is in line with what a box actually is, but It can be pretty handy sometimes, e.g.

image

(Ok, I could have done this another way without using a rule depth for the longer barcode lines, but heh...)

On the other hand, it all depends whether we want to be be similar to TeX's `\hrule` (which has optional width, height, depth parameters) or to LaTeX's `\rule` (which has width, thickness and raise options, the latter effectively being a depth below baseline if negative). It seems to me the first case is more general...

Anyway, the fix is simple - pick up your choice:
- Remove the depth option...
- Or rather restore the Y cursor after drawing the rule (and update documentation to mention that depth option and rephrase the weird wording).

(EDIT: Fixed the TeX and LaTeX similar command descriptions)

Omikhleia:
Closes #1382 

That would be my pick, at least.


alerque:
Openresty has been segfaulting sometimes in CI. It is a memory problem on the runner and not our deal. I sometimes re-run it a few times just to check, but if both 5.4 and the 5.1 based plain luajit pass it is safe to assume most changes probably wont' actually be a problem in openest either.

alerque (approved):

Issue #1114, opened 2020-11-30
Make sure class, frame, and typesetter initialization sequence isn't co-dependent


alerque:
While working on #1065 it has become apparent that there is a bit of a chicken & egg problem going on with classes vs. typesetters vs. frames. Ostensibly the class initialization is called first, then the frame, then the typesetter. This makes some sense, but the class sometimes actually tries to do things with the typesetter. This race condition is/was being papered over by `stdlib` and poor variable scoping in SILE. The class initialization was actually causing a new typesetter to initialize, then when the frames got going another one was started.

I think the whole init sequence and relationship of typesetters to frames and classes should be reviewed, but I can't even load that into my brain while refactoring the object model for the document classes, so I'm leaving this issue to track addressing this later.

alerque:
Will be largely addressed by other refactoring in v0.13.0, but still needs review.

alerque:
Yes, yes they are co-dependent. BAH. So much untangling. #1399 picked off a few more bits but we're not out of the woods here yet.

alerque:
This isn't going to make the v0.13 cut. I've made a lot of progress on it, but there are some scoping issues that we are currently solving with the global `SILE` variable and just blindly trusting that some other thing will be done first. I believe the situation is solvable, but it will require a little bit more isolation and passing of state. Not much actually, but a little.

At least one of these places is going to be a breaking change. I suspect I will be able to shim it for at least one release cycle, but I can't put that shim in place with the stdlib shim in place for classes. Hence a two stage deprecation. The v0.13.x series will get us on the right libraries and class inheritance model, then when we remove that shim in v0.14.0 we can add the shim for access to the typesetter from a global.

We could make both breaking changes at once, but I think the shims make it easier to migrate projects. When you run them they still build (for at least one release cycle) and spit out appropriate warnings with hints about how they need to be re-coded. As long as we *can* do that I want to keep trying to ease the pain of breakage.

alerque:
This will never be testable until we can stop assuming a single global SILE instance. A lot of this sequence stuff is untangled, but as long as SILE cannot properly be used as a library (with potentially multiple instances) because of its assumptions about the global scope we will never be able to test co-dependencies in a sane way. I dropped the version milestone to reflect that and added it to the relevant topical project instead.

Issue #683, opened 2019-10-16
Refactor twoside package


alerque:
1. The use of a module variable is going to produce a stateful action that might trip us up later in the event two or more typesetters are working on different places in a document. Rather than wait for bugs to crop up in usage this should be moved into the current typesetter or class state to just avoid those surprises.

2. As noted in comments the implementation reaches far too deep into other code for comfort.

Omikhleia:
Instead of opening a new issue, I'll provide my experimenting results here, since it seems pretty well related with the issues I am facing and the things a refactor might need to address. I hesitated with #385, also somehow related, but which seems more specific.

I am trying to have 3 (two-side) master sections than can be switched anytime in the course of a book:
- The main (right, left) masters which correspond to the default regular book pages
- The (tworight, twoleft) masters for a two-columns layout: the use case for it is, after regular pages, to have a two-column lexicon or an index.
- The (landright, landleft) masters where the content (and ideally the footnotes, but see below) are rotated: the use case for it is for wide (possibly multi-page) tables (chronologies etc.) that wouldn't fit will on a portrait page.

Expected layout design illustration:
![image](https://user-images.githubusercontent.com/18075640/159121553-59d834e1-f51c-4e72-97bc-7010089518d5.png)

I can alleviate some constraints, for clarity:
- Just to keep it clear, it is acceptable to change masters on a new page (i.e. I don't need, esp. regarding the two-column case, switching between these layouts in a same-page flow).
- In the two-column case,
  - It is acceptable for footnotes to be in the second column only. In this scenario, footnotes would not be recommended (and endnotes are indeed preferable in such cases), so a "degraded" behavior is ok for _minimal_ support of a few footnotes. I.e. I don't really care what happens at the end of that section or if footnotes are so long that they fill the column.
  - Eventually, balanced columns would be ideal, but I propose ignoring them for now (I tried and failed, but this is still minor regarding the expected layout and I understand the balanced-frames package is still experimental).
- In the landscape case,
  - I have no idea if the footnote frame can be rotated as shown in the above design and if the insertions module can "steal" space as expected. I propose to alleviate that requirement for now and let the footnotes live at their usual "bottom" place above the folio...
  - The mirrored content frames in the design are rotated so that too be always oriented towards the spine of the book. Let's admit this is debatable and accept alleviating that requirement.

[...]

Issue #630, opened 2019-06-04
Parallel not working with book class


fhardison:
I'm working on getting a two column layout where one of the columns is wider than the other. Per Simon's advice, I tried to tweak the diglot class to inherent the book class rather than plain. When I setup the parallel package to use one of the columns defined in master template, it produces a bunch of overflow errors and each word is typeset into it's own page (i.e. a two page document becomes apx. 500 pages with one word on each). 

Bellow are the changes I made to the book class which I call `grbook` and the custom diglot class is below that.

## Changes made to book class

```
-- grbook
book:loadPackage("masters")
book:defineMaster({
    id = "right",
    firstContentFrame = "a",
    frames = {
      a = {
        left = "8.3%pw",
        right = "66%pw",
        top = "11.6%ph",
        bottom = "80%ph"
      },
      folio = {
        left = "left(a)",
        right = "right(b)",
        top = "bottom(a)+3%ph",
        bottom = "bottom(a)+8%ph"
      },
      runningHead = {
        left = "left(a)",
        right = "right(b)",
        top = "top(a)-8%ph",
        bottom = "top(a)-3%ph"
      },
      b = {
        left = "71%ph",
        right = "100%pw-left(a)",
        top = "top(a)",
        bottom = "bottoom(a)"
      }
    }
  })

```

## My custom diglot class

```
local book = SILE.require("grbook", "classes")
local grdiglot = book { id = "grdiglot" }

SILE.require("packages/counters")
SILE.scratch.counters.folio = { value = 1, display = "arabic" }

grdiglot:loadPackage("parallel", { frames = { left = "a", right = "b" } })

return grdiglot
```

simoncozens:
OK, I can replicate this and it looks like the frames being fed in to the `parallel` package don't have any constraints on them. It further looks like the `newPage` routine is not being correctly called. Will keep looking into it.

simoncozens:
Various things are wrong here, some yours and some mine! In your definition of frame `b`:

```lua
      b = {
        left = "71%ph", -- This should be pw
        right = "100%pw-left(a)",
        top = "top(a)",
        bottom = "bottoom(a)" -- This should be bottom
      }
```

In `packages/parallel.lua`, there are some bugs:

```diff
@@ -71,18 +73,18 @@ local setupParallel = function (klass, options)
   local o = klass.newPage
   klass.newPage = function(self)
     allTypesetters(function (n,t)
       calculations[n] = { mark = 0 }
     end)
[...]

PR #1392, opened 2022-05-19
Refactor typesetter, removed last usage of Lua stdlib

alerque:
This still leaves the core `requrie("std")` and dependency, but they now do nothing. We can pass all our tests without them and I plan to drop the dependency entirely in a later release. For now we leave it for the ease of downstream packages that aren't expecting it to be removed yet, and the deprecations shims for class and typesetter inheritance rely on them. When we drop those shims we can drop the dependency too.

Partially addresses #1114
Partially addresses #630
Partially addresses #683


simoncozens:
I was trying to look at that. :-) It looks fine, but I can't build SILE right now anyway.

PR #1393, opened 2022-05-19
Stop depending on & providing std

alerque:
To be released no earlier than v0.14.0


alerque:
Gah, no wonder I was having dejavu while working on #1420.

alerque:
Obsolete. I merged the good ideas about deprecation warnings from this into #1420.

PR #1414, opened 2022-06-01
Move special callbacks in TOC package out of class into package

alerque:
- chore(packages): Work on making footnote package idempotent
- feat(classes): Add hook system for more versatile packages
- refactor(packages): Get TOC package stuff out of core classes


packages/tableofcontents.lua
```diff
@@ -159,8 +92,96 @@ SILE.doTexlike([[%
...
+    SILE.call("fluent", {}, { "toc-title" })
+  end)
+
+end
+
+return {
+  init = init,
+  registerCommands = registerCommands,
+  exports = {
+    writeToc = writeToc,
```
Omikhleia, line 182: Why do we need to still export these functions (and pollute the class with them), if they are to be called via the new hooks we registered above?

packages/tableofcontents.lua
```diff
@@ -159,8 +92,96 @@ SILE.doTexlike([[%
...
+        SILE.call("tableofcontents:item", {
+          level = item.level,
+          pageno = item.pageno,
+          number = item.number,
+          link = linking and item.link
+        }, item.label)
+      end
+    end
+    SILE.call("tableofcontents:footer")
+    _tableofcontents = toc
```
Omikhleia, line 124: Somewhat unrelated, but since you are on refactoring things here and adding hooks, etc...

Since `_tableofcontents` is a local variable, it means that one cannot easily just redefine the tableofcontents command to process and display it (possibly very) differently. I had, typically this problem in my "fancytoc" package where I just want to show the TOC in a different way (= hierarchical, with braces, rather than line by line).
There could be two approaches:
- Moving that to a scratch global variable (though it does expose some of the internal trickery) so that a package wishing to overwrite the `\tableofcontent` command could fully reimplement it.
- Or split the whole TOC loading in an exported function. I used that strategy for my fancytoc package in my repo, not yet in main but [here it is for the record](https://github.com/Omikhleia/omikhleia-sile-packages/blob/book-project-iteration/packages/experimental/fancytoc.lua)

packages/tableofcontents.lua
```diff
@@ -159,8 +92,96 @@ SILE.doTexlike([[%
...
+return {
+  init = init,
+  registerCommands = registerCommands,
+  exports = {
+    writeToc = writeToc,
+    moveTocNodes = moveNodes
+  },
   documentation = [[
 \begin{document}
 The \autodoc:package{tableofcontents} package provides tools for class authors to
```
Omikhleia, line 187: The documentation afterwards still refers to "At the end of each page, the exported Lua function \code{moveTocNodes}
should be called to collate the table of contents entries and store which page they’re on." (etc.)
With the proposed hook, I am not sure we should document things that are not going to be used by class developers any longer, and now pertain to low-level internal details.

packages/tableofcontents.lua
alerque: Good point, maybe we don't...

Unless it's possible some other package would need to call it directly.

packages/tableofcontents.lua
alerque: Agree.

alerque: I rewrote the package docs to reflect the new usage where the end user only needs to consider where `\tocentry` is run, not the functions now covered by class hooks.

packages/tableofcontents.lua
alerque: I agree this is less than ideal but am going to leave it for a future refactor to take care of. We have a huge mix of private module vars vs. random scratch globals with inconsistent names. Also it isn't clear what things in scratch are "public" in the sense that changing our use of them should be considered breaking vs. those being private implementation details to use at your own risk.

alerque: c.f. #1417

PR #1425, opened 2022-06-12
Add hrulefill command and redo fullrule so as not to overflow lines

### LINKED ISSUE #845: Sometimes indent in unexpected and unwanted
I hate to open up the can of worms of context-sensitive behavior, but actually @simoncozens opened it himself in [this CaSILE issue](https://github.com/sile-typesetter/casile/issues/46). In that example the situation is pretty cut and dry. Given the _definition_ of `\fullrule()` as being 100%lw it is unexpected and unwanted that the rule should be indented.

Or is it?

One of the nightmares of TeX is never knowing how something is going to behave because different commands work, don't work, or work differently depending on what environment they happen to be nested in.

Fullrules won't be the only thing that expect to use 100%lw. Tables, images, etc. might all expect this. Should we add `\noindent` as part of their definitions? Does that really help us over expecting the user to have full control over when they do/don't want indent?

Omikhleia:
Closes #845 

Another PR on the **rules** package that should likely be considered before #1334, as an attempt to address the discussions at #845:

1. Add an `\hrulefill` command that creates an infinite-strech line, similar to `\hfill` and `\dotfill`
   - It's kind of similar to TeX's `\hrulefill` and  and LaTeX's `\rule`, which are both kind of overlapping in some ways, but doesn't do exactly the same thing. I don't think we can avoid the naming similarity anyway.
   - As for options,
     - `thickness` allows specifying the drawn line thickness
     - `raise` or `position` are exclusive and allow specifying the position relative to the baseline.
     - The former is a measurement. The latter can be set to `underline` or `strikethrough`, to use the corresponding positioning (and the thickness, if not specified).
       - I think `\hrulefill[position=underline]` is interesting e.g. for input fields (`Name: _____`) where one could expect that, rather than a more or less random placement.
       - No idea if the `\hrulefill[position=strikethrough]` can have any use, but it comes at nearly no cost now that we have proper strikethrough support.

2. Try to fix the existing `\fullrule`, while maintaining compatibility for now, but with warnings:
   - Deprecate option `width`: it's a "full" rule, so shouldn't have a width.
   - Deprecate option `height` in favor of `thickness`: it seems more correct on a semantic viewpoint, we are not dealing with the usual height, depth, width concepts here
   - Deprecate usage in horizontal mode (and when the compatibility will be removed, it will force vertical mode)
   - Clears indent before it: that could be considered a breaking change, _but_ the old behavior was wrong anyway with its "100%lw" default, which caused a line overflow if there was an indent, so I think it's safe to consider this a fix anyway.
   - Internally, it doesn't use the frame and the r/l skips, but just uses a fill, so should be more robust and simple (?)
   - All in all, the aim is to have this command as a mere convenience, similar to HTML's `<hr/>` or (Pandoc-)Markdown `---` standing on a line of their own - nothing more, nothing less.


packages/rules.lua
```diff
@@ -1,12 +1,48 @@
...
+  local face = SILE.font.cache(fontoptions, SILE.shaper.getFace)
+  local font = ot.parseFont(face)
+  local upem = font.head.unitsPerEm
+  local underlinePosition = font.post.underlinePosition / upem * fontoptions.size
+  local underlineThickness = font.post.underlineThickness / upem * fontoptions.size
+  return underlinePosition, underlineThickness
+end
+
+local function getStrikethroughParameters()
+  local ot = SILE.require("core.opentype-parser")
```
Omikhleia, line 13: Oops, should be a mere `require` here

Omikhleia: Done

packages/rules.lua
```diff
@@ -31,23 +67,80 @@ local function registerCommands (_)
     })
   end, "Draws a blob of ink of width <width>, height <height> and depth <depth>")
 
+  SILE.registerCommand("hrulefill", function(options, _)
```
alerque, line 70:
```suggestion
  SILE.registerCommand("hrulefill", function (options, _)
```

As a minor style point and 100% my preference, not a requirement, I try to format function *definitions* with a space after the function keyword and function *calls* without, so

```lua
-- define
local foo = function (bar) end
-- call
foo(bar)
```

Similarly I keep stuff in parens tight (`foo(6)`) but stuff in braces loose (`foo({ bar = 6 })`). Again a super minor issue but I do periodically normalize all SILE's code to this convention to make it easier to search for things & use my own editor shortcuts ;-)

packages/rules.lua
```diff
@@ -31,23 +67,80 @@ local function registerCommands (_)
...
+        end)
+      else
+        -- This was very broken anyway, as it was overflowing the line.
+        -- At least we try better...
+        SILE.call("hrulefill", { raise = raise, thickness = thickness })
+      end
+     return
+    end
+    if options.width then
+      SU.deprecated("\\fullrule with width", "\\hrule and \\raise", "0.13.0", "0.14.0")
```
alerque, line 126:
```suggestion
      SU.deprecated("\\fullrule with width", "\\hrule and \\raise", "0.13.1", "0.14.0")
```

We can't deprecate retrospectively, the first version argument should be the version where we start warning and the second one where we stop shimming and error.

I would prefer to keep shimming at least one major version cycle when feasible. Not it was *not* feasible to do with the stdlib / document class stuff so that is set to error after just one major version.

packages/rules.lua
```diff
@@ -31,23 +67,80 @@ local function registerCommands (_)
...
-    SILE.call("raise", { height = options.raise or "0.5em" }, function ()
-      SILE.call("hrule", {
-          height = options.height or "0.2pt",
-          width = options.width or "100%lw"
+    local thickness = SU.cast("measurement", options.thickness or "0.2pt")
+    local raise = SU.cast("measurement", options.raise or "0.5em")
+
+    -- BEGIN DEPRECATION COMPATIBILITY
+    if options.height then
+      SU.deprecated("\\fullrule[…, height=…]", "\\fullrule[…, thickness=…]", "0.13.0", "0.14.0")
```
alerque, line 105:
```suggestion
      SU.deprecated("\\fullrule[…, height=…]", "\\fullrule[…, thickness=…]", "0.13.1", "0.14.0")
```

alerque (approved):

packages/rules.lua
alerque: If there isn't an urgent *need* to have an earlier EOL for the shim, lets make it 2 major bumps by default, so v0.15.0. If there is some reason to hurry to drop the shim (e.g. if some other refatoring is waiting on it being completely deprecated) then 1 major version is fine by me.

Omikhleia: I fully agree on the "2 major bumps" - so re-targeted to 0.15.0.
Done.

PR #1438, opened 2022-06-15
Work on making SILE functional as a Lua library

alerque:
- test(packages): Refactor unit tests for package loading
- chore(packages): Move all packages to dedicated directories
- chore(packages): Work on making color-fonts package idempotent
- chore(packages): Work on making font-fallback package idempotent
- chore(packages): Work on making pdfstructure package idempotent


Omikhleia:
This old PR never came to fruition as-is and now has **many** conflicts, an **empty** description and no **link** to any open issue.

_I do know that recent changes (the Rust wrapper, etc.) are part of the solution towards the goal mentioned here, but the path from this PR to it are not clear._

alerque:
In spite of appearances I'm actively working on this. Last week in fiddling with [decasify](https://github.com/alerque/decasify/) I spent extra time playing with build arrangements as a PoC for *this* problem (and succeeded in releasing it). In the mean time it is not practical to pursue directly until after #2072 lands.

PR #1443, opened 2022-06-18
Cleanup font fallback handling

### LINKED ISSUE #716: Unexpected 'glyph(s) not available' errors from font-fallback package
I tested this with SILE c0542ca6715946192fdfe4abb239a3ea6705d784

```
\begin[class=jbook]{document}

\script[src=packages/font-fallback]

\font[family=Roboto]
\font:add-fallback[family=Noto Serif CJK JP]

こんにちは!

\end{document}
```

```
$ sile test.sil
This is SILE v0.9.5.1-413-gda03715
<test.sil>

! Some glyph(s) not available in any fallback font, run with '-d font-fallback' for more detail at in "␤␤こんにち…"

! Some glyph(s) not available in any fallback font, run with '-d font-fallback' for more detail at in "␤␤こんにち…"

! Some glyph(s) not available in any fallback font, run with '-d font-fallback' for more detail at in "␤␤こんにち…"

! Some glyph(s) not available in any fallback font, run with '-d font-fallback' for more detail at in "␤␤こんにち…"
[1] 
```

### LINKED ISSUE #1442: Font fallback size does not accept units
SILE 0.12.5

Reading its code, `\font:add-fallback` accepts all font options, so I was tempted to do this:

```
\script[src=packages/font-fallback]
\language[main=en]
\font[family=Libertinus Serif]
\font:add-fallback[family=Symbola, size=0.8em]

Ballot box with check (U+2611) is not in Libertinus, so fallback to Symbola: ☑

\end{document}
```

Expected result:
I would have expected the checkbox to be show at 0.8em of the font being replaced.
The rationale is that Symbola characters such as this one are somewhat bigger than the main font, so it would be neat to have them scaled down a bit to better match it -- expectedly with a relative unit, relative to the current size of the font being replaced.

Observed result:
```
SILE v0.12.5 (Lua 5.2)
<testfb.sil>

! Some glyph(s) not available in any fallback font, run with '-d font-fallback' for more detail at in "␤␤Ballot box with …"

Error detected:
	/usr/share/lua/5.2/std/string.lua:48: bad argument #2 to '_format' (number expected, got string)
```

(N.B. The warning is like the same issue as #716)

Additional tests:
- The same error is observed too, say, with "8pt"
- It does work with just "8" without units:
```
\script[src=packages/font-fallback]
\language[main=en]
\font[family=Libertinus Serif]
\font:add-fallback[family=Symbola, size=8]

Ballot box with check (U+2611) is not in Libertinus, so fallback to Symbola: ☑

\end{document}
```
image

N.B. without specify a size in the fallback, one gets:
image

So the size is actually taken into account - it's just it doesn't work with units being provided (regardless whether it would do it properly as expected with a _relative_ unit).

EDIT: Full stack trace if it helps:
```
Error detected:
	/usr/share/lua/5.2/std/string.lua:48: bad argument #2 to '_format' (number expected, got string)
stack traceback:
	[C]: in function '_format'
	/usr/share/lua/5.2/std/string.lua:48: in function 'format'
	/usr/share/sile/core/font.lua:73: in function '_key'
	/usr/share/sile/core/font.lua:109: in function 'cache'
	/usr/share/sile/core/harfbuzz-shaper.lua:31: in function 'shapeToken'
	/usr/share/sile/packages/font-fallback.lua:82: in function 'shapeToken'
	/usr/share/sile/packages/font-fallback.lua:135: in function 'createNnodes'
	/usr/share/sile/core/nodefactory.lua:257: in function 'shape'
	/usr/share/sile/core/typesetter.lua:283: in function 'shapeAllNodes'
	/usr/share/sile/packages/bidi.lua:237: in function 'boxUpNodes'
	/usr/share/sile/core/typesetter.lua:568: in function 'leaveHmode'
	/usr/share/sile/core/typesetter.lua:255: in function 'endline'
	/usr/share/sile/core/typesetter.lua:239: in function 'typeset'
	/usr/share/sile/core/inputs-common.lua:51: in function 'process'
	/usr/share/sile/core/inputs-texlike.lua:185: in function 'process'
	/usr/share/sile/core/sile.lua:263: in function 'readFile'
	/usr/bin/sile:60: in function </usr/bin/sile:60>
	[C]: in function 'xpcall'
	/usr/bin/sile:60: in main chunk
	[C]: in ?
```

alerque:
Closes #716
Closes #1442

So far this gets the font size options passed through per #1442, but this only affects the space allocated (i.e. the hbox in the outputter adjusts to the right size but the glyph is still shaped wrong). The logic for fallback handling is totally borked and I think there is an off by one error or something like that causing the previous iterations options to be used. In any case by the time the fallback font options are applied to the node it is already shaped with the original fonts options. Unraveling this logic should fix both issues.


alerque:
Hey @simoncozens the fallback font shaper chunking makes no sense to me, maybe you can shet some light on it. I'm using this MWE:

```sile
\script[src=packages.font-fallback]
\language[main=en]
\font[family=Libertinus Serif]
\font:add-fallback[family=Symbola, size=.5em]

☑

b☑

c☑d

\end{document}
```

It throws some warnings about some glyphs not being found in any fallbacks, but all the glyphs we care about *are* found and rendered correctly. The glyph it seems to have trouble with is a newline!

https://github.com/sile-typesetter/sile/blob/master/packages/font-fallback/init.lua#L74

If I debug (e.g. add `SU.dump(chunk)`) it shows the chunks at this point have leading and/or trailing newlines. They are also not being shaped in the same order they were input, but that might be okay.

Note the trouble cases are the ones where the glyph not found in the base font are at the end of a paragraph, it works as expected if they are embedded in the middle of an hmode run.

Ideas?

alerque:
Ignore my last comment, I found that issue. The fallback queue was iterating over native Lua's idea of the length of a string, but the string is UTF8 and we need to iterate over the UTF8-aware length. There are other bugs too...

alerque:
@Omikhleia Any chance you could check this out and see if if does what you expect? I have it passing options and the shaping does adjust to sizes now, but I think there is a scope issue with when they get absolutized still. Closer but not quite there maybe...

@simoncozens If you get a chance to comment on what @24cc4a2f was about it would help a lot. I've refactored the entire fallback handling to get sane error messages and pass options. One major difference is it now only processes byte runs that actually fell-through as `.notdef` instead of running the whole input text through all the fallbacks and extracting the relevant segments. I couldn't come up with an MWE that exibits the need for the multi-glyph bit so that hasn't been refactored back in yet.

alerque:
@simoncozens I blindly refactored the unknown code so (right or wrong) I think it will keep doing the same thing, but this bit was never being tested before and even mucking around with fonts and example sequences from the Harfbuzz sources I haven't come up with anything that tests it now. Since this the upcoming patch release is seriously holding up work on other projects I'm going to merge and move on, but we still need a test for this and the logic may or may not be correct. I think the bulk of the fallback handling is substantially improved, but I know there are still unfixed/untested edge cases.

PR #1444, opened 2022-06-18
feat(languages): Add full Esperanto language support

ctrlcctrlv:
Fliko ankaŭ enhavas la verkon de @pok49 en lia projekto "vortdivido".

# Enigo

```lua
SU.debug("eo", eo_string(3333300030))
SU.debug("eo", eo_string(3000003))
SU.debug("eo", eo_string(1000000))
SU.debug("eo", eo_string(1000001))
```

# Eligo

```ini
[eo] tri miliardoj kaj tri cent tri dek tri milionoj kaj tri cent mil tri dek
[eo] tri milionoj kaj tri
[eo] miliono
[eo] miliono kaj unu
```

Direndaĵojn mi ne havas plu. :wink:


languages/eo.lua
```diff
@@ -1,4 +1,81 @@
...
+SU.formatNumber.eo = {
+  string = function (num)
+    return eo_string(num)
+  end,
+  nth = function (num)
+    return num .. 'a'
+  end
+}
+--[[
+SU.debug("eo", eo_string(3333300030))
```
alerque, line 77: Can we add a `busted` style unit test for these? If there isn't one already I'll work one up for Turkish so there is an easy copy/edit model to follow.

languages/eo.lua
```diff
@@ -1,4 +1,81 @@
...
 }
+
+SILE.doTexlike([[%
+\define[command=tableofcontents:title:eo]{Enhavoj}%
+\define[command=book:chapter:pre:eo]{Ĉapitro №}%
```
alerque, line 12: This command form of adding string localizations has been completely deprecated in SILE v0.13.0 in favor of Fluent format l10n files. Please move these to `i18n/eo.ftl`; the other files in that folder should make it pretty obvious how to format these strings.

If the № is mandated by or at least normative in Esperanto then of course it's fine; if is optional and just your preference then lets stick with just the plain variant and let users set their preferred localization via `\ftl{}`. Also note if dropping the № this should probably then end in a space.

alerque (changes requested):

languages/eo.lua
ctrlcctrlv: It's required unless I call nth(…), but I wasn't sure how

ctrlcctrlv: Basically you can write it either "Ĉapitro №*n*" or "Ĉapitro 1a" (2a…*n*a). I honestly don't know which is more common, probably the latter, but it's not like books abound in Esperanto lol. I've seen it before with the ª as well. (Hmm, maybe my fonts should have an <sup>aj</sup> ordinal indicator…)

languages/eo.lua
ctrlcctrlv: I don't think there is one :) the English versions aren't tested either fyi

languages/eo.lua
alerque: Okay, then leave it for now. I haven't implemented function callbacks in fluent-lua yet. When I do we can update the localization with a callback to the number formatting utility.

The strings still need moving to Fluent though.

languages/eo.lua
alerque: I see that! I just added one to this PR using the strings from the body of this PR. It looks like there is some extra whitespace being added, perhaps the `eo_string()` function could be refined to not pad everything, or (a bit lazier) at least trim it all off after assembly.

languages/eo.lua
ctrlcctrlv: I was able to fix both issues. I misunderstood the `*:post` macros.

languages/eo.lua
ctrlcctrlv: > It looks like there is some extra whitespace being added

Really? It does trim though

languages/eo.lua
```diff
@@ -1,4 +1,87 @@
+-- Verkita originale por Fredrick BRENNAN (copypaste@kittens.ph) 2022-jare
+
 SILE.hyphenator.languages["eo"] = {exceptions={}}
 SILE.hyphenator.languages["eo"].patterns = {
-"a1", "b1", "c1", "ĉ1", "d1", "e1", "f1", "g1", "ĝ1", "h1", "ĥ1", "h1", "i1", "j1", "ĵ1", "k1", "l1", "m1", "n1", "o1", "p1", "r1", "s1", "ŝ1", "t1", "u1", "ŭ1", "v1", "z1",
+-- La fonto de la ŝablonoj ĉisubaj estas Sergio Pokrovskij (@pok49)'s GitHub-projekto «vortdivido».
+-- Ili uzendas LaTeX-permisile.
```
Omikhleia, line 6: I have not idea what this means. I don't mind the native language, but comments should IMHO always have an English version, for developers to understand what's at stakes.

ctrlcctrlv: Oh, that's fine. I thought only E-o speakers would care, it's data sourcing information.

languages/eo.lua
```diff
@@ -1,4 +1,87 @@
...
+}
+
+SILE.settings:declare({ parameter = "languages.eo.use_raised_ordinal_suffix", type = "boolean", default = false })
+SILE.settings:declare({ parameter = "languages.eo.hyphen_before_ordinal_suffix", type = "boolean", default = false })
```
Omikhleia, line 11: SILE manual has a chapter about language-specific settings. Would need to be updated (in a separate "docs(manual)" commit likely.

languages/eo.lua
```diff
@@ -1,4 +1,87 @@
...
+  a = SILE.settings:get("languages.eo.use_raised_ordinal_suffix") and "ª" or "a"
+  if SILE.settings:get("languages.eo.hyphen_before_ordinal_suffix") then
+    a = "-" .. a
+  end
+  return n .. a
+end
+
+SILE.registerCommand("book:chapter:post:eo", function()
+  SILE.typesetter:typeset(eo_nth())
+  SILE.call("medskip", {})
```
Omikhleia, line 32: Typesetting decisions in a language-specific file seem wrong to me. We'd perhaps need another discussion about these ugly hooks), still something looks amiss (Why a medskip? That should be a class decision, no?)

languages/eo.lua
Omikhleia: Also these settings have overly long names ^^

They both seem to refer to ordinals, so perhaps `languages.eo.ordinal.raisedsuffix` and `languages.eo.ordinal.hyphenbefore` ? (just an idea, to keep some namespacing and shorter flag names)

languages/eo.lua
ctrlcctrlv: https://github.com/sile-typesetter/sile/blob/b211d9d842276f66b88e2107aea52249fd7c63d1/languages/ja.lua#L183

ctrlcctrlv: https://github.com/sile-typesetter/sile/blob/b211d9d842276f66b88e2107aea52249fd7c63d1/languages/fr.lua#L25-L51

Omikhleia: I've no idea if the first mention for japanese is correct, and no reason to believe it....

The second mention regarding the French stuff is wholly unrelated: that's typography according to the rules, not decisions how chapter numbers could be presented.) Also note that they are customizable (as Canadian and Swiss, for instance, have some variations of these.)

Unless there are normative references for the presentation of chapters, that say a medskip (with a given definition) is the only typography-respecting way for this, it's at best an author decision...

ctrlcctrlv: dude whatever

ctrlcctrlv:
Merger of this PR is blocked on merger of #1445.

`.ftl` files do not get put into data directory.

languages/eo.lua
Omikhleia: > dude whatever

?

Sorry I don't get your reply.

Maybe I have just misunderstood the very idea of introducing the "fluent" things, but with its, the **book** class now defines by default:
```
  SILE.registerCommand("book:chapter:post", function (_, _)
    SILE.call("fluent", {}, { "book-chapter-post" })
    SILE.call("par")
  end)
```

It's true that `\book:sectioning` in that class still has the existence check on a command `\book:chapter:post:xxx` (where xxx is the language), but only the Japanese language file has it. It seems to me we should deprecate it there.

The _only_ difference the "ja" (and now the "eo") have is a medskip instead of a par. It doesn't seem logical -- spacing between a "book" chapter header and its title is something that should only depend on the so-called "book" class, no? That's the only thing I am saying: skips are a class-decision, not a translation / language support issue.

alerque: The `\medskip` in the *j* classes/languages is leftover from a very early stage of SILE development.

At this point language support files shouldn't not be registering commands of any sort unless absolutely necessary, and the only thing that needs changing here is the `book-chapter-post` key in the Fluent sources. It looks like that may be incoming after this commit anyway...

ctrlcctrlv:
This PR is blocking again. This time on #1447.

alerque:
My `pcall()` solution may not have worked. Oops. I'll fix my mistake.

alerque (approved):

Omikhleia:
Mi ja esperas, ke la kandidato havas seriozajn kialojn por proponi ĉi tiun funkcion...

ctrlcctrlv:
@Omikhleia: Ne.



Kotizoj al malfermit´fontaĵoj vere estas (almenaŭ por mi) ega malŝparfonto. Sed eble tio ĉi estas por la (ankaŭ ega) malŝpareco de la projektoj miaj.

Good luck translating that with Google, _kara_.

alerque:
I don't really care whether the intended use is serious or comical, adding Esperanto support uncovered several bugs that affected everybody and forced us to clean up several aspects of the language interface. I say thanks for the contribution.

Also I like the throwback to 19<sup>th</sup> century with *ALL THE STUFF* on the title page.

Omikhleia:
Such things would actually be real cool in a "showcase" (different from sample examples that cover a feature, just for the show and the hype)

alerque:
Yes. See https://github.com/sile-typesetter/sile-typesetter.github.io/issues/2 and https://github.com/sile-typesetter/sile-typesetter.github.io/issues/31, but even without that reorganization the top level "examples" category is already open to "show and tell" stile documents that do any combination of cool things as long as they do them in a reasonably idiomatic and robust way.

The new package manager coming in v0.14.0 should make it relatively trivial to build website showcase documents with 3<sup>rd</sup> party package dependencies, something we haven't been able to do yet. The website only has things that work in bone-stock SILE. I'll be happy to see it showcase things that require classes / packages not included in core.

ctrlcctrlv:
> Also I like the throwback to 19th century with ALL THE STUFF on the title page.

All my SILE title pages are like that lol

image
image

PR #1445, opened 2022-06-18
fix(build): Make sure `i18n/` dir is actually distributed

ctrlcctrlv:
Fixes this error:

```
! Error loading localizations eo: no localizations for this language at why.sil:2:1: in \languagetable: 0x41510d90
[1] [2]
! No localized message for book-chapter-title-pre found in locale eo at why.sil: in \fluent near 9:1: in \book:sectioningtable: 0x413e6178
```

#1444 is blocked on this being merged.


alerque:
Ouch, what a goof.

Goes to show I do too many project builds using `~/projecs/sile-typesetter/sile/sile` instead of my system default `sile`! Also we need to run regression tests against DISTDIR not the git repo directly...

alerque (approved):

Issue #1245, opened 2021-09-21
Entire CJK stack upside down


alerque:
I'm monkey-patching some stuff through to get the manual to build for v0.12.0, but the entire Japanese language support and Tate / Ruby package scene is chaos. I'm adding fallback fonts early everywhere so the units and things work if called from anything other than an already working environment, but the order is all wrong. You can't measure glyphs before you load fonts, which means you can't measure glyphs in class init.

alerque:
Basically see everything in #1215 except the first commit, particularly d76fd7b..ceba809.

ctrlcctrlv:
I totally agree with you. I have a better version of the Ruby package which I will be PR'ing _soon_…it's just a matter of finding a charger, fixing an old laptop etc. Maybe I'll get to it now. But yes, really `tate.lua`/`ruby.lua` is not so good but I find myself quite outnumbered in terms of caring about doing anything as I seem to be the only user. Furthermore, I only am interested in Japanese typesetting (so the J in CJK) making me an unsuitable general expert.

Omikhleia:
I don't know if it fits here in this issue, but there seem to be some code smell in the `tate`/`break-firstfit`/`ja` logic.

- `languages/ja.lua` [marks some nodes](https://github.com/sile-typesetter/sile/blob/2888d6c6cc931dae31b72cd395d8cfef898a1a8f/languages/ja.lua#L163) as "hangable", it seems to be the only occurrence in the sources.
- `packages/break-firstfit.lua` [has a test](https://github.com/sile-typesetter/sile/blob/2888d6c6cc931dae31b72cd395d8cfef898a1a8f/packages/break-firstfit/init.lua#L14) on `isHangable`, also a unique occurrence in the sources
- `packages/tate` [loads and uses the firstfit method](https://github.com/sile-typesetter/sile/blob/2888d6c6cc931dae31b72cd395d8cfef898a1a8f/packages/tate/init.lua#L15)

I am illiterate as far as Japanese is concerned, and have know idea what a "hangable" thing here in this language, but anyhow I can't see how the `isHangable` would ever be valued. It doesn't seem we have some logic for it (as opposed, say, to the `is_box` etc. constructs, which are valued via the node's type at construction).

My apologies if the above assumption is wrong -- It's just something I saw by accident and which left me perplexed.

ctrlcctrlv:
[...]

Issue #1327, opened 2022-02-26
Re-scope settings stack to typesetter, document, or frame as appropriate


alerque:
Per [this comment and following](https://github.com/sile-typesetter/sile/issues/1317#issuecomment-1051411302), the relationship between settings and typesetters is a bit conflated. I think we should refactor it entirely. This issue is just a placeholder, but we have a jumble of:

* `SILE.documentState`, very assorted
* `SILE.typesetter.state` (has stack)
* `frame.state` (includes direction)
* `SILE.settings` (has stack)
* assorted completely independent state vars

For starters, I don't think we should have a single global `SILE.typesetter`, we should be able to have several instances of this easily without using a function shuffle dance every time.

Next I think the things in settings need to be scoped somehow such that the stack could be managed by scope (`document.*` should be scoped to documentState, most of the settings defined in `typesetter.lua` should be scoped to the typesetter instance state, etc.

alerque:
c.f. #1367 and #1368. All three of these issues have a lot of overlap. Also with

PR #1447, opened 2022-06-18
fix(core): Move the zenkaku width (全角幅; `zw`) unit into core

ctrlcctrlv:
Where it belongs.

This touches on #1245.

This is blocking #1444, somehow (the manual is extremely brittle as all know well):

```
[font-fallback] Glyph あ not found in Gentium Book Basic
! Some glyph(s) not available in any fallback font, run with '-d font-fallback' for more detail at documentation/c08-language.sil:244:1: in \subsection after in \book:subsectionfont

! Unable to measure character at:
	documentation/c08-language.sil:244:1: in \subsection after in \book:subsectionfont
	documentation/c08-language.sil: in <file> (\)
	documentation/c07-settings.sil:51:1: in \includetable: 0x40ddce50
	! debug.traceback() did not identify code location
make: *** [Makefile:1451: documentation/sile.pdf] Error 1
make: *** Deleting file 'documentation/sile.pdf'
```

After much ado I discovered that the problem is that the unit `zw` (which stands for the width of a zenkaku—i.e. 全角幅) was popping in and out of existence based on the language setting of the document. Of course this won't work.

All units must be permanent. The concept of a zenkaku is not even Japanese-only.

### DIFF
--- a/core/base-shaper.lua
+++ b/core/base-shaper.lua
@@ -59,21 +59,14 @@ SILE.shapers.base = pl.class({
       return shapespace(width and width.length or items[1].width)
     end,
 
-    measureChar = function (self, char, fallible)
+    measureChar = function (self, char)
       local options = SILE.font.loadDefaults({})
       options.tracking = SILE.settings:get("shaper.tracking")
       local items = self:shapeToken(char, options)
-      local error_func
-      if fallible == nil or fallible == false then -- default
-        error_func = SU.error
-      else
-        error_func = SU.warn
-      end
       if #items > 0 then
         return { height = items[1].height, width = items[1].width }
       else
-        error_func("Unable to measure character", char)
-        return nil
+        SU.error("Unable to measure character", char)
       end
     end,
 
--- a/core/base-shaper.lua
+++ b/core/base-shaper.lua
@@ -59,14 +59,21 @@ SILE.shapers.base = pl.class({
       return shapespace(width and width.length or items[1].width)
     end,
 
-    measureChar = function (self, char)
+    measureChar = function (self, char, fallible)
       local options = SILE.font.loadDefaults({})
       options.tracking = SILE.settings:get("shaper.tracking")
       local items = self:shapeToken(char, options)
+      local error_func
+      if fallible == nil or fallible == false then -- default
+        error_func = SU.error
+      else
+        error_func = SU.warn
+      end
       if #items > 0 then
         return { height = items[1].height, width = items[1].width }
       else
-        SU.error("Unable to measure character", char)
+        error_func("Unable to measure character", char)
+        return nil
       end
     end,
 
--- a/core/settings.lua
+++ b/core/settings.lua
@@ -68,6 +68,13 @@ function settings:_init()
     help = "Skip to be added to left side of line"
   })
 
+  self:declare({
+    parameter = "document.zenkakuchar",
+    default = "あ",
+    type = "string",
+    help = "The character measured to determine the length of a zenkaku width (全角幅)"
+  })
+
   SILE.registerCommand("set", function(options, content)
     local parameter = SU.required(options, "parameter", "\\set command")
     local makedefault = SU.boolean(options.makedefault, false)
--- a/core/units.lua
+++ b/core/units.lua
@@ -201,4 +201,24 @@ units["en"] = {
   definition = "0.5em"
 }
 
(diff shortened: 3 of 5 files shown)


core/settings.lua
```diff
@@ -68,6 +68,13 @@ function settings:_init()
     help = "Skip to be added to left side of line"
   })
 
+  self:declare({
+    parameter = "document.zenkakuchar",
```
alerque, line 72: Setting scopes are an open issue (#1327), but should this be a document level setting or might it be better off scoped to the typesetter or ... something else? I'll probably go ahead with this merge this way since rearranging that mess is going to be a breaking change anyway, but just looking for your idea on what this *should* be.

core/base-shaper.lua
```diff
@@ -59,14 +59,21 @@ SILE.shapers.base = pl.class({
       return shapespace(width and width.length or items[1].width)
     end,
 
-    measureChar = function (self, char)
+    measureChar = function (self, char, fallible)
```
alerque, line 62: This was a good idea and I see why you took it this direction, but I do want to think whether it might not be better to leave this function alone and just call it with a `pcall()` wrapper for the rare cases we want to let it fail. I need more sleep to think through those ramifications though...

alerque (commented):
Thanks for taking a stab at this. My (lack of) expertise on CJK has definitely shown through in the hacks in the manual. Speaking of which there is at least one more than needs removing if this works right...

core/settings.lua
ctrlcctrlv: We should expect the user to swap it mid-document. I guess that means scoped to the typesetter.

I'm a Unicode nerd so I tend to think of the rarer languages of the world, and I can imagine documents where you'd want to flip to Ainu or Taiwanese Kana or Okinawan. All three of those are written in Japanese kana and with traditional Japanese typesetting rules including _zenkaku_. But we shouldn't expect all fonts made _just_ for e.g. Ainu or Okinawan to support even あ, as the Ainu orthography (example: アイヌ・イタㇰ) doesn't include hiragana.

core/base-shaper.lua
Omikhleia: Or scoped to the font?

core/settings.lua
Omikhleia: Or scoped to the font? (perhaps it could even be something extracted from font switching commands, rather than via settings?)

Omikhleia: I know nothing about eastern languages, but isn't U+3000 IDEOGRAPHIC SPACE supposed to provide that width?

ctrlcctrlv:
@alerque I still want to rewrite `ruby.lua`, it's a longterm goal of mine. My private version is pretty cool:







I call these bizarre rubies 重ねルビ (_kasane-rubi_) which means "stacked ruby", or 𝑛重ルビ (𝑛-stacked ruby).

But historically they are not rare.

### [法令全書　明治７年　第６３ページ](https://dl.ndl.go.jp/info:ndljp/pid/787954/63)

image

### 台湾語仮名
image

### 韓国語での曲の翻訳
image

But my private version fails at some very important cases. I think the recent work on math.lua, and patches by @Omikhleia, will help a lot. I wouldn't want to release a version that doesn't meet most needs.

core/settings.lua
ctrlcctrlv: Scoping to the font won't work. The OT feature `ruby` often has a different zenkaku width. It's typesetter scope if anything.

U+3000 would be more workable than あ and is an interesting suggestion of a better default.

alerque: Scoping to the font would only make sense if the unit was something only used to scale the font itself. In this case it's a general purpose unit that get used for all sorts of things relative to much larger scopes including frame properties. Typesetter makes a lot of sense I think. Although we do have a bunch of document level properties that are actually at the same scope that are not immutable class options, so `document._` for now until we sort that out sounds like a fair plan.

core/base-shaper.lua
alerque: I refactored this going back to throwing an error but using `pcall()` to wrap the one case where we want to allow that and output the warning there. The overall logic should be the same.

alerque:
Oops. My modifications would have been smart if `SU.error()` was raising an error. It is not. Furthermore it is failing to do what it was supposed to do (finish off a valid PDF file on an aborted run). I'm looking into that...

Issue #1059, opened 2020-09-14
How to contribute with packages/classes to sile?


andresantosjor:
I wrote some small packages in an attempt to complete my first Sile class. However, my code is subpar (at best) and I deemed it unworthy to make a pull request and integrate it to the official code. However, as they are very useful (at least for me), I wanted to share them with our small community. I visited the "Packages" page in the website but couldn't find any listing. Is there some way to link my packaging attempts where someone else could see it?

alerque:
Probably the best thing to do for now is split your packages up into Git repositories and send a PR to the [sile-packages](https://github.com/sile-typesetter/sile-packages) repo linking to them. This will is basically a directory of people's assorted packages that they use but haven't been upstreamed into SILE's core. Some are useful to others, others may not be, but it's a good place just for your own convenience and also to get some exposure / discoverability so others could find them.

No this isn't well documented yet. Nor is the package manager really all that fully developed yet, but it's a start.

rjmunro:
Could you maybe leverage luarocks, rather than develop your own package manager?

Cordova used to have it's own plugins, but then they switched to making plugins installable with npm and it was a huge improvement.

ctrlcctrlv:
The packages as it were are all quite simple. You just copy the Lua file(s) into SILE's share directory. I imagine Luarocks can do this... And anyway, more complex packages might rely on Luarocks packages.

alerque:
@rjmunro I actually suggested that (cf #538) and got vetoed before Simon implemented the current system. I've thought about it a lot more since and have more pros and cons for doing that. Luarocks does actually allow use against a third party package server (so we wouldn't have to populate the actual Luarocks site with stuff not of general interest) and quite handily it does cover installation to system, user, or local trees (a major shortcoming of the current system) as well as for multiple Lua versions. At the same time there are some downsides to using –neh hijacking– a 3rd party package manager and conscripting it to a different purpose.

Perhaps this issue isn't the best one to discuss this on though. Could be move over to №538 or a new one? See also the [Package Manager](https://github.com/sile-typesetter/sile/projects/5) project I started today to organize related issues.

alerque:
c.f. My last comment [on related issue](https://github.com/sile-typesetter/sile/issues/324#issuecomment-1169873737). The TL;DR version is that packages and classes can now be easily be published via LuaRocks, including on luarocks.org or any other package registry.

Issue #690, opened 2019-10-17
Package manager code needs help


alerque:
Working through linting errors, `core/packagemanager.lua` has some pretty egreious stuff.

- [ ] First we should probably stash our path in the SILE object, not as a global but that's minor.

- [ ] Second the check for Lua 5.2 will not work because some environments have a poly-fill that erases that apparent difference. You must check the actual version. I'd fix it but I can't even tell which way the conditional is supposed to match.

- [x] Third there are typos in variable so I'm not sure the paths ever worked. (fixing in #681)

- [x] 3 space indent? Are you kidding me? (fixing in #681)

- [x] Some things aren't even functions (e.g. `execute()` should be `os.execute()` ... it doesn't seem like this was very well tested.

- [ ] Commands we want on the REPL probably need some specific handling besides just jamming them into the global namespace. Some sort of register good like we have for document commands perhaps.

Issue #324, opened 2016-06-07
Package manager


suhr:
To install SILE packages somewhere into `$HOME/.sile/` (or `$HOME/.local/share/sile`). Also SILE should search for packages in that location.

PS: there's some useful [article](https://medium.com/@sdboyer/so-you-want-to-write-a-package-manager-4ae9c17d9527).

alerque:
I agree sooner or later this is going to be necessary. But I also know implementing it is going to be painful, no matter how right we try to do it. Surely we can do better than TeX's CTAN or Lua's rocks, but it might take a couple iterations before we get something flexible and really useful.

Obviously  the biggest hassle is going to be versioning and the interaction between SILE versions, package versions, and documents. I've been putting together a library of things I use for various books and still can't get my head around what bits need to be versioned alongside the document and what bits can be independent. At the moment every upgrade to SILE means a good deal of manually checking, comparing, and usually a few tweaks to by code to interact nicely with upstream changes. Doing this for my own personal collection is enough of a mind bender, sorting it out to work for others is not going to be a  walk in the park.

simoncozens:
You know what, right now, I'm thinking that all user-contributed packages and classes should just go in the SILE distribution. Avoids all the package manager madness, and we're talking about a bunch of tiny Lua files. Right now the packages and classes directories take up 300 kilobytes. If that grew by a factor of 100, I still wouldn't be worried.

alerque:
I wouldn't be worried about the size either—and I even agree that right now the best solution for contributed packages is to just accept them as part of the upstream packages repo and version them along with SILE.

However that isn't the only problem here. The two issues I see that call for a separate package manager (eventually) are:

* Private packages that won't be contributed back upstream. I'm in this scenario already. I have a toolbox for our publishing house that has classes, functions and all sorts of gizmos that are specific to our projects. I'm keeping an eye on it and when I do things that would be useful to more than just us I try to refactor it as a package and contribute it to SILE, but not everything in there makes sense to push out to other people.

   Not having a package management system that SILE knows about leaves me a build system reminiscent of MacGyver's handiwork. It's serving me so far, but versioning is already an issue. Once a book is released it may or may not build the same with the next version of the toolkit, so I have it set as a versioned git submodule in each book or series repository and a Makefile that runs sile with the right paths set.

   This setup would hardly be intuitive to some green horn that comes along and wants to get started. An established package management system would give other publishing houses a path to follow rather than having to re-invent the wheel.

[...]

PR #1450, opened 2022-06-21
Enable Luarocks to install SILE packages

### LINKED ISSUE #538: Consider using luarocks as a package manager for SILE extensions
Today I noticed [some discussion](https://github.com/awesomeWM/awesome/issues/1171#issuecomment-254629024) in regard to the [awesome window manager](http://awesome.naquadah.org/) that is also build in and can be extended in Lua. I know the idea of contributed packages has come up before (see #60) and that for now most stuff is making it's way into core, but I've also noticed that this isn't going to work indefinitely. Even as a SILE developer and managing my own release branch so that I have just the patches I want running on my CI servers that build SILE projects, the pace of developing helper libraries that get used across different projects is a different page that what is comfortable in core. Likewise having the same code duplicated on a per-project basis is undesirable.

I think we should consider normalizing a file format for packaging extended features not shipped in the SILE core but possibly useful as building blocks in other projects and also way of interfacing with them that allows a document to specify dependencies, not just as in a local file path but to request that SILE have access to a luarock.

### LINKED ISSUE #843: Package manager borked
Inserting the line: `print("In("..lfs.currentdir()..") git pull")` appropriately in core/packagemanager.lua and then run with `sile -e 'installPackage("fontproof");os.exit()'` gives the output: `In(/usr/local/share/sile/packagemanager/fontproof) git pull`.

Please fix this so that we can actually use this sile. I have no idea what you think you want it to do, so I have to leave this to you.

alerque:
Closes #538
Closes #843
Obsoletes #690
Answers #1059, will close with specific comment later.
Answers #324, although we can consider the automatic searching aspect separately.

This fixes several existing issues with Lua path handling, then adds a feature to SILE's paths that look in directories prefixed with `/path/sile/?` instead of just `/path/?`. This should enable `luarocks` to install and manage packages that are loaded by SILE (and only SILE) without conflicting with the LuaRock root namespace.

Compliant packages should need no change to their current `classes/name.lua` and `packages/name.lua` paths, but should include a rockspec file that installs all of these under the global sile root path.

Installing packages should be possible without root access using either user (`--local`) or project (`--tree=./`) scoped installations, then added that to the user shell path with `eval $(luarocks path)` with the same scope args as used for installation.

```console
$ luarocks --local install <package>
$ eval $(luarocks --local path)
$ sile file.sil
```

Todo:

- [x] Write or convert an existing external plugin to use this as a POC.
- [x] Document usage in manual
- [x] Add deprecation warnings to old package manager


alerque:
After 4 years (to the day) stressing about this it turned out to be about 10 lines of code and worked on the first try. I'm not sure weather to release this in a patch release or a minor bump. The old system will keep working for at least one cycle either way...introducing it is not a breaking change, only finally deprecating it will be. On the other hand it's a bit weird for plugins to be compatible with only an patch release.

ctrlcctrlv:
@alerque You could use #1456 as an excuse to make a major breaking release. Do this:

* Merge #1456 as 0.13.2 (which activates the "include \language[main=ja] dummy" warning)
* Merge this as 0.14.0 (which makes that warning into an error)

alerque:
My new plan is to do something like that with a little bit different rational. The tate/jplain refactoring looks like something we can shim for a long time and deprecate slowly (two major releases not one). As such I'm going to target it for 0.14.0 as a breaking change (already rebased that commit in the other PR to mark it as such)  and also this luarocks stuff will be in 0.14.0. I'l ldo a quick 0.13.2 release just before that as a fallback in case people have issues with all the things that will stop working in 0.14.0 (all the legacy class and stdlib stuff).

ctrlcctrlv:
Works for me :o)

Issue #1452, opened 2022-06-22
[tate] [CJK] Unusable in master because よっ runs together


ctrlcctrlv:
![image](https://user-images.githubusercontent.com/838783/175027066-07d11ae4-cf7b-4165-a059-712a7e90e710.png)

alerque:
Can you include some MWE code of the shortest document string you see failing? I'm chasing gremlins...

ctrlcctrlv:
Sorry for late reply, I finished what I wanted this for in upLaTeX so was distracted :blush: 

![image](https://user-images.githubusercontent.com/838783/175122291-e1cc6b75-89d5-420b-b740-a0dd85c70608.png)
![image](https://user-images.githubusercontent.com/838783/175122311-be8c598f-4785-4968-9da0-0a5819ed8de6.png)
![image](https://user-images.githubusercontent.com/838783/175122339-76babd0c-16ce-4775-9c36-8eb0c0c5aacd.png)

It'd be great if I could do something like that in SILE but SILE is still has way too glitchy CJK stack.

ctrlcctrlv:
Wait, no, I was testing with the non-master #1447.

I just pulled your master and got:

![image](https://user-images.githubusercontent.com/838783/175123107-bcfff6d2-a5a3-4b5e-9a5f-d1b42590c2b6.png)

Now I'll try w/#1451.

ctrlcctrlv:
Sorry @alerque #1451 makes no difference to this bug.

![image](https://user-images.githubusercontent.com/838783/175123477-0f7dda01-9789-47a6-bd47-df81b820af28.png)

alerque:
Pardon my ignorance here, but what is this MWE *supposed* to look like? Is that character in it's own column supposed to be hanging out inline with the others?

I just spun up some old versions via Docker and I don't see anything different in this document rendered in v0.11.1, v0.12.0, v0.12.5, or v0.13.1.

ctrlcctrlv:
If that's the case it's always been broken. Tate has many issues so I haven't used it in my projects extensively.

![image](https://user-images.githubusercontent.com/838783/175130475-550123d6-f479-4d81-b3a7-0fe65a4dbc43.png)

Omikhleia:
FWIW, @ctrlcctrlv 's MWE

> ```
> \begin[papersize=a4,class=jplain,layout=tate]{document}
> \set[parameter=font.family, value=Noto Serif CJK JP]
[...]

Issue #1171, opened 2021-05-30
Multi-layer color fonts (so, almost all of them…) broken in TTB mode


ctrlcctrlv:
Don't bother patching/looking into, this is a tracking issue, I have a very simple PR that took more than an hour to come up with!! :laughing: 

# Expected display
![image](https://user-images.githubusercontent.com/838783/120113885-9109a080-c14a-11eb-886e-db1119d51d55.png)

# Actual display
![image](https://user-images.githubusercontent.com/838783/120113863-76cfc280-c14a-11eb-8158-d9d4559ffad6.png)

# Code
```latex
\begin[class=jplain,layout=tate]{document}
\set[parameter=font.family,value=Noto Serif CJK JP,makedefault=true]
\define[command=emoji]{\font[family=Twemoji Mozilla]{\process}}
いいですよ。\emoji{😉} \emoji{🤣}
\end{document}
```

PR #1456, opened 2022-06-24
fix(classes): Clarify the scopes of `tate` and `jplain`

ctrlcctrlv:
As I've been attempting to enforce through my commits, counter to how
@simoncozens did things:

* `tate` and `ruby`: General purpose libraries that *should not be loading
  any fonts* or *assuming a Japanese document*.
* `jplain`: The Japanese language class, *may* enforce
  `\language[main=ja]` but I didn't for backwards compatibility.

#1452 shows the above backwards compatibility preservation to be ill
advised. Therefore this commit breaks it, closing https://github.com/sile-typesetter/sile/issues/1452.

---

# Note re tests
Also adds `\language[main=ja]` to all jplain tests.

Also fixes a test that was returning bad (tests/bug-1171.sil which tests
for regressions to the issue I reported in #1171) because its behavior
is different in the current Noto font w/ja language. I confirm the new
behavior is not worse, just slightly different.

### DIFF
--- a/classes/jplain.lua
+++ b/classes/jplain.lua
@@ -17,13 +17,21 @@ function jplain:_j_common ()
   self:loadPackage("font-fallback")
   self:loadPackage("hanmenkyoshi")
   self:registerPostinit(function (class)
-      class:bidiDisableTypesetter(SILE.typesetter)
-      class:bidiDisableTypesetter(SILE.defaultTypesetter)
-      SILE.call("font:add-fallback", { family = "Noto Sans CJK JP" })
-    end)
+    class:bidiDisableTypesetter(SILE.typesetter)
+    class:bidiDisableTypesetter(SILE.defaultTypesetter)
+  end)
   self.defaultFrameset.content.tate = self.options.layout == "tate"
   self:declareHanmenFrame("content", self.defaultFrameset.content)
   SILE.settings:set("document.parindent", SILE.nodefactory.glue("10pt"))
+  if SILE.settings:get("document.language") ~= "ja" then
+    SU.deprecated("document.language ≠ \"ja\" & jplain:…", nil, "0.13.2", "0.14.0", "Prior to SILE v0.13.2, `jplain`, despite its name, did not enforce the use of Japanese in its documents. To use a class like `jplain` for other languages, please base a  *new* custom class off of `jplain`; do not use `jplain` itself. To use `jplain` for Japanese, you *must* specify \\language[main=ja]. This will become a hard error in SILE v0.14.")
+    --[[ @alerque — Remove these lines post v0.14.0! ]]
+    self:registerPostinit(function (_)
+      SILE.call("font:add-fallback", { family = "Noto Sans CJK JP" })
+    end)
+    SILE.settings:set("document.language", "ja")
+    SILE.languageSupport.loadLanguage("ja")
+  end
 end
 
 function jplain:_init (options)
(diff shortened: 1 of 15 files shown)


ctrlcctrlv:
@alerque I'm sorry I have no clue wtf even is openresty so idk how to repro the test fail

classes/jplain.lua
```diff
@@ -17,13 +17,21 @@ function jplain:_j_common ()
...
-      class:bidiDisableTypesetter(SILE.defaultTypesetter)
-      SILE.call("font:add-fallback", { family = "Noto Sans CJK JP" })
-    end)
+    class:bidiDisableTypesetter(SILE.typesetter)
+    class:bidiDisableTypesetter(SILE.defaultTypesetter)
+  end)
   self.defaultFrameset.content.tate = self.options.layout == "tate"
   self:declareHanmenFrame("content", self.defaultFrameset.content)
   SILE.settings:set("document.parindent", SILE.nodefactory.glue("10pt"))
+  if SILE.settings:get("document.language") ~= "ja" then
```
alerque, line 26: Nice work with the deprecation warning for documents using the old style. I wonder though if we might be able to clarify the scopes here even further. I would be among the many to assume `jplain` probably defaulted to Japaneese. What if we renamed this class to `cjkplain` or something like that (you might know better what an appropriate umbrella name would be) and then make a new `jplain` class that inherits from it and just adds the default to Japanese language & fonts?

classes/jplain.lua
```diff
@@ -17,13 +17,25 @@ function jplain:_j_common ()
...
+  Prior to SILE v0.13.2, `jplain`, despite its name, enforced the
+  language being set te Japanese. It no longer makes this assumption.
+  To use a class like `jplain` for other languages, please base a
+  *new* custom class inheriting from `jplain` rather than `jplain`
+  directly. To use `jplain` for Japanese, you *must* specify
+  \\language[main=ja]. (Also you'll probably want to set a font!)
+    ]])
+    SILE.languageSupport.loadLanguage("ja")
+    SILE.settings:set("document.language", "ja", true)
+    SILE.settings:set("font.family", "Noto Sans CJK JP", true)
```
alerque, line 37: I refactored this just a bit to set the default font instead of the fallback font to support Japaneese. I realize this is just a shim about to go away, but this seems like a better example for somebody that wanted to know how to do this in their own class later. Does this make sense or am I missing something and leaving Gentium as the default font while the language is set to `ja` actually makes sense?

ctrlcctrlv: You're not missing anything.

classes/jplain.lua
ctrlcctrlv: That's a good idea. I suggest `tplain`. This has precedent from pLaTeX, the t is for `tate`.

That way in the far future this can be used for Mongolian too!!

ctrlcctrlv: PS I'm aware that this layout isn't always used to make TTB docs of course. I am just suggesting the name to mean "`tate`-_capable_ plain document".

I'd like to see this file go away as much as anyone, I think plain old `plain` should support `tate`, but that'll require getting on board with my thoughts re what to do about BiDi and `latin-in-tate`  :) so this class'll be around a while.

alerque: I'm not knowledgeable enough to weigh in on BiDi in TTB, so that decision might get put off a bit. Honestly I do think loading the tate package being viable in the default class makes sense too, but in the mean time lets shuffle the cheese and go with `tplain` that makes no language assumptions and `jplain` that does. That will also make this whole PR a *non* breaking change, right?

alerque:
With the addition of `t-` series classes the `j-` series is now back to doing effectively what they were before, just doing it more correctly. Since this is now *not* a breaking change I've rewritten that commit message and am re-targeting this for the next patch release since.

classes/jplain.lua
alerque: As far as I can see most of this is resolved and for the bit that isn't we have other issues tracking them.

PR #1457, opened 2022-06-24
Make papersize parser more tollerant

### LINKED ISSUE #1433: Document papersize parameter case sensitivity
Sometimes it is natural to write the papersize for ISO 216, ISO 217 and ISO 269 in caps (A, B and C series), however the _papersize_ parameter for a document is lowercase (defined in _core/papersize.lua_). See https://github.com/sile-typesetter/sile/blob/5e95ff4fb93ae675c0812cb6286957e89c4b9e51/core/papersize.lua#L13 and 
https://github.com/sile-typesetter/sile/blob/5e95ff4fb93ae675c0812cb6286957e89c4b9e51/core/papersize.lua#L59

Given this document input
```
\begin[papersize=A0]{document}
\end{document}
```
Gives the output
```
$ sile test.sil 
SILE v0.12.5 (Lua 5.2)
<test.sil>

! Unknown paper size A0 at test.sil: in <file> (\)

```

alerque:
Closes #1433


core/papersize.lua
```diff
@@ -58,13 +58,14 @@ setmetatable(papersize, {
     __call = function (self, size)
       local _, _, x, y = string.find(size, "(.+)%s+x%s+(.+)")
```
ctrlcctrlv, line 59: Why %s+ and not %s* ?

alerque: Because some paper size presets have an `x` in them such as "executive" and "halfexecutive". Unless we switch to a PEG based parser that detects valid iterations of our units format we need something more definitive than just an X to split this field.

ctrlcctrlv: It can wait for sure. I only asked as I've tried `5cmx5cm` before.

Issue #472, opened 2017-06-10
Page frame doesn't update when setting paper size


alerque:
For various reasons, I don't always know the papersize something will be rendered at when instantiating the document class. When adding crop marks or changing papersizes due to programmatic input, the PDF renders fine but the `page` frame doesn't update to the new dimensions. This has the site effect that things such as the `background` package sets the background on the original papersize, not the current one.

See also #430.

Issue #430, opened 2017-01-02
Support changing papersize per page in PDF output


alerque:
While not common, it is valid to have a PDF with different page sizes. In some rare cases it's even useful and (in my case) required for an a automated print job. I can make both parts in SILE and mash them together in post processing, but it would make a lot more sense if SILE could just do this in the first place.

For starters it looks like some of the code from the crop package needs to be make re-usable so that it can be called on any papersize change. Beyond that I'm a little fuzzy on what would have to happen, but something needs to keep track of which pages are what size and pass that to the outputter.

PR #1470, opened 2022-07-03
Full bleed printing and page size options

### LINKED ISSUE #1285: cropmarks package is broken
Document:
```
\begin[class=book, papersize=6in x 9in]{document}
\script[src=packages/cropmarks]
%\define[command=crop:header]{crops}
\crop:setup[papersize=a4]

Test page 1

\eject

Test page 2

\end{document}
```

Result = error

```
SILE v0.12.0 (Lua 5.2)
<examples/croptest.sil>
[1] 
Error detected:
	/usr/share/sile/packages/date.lua:7: attempt to index local 'options' (a string value)
```

Removing the comment to override the default crop header = page 1 is wrong.

image

(Page 2 is ok)

### LINKED ISSUE #1439: A thought on full bleed printing, trim areas and cropmarks
I was about to add this comment to #1285, but it's a more general topic, which (beyond mere cropmarks) requires that we think about a more general solution to the underlying question, though these things are somehow related anyway (IMHO)...
  1. Sometimes, as show in the example in #1285, one wants to produce, say, a 6x9in book but only has a regular printer, with A4 paper: the expectation is to have, usually, the "intended" page to be (more or less) centered. That's the "poor man" solution (I want to see print my 6x9 in book as it would appear in print, but centered on the only paper available to me... Or: I am going to distribute an A4 version to proofreaders, so they have the book as intended, but lots of margins for annotations, etc.)
  2. But more seriously, professional printers nowadays rather expect an extra "trimming" margin around the target page -- for what is usually called "full bleed" printing. E.g. the extra space around the target page could be 3mm (0.125in), and the assumption then is that "full page" images or background color would extend past the target page size so as to cover the bleed area -- when the pages are "cut", one wouldn't want blank white space to occur if for some technical reason the cut is not perfect...
  3. Cropmarks are a bit orthogonal to the above points, but somehow related: Some printers want them (honestly, AFAIK, it's seldom used nowadays in the PDF printing era...), as visible "cut" indications; but others would reject them and just ask for proper bleed areas, e.g. in my example, directly ask for a 6.125x9.125in PDF as their input, without any visible cropmarks...

All of the above, as the "papersize", is something one usually decides globally for a whole document, not when loading (on-demand, at any point in a document, a package...)

Summarizing:
- Document level options (i.e. as the papersize) rather than when some package is loaded at some point... (We wouldn't want to change that, usually, in the course of a document...).
- Actual paper size vs. target page size
- Bleed areas
- Optional cropmarks, for those who want such things
- Ensuring full page content actually extends over the bleed area (= e.g. `background` package, I'm looking at you...)

image

Kind of a proposal for discussion...

- `begin[pagesize=6in x 9in, bleed=0.125in}{document}` = leads to a 6.125x9.125in PDF, with bleed distributed evenly.
- `begin[pagesize=6in x 9in, bleed=0.125in, papersize=a4}{document}` = leads to an A4 PDF, with a 6x9 centered page, and bleed distributed evenly, applying to background color/images. Obviously, error if pagesize + bleed > papersize, and bleed defaults to 0 if unspecified.
- `begin[..., cropmarks=true]{document}` = show crop marks around the pagesize, for any of the above cases.

The current way (with `begin[papersize=...]{document}` would merely mean the page size is the same as the paper size, with no bleed area.

And the `%pw` units, or page-related frame dimensions, of course apply to the pagesize.

Whether one uses print-on-demand services (such as Lulu, Blurb, KDP, etc.) or professional printers, it seems to me that "full bleed" printing with bleed areas is the "standard" today, without the (old-fashioned?) cropmarks. Whether the above proposal is sound or not, we should likely address it and make it easier for users.

Omikhleia:
This is my proposed take on #1439. It is a (slightly) breaking change with deprecation warnings set to **0.14.0** (to adjust if/when needed).

**What it does**

- Adds global class options `pagesize` and `bleed` (options documented in the manual + see illustration below)
- Adds support for background images (**background** package)
- Ensures background images extends to the trim/bleed area (**background** package)
- Fixes crop marks (with changes) and ensures they also respect the trim/bleed area (**cropmarks** package)

One can now do such things:

image

(A6 paper, A7 page, 0.25in bleed, crop marks. First page is using a background color, subsequent ones an image background, except the last which reverts to no background. The sample is run with `sile -d frames` to show the actual frame extents.)

Closes #1285, #1439 

**What it changes and may break**

- For this to work well, all frames need to be related to the `page` (e.g. `left = left(page) + 8.3%pw` instead of the previous `left =  8.3%pw`). If this is not done, things will work as usual if the paper size and page size are the same and no trim/bleed is defined, i.e. it has normally no adverse effect on existing documents (except for those using crop marks, see below). Personally, I do find it a bit more clear that the previous absolute values that implied a target page starting at (0,0)... N.B. The old crop marks package would have re-constrained the frames relatively to the page, but note that this was still anyhow kind of broken (It would have re-constrained the know frames at that point, but frames declared latter would obviously not be taken into account, so would have been to be page-relative to work well... This is kind of just generalizing that fact now.
- 3rd party packages that would try to hack `SILE.documentState.pageSize` will likely break (but this is unlikely to exist: as noted long ago e.g. in #430 or #472, we are not really here yet...)
- The **cropmarks** package no longer redimensions the page, it just does what its name implied (= drawing crop marks).
- The **background** package no longer loads the **color** package: it never needed it as a dependency. Separation of concerns is the reason for the dependency removal, but documents that would have expected `\color` to be defined will have to load it explicitly one way or another.

Closes #1439


Omikhleia:
Update: Rebased on current master at this date (7925d45257d2fb228d1c569eabeb5f7ade35d9ef) and upgraded to 0.14 API.

packages/twoside/init.lua
```diff
@@ -15,10 +15,10 @@ local mirrorMaster = function (_, existing, new)
   for name, frame in pairs(SILE.scratch.masters[existing].frames) do
     local newframe = pl.tablex.deepcopy(frame)
     if frame:isAbsoluteConstraint("right") then
-      newframe.constraints.left = "100%pw-("..frame.constraints.right..")"
+      newframe.constraints.left = "2 * left(page) + 100%pw-("..frame.constraints.right..")"
```
alerque, line 18: Good catch here. I've been holding off merging this because it is a breaking change that I haven't had time to regression test and figure out how to transition people onto yet, but this fix should go in a patch release even without a major version bump. I'm going to stuff it in #1562.

packages/cropmarks/init.lua
```diff
@@ -7,107 +7,91 @@ local outcounter = 1
...
+  SILE.typesetter.frame.state.cursorX = page:left() + offset
+  SILE.typesetter.frame.state.cursorY = page:top() - offset - 3
   outcounter = outcounter + 1
 
   if hbox then
     for i = 1, #(hbox.value) do hbox.value[i]:outputYourself(SILE.typesetter, { ratio = 1 }) end
   end
 end
 
-local function reconstrainFrameset (fs)
```
alerque, line 38: Why remove this? I realize you stopped using it, but couldn't we repurpose it to shim existing frameset configurations that are not setup relative to the page frame? This would ease the transition to the split paper/page setup if people didn't have to at the same time redefine all their framesets.

Omikhleia: Perhaps indeed. I just removed it because it was a local function, not exposed by the package, and now no longer used internally, but perhaps it could be moved e.g. to utilities (`SU.reconstrainFramesetToPage()` ?)

packages/twoside/init.lua
Omikhleia: >  it is a breaking change that I haven't had time to regression test and figure out how to transition people onto yet

It's indeed slightly breaking (people who did neither use cropmarks nor redefined the papersize "in the wild" wouldn't even notice it - That's possible the general case). There's no urge on this PR, as long as we know it exists and might be a way to address full-bleed printing.

Actually I was kind of tempted to split the `\background` fixes and improvements (dependency on color, ability to use an image) and the crop header fix (date issue) separately, would we want them earlier (true, it wouldn't account for page bleeds in print, but that's already the case for a colored background). Not sure, but this could have an interest for people just producing a PDF and not requiring page bleeds. It's a bit of extra work, but if you'd consider the approach more reasonable, I could do such a split (and thenlater rebase this PR over it).

alerque: I had to pull this fix from #1562 because it breaks the default book class. It only works correctly in the context of all frames being relative as you've done in this PR ... which makes it a breaking change that shouldn't go in a patch release.

I'd wait to split this up, I'm going to take a stab at it along with a compatibility shim, but I'm going to have to ship my next book tomorrow with just a hacked up mirror function.

packages/cropmarks/init.lua
alerque: I've been thinking about just automatically constraining all frames to the page *on instantiation* rather than "0,0" as the assumed default. Would that break anything? Then define "absolute" vs. "relative" as anything linked to the page/paper/trim/bleed frames rather than just page.

packages/twoside/init.lua
Omikhleia: I have more doubts than ever, having dabbled into dark horizons: the typesetter, pagebuilder, frame and insertion mechanisms are very entangled and touchy, with plenty of assumptions.

For instance, the X,Y cursors in a frame are not initialized when the frames are (re-)solved. Playing with the (undocumented) `steal-position` insertion for top-growing insertions (for #1559) causes much havoc :)

In other terms, I start thinking that this PR comes too early, in a way, and might be premature.

As noted, the `\background` fixes and improvements (dependency on color, ability to use an image) and the crop header fix (date issue) could still make sense outside of this picture.

alerque:
Would you mind unlocking this PR so I can push a merge from master with conflict resolved? Otherwise I'll have to work on handling this in a new PR or offline.

Omikhleia:
@alerque 
> Would you mind unlocking this PR so I can push a merge from master with conflict resolved? Otherwise I'll have to work on handling this in a new PR or offline.

Sorry, I failed to notice that message earlier. It should be unlocked now.

alerque:
This is *probably* ready for prime time, but I don't have the bandwidth to review it again cross referencing my own full bleed implementation, and I'm in a hurry to get v0.14.9 out for various other reasons so I'm bumping this.

packages/twoside/init.lua
Omikhleia: Nah. Some ideas here might be reusable, but definitively, a different approach is needed.
Possibly the whole bleed question is an outputter thing only...

Omikhleia:
Removing the milestone - I came to the conclusion that there's a simpler path for this. Back to experimenting (some ideas will stay, but we can avoid the frame impacts)

Omikhleia:
Closing (rejecting) the PR in favor of the better approach proposed in #1853

Issue #1336, opened 2022-03-05
markdown class is broken


Omikhleia:
SILE 0.12.2

_First feeling,_ the **markdown** class is nowhere documented properly, this might be bad sign...

Input document is a random markdown sample taken [here](http://www.unexpected-vortices.com/sw/rippledoc/quick-markdown-example.html).

Preamble is:
```
\begin[papersize=7.44in x 9.68in, class=markdown]{document}
\language[main=en]
\font[family=Libertinus Serif, size=10pt]

\end{document}
```

So going for: `sile  -I markdown-preamble.sil markdown.md` 

I made an early attempt before and got an error. I didn't record it at the time, and decided to give it another fresh try.
```
Error detected:
	./lua-libraries/lunamark/reader/markdown.lua:23: attempt to call field 'lower' (a nil value)
```
That is different that what I remembered, but along the same lines (UTF8 related)... Then I realized I now have on my computer an "utf8" package in some external dependency which I hadn't before... So SILE's bundled lunamark is requiring an utf8 package it doesn't provide...

I know we have "lua-utf8", though... Out of curiosity, I Hacked two occurrence of `require("utf8")` to `require("lua-utf8")` in the lunamark code... And got a document... Yeah... Ah wait, it's a joke. Paragraphs and text contents are _mostly_ there, but without any formatting...

![image](https://user-images.githubusercontent.com/18075640/156878652-4a19f874-73ee-4854-a6a8-a853343fb9cb.png)

What have I done wrong? How is this supported to work.

_Second feeling,_ this is very probably highly broken... Get me wrong. But if this is so broken, why is it here in the classes folder, without any caution? Oh well (take this with a grain of salt and humor), I _already_ know that all default classes are broken, the book class in the first place. Why should I be surprised ? :clown_face:

Omikhleia:
Second thought: if the issues are confirmed (and a mistake on my side), then move this class in something such as `classes/unsupported` or whatever. Don't let it live with supposedly working classes until it's not production-ready and feature-full.

alerque:
I do kind of agree. This and the `docbook` class (#1338) are both basically *Proof of Concept* stage classes and are better suited to be examples than packaged and distributed implementations. At least that's my first gut level reaction.

[...]

PR #1475, opened 2022-07-12
textsubsuper package (real or fake super/subscripts)

### LINKED ISSUE #1258: Detecting font features for fallback?
This is kind of an enhancement suggestion, but also a request for advice on how to best address this idea: could SILE provide a way to detect available font features, so one could either use them or implement a fallback?

I have this case where I'd like to use the superscript characters in a font, when available, instead of raising and scaling a character, because most of the time they'll obviously look better. That could for a number (e.g. in footnote calls), but also for letters (e.g. in century references in French, "IVe" with superscript e, or like with in sequences in English, 12th with superscript th).

Of course, one can check prior-hand if a font has the necessary features and adapt one's own code. E.g. for now, I have a very simple command:

```
SILE.registerCommand("style:superscript", function (options, content)
  if SU.boolean(options.native, true) then
    SILE.call("font", { features="+sups" }, content)
  else
    SILE.call("raise", { height = "0.77ex" }, function ()
      SILE.call("font", { size = "1.414ex" }, content)
    end)
  end
end, "Typeset in superscript.")
```

And I just use `SILE.setCommandDefaults` to set it to native or not, depending on the font I decided to use for a given text. On the other hand, I would find it nicer to be automatically detected.

I guess such a thing could rely on the shaper (harfbuzz's `hb_ot_layout_collect_features` probably) or more likely on the recent opentype parser. Obviously, the initial question may actually have two sides: (1) is a font feature supported? (2) but also: are the characters supported for that feature? (that might be harder).
 
Understandably, there are not many font features, probably, where this could be worth having. Besides `sups`, probably `subs` or `sinf` (for subscripts), and `smcp` (for a scaled fallback for smallcaps). So I understand this "convenient" feature might not be that needed. 

Moreover, some fonts. such as the default (Gentium Plus) do not have the sups/subs feature but do include however some of the [Unicode](https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts) super/subscript characters... So the whole idea of doing it via font feature detection might not be as general as I thought... And perhaps the whole point is moot?

Omikhleia:
Closes #1258 (or rather, addresses the core need expressed in it).
Rationale:
- It uses @simoncozens 's suggestion for "real" super/subscripts,
- It uses clever inputfilter logic and a bit of PDF re-scaling for "fake" ones - perhaps another nice example for class/package designers of things that may be done with SILE.
- There's a bit of Bringhurst's interesting comments and ideas taken into account there (see package documentation).

Short story: In the course of making my book with SILE, I wanted nice footnote calls etc. I used a lighter version of this proposal, based on the code mentioned in the original issue. Libertinus' real superscripts are somewhat decent (esp. compared to those in other fonts...), time was running short, so I ended up using them. They were still looking better (IMHO) than the naively scaled and raised version from the footnotes package... But still, I wasn't fully satisfied and I made several other attempts (including trying the OS2 table parameters, which proved to be wrong on almost all fonts I checked...). I was in kind of a rush however and, at some point, I thought I had lost those attempts... Until I found them just lying in a some lost git stash. Doh. A bit of clean-up and here we are. (So it's a greatly enhanced version of my own current package by the same name in my repo).

Should this go to SILE, though? Oh, after all, why not... I am proposing it, anyway...

The real question is:

image

Which line is the best?

And I would have to be fully honest, so...

image

(The Brill font has pretty good real superscripts... It's hard to compete here :baby_chick: )

Would you accept it,
- I have not updated the footnote package to use it... It breaks several tests involving footnotes, so the effort has to be worth it. **N.B.** This could be addressed separately later, anyway.
- Likewise, inclusion in the SILE manual would have to be addressed.

### DIFF
--- a/packages/textsubsuper/init.lua
+++ b/packages/textsubsuper/init.lua
@@ -1,0 +1,366 @@
+--
+-- Text superscript and subscript package for SILE
+-- using OpenType features when available.
+-- 2021-2022, Didier Willis
+-- Derived from an idea sketched by Simon Cozens
+-- in https://github.com/sile-typesetter/sile/issues/1258
+-- License: MIT
+--
+local inputfilter = require("packages/inputfilter").exports
+
+local textFeatCache = {}
+
+local _key = function (options, text)
+  -- We don't use the size in the cache key, as we don't expect it to
+  -- change whether features are supported or not...
+  return table.concat({
+      text,
+      options.family,
+      ("%d"):format(options.weight or 0),
+      options.style,
+      options.variant,
+      options.features,
+      options.filename,
+    }, ";")
+end
+
+local function textFeatCaching (options, text, status)
+  local key = _key(options, text)
+  if textFeatCache[key] == nil then
+    textFeatCache[key] = status
+  end
+  return status
+end
+
+local function checkFontFeatures (features, content)
+  local text = SU.contentToString(content)
+  if tonumber(text) ~= nil then
+    -- Avoid caching any sequence of digits. Plus, we want
+    -- consistency here.
+    text="0123456789"
+  end
+  local fontOptions = SILE.font.loadDefaults({ features = features })
+  local supported = textFeatCache[_key(fontOptions, text)]
+  if supported ~= nil then
+    return supported
+  end
+
+  local items1 = SILE.shaper:shapeToken(text, fontOptions)
+  local items2 = SILE.shaper:shapeToken(text, SILE.font.loadDefaults({}))
+
+  -- Don't mix up characters supporting the features with those
+  -- not supporting them, as it would be ugly in most cases.
+  if #items1 ~= #items2 then
+    return textFeatCaching(fontOptions, text, false)
+  end
+  for i = 1, #items1 do
+    if items1[i].width == items2[i].width and items1[i].height == items2[i].height then
+      return textFeatCaching(fontOptions, text, false)
+    end
+  end
+  return textFeatCaching(fontOptions, text, true)
+end
+
+local function getItalicAngle ()
+  local ot = require("core.opentype-parser")
+  local fontoptions = SILE.font.loadDefaults({})
+  local face = SILE.font.cache(fontoptions, SILE.shaper.getFace)
+  local font = ot.parseFont(face)
+  return font.post.italicAngle
+end
+
+local function getWeightClass ()
+  -- Provided we return it in the font parser, maybe we could
+  -- rather do:
+  -- local ot = require("core.opentype-parser")
+  -- local fontoptions = SILE.font.loadDefaults({})
+  -- local face = SILE.font.cache(fontoptions, SILE.shaper.getFace)
+  -- local font = ot.parseFont(face)
+  -- return font.os2.usWeightClass
+  return SILE.settings:get("font.weight")
+end
+
+local function rescaleFilter (node, content, args)
+  if type(node) == "table" then return node end
+  local result = {}
+  local chars = SU.splitUtf8(node)
+  for _, char in ipairs(chars) do
+    if not tonumber(char) then
+      result[#result+1] = inputfilter.createCommand(
+        content.pos, content.col, content.line,
+        "textsubsuper:scale", {
+          xRatio = args.xScale,
+          yRatio = args.yScaleOther
+        }, { char }
+      )
+    else
+      result[#result+1] = inputfilter.createCommand(
+        content.pos, content.col, content.line,
+        "textsubsuper:scale", {
+          xRatio = args.xScale,
+          yRatio = args.yScaleNumber
+        }, { char }
+      )
+    end
+  end
+  return result
+end
+
+local function rescaleContent(content)
+  local transformed
+  if SILE.outputter ~= SILE.outputters.libtexpdf then
+    -- Not supported
+    transformed = content
+  else
+    transformed = inputfilter.transformContent(content, rescaleFilter, {
+      xScale = 1,
+      yScaleNumber = SILE.settings:get("textsubsuper.vscale.number"),
+      yScaleOther = SILE.settings:get("textsubsuper.vscale.other"),
+    })
+  end
+  SILE.process(transformed)
+ end
+
+local function declareSettings (_)
+  SILE.settings:declare({
+    parameter = "textsubsuper.scale",
+    type = "integer",
+    default = 0.66,
+    help = "Size scaling ratio of a fake superscript or subsscript"
+  })
+
+  SILE.settings:declare({
+    parameter = "textsubsuper.bolder",
+    type = "integer",
+    default = 200,
+    help = "Weight increase of a fake superscript or subsscript (e.g. 200 for normal to semibold)"
+  })
+
+  SILE.settings:declare({
+    parameter = "textsubsuper.vscale.number",
+    type = "number",
+    default = 0.90,
+    help = "Vertical ratio applied to numbers in fake superscript or subsscript"
+  })
+
+  SILE.settings:declare({
+    parameter = "textsubsuper.vscale.other",
+    type = "number",
+    default = 0.95,
+    help = "Vertical ratio applied to numbers in fake superscript or subsscript"
+  })
+
+  SILE.settings:declare({
+    parameter = "textsubsuper.offset.superscript",
+    type = "measurement",
+    default = SILE.measurement("0.70ex"),
+    help = "Offset of a fake superscript above the baseline (logically in a font-relative unit such as ex)"
+  })
+
+  SILE.settings:declare({
+    parameter = "textsubsuper.offset.subscript",
+    type = "measurement",
+    default = SILE.measurement("0.25ex"),
+    help = "Offset of a fake subscript below the baseline (logically in a font-relative unit such as ex)"
+  })
+end
+
+local function registerCommands (_)
+  -- REAL SUPERSCRIPT / SUBSCRIPT WHEN AVAILABLE
+
+  SILE.registerCommand("textsuperscript", function (options, content)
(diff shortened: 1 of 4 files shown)


Omikhleia:
N.B. These `\textsuperscript` and `\textsubscript` are also possibly useful for (Pandoc-extended) Markdown, as in `H~2~O is a liquid, 2^10^ is 1024` -- Switching to math is another alternative, but it has other drawbacks (e.g. switching to math font, which would be problematic in most cases e.g. with`16^th^`) - So I am using them:
- In my Lua-based Pandoc writer
- In some tentative at improving **lunamark** for #1336 but I'll comment there.

Omikhleia:
Re-enabling "Allow edits and access to secrets by maintainers" on this one.
Removed the task list - I am not going to address the two points mentioned there.
It's now all up yo you.

Omikhleia:
Update:
- Updated to 0.14 API
- Added a non-regression test - Not sure yet why it is failing (seems to be a font (version or style) issue, even though my dockerized master uses Makefile.fonts to install the fonts), will investigate later...

packages/textsubsuper/init.lua
```diff
@@ -0,0 +1,366 @@
...
+      end
+    })
+  end, "Scale content by some horizontal and vertical ratios")
+end
+
+return {
+  declareSettings = declareSettings,
+  registerCommands = registerCommands,
+  documentation = [[\begin{document}
+\script[src=packages/textsubsuper]% SILE 0.13 needs a big big FIXME, it's 0xDEADBEEF
```
alerque, line 296: I don't understand what you think is the matter here. The manual *partially* loads packages just enough to flesh out a list of valid commands and load documentation. It does *not* initialize packages automatically when outputting their documentation because some of them should not be loaded and would screw up the manual. If you dogfood the package in it's own documentation (as this does by using `\textsubscript` etc. in the documentation) then you have to explicitly use it.

I'll update the API here for 0.14, but if there is something else going on I don't understand let me know.

ctrlcctrlv:
What do `oj`/`aj`/`ojn`/`ajn` look like?

packages/textsubsuper/init.lua
```diff
@@ -0,0 +1,374 @@
...
+        yScaleNumber = SILE.settings:get("textsubsuper.vscale.number"),
+        yScaleOther = SILE.settings:get("textsubsuper.vscale.other"),
+      })
+    end
+    SILE.process(transformed)
+   end
+
+  -- REAL SUPERSCRIPT / SUBSCRIPT WHEN AVAILABLE
+
+  self:registerCommand("textsuperscript", function (options, content)
```
alerque, line 179: Why prefix these commands with "text". Would `\superscript` be okay? I can potentially see more than one package doing these things in a different way, but I would also expect them to be mutually exclusive. No?

ctrlcctrlv: I think due to the maths package it's a no

alerque: Math mode has a completely different parser and command space. No regular SILE commands are available in the math language and vice versa.

alerque: Fair enough. I'm not sure whether it is the TeX feel or sheer verbosity that bothers me, but `sub`/`sup` doesn't sound bad. Opinion on that @Omikhleia?

(and pinging @OlivierNicole for kudos on there *being* a math mode)

packages/textsubsuper/init.lua
Omikhleia: I'll refactor this comment. I can't even remember what it was about (I think the original rebase where I had to add this line occurred in between the time we had `\script` and not yet the new `\use` and things were not that clear about what each would do).

packages/textsubsuper/init.lua
Omikhleia: I'd advocate that there are _good_ reasons for these commands to be called `\textsub/super` rather than `\sub/super`, IMHO.
- They are not _general_ sub/super (à la HTML) that would work on any type of content (e.g. boxes, etc. - for that, we already have "raiselower"...).
- ... but rather, they work on "text" strings:
  - Trying to use font features when available...
  - Trying to redimension/scale some things e.g. digits...
  - Trying to use semibold (following Bringhurst) when they can etc...
- I.e., overall, trying to get a "decent typographic" output for e.g. 12th century, issue n° 2 (without a degree), Mlle ("mademoiselle"), etc. - That's again text sub/superscripted text bits, but not general super/subscript (you can't even expect them to nest with several levels...)

N.B. 
>  (It's really impressive, @Omikhleia et al. did a great job.) 
I had not involvement in math.... But the math stuff is unrelated, here we are dealing with pure text things, i.e. typography for raised/lowered text outside the context of math typesetting (again: centuries, abbreviations, numbers, etc. - topics that don't mandate math to even be supported and have other, albeit similar, constraints).

Omikhleia: BTW: Who cares about (La)TeX possibly similar names? There are certainly dozens of existing TeX/LaTeX packages that use similar names (_confer_, I even discovered an "enumitem" package I had no prior knowledge of, and anyway we ended up calling this "lists" - Wait, I am sure some LaTeX stuff already used those names for something else...). Sometimes we just have to stop running after TeX, no? (Esp. with regards to math, when we claim compliance to MathML over TeX, so why bother?)

packages/textsubsuper/init.lua
Omikhleia: Done.

Omikhleia:
> What do `oj`/`aj`/`ojn`/`ajn` look like?

@ctrlcctrlv What do you mean or expect? As sub or superscripts (faked or not, depending on font). I have the uncertain feeling that these look like Esperanto endings, but I can't recall if Esperanto uses ever superscripts at all or has specific typography rules about them - admittedly, my knowledge of Esperanto is very limited: though a conlanger, I have always been more interested in artlangs than auxlangs :p

packages/textsubsuper/init.lua
```diff
@@ -0,0 +1,374 @@
...
+default Gentium font, which does not have these features\footnote{Though it does
+include some of the Unicode superscript and subscript characters, but this very
+package does not try to address such a case.}). In that case, it relies on
+scaling and raising (or lowering) the characters, so as to build “fake”
+superscripts (or subscripts).
+
+By nature, this package is \em{not} intended to work with multiple levels of
+superscripts or subscripts. Also note that it tries not to mix up characters
+supporting the features with those not supporting them, as it would be somewhat
+ugly in most cases. Fake superscripts or superscripts will also be used if such
```
Omikhleia, line 330: Read: fake superscripts or subscripts

packages/textsubsuper/init.lua
```diff
@@ -0,0 +1,374 @@
...
+    type = "integer",
+    default = 0.66,
+    help = "Size scaling ratio of a fake superscript or subsscript"
+  })
+
+  SILE.settings:declare({
+    parameter = "textsubsuper.bolder",
+    type = "integer",
+    default = 200,
+    help = "Weight increase of a fake superscript or subsscript (e.g. 200 for normal to semibold)"
```
Omikhleia, line 103: Here and later: subsscript > subscript ;-)

OlivierNicole:
> The real question is:
> 
> image
> 
> Which line is the best?
> 
> And I would have to be fully honest, so...
> 
> image
> 
> (The Brill font has pretty good real superscripts... It's hard to compete here baby_chick )

Regarding the first picture, I prefer the first line, in the second the superscripts seem thinner than the rest of the text (I suppose these are the fake ones).

In the second picture I can hardly see any difference. In fact, in both cases the superscript “t” does not seem to be a scaled-down version of the normal t. Out of curiosity, is there a reason for this?

Omikhleia:
@OlivierNicole :
> Regarding the first picture, I prefer the first line, in the second the superscripts seem thinner than the rest of the text (I suppose these are the fake ones).

Hehe, wrong guess ;-)
Font: Libertinus Serif
First line: fake super/subscripts = scaled, raised... and semibold applied (hence the slightly heavier look), plus another trick (see further below)
Second line: font native super/subscript features.

As I wrote, Libertinus' native superscripts are somewhat decent (esp. compared to those from other fonts). Still, they might be a bit too thin and a bit positioned too high IMHO ;)

> In the second picture I can hardly see any difference. In fact, in both cases the superscript “t” does not seem to be a scaled-down version of the normal t. Out of curiosity, is there a reason for this?

Font: Brill
First case in each example line: font native super/subscript features.
Second case in each example line: fake super/subscript. 

The question is sound, though. There's a small additional trick (with documented tunings): When faking super/subscripts, besides scaling and raising, the package also applies some small vertical shrinking with a transformation matrix (CTM) -- when the outputter is libtexpdf -- With a bit more scaling on digits (90%) and a bit less on other characters (95%). These ratios are kind of empirical, but try to mimic how such characters were _usually_ designed. (That's the "clever PDF-rescaling alluded to in my second point in the PR description)

The real math would be much harder, re-designing each character specifically (of course), but also checking the max ascender of a regular line and other subtle details... So it's only a naive first-hand approximation here. Still, I'd be quite happy if the fake ones with Brill do not look that different from the native ones, which are really well designed. BTW, one feature I briefly considered is also inserting a bit of inter-letter spacing (as can be seen in the Brill native case, the superscript letters are slightly less cluttered than in the fake examples), but I renounced it -- it proved quite impossible to provide a decent default setting for any font and/or in combination with digits, and for an extra perfectionism which would have been, at best, quite debatable (I suspect it would kind of kill natural kerning, so the trade-off is uncertain).

OlivierNicole:
Haha I see, well, good to know.

ctrlcctrlv:
@Omikhleia Yes they are Esperanto. I use them, but my Esperanto typogaphy advice is decidedly unorthodox, it's gotten me in trouble with @alerque before even. :joy:

Omikhleia:

PR #1482, opened 2022-07-19
Add explicit \use command

### LINKED ISSUE #1385: De-conflate script & require
On the Lua side of this we are fine, but on the SILE `\script` side we have conflated two different operations: `require` and `dofile` are fundamentally different and we should have both. So far we keep using `\script[src=package/foo]` as if running a script was the same as loading a library, but it is not. Loading a library happens once and then even running it again will have (or should have) no effect. On the other hand a document **should** be able to execute an external script more than once and have it run all over again.

I think we need to unscramble this and make the `\script[src=...]` always re-execute the external code (previously loaded or not) and add a new `\require` or `\package` or something to that effect that runs `SILE.require()`.

### LINKED ISSUE #1486: Inputter options
The `\include` command is implemented as follows:
```
  self:registerCommand("include", function (options, content)
    if options.src then
      SILE.processFile(options.src, options.format)
    else
      SILE.processString(content[1], options.format)
    end
  end, "Includes a content file for processing.")
```
(I am quoting, on current master at this date, that is after the nice changes added by @alerque, but the remark below also applies to earlier versions of it AFAIK).

Would it be possible for it to have extra options, passed to the underlying inputter's `process` method (which only takes the content data doc currently), which could them use them to affect its parsing logic?

I am foreseeing at least two use cases for this.

- For XML, ability to enforce a namespace/prefix on tags. I have some documents which have tags corresponding to SILE commands, but used for other purposes. It was workable by saving the original command to something else, but a nice option could be to enforce some sort of namespacing. For instance `\include[src=myfile.xml, prefix=myscheme]` and all tags would be prefixed (e.g.) with `myscheme:`, so that I can separate those XML tags from SILE commands, avoiding clashes (and implementing them all under that appropriate naming scheme). Say, e.g. `<comment>xxx</comment>` would become `\myscheme:comment{xxx}` and I don't have to temporarily switch/restore the usual `\comment`.
- For Markdown, there are many flavors of it (read: extensions), without obvious ways to know in advance, and this could be a way to select some of them, e.g. `\include[src=myfile.md, smart=false, startnum=false]`

This is just a low-priority "convenience" remark in passing, none of this is absolutely necessary: in the Markdown cases, these are mostly Pandoc-like extensions that seldom affect the writer's intentions, AFAIK; in the XML case, it is rather easy, as noted, to workaround it (though a bit of a challenge sometimes), and in the worst case, preprocessing the input with an XSLT stylesheet (or other solutions) is quite doable too.

alerque:
Closes #1385
Closes #1486


alerque:
Two bonus points here besides the other discussion:

1. This will fix the upcoming problem of using SILE as a library and potentially having more than one SILE instance active. In that case running `require()` which is global and handled by Lua will very much not be the same thing as `class:loadPackage()` which will tie the package to the class scope.
1. Once we get the internal APIs for it this will lend itself to an `\unload-package` command ;-)

alerque:
See discussion on naming this function in relevant #1385.

OlivierNicole (approved):
For the math package, this looks good to me (I only looked at that), since the changes seem pretty systematic and the tests pass.

Issue #1494, opened 2022-07-24
math: Modulus operator


matteodelabre:
As far as I can tell, SILE’s `math` package doesn’t currently have support for the various forms of the modulus operator. I’d be happy to provide a patch, but I wanted to first discuss this here to make sure I’m not missing something.

Expected presentation
---------------------

The modulus has two main forms.

1. As an operation that refers to the remainder of a division, it is usually written in infix notation, using the `mod` or `%` signs, or in prefix notation. In LaTeX,

```tex
\documentclass{minimal}
\begin{document}
    \[42 \bmod 11 = 9\]
    \[42 \mathbin{\%} 11 = 9\]
    \[\mathrm{mod}(42, 11) = 9\]
\end{document}
```
renders as

![Screenshot from 2022-07-24 16-49-30](https://user-images.githubusercontent.com/1370040/180665338-3692eb99-f66c-4b40-8700-3647a6735409.png)

2. As part of a modular congruence relation, it is usually written at the end of the line, separated by a quad space, optionally within parentheses. In LaTeX,

```tex
\documentclass{minimal}
\usepackage{amsmath}
\begin{document}
    \[42 \equiv 9 \pmod{11}\]
    \[42 \equiv 9 \mod{11}\] % \mod provided by amsmath
\end{document}
```

renders as:

![Screenshot from 2022-07-24 16-48-32](https://user-images.githubusercontent.com/1370040/180665300-e5fc44e8-d190-4bb6-bf31-a7603f4fe1e4.png)

In SILE
-------

Trying to set the forms listed above in SILE leads to incorrect spacing, and the form that uses the `%` symbol cannot be written in `\math` since the `\%` command does not exist there.

```tex
\begin{document}
    \script[src=packages/math]
    \math[mode=display]{42 \mo{mod} 11 = 9}
    \mathml[mode=display]{\mrow{\mn{42}\mo{\%}\mn{11}\mo{=}\mn{9}}}
    \math[mode=display]{\mi{mod}(42, 11) = 9}
    \math[mode=display]{42 ≡ 9 (\mo{mod} 11)}
    \math[mode=display]{42 ≡ 9 \mo{mod} 11}
\end{document}
```

This renders as:

![Screenshot from 2022-07-24 16-55-11](https://user-images.githubusercontent.com/1370040/180665530-b6d17411-007b-40dd-b50e-f6cd2e0cbc5a.png)

Solutions
---------

[...]

Issue #1361, opened 2022-04-14
Font-relative parindent and line justification issue


Omikhleia:
And after #1358 and #1359, a train can hide yet another again ;) Here is another source of my justification issues.

![image](https://user-images.githubusercontent.com/18075640/163291209-3ee4b6df-d033-44bb-b212-3a564abbbb11.png)

I have this small "block quote" environment that leaves Hmode, sets a smaller font and reduced lskip and rskip margins. As can be noticed, the lines beginning a paragraph within that environment have a small offset.

FWIW, my input code looks as follows.

```
\begin[class=omibook, papersize=6in x 9in]{document}
\language[main=fr]
\set[parameter=linebreak.emergencyStretch, value=15%lw]%

L’auteur de poursuivre :

\begin{blockquote}
C’est une tradition de toute la région, m’avait-on dit, que de tenir,
au mois de février, une fête de l’ours, ou \em{festa de l’ós}. Bien que
ce ne soit pas propre à la Cerdagne, mais aussi au Vallespir &c.

Je ne m’étonnai pas sur le moment de cette étrange comparaison. Notre
discussion dériva bientôt vers d’autres sujets. Lorsque le souvenir m’en
revint ensuite sur la route, je ne pus me retenir de penser que mon
interlocuteur avait en réalité voulu évoquer (...)
\end{blockquote}

Lorem ipsum (...)

\end{document}
```

But nothwithstanding how I implemented it, I think the important things here is that the `document.parindent`  was expected to be font-relative, i.e. it is set to `1.2em` in my class - because smaller text in such quotes ought to have a smaller indent, ideally...

Back to default plain class and a simpler test case, the observed behavior can be simplified as follows:

```
\begin[papersize=6in x 9in]{document}
\script[src=packages/rules]
\script[src=packages/lorem]
\nofolios
\font[size=30pt]
\set[parameter=document.parindent, value=1em]

\noindent{}\hrule[width=1em]% This rule just to show a 1em width at 30pt

\font[size=10pt]{\par
\lorem[words=35]

\lorem[words=35]
\par}

\end{document}
```

![image](https://user-images.githubusercontent.com/18075640/163292310-483b96a2-1226-48bc-afb0-0de5713a85dd.png)

[...]

PR #1496, opened 2022-07-28
math: Modulus support, operator atom type override, space commands, and macro fix

matteodelabre:
This set of patches implements what was discussed in #1494.

Summary of changes:

* Improvements in the debug logs in the `math` and `texmath` scopes. The debug statements previously printed a bunch of unhelpful ″table: <address>” lines. They now print the TeX parse tree, compiled MathML, and generated mbox tree.
* Macros no longer wrap their replacement in `mrow` elements. This was unexpected as macros are supposed to behave as simple replacements, plus this caused some spacing errors with operators inside macros.
* Add TeX-like space commands in math mode, with relative space dimensions.
* Add support for the modulus operator, define `\bmod` and `\pmod` for commonly used forms.
* Allow authors to override the atom type of an operator with the `atom` option, e.g. `\mo[atom=open]{/}` makes a forward slash behave as an opening parenthesis.

The output of regression tests `math-fractions`, `math-bigops` and `math-macros` changed (the changes improve the spacing around operators in those tests). New tests `math-spaces` and `math-ops` were added to test some features of this PR.


matteodelabre:
(Rewrote commits to fix lint issues.)

packages/math/base-elements.lua
```diff
@@ -744,24 +752,37 @@ elements.terminal = pl.class({
...
   end,
   _init = function(self, width, height, depth)
     elements.terminal._init(self)
     self.width = width
     self.height = height
     self.depth = depth
   end,
-  shape = function (_) end,
+  shape = function (self)
+    self.width = self.width:absolute() * self:getScaleDown()
```
Omikhleia, line 767: Do you confirm the `absolute()` is required here for some reason? For operations on length, these need to be absolute indeed, but t's a bit dangerous to absolutize "in the wild" without e.g. using the current font options at the time the node was pushed, but not yet shaped. Just asking because we already have, IMHO, plenty of cases due to "too early or too late" absolutizing of relative units (e.g. "1em").

matteodelabre: Without the call to `absolute()`, the parent stackbox [will try to add relative and absolute lengths together](https://github.com/sile-typesetter/sile/blob/10b333f24f142673ee0dc2bfc23a75b4aecea253/packages/math/base-elements.lua#L462), which causes an error. Would there be a better way to handle this?

alerque: The issue is less the "way" to handle this as the "when". Is when this function gets called (and the scope it is called from) the right time in the process and does it have the right context to do so?

Omikhleia: Unless mistaken, `self.options` (inherited from the mbox) contain the options that were in effect when the node was created (which could be a different scope from that where it is pushed, but anyway). So could look as something such as:
```
  SILE.settings:temporarily(function ()
    SILE.settings:set("font.size", self.options.size)
    xxx = xxx:absolute()
  end)
```
(But I don't know that part of the code well, just seeing too that other places refer to `self.options`, e.g. to evaluate the nnodes, etc.)

Omikhleia: > which could be a different scope from that where it is pushed, but anyway
As far as I checked, the only things that are created and pushed at different place are lrskips, parskips, parindents, etc. Hence my "but anyway" - usual box contents are usually built and pushed to the typesetter queue in a row.

Omikhleia: (The mess not absolutizing in right context lead to things such as #1361)

matteodelabre: Thanks for the pointers. It does seem to me like this is the right place to do the conversion, since other `shape()` functions also use the current font size in their calculations. However it hadn’t occurred to me that the math font size is a separate setting from the main text font size. So your solution seems correct @Omikhleia, if we also take into account the scale down factor from `getScaleDown()`.

matteodelabre: I think a cleaner way would be to define a math unit as in TeX, that is relative to the math font size, while the em stays relative to the main font size. Distinguishing between the two will be especially important when we later add non-math text inside math formulas.

alerque:
By the way there may be some merge conflicts showing up here from unrelated package coding cleanup. Don't worry about that, just ping me when you're done with the functional end of this and I'll fix the conflicts and review.

matteodelabre:
Will do, thanks for the heads up.

packages/math/base-elements.lua
matteodelabre: I implemented this strategy in the latest commit. I kept the original commit that implemented relative spaces, to retain the context of our discussion for now.

matteodelabre: While working on this, I noticed that [`getMathMetrics()`](https://github.com/sile-typesetter/sile/blob/5a6de0ca3a71163b56b5bf1ebb0eac1deb92c420/packages/math/base-elements.lua#L129) caches its result, but the cache doesn’t seem to be invalidated when the font family or size is changed. Am I missing something here @OlivierNicole?

OlivierNicole: No, you are right, that was one of the things I was supposed to fix “later”…

matteodelabre: I added a fix for this in the latest commit.

OlivierNicole:
Wow, that is a big diff and certainly a lot of work, thanks a lot for that. I'll do my best to review it in a reasonable time this beginning of week.

packages/math/base-elements.lua
```diff
@@ -809,10 +809,6 @@ elements.text = pl.class({
       if self.text == "-" then
         self.text = "−"
       end
-      if symbolDefaults[self.text] then
-        self.atom = symbolDefaults[self.text].atom
-        self.stretchy = symbolDefaults[self.text].stretchy
-      end
```
OlivierNicole, line 794: So that was the reason for the incorrect spacing of integrals… cheers to you, this was a 2-year-long standing bug :sweat_smile:

packages/math/texlike.lua
```diff
@@ -401,6 +401,9 @@ local function convertTexlike (_, content)
   return ret
 end
 
+registerCommand("%", {}, function ()
```
OlivierNicole, line 404: I haven't seen the changes in the texlike parser that allow to parse `\%`. Am I missing something?

OlivierNicole (changes requested):
Except for a few minor comments, this looks good to me. This patch not only introduces new functionality but also fixes many bugs, so it's very pleasing.

packages/math/base-elements.lua
matteodelabre: Well thank you for putting me on the trail!

packages/math/texlike.lua
matteodelabre: `\%` was already parsed as a command. If I read the grammar correctly, this is because the `ctrl_sequence_name` rule has two cases, one of them (`ctrl_word`) accepts alphabetic command names of any length, the other (`ctrl_symbol`) accepts any one-character-long command name except from `{}\`, therefore including `\%`.

packages/math/typesetter.lua
```diff
@@ -40,6 +57,29 @@ function ConvertMathML (_, content)
         attributes[attribute] = value
       end
     end
+    if content.options.atom then
+      if tonumber(content.options.atom) then
```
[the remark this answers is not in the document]
matteodelabre: This is indeed probably too confusing for little added value. I’ll take it out.

packages/math/typesetter.lua
```diff
@@ -40,6 +57,29 @@ function ConvertMathML (_, content)
...
+        local atom = tonumber(content.options.atom)
+        local exists = false
+        for _, value in pairs(b.atomType) do
+          if atom == value then
+            exists = true
+            break
+          end
+        end
+        if not exists then
```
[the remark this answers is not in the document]
matteodelabre: Good point. I’ll be removing this code, but for the record Penlight [does provide such a function](https://stevedonovan.github.io/Penlight/api/libraries/pl.tablex.html#find).

matteodelabre:
Thank you Olivier for taking the time to review this, and thanks for your comments. I believe I have addressed all of them.

I guess what remains to make this PR complete is to document the new features (space commands and atom type override). I will tackle this tomorrow.

packages/math/texlike.lua
OlivierNicole: OK, I wasn't sure.

OlivierNicole (approved):

packages/math/texlike.lua
```diff
@@ -362,6 +412,26 @@ compileToMathML(nil, {}, convertTexlike(nil, {[==[
...
+  \def{!}{\negthinspace}
+  \def{medspace}{\mspace[width=med]}
+  \def{negmedspace}{\mspace[width=-med]}
+  \def{>}{\medspace}
+  \def{thickspace}{\mspace[width=thick]}
+  \def{negthickspace}{\mspace[width=-thick]}
+  \def{;}{\thickspace}
+  \def{enspace}{\mspace[width=1en]}
+  \def{enskip}{\enspace}
+  \def{quad}{\mspace[width=1em]}
```
Omikhleia, line 429: Just to be sure: these are only in math context and won't override the `\quad`, `\qquad`, `\enskip` etc. commands from the plain classes (that are available in non-math context), right?

OlivierNicole: Yes, those are TeX-like macros, they are only defined in the context of TeX-ilke `math`.

Omikhleia: You rock, folks!

alerque (approved):
Fantastic work all round folks. Thanks a lot for this contribution @matteodelabre and besides the bulk of the work I also appreciate the extra touches getting all the tests, commits, etc. aligned to the project. You made the work of reviewing and releasing this easy even without my knowing much about math typesetting! Thanks @OlivierNicole for dropping by with pointers and reviews, and @Omikhleia for keeping an eye on things while I've been otherwise occupied.

alerque:
@matteodelabre Please do post documentation updates if you have them in a new PR. I have some other stuff I wanted to mess with while testing everything we have so far together so I went ahead and merged, but I'll actually be holding up the v0.14.0 release for documentation updates on this and one other thing too...

matteodelabre:
Happy to help! I’m working on the documentation update right now.

PR #1499, opened 2022-07-29
Pass document direction info to font shaper by default

### LINKED ISSUE #1472: Mongolian shaping is foo-bared
I noticed while reviewing the Jekyll→Zola transition on the website that our Mongolian example output has shaping that is is pretty much borked to pieces:

image

A couple quick tests with docker suggest it was last working properly in v0.11.1 and anything v0.12.0–HEAD turns up the same scrambled output. Not a proper bisect yet, but at least a range to look into.

The website file `content/examples/mongolian.sil` is suitable as a regression test, although it could probably use cleanup.

alerque:
Fixes #1472

### DIFF
--- a/classes/plain.lua
+++ b/classes/plain.lua
@@ -42,6 +42,7 @@ function class:declareOptions ()
   self:declareOption("direction", function (_, value)
     if value then
       SILE.documentState.direction = value
+      SILE.settings:set("font.direction", value, true)
       for _, frame in pairs(self.defaultFrameset) do
         if not frame.direction then
           frame.direction = value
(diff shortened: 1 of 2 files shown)


alerque:
These changes *feel right* to me, but an opinion from somebody that actually publishes stuff in directions other than TTB-LTR would be welcome. @ctrlcctrlv @khaledhosny @simoncozens ...

As of this PR:

* When setting the direction option of a document class, the `font.direction` setting's **default value** is changed to match.
* When setting the direction of the current frame, the `font.direction` settings **current value** is changed to match.

Does this feel right over the prior behavior of being expected to set this manually (via either `\set[parameter=font.direction]` or `\font[direction=]`) in addition to setting the class or frame properties so things other than shaped text advance correctly?

alerque:
Baring feedback I'm moving ahead with what makes sense to me here, but I'm happy to review again if this turns out to be unexpected behavior.

Issue #1346, opened 2022-03-18
Image resolution and converters enhancement


Omikhleia:
This is a proposal for discussion about improving image inclusion and converters.

First, my assumptions (which may need some more informed check) from a brief look at the libtextpdf C code:
- It cleverly uses a PDF XObject for a given image, so if the same image is reused multiple times, its raw data is only included once in the PDF.
- Would the image be sized explicitly (e.g. with `\img[width=...]` relatively to its _natural_ (claimed) size, it uses a transformation matrix. The image is not up/down-sampled, just scaled.

Now consider I got that nice PNG image which, at its natural size, takes a 2 inches (5.08 cm) width on my page. Since the natural size of the image (see #1267) depends on the resolution set in its metadata, it could actually mean several things, e.g. (1) 1200 pixels and a resolution 600 px/inch, (2) 600 px and a resolution 300 px/inch, or (3) 300 px and a resolution 150 px/inch.

_Use case 1_
These all end up, on my page, occupying the same width... but they do not have the same quality in print. Now, say my printer supports 300 dpi (ink dots per inch). Notwithstanding Shannon's law here, one could consider (1) to have a negative useless impact on the size of my PDF, (2) to be just fine and (3) to be real bad for print, with a blurry image...

Could SILE warn me about it? Actually, it would need but a _small_ modification to the C code. The natural size is retrieved via `SILE.outputter.getImageSize`, which actually relies on `get_image_bbox` (from `src/imagebbox.c`). The latter relies on libtexpdf, but at this point it has the "density", here supposedly expressed in pt/inch. Note the pt (point): libtexpdf retrieves the metadata (which can already be in various units depending on the image format). For [a PNG](https://github.com/sile-typesetter/libtexpdf/blob/ce78dd4913bd77271c07cd5a546ebcfae153c8e8/pngimage.c#L1094), it gets them as px/m (meters, yes, thanks `libpng`... er... thanks actually [PNG format](http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.pHYs)), converts that to px/inch and multiplies by 72 (1pt = 1/72in), so we actually end up with a _density_ in pt/inch. **Anyhow**, we could return that density up to SILE's Lua (we just need to either to convert it back to a resolution, or call it a density indeed but avoid using a misleading "ppi" unit in its description!).

[...]

PR #1500, opened 2022-07-31
Image resolution and PDF page selection with libtexpdf backend

Omikhleia:
Three commits touching quite the same bits of code:
* This is the required C support for the possible enhancements evoked in #1346
  Outputters using the libtexpdf backend to compute an image size, such as "debug" and "libtexpdf", also return, as additional results on the `getImageSize()` method, the image resolutions in dots (pixels) per inch.
  This may be used by packages/classes to check images have an appropriate (e.g. print) quality regardless of their claimed size (in pt), or to downsize too resolved images, etc. - but that is a separate concern ; this PR just offers the tooling.
  Ex. with `documentation/gutenberg.png` (200 x 243 pixels) unscaled = 144pt x 174.96 pt at resolution 100 dpi x100 dpi.
  N.B. Libtexpdf's density is in pt/in. It's converted back to dpi, here, as it is more usual for users (who tend to know that decent print quality in the range of 300-600 for printed books, for instance)
* Add ability to select the page on a PDF image, i.e. `\img[src=somepdf, page=2]` (the default is obviously 1, with same behavior as before) = my misunderstanding of question #1515, but the idea is sound nevertheless.
* Additionally, image height and width parsed as `SILE.measurement` rather than complex frame dimensions (as discussed in #1267 and done on the raiselower package in #1506.

### DIFF
--- a/documentation/c12-tricks.sil
+++ b/documentation/c12-tricks.sil
@@ -186,7 +186,7 @@ One SILE project needed two different kinds of sidenotes, margin notes and gutte
 notes.
 
 \line
-\img[src=documentation/discovery.png,width=width(content)]
+\img[src=documentation/discovery.png,width=100%fw]
 \line
 
 Sidenotes can be seen as a simplified form of
--- a/outputters/debug.lua
+++ b/outputters/debug.lua
@@ -139,8 +139,8 @@ end
 
 function outputter.getImageSize (_, src)
   local pdf = require("justenoughlibtexpdf")
-  local llx, lly, urx, ury = pdf.imagebbox(src)
-  return (urx-llx), (ury-lly)
+  local llx, lly, urx, ury, xresol, yresol = pdf.imagebbox(src)
+  return (urx-llx), (ury-lly), xresol, yresol
 end
 
 function outputter:drawSVG (figure, _, x, y, width, height, scalefactor)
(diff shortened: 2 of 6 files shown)


alerque (approved):

alerque:
Thanks for these contributions. Sorry it is taking me so long to review stuff these days. It's hard to get my own production work done, have a bunch of my own refactoring and development projects loaded up in my head, and reason well about contributions along other lines all at the same time.

ctrlcctrlv:
I think that we should have a meta discussion about what's a _third-party package_ and what's supposed to be PR'd here, as @Omikhleia PR's a lot of stuff I consider packages.

Notably, SILE and (Xe|Lua|p)?(La)TeX(2e)? disagree on this issue:

https://github.com/ctrlcctrlv/sile-logos
image

Precedent says that in SILE this is firmly a third party package, while in (Xe|Lua|p)?(La)TeX(2e)? it's a core macro.

alerque:
Yes, we do. We have several PRs here (including some of my own) that I plan to repurpose as packages. This PR however provides core tooling that would be needed by packages…

ctrlcctrlv:
Oh, absolutely. I'm commenting on the all of @Omikhleia's PR's as a group—not this particular one, didn't even read the title, just your comment :joy:

PR #1501, opened 2022-07-31
Minor href regression and grid fix

Omikhleia:
- Seen in passing, an issue introduced in 0.13
  When pdf is supported, `\href[src=uri]{text}` outputs the text in current font and `\href{uri}` delegates the content to `\url`, with additional breakpoints and possibly a different font. The formatting behavior shall be the same when pdf is not supported (i.e., same visual output, whether the hyperlink is active or not).
- From [discussion here](https://github.com/sile-typesetter/sile/pull/1334#issuecomment-1200125592)
  Incorrect detection of the backend.
- From [analysis here](https://github.com/sile-typesetter/sile/issues/1174#issuecomment-1173141699)
  Ensuring the grid frame hook is ineffective when the grid is off. That might not be a true fix, rather a workaround, but the immediate benefit is that some long-reported vertical issues in the SILE manual are solved.

### DIFF
--- a/packages/grid/init.lua
+++ b/packages/grid/init.lua
@@ -53,6 +53,7 @@ local function pushExplicitVglue (typesetter, spec)
 end
 
 local function startGridInFrame (typesetter)
+  if not SILE.typesetter.state.grid then return end -- Ensure the frame hook isn't effective when grid is off
   local queue = typesetter.state.outputQueue
   typesetter.frame.state.totals.gridCursor = SILE.measurement(0)
   if #queue == 0 then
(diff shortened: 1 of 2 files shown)


alerque (approved):
Thanks for the fixes!

A super minor nitpick that I just fixed: when considering commit message types, `fix:` only applies when fixing bugs that existed in the last tagged release. When fixing bugs introduced in development work since the last release, just use `chore:`. That way they don't make their way into the changelog when I release. It doesn't make sense for people following along with release news to see things that aren't relevant to released code—its' more a "how the sausages are made" kind of thing ;-) It can still be `chore (scope): Fixed blah blah` the "Fixed" verb in the message is fine just not the commit type callout.

PR #1503, opened 2022-08-03
EAN-13 (ISBN etc. barcodes)

Omikhleia:
I just ported my "EAN-13 barcodes" package to (pre)0.14 API, augmented with new features compared to the April 2022 version that revealed bug #1382. I am making a PR here in case you could be interested, ~~but as a draft PR as testing revealed another issue (which I am going to open separately)~~ (fixed).

This is the kind of things it can do:
image

(So now also supporting EAN-13/2 and EAN-13/5 supplemental parts)

_Rationale_
It's cool to be able to typeset an ISBN in a book or its cover with SILE without external program ;)


Omikhleia:
Update, in passing:
- A minor tweak to just make this play a bit nicer with other fonts than OCR B or FreeMono, in case they end up being less wide than expected, so e.g. the sequences of digits below look "decently" centered (instead of being somewhat flushed left in the initial implementation), despite the "poor" font choices:
image
(Here using the "Hack" font, which I had set in the package documentation (to avoid hazardous substitutions), which I have not added to the SILE manual here, but I just did it in my local working branch, and I was not that satisfied with the look of it ;)
- The non-reg. test had indent and folios, just removed them avoid the useless noise.

ctrlcctrlv:
I think this belongs as a PR to sile-typesetter/packages, not here :)

https://github.com/sile-typesetter/sile-packages/blob/master/packages.lua

Has everyone forgotten about this file? Seems I'm its only recent-ish user…

alerque:
I was kind of thinking to make this an external package too. But not with that packages system, it is completely obsolete. I need to demo how to publish packages anyway…

ctrlcctrlv:
Apologies, I don't use SILE in a really normal way&dagger; so I didn't know it had been deprecated for over a year.

&dagger; I use it to test fonts, make silly documents that are overkill by any standard, and every few years write Unicode papers with it.

Omikhleia:
@ctrlcctrlv :
> I think this belongs as a PR to sile-typesetter/packages, not here :)

This "old" package manager thing died long ago, no? And was neither maintained nor kept up with SILE -- as the few linked packages (Take @simoncozens endnotes... Guess what, I also need end notes. I wonder what my best course of action would be, here? Not to mention the other unmaintained packages listed there... *gentle smirk*).

But why "not here"? Such a small thing that assemble hrules? Please explain, why make it a "second-class" citizen, however, in the form of an "external" package, whatever this really means (in terms of packaging, maintenance, non-regression, etc.)?

On one hand, I could advocate, being obviously biased towards it:
- Real books usually have an ISBN (so here, it's all about what SILE targets, or actually doesn't...)
- AFAIC:
  - My own book class will always load it, because an ISBN is expected in front or end matters...
  - My own "cover designer" will always load it, because it's usual for an ISBN to occur in back cover (yup, my current work-in-progress is a cover designer class - That's kind of fun after viewing @simoncozens FOSDEM 2015 talk again recently, it all started with a cover design issue, lol... I am _still_ lagging behind him, with my weird needs :rotfl: ).
- When writing, then updating, this "small" package, I found two issues (#1382, #1504). How worse could have this been? What are the risks with having it? ;)

On the other hand, I don't *really* care.
I wrote it, refined it, so eventually proposed it here, that sounded logical -- but that's all. It's MIT-licensed, so do with it what you deem is right :)

Still, in 2016, @simoncozens once wrote (in #324):
> You know what, right now, I'm thinking that all user-contributed packages and classes should just go in the SILE distribution. Avoids all the package manager madness, and we're talking about a bunch of tiny Lua files. Right now the packages and classes directories take up 300 kilobytes. If that grew by a factor of 100, I still wouldn't be worried.

That was 6 years ago. Still, there isn't yet any kind of working ecosystem outside the core distribution, with so many new classes or packages that it couldn't cope with it. Just questioning: what do we expect?

Omikhleia:
But then I realize I missed an earlier discussion there https://github.com/sile-typesetter/sile/pull/1500#issuecomment-1216478671 on an unrelated PR... Quoting @ctrlcctrlv there:
> I think that we should have a meta discussion about what's a third-party package and what's supposed to be PR'd here, as @Omikhleia PR's a lot of stuff I consider packages. (...) I'm commenting on the all of @Omikhleia's PR's as a group—not this particular one (...)

Maybe a clarification is indeed required, with clear guidelines.

But it would have to be more specific than "on all of Omikhleia's PR"... Out of my 9 current **open** PRs:
- 1 is a draft attempt touching on C code (but maybe the whole "svg" package ought to be an external package outside of SILE's scope?) (1517)
- 1 is related to fluent and was partly taken into account (and is likely kept open for future reference) (1474)
- 2 are touching upon the typesetter logic and/or the base class (1470, 1334)

This "only" leaves us with 5 PRs that could be in that discussion (= it's not as if I had been proposing all my classes and packages, right?):
- This very one on EAN-13 ISBNs = Maybe because each time I tackled with it, it raised issues ;)
- Cross-references (1366) = I _only_ proposed it because there such a request issue (your own, @ctrlcctrlv ! From now 2+ years !) _and_ it was listed in the Wiki too. I wouldn't mind if it's an "external" package (or even if it's ditched here), but for such things, then, the Wiki would have to be clarified, and related issues moved out.
- Boxes and tables = Again tackles with several open issues and requests... Yet, it could be an external package too - but are we going to keep "simpletable" in the core distribution then?
- The textsubsuper thing (1475)
- Markdown = The whole thing (inputter, class, packages) could entirely be moved out. Including the previous "support" we had for it, that it.

But it could go further then... Plenty of things could be moved out as non-depending on core, being too specific etc.
- too specific classes: bible, letter, docbook...
- too specific packages: bibtex, boustrophedon, chapterverse, chordmode, dropcaps, footnotes and tableofcontents (not even kidding on those two: I use replacements), hanmenkyoshi, indexer, pandoc, pullquote, specimen, xmltricks...

Omikhleia:

Omikhleia:
There it goes https://github.com/Omikhleia/barcodes.sile

PR #1505, opened 2022-08-03
fix(package): Unit conversion issues in raiselower package

### LINKED ISSUE #1504: Programmatic use of lower breaks (raiselower package)
**Test scenario**
```
\use[module=packages.raiselower]

Text\lower[height=0.5ex]{lowered}% OK

Text\lua{SILE.call("lower", { height = "0.5ex" }, { "lowered" })}% OK

Text\lua{SILE.call("lower", { height = SILE.length("0.5ex") }, { "lowered" })}% OOOPS
\end{document}
```
**Observed results**
When compiling, the last case (marked OOOPS) generates:
```
Error detected:
	./packages/raiselower/init.lua:28: attempt to concatenate a table value (field 'height')
```
**Analysis**
At some point in 0.13 (IIRC), some "string" shims were removed, that allowed `SILE.length` (or `SILE.measurement` etc.) to support automatic conversion to string -- leading to this regression (though the code concatenating a minus char to the provided height option had some smell ;) 

_Issue found when upgrading some 0.12.5 code (#1503) where I use computed lengths in Lua code - so it likely means there's no existing test coverage for it._

Omikhleia:
Suggested fix
- Closes #1504 on `\lower`
- For consistency, addresses too the controversial frame-related units in `\raise`, discussed in #1267 

Some additional unit tests would be required, though.

### DIFF
--- a/packages/raiselower/init.lua
+++ b/packages/raiselower/init.lua
@@ -6,8 +6,7 @@ package._name = "raiselower"
 function package:registerCommands ()
 
   self:registerCommand("raise", function (options, content)
-    local height = options.height or 0
-    height = SILE.parseComplexFrameDimension(height)
+    local height = SU.cast("measurement", options.height or 0):absolute()
     SILE.typesetter:pushHbox({
         outputYourself = function (_, typesetter, _)
           typesetter.frame:advancePageDirection(-height)
@@ -25,7 +24,8 @@ function package:registerCommands ()
   end, "Raises the contents of the command by the amount specified in the <height> option")
 
   self:registerCommand("lower", function (options, content)
-    SILE.call("raise", { height = "-" .. options.height }, content)
+    local height = SU.cast("measurement", options.height or 0)
+    SILE.call("raise", { height = -height }, content)
   end, "Lowers the contents of the command by the amount specified in the <height> option")
 
 end


Omikhleia:
Oh so you went for another fix in #1506 - This PR can be rejected then.

Except I think there might be something wrong in #1506, the absolute() seem to be on lower but not on raise (discrepancy arising?). I dunno for sure, it was merge before any review.

Omikhleia:
> Oh so you went for another fix in #1506 - This PR can be rejected then.
> 
> Except I think there might be something wrong in #1506, the absolute() seem to be on lower but not on raise (discrepancy arising?). I dunno for sure, it was merge before any review.

Ignore that. Just checked #1506 and it seems allright. So you can reject this PR as duplicate, we arrived at the same conclusions. :+1:

alerque:
Sorry about the duplicated work, somehow I completely missed the PR when I first read the issue. Yes it seems we arrived at the same conclusions independently.

Omikhleia:
> Sorry about the duplicated work, somehow I completely missed the PR when I first read the issue. Yes it seems we arrived at the same conclusions independently.

No problem, these were just a few lines (quickly written so I could proceed with my 0.12 to 0.14 ports) -- and it's pretty cool we got the same conclusion on that issue.

PR #1508, opened 2022-08-04
Build back Babel

alerque:
This is cherry picks from #1474 with some rehashing to:

* Not include messages we don't utilize yet, providing ones people *may use* is something I want to think on more.
* Prefix messages with their module scope.
* More idiomatic Fluent usage, especially no term reuse with no context.

Also:

* Adds support for SILE commands in FTL messages via XML syntax.


alerque:
Important to address before merging: point 6 from original PR. The Fluent messages need to fall back to something, probably the `und` local for messages that don't exist.

i18n/ar.ftl
```diff
@@ -0,0 +1,5 @@
+bibliography-et-al = et al.
```
alerque, line 1: Something doesn't feel right about this `bibliography-et-al` key having the same Latin phrase repeated it all the localization files. Sure academic texts in many/most Western languages probably *do* actually use this phrase. But at the same time RTL languages are going to hit BiDi here, and while many Arabic fonts have been including Latin glyphs for end user convenience, for example most Amharic fonts do not (see `am.ftl`).

@Omikhleia Did this actually come out of Babel/Polyglossia/somewhere indicating this string is used in those languages? If not I'm seriously inclined to move this to a fallback somewhere like 'und' plus languages like 'en' that we can confirm as actually using. If so then maybe the least we can do is wrap it in a language tag like:

```fluent
bibliography-en-al = <language main="la">et al.</language>
```

Thoughts?

Omikhleia: @alerque No, this `bibliography-et-al` is all ours (from your initial work on "fluentizing" this package). I haven't found anything looking like it in the LaTeX Babel templates, so it doesn't seem to support it. (There might be other LaTeX packages for such things, I haven't investigated).

Sure thing is that some non-latin languages might indeed have ways to translate it, e.g. arabic https://context.reverso.net/translation/english-arabic/et+al --- but I'm afraid we would have to take our shovels to check which and how! Russian for instance https://context.reverso.net/translation/english-russian/et+al seems to use _et al._, _и др._ (just a transcription, I assume) or _и соавт._ depending on quoted sources...

Omikhleia: BTW, I'm told the Russian и соавт. might be the abbreviation for "and co-author(s)"

alerque: No, *и др.* is not a transcription of *et al.*, it is short for *и другие.* with the meaning of the Latin translated. But yes that does look promising. I think I'll drop all the ones we can't directly confirm as it does not seem helpful to stuff Latin into languages that may or may not use it.

PR #1509, opened 2022-08-04
Update math documentation following #1496

matteodelabre:
This patch adds documentation for atom types and new math spacing features added in #1496. I omitted documenting the `inner`, `over`, `under`, `accent`, `radical`, and `vcenter` atom types since they seem to have been carried over from TeX but not actually used in the current package code (please correct me if I’m wrong @OlivierNicole).

The `math` package documentation was getting long, so I added some sectioning. For lack of a `\subsubsection` level, I reused a paragraph-like sectioning style from the `lists` package documentation.

I also added a line mentioning the `math.font.size` setting.

### DIFF
--- a/packages/math/init.lua
+++ b/packages/math/init.lua
@@ -80,6 +80,8 @@ package.documentation = [[
 \use[module=packages.font-fallback]
 \font:add-fallback[family=Symbola]
 
+\define[command=paragraph]{\smallskip\em{\process.}\novbreak\par}
+
 This package provides typesetting of formulas directly in a SILE document.
 
 \note{Mathematical typesetting in SILE is still in its infancy.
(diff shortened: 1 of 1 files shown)


alerque (approved):

OlivierNicole:
> I omitted documenting the `inner`, `over`, `under`, `accent`, `radical`, and `vcenter` atom types since they seem to have been carried over from TeX but not actually used in the current package code (please correct me if I’m wrong @OlivierNicole).

Mm, we do support the MathML `<munderover>` tag. It behaves like `\stackrel{...}{...}` in TeX.

matteodelabre:
Oh, that’s right, but I didn’t see the `under`/`over` atom types actually used in the code. For what I could find, it’s only assigned to the `\underbracket`, `\underparen` and `\underbrace` symbols (and their `\over…` counterparts). If I’m not mistaken, those are meant to set a stretchy bracket/parenthesis/brace under/over part of a formula, which is distinct from the `<munderover>` feature.

PR #1511, opened 2022-08-04
Address discarded math

### LINKED ISSUE #1495: math: Equation disappears when immediately preceded by a line break
When the following document is compiled, the second equation (`y`) is not shown.

```tex
\begin{document}
    \script[src=packages/math]
    \math{x}\cr\math{y}
\end{document}
```

Adding just a single non-whitespace character between `\cr` and `\math{y}` is enough to make it display the `y`. This also happens with the `\mathml` command.

I’m using the latest commit from master (c7db5d5b3c5c6756ffe7e923ea3b938148167b2b). Here’s the output from `--debug=versions`:

```console
$ ./sile --debug=versions ./wraptest.sil
SILE v0.13.3.r182-gc7db5d5-dirty (Lua 5.4)
<./wraptest.sil> as sil
[1]
Harfbuzz version: 4.4.1
Shapers enabled: graphite2, ot, fallback
ICU support enabled

Fonts used:
/usr/share/fonts/libertinus/LibertinusMath-Regular.otf:0        Version 7.040;RELEASE
/usr/share/fonts/gentium/GentiumPlus-Regular.ttf:0      Version 6.101
```

alerque:
Closes #1495


alerque:
Does this look right to you guys @matteodelabre and @Omikhleia? The refactor commits are close to being a noop functionally but they both actually add some missing `_init()` functions (note in penlight classes you should always redefine `:_init()` methods even if you make them pass-through to the parent or you break multiple level inheritance — likewise using `self:super()` *should* work but breaks multi-level inheritance in some Lua/Penlight version combinations so I try to avoid it).

The only change directly affecting the bug we setup to fix is 0279556. This doesn't address the general problem such as info nodes being discarded yet, but it seems like using `hbox` as a starting point for math nodes makes sense anyway.

alerque:
@matteodelabre If you have further fixes to this feel free to open a PR against [my branch here](https://github.com/alerque/sile/tree/un-disapear-math) or just link me to a branch and I'll cherry pick it, and don't worry about merge/rebase stuff I cana take care of that.

matteodelabre:
Thanks for taking care of this. The changes look good to me. Perhaps regression tests would be helpful here, I opened a PR to add some.

PR #1512, opened 2022-08-04
Fix and improve error reporting

### LINKED ISSUE #1507: Fix stack trace source locations
Some work on inputters seems to have broken some cases of the source locations reported in the stack trace. Better review some cases in each language, especially adding `\sil{}` input to Lua  input `\lua` input to SIL.

alerque:
Closes #1507

### DIFF
--- a/classes/base.lua
+++ b/classes/base.lua
@@ -324,7 +324,8 @@ function class:registerCommands ()
   self:registerCommand("include", function (options, content)
     local packopts = packOptions(options)
     if SU.hasContent(content) then
-      return SILE.processString(content[1], options.format, nil, packopts)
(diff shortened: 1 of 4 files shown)


alerque:
We don't have a test framework that works to capture CLI output very well yet, so this isn't getting tested properly. Here's the kind of pathological nesting I was messing around with:

```sile
\begin{document}
\font{
foo
\begin{raggedleft}
bar
\begin{script}
SILE.typesetter:typeset("baz")
SILE.processString([[
\em{qiz}
lua{print(6)}
\xml{<em>bob<lua>print(7);SU.error("bob")</lua></em>}
]], "sil")
\end{script}
\end{raggedleft}
}
\end{document}
```

PR #1519, opened 2022-08-06
Fixup little issues (mostly with compatibility shims) found in production

alerque:
The internal uppercase() function was exported using the deprecated
export mechanism, but lowercase() and titlecase() were tripping over
their own shim logic.

This fixes the shim logic but also exports them as package functions
which should be used going forward.


alerque:
Hmmm, since the last release CI has started deducing a bogus version number which is causing it to error instead of warn on some deprecations. For example `SILE v0.r4-g3b86f97` should be `SILE v0.14.0-r4-g3b86f97`.

PR #1524, opened 2022-08-10
Extend the infonode package to allow recording node positions

matteodelabre:
Knowing the position of an infonode—in addition to which page it landed on—might be useful for example to add “post shipout” content to the page that takes advantage of this information. In the test included in this PR, a polyline links together all the nodes of a given category after they have been positioned.

This patch adds a new `savepos` parameter to the `\info` command and makes the existing `value` parameter optional. When `savepos` is set to false for a node (the default), the command behaves exactly as it did before, i.e., its value is added directly to the `SILE.scratch.info.<category>` list. When the new option is activated, an object with the shape `{ value = <value>, x = <x-pos>, y = <y-pos> }` is added to the list, where `x` and `y` are the node’s coordinates on the page, and `value` is the node’s value, if it has any.


alerque:
This looks useful, but it also seems like a bit of bolt-on. For many use cases one is going to want the location of an *existing* node with actual content in it, not a virtual one stuffed in between the rest. Does attaching this to *info nodes* in specific make sense for a specific workflow or should we work on a package that extends the outputter to log the position of any node (also including `w` and `h` along with `x` and `y`)?

matteodelabre:
Thanks for the feedback!

> This looks useful, but it also seems like a bit of bolt-on. […] Does attaching this to _info nodes_ in specific make sense for a specific workflow […]?

I needed something like this for a pseudocode package I’m working on, where I wanted to draw some vertical rules from one line to another to mark blocks depth. Something like this (from the LaTeX algorithm2e package documentation):

Screenshot from 2022-08-11 21-04-59

The idea was to record the location where each block starts and ends, and then add rules during the `endpage` hook. Since `infonodes` already had the logic for collecting nodes on a page, it seemed fit for that task. But perhaps a separate package would make sense.

> For many use cases one is going to want the location of an _existing_ node with actual content in it, not a virtual one stuffed in between the rest. […] should we work on a package that extends the outputter to log the position of any node (also including `w` and `h` along with `x` and `y`)?

Being able to track the location and size of any node does seem useful as well. Perhaps something like a `\savepos[id=…]{<contents>}` tag that marks the nodes it contains to be tracked, and a way to be called back when those nodes have been placed on the page. Not sure how to handle the case where there is a page break somewhere inside that set of nodes.

Omikhleia:
> Not sure how to handle the case where there is a page break somewhere inside that set of nodes.

That were things get tricky.

While I understand the idea here, I would be very favorable to adding features to the infonode which are not demonstrated to be really necessary and the only way to proceed so as to implement another feature.

Omikhleia:
Unless mistaken, this is not going to happen the way this PR proposes.
After 1+ year of inactivity on the topic, I am closing it.
Please feel free to suggest another implementation and/or a 3rd-party package achieving that specific need.

Issue #583, opened 2019-04-22
Discressionaries in \href aren't counted towards width


alerque:
Our temporary hack for #355 has surfaced another bad side effect. Having such a discretionary character inside a `\pdf:link{}` causes SILE to miscalculate the width of the hbox and the layout of the rest of the line is off by the width of the apostrophe that was included (but not calculated in the width).

```TeX
\begin[papersize=a6]{document}
\script[src=packages/url]
\begin{script}
SILE.registerCommand("ah", function ()
  SILE.call("discretionary", { prebreak = "-", replacement = "’" })
end)
\end{script}
\language[main=tr]
\font[language=tr]
(\href[src=""]{Yasa\ah{}nın Tekrarı 19:21})
\end{document}
```

![image](https://user-images.githubusercontent.com/173595/56499512-62528200-650f-11e9-83ed-3d626f44eb19.png)

Oddly enough, another different breakage occurs in the event of an `\href` without the `src` attribute:

```TeX
(\href[]{Yasa\ah{}nın Tekrarı 19:21})
```

![image](https://user-images.githubusercontent.com/173595/56499614-c5441900-650f-11e9-9b5d-9a171af71370.png)

Either way the  width is miscalculated, but in one iteration the discretionary doesn't even get drawn until after the link!

Note the MWE is in Turkish because that's what I paired it down from and where it finds a real-world manifestation, but it's not actually specific to the language.

alerque:
c.f. #1362

It isn't just inside hrefs that is a problem, discretionaries aren't being added up right across a whole line.

Omikhleia:
First, assuming the `hbox` logic (currently used in `\href` via `pdf:link`) might be the culprit, I simplified the MWE as follows:

```
\begin[papersize=a6]{document}
\neverindent
\define[command=ah]{\discretionary[prebreak=-, replacement=’]}
\language[main=tr]

(\hbox{Yasa\ah{}nın Tekrarı 19:21}) with hbox

(Yasa\ah{}nın Tekrarı 19:21) without hbox

\end{document}
```

The issue is indeed observed:

![image](https://user-images.githubusercontent.com/18075640/227813014-a946fafe-8737-4678-8edd-228426b0b3c2.png)

Then, trying this fix (on the hbox building logic, from the revised implementation in #1702)

```
diff --git a/typesetters/base.lua b/typesetters/base.lua
index 22651ad0..07004095 100644
--- a/typesetters/base.lua
[...]

Issue #1367, opened 2022-04-16
Bable strikes back — chaos in language settings


alerque:
There are times I want to throw up my hands.

* Languages are messy.
* SILE's language settings are scrambled.

In working on #665 (see progress in #675) I've run into the fact that we don't handle settings sanely. We have the `document.language` setting which can be set directly, but far more often it gets set as a side effect of `\font[language=...]{}`. The trouble is actual SETTINGS are scoped with a stack system where temporary closures can bring them in and out of scope. These don't necessarily correspond to where other font settings are stopped and started. One current example is any use of the `\verbatim` environment between to `\chapter` headings will change the language use to localize  the "Chapter" text. The environment is supposed to be protected with a temporary settings block, but changing the font *and* changing the document language result in an imbalance in the settings stack.

Confounding the problem is the possibility of multiple typesetters. If I have one typesetter working on one frame in one language and another working on another frame in another language, the `document.language` setting is going to get cross-polluted constantly.

Personally I think the `document.language` setting is over 50% bogus. Having a document default might make sense, but it in that case it should not be a **running** value that changes constantly. It should be a fallback for when no other settings are in scope. Then each typesetting should probably have a language setting that follows the typesetter state, not the document settings state.

See also related discussion in #74 and #83 (mostly bidi, but some document expectations).

Omikhleia:
You seem to mention two separate issues: one with a case with a single typesetter, another with multiple typesetters. Do you have a minimal example for the first case? (As for multiple typesetters, things cant get more complex but I suspect it depends how one accesses them). (By the way, verbatim is a bit bogus, no?)

simoncozens:
You’re right, it’s a mess, and I apologise for that. A document should have a fixed default language, but the typesetter should know what language it’s currently working with. So we do need document.language and font.language, but only font.language should shuffle about.

alerque:
...and the setting state of font.language should follow the typesetter state not the document state. As should a lot of other settings.

Omikhleia:
... and fonts, intrinsically, shouldn't have a a language :p

alerque:
Hence why I mentioned in #1368 that we've been using "language" internally pretty much across the board. In some cases we really do mean "language", but in some places we actually mean "locale" and in others "script". Untangling them without breaking everything for everybody is going to be ... fun.

PR #1527, opened 2022-08-13
Discretionary glyphs vs. hyphenation

### LINKED ISSUE #355: Drop characters when hyphenating

alerque:
Assuming I can sort out concessionaires as hyphenation content...

Fixes #355
Does not touch #583 since that uses manual discretionaries and this uses hyphenation patterns.
Exacerbated by and might further exacerbate #1362, but will be more correct and make that bug easier to squash.

~~Not ready for merge yet (since it currently just blindly assumes all languages need this treatment) but some preliminary hacking on hyphenation for Turkish that handles apostrophes as discretionaries if they fall at hyphenation points (which they are by definition).~~


core/nodefactory.lua
```diff
@@ -276,6 +275,11 @@ function nodefactory.discretionary:toText ()
 end
 
 function nodefactory.discretionary:outputYourself (typesetter, line)
+  -- TODO this is an indication of a deeper bug. If we were asked to output
+  -- discretionaries but the parent nnode is not hyphenated then we're
+  -- outputting things that were only used for measuring "what if" scenarios in
+  -- break point calculations. These nodes shouldn't exist at this point.
+  if self.parent and not self.parent.hyphenated then return end
```
Omikhleia, line 282: Re: "These nodes shouldn't exist at this point": There's nothing (currently) to remove unused discretionaries. The last one in a line is marked as used (the next test below), in `computeLineRatio` which also sets the parent's hyphenated to true in that case.

alerque: I wasn't expected them to be *removed* necesesarily, but they are getting added at some points they should not be. For example somehow after typesetting a line we're ending up with hyphenation discretionaries *after* each word. A word with 3 hyphenation points that isn't even a candidate for hyphenation will end up with 3 discretionaries *after* the word (not in their places). The actual words near the end of a line that are hyphenation candidates based on having filled up a line and gone over-length will have the correct placement.

alerque:
This gets the job done, but one thing I'm not happy with is the blunt-force special casing for Turkish. AT the moment Turkish is the only language with hyphenation patterns with apostrophes but I have to imagine it shouldn't be. At the very least some other Turkic languages are likely to need similar treatment. I have to imagine there are other languages with...

* intraword apostrophes or other marks
* that can have segments longer than 2 letters on the right hand side
* and the marks need special treatment if landing on a (used) hyphenation point.

That being said we don't have a good ergonomic place to hang such a setting on yet. I suppose eventually a per-language setting and a couple of hooks are in order.

I think I'm going to move on without those things at this point because I'm not in a good position to come up with something ergonomic and just hacking one in before refactoring the setting scope (#1367 and friends) and making the typesetter, line breaker, and language support bits proper modules is going to materially slow down the line break loops for all languages for no gain.

core/nodefactory.lua
alerque: I split out my findings on this issue in #1528.

PR #1529, opened 2022-08-16
Add API so packages know their basedir and can load relative assets

alerque:
Some packages may bundle non-Lua resources (images, SIL, XML, etc).
Given the packages themselves are loaded with Lua's module loader using
`package.path` it can be tricky to then find the other assets. Rather
than letting ever package author re-invent this wheel this just gives
a little API that can be used any time later to recover the path where
the package was loaded from.

### DIFF
--- a/packages/base.lua
+++ b/packages/base.lua
@@ -8,11 +8,18 @@ package.class = nil
 -- For shimming packages that used to have legacy exports
 package.exports = {}
 
+local function script_path ()
+  local src = debug.getinfo(3, "S").source:sub(2)
+  local base = src:match("(.*[/\\])")
+  return base
+end
+
 function package:_init (_)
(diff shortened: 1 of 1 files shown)


ctrlcctrlv:
Good idea, this will be especially helpful for packages that call external programs, as I was thinking about doing for syntax highlighting. I actually have put my syntax highlighting debug functions up here:

https://github.com/ctrlcctrlv/debug.bashrc/blob/14e942358992fb30d29d22955f5037afe7a1efc8/debug.bashrc#L15-L45

The external program can be bundled via `git submodule` this way :-)

alerque:
To go along with that also note the little trick I did in [the sile-logos rockspec](https://github.com/ctrlcctrlv/sile-logos/blob/master/sile-logos-dev-1.rockspec) for getting LuaRocks to *install* random non-Lua files inside the same directory as a module.

PR #1538, opened 2022-08-25
feat(packages): Add optional brackets to url.lua

ctrlcctrlv:


packages/url/init.lua
```diff
@@ -147,6 +153,12 @@ function package:registerCommands ()
         secondaryPenalty = secondaryPenalty,
         worsePenalty = worsePenalty
       })
+      local brackets = SILE.settings:get("url.brackets")
+      brackets = SU.splitUtf8(brackets)
+      if #brackets == 2 then
+        result[1] = brackets[1] .. result[1]
+        result[#result] = result[#result] .. brackets[2]
+      end
```
Omikhleia, line 161: FWIW, another possibility, rather than using settings, would be to call a "urlstyle" hook command (name based on "foliostyle"), instead of directly calling "code" at line 162 just below. The default implementation would just could "code" as of today, but that would allow easily changing it:
- Allowing one to use other font choices (or possibly none - my own use case, which I am solving by temporarily saving and changing "code", which is awkward and cumbersome...)
- Allowing one add brackets (your own use case here)
- Or even allowing other things for other weird use cases (e.g. we could pass the url options through to this urlstyle command, and e.g. people could come with version of it that add small pictures based on a type parameter, etc.)

packages/url/init.lua
```diff
@@ -147,6 +153,12 @@ function package:registerCommands ()
         secondaryPenalty = secondaryPenalty,
         worsePenalty = worsePenalty
       })
+      local brackets = SILE.settings:get("url.brackets")
+      brackets = SU.splitUtf8(brackets)
+      if #brackets == 2 then
+        result[1] = brackets[1] .. result[1]
+        result[#result] = result[#result] .. brackets[2]
```
Omikhleia, line 160: This is wrong / not robust, breaking current behavior:
It's far too bold to assume that `result[#result]` is a string. Here it can be a penalty, since this is invoked _after_ the inputfilter has been applied. You do need to add a test scenario here, to check what happens if the URL ends with a "/" or other characters after which an extra penalty might have been inserted... I did a quick check and it oops:
```
error summary:
	Processing at: experiments/url.sil:4:5: in \url
	Using code at: ./packages/url/init.lua:150: attempt to concatenate a table value (field '?')
```
So `result[#result+1] = bracket[2]` would be the solution here...

But that could be true to for the first element, it a penalty was inserted _before_ the actual text. Currently it only occurs for URL fragments, so it's unlikely at start position, but for the sake of robustness you should probably handle it too... 

... Because the content provided by the user might not even be a string! E.g. all of the following URL currently works:

```
\url{\img[src=google.png] http://google.com/}
\url{\strong{https:}//google.com/}
\url{\underline{https://google.com}}
```
Whether the above is fully legit or not, it does work as of today... And breaks when you start to consider the content is a string.

ctrlcctrlv: Thank you. I wasn't too sure about this one. I appreciate the thorough review. Your comment above seems a better solution than mine and I'll implement it. :+1:

ctrlcctrlv: (The `urlstyle` comment in case not clear.)

packages/url/init.lua
ctrlcctrlv: Very good idea. I often want to make all my links blue for example.

packages/url/init.lua
Omikhleia: You are welcome. I actually stumbled on the `\code` stuff in the url package in #1194 (that's one of the unaddressed tick box at this date), but I didn't knew as much at the time. Would it be now, by the way, I'd propose to consider moving the default definition of "code" to the plain class (besides "strong" and "em") - it's used in plenty of places (manual, package docs, markdown, etc.) and while assuming the url package to be generally loaded is quite reasonable for most scenarios, it still does feel out of place, IMHO.

packages/url/init.lua
alerque: I agree a style wrapper function would probably be better for this. That would give a lot more flexibility. Even start/end brackets may not always be a single character, what if wanted to include a thin space for example? A style wrapper that by default is a passthrough sounds like the best plan to me.

packages/url/init.lua
alerque: I agree `\code` is out of place in the *url* package and should probably be moved to the *plain* class.

A minor side thought is that we don't currently request a monospace font be installed by packages. We do request Gentium Plus as default and we also suggest Libertinus Math (math package default) and Noto Sans CJK JP (*jplain*/*jbook* class default) but we don't similarly suggest Hack (verbatim package default).

A default mono font family should probably be a top level setting and that wrapper function in verbatim should use the setting instead of setting it directly.

packages/url/init.lua
```diff
@@ -46,8 +46,11 @@ function package.declareSettings (_)
 end
 
 function package:registerCommands ()
-
   self:registerCommand("href", function (options, content)
+    if (not content or #content == 0 and not content[1]) and options.src then
```
Omikhleia, line 50: This going to break I think, if content is not an AST table but (for instance) a function - e.g. when this is used from La code as:
```
SILE.call("href", { src = "..." }, function ()
   -- Some logic formatting the content
end)
```
This would now break with  `attempt to get length of 'content' (a function value)`, I'm afraid.
We already support
- `\href[src=URL]{Some content}`
- `\href{URL}
Not sure we should try something to `\href[src=URL]` (which I guess is the idea here), but be conscious its a bit tricky.

Omikhleia: Anyway, if we go for it, then use `SU.hasContent(content)` - it's the way to go I guess.

packages/url/init.lua
```diff
@@ -147,11 +150,12 @@ function package:registerCommands ()
         secondaryPenalty = secondaryPenalty,
         worsePenalty = worsePenalty
       })
-      SILE.call("code", {}, result)
+
+      SILE.call("urlstyle", {}, result)
     end)
   end, "Inserts penalties in an URL so it can be broken over multiple lines at appropriate places.")
 
-  self:registerCommand("code", function (options, content)
```
Omikhleia, line 154: So we need to move `code` to the plain class, or plenty of packages documentation will break, etc. It's non visible in the manual, because it redefines it in its macros, but several packages make use of it.

I wonder whether `urlstyle` should call that `code` or directly `verbatim:font` (which is from the **verbatim** package, which url loads... Hmm that makes a possible reason to call it directly here.. The implementation of code becomes trickier, though... Er... Maybe `verbatim:font` should _only_ be used in the context where verbatim is actually really used (One can imagine a different font being used for code in text flow vs. code in verbatim blocks).

tests/url.sil
```diff
@@ -0,0 +1,10 @@
+\begin{document}
+\nofolios\noindent
```
Omikhleia, line 2: `\neverindent` in tests so it kills it all, I think.

tests/url.sil
```diff
@@ -0,0 +1,10 @@
...
+\script[src=packages/url]
```
Omikhleia, line 3: Be in 0.14 mood: `\use[module=packages.url]`

tests/url.sil
```diff
@@ -0,0 +1,10 @@
+\begin{document}
```
Omikhleia, line 1: In tests, papersize a6 for small tests, @alerque once told me (default is a4). Quoting his words from another review:
> Another test detail is I really prefer tests to use small pages where possible. I sometimes generate PDFs of everything and make a montage of them for review. Big A4 pages waste a lot of space and make these hard to review. A6 for singe pages tests and A7 for multi-page tests are preferred if possible. This way the default font sizes make the tests easier to review en masse.

tests/url.sil
```diff
@@ -0,0 +1,10 @@
...
+\nofolios\noindent
+\script[src=packages/url]
+\href[src=https://sile-typesetter.org]{Test.}
+\medskip
+\href{https://sile-typesetter.org}
+\medskip
+\define[command=urlstyle]{<\verbatim:font{\process}>}
+\href{https://sile-typesetter.org}
+\end{document}
```
Omikhleia, line 10: No test covering the `\href[src=...]` case introduced above (if we stick with it)?

Omikhleia:
Regardless of the details, the PR has switched subjects, it's not longer "Add optional brackets to url.lua" but rather something such as "Add style hook to the url package" (or to that extent).

packages/url/init.lua
```diff
@@ -46,8 +46,11 @@ function package.declareSettings (_)
 end
 
 function package:registerCommands ()
-
   self:registerCommand("href", function (options, content)
+    if (not content or #content == 0 and not content[1]) and options.src then
+      content = {options.src}
```
Omikhleia, line 51: So this would lead to have both a content and a src, which in this case will bypass the inputfilter for extra breakpoints. Shouldn't be `options.src` set to nil if this is moved to the content, so that your new `\href[src=URL]` becomes really identical to `\href{URL}`?

Omikhleia: Additionaly, what happens if one just calls `\href`? (no content, no src option)

packages/url/init.lua
ctrlcctrlv: @Omikhleia Are you OK with me dropping it? It's too complex to keep I  think and is a holdover from when I was still trying to put the feature into `\url` and not `\href`.

Omikhleia: Sorry for the long delay. No problem with me for dropping `\href[src=...]` without arg. As I said, I wasn't even sure it was needed in the first place - And indeed you are right, this is a different topic. It will be cool enough to get the hook so one can have, say, color-bracketed URLs in upright font :)

Omikhleia:
@ctrlcctrlv Just to give a heads up - Are you still working on this topic?

Omikhleia:
Thanks for the exchanges and incentives to address URL styling.
In 0.14.9, we went for #1709 as a way to alter the style the URL, based on some discussions above.
--> Closing this PR, as proposed change now overlap.
Feel free to propose other improvements if felt necessary.

PR #1540, opened 2022-08-26
fix(justenoughlibtexpdf): pdf_metadata

Pi-Cla:
The function is non-void, but it doesn't return anything in the end. Without this, gcc complains with a "return reaches end of non-void function" warning, which is treated as an error on certain Linux distros such as openSUSE.

### DIFF
--- a/src/justenoughlibtexpdf.c
+++ b/src/justenoughlibtexpdf.c
@@ -331,6 +331,7 @@ int pdf_metadata(lua_State *L) {
   texpdf_add_dict(p->info,
                texpdf_new_name(key),
                texpdf_new_string(value, len));
+  return 0;
 }
 /* Images */
 


src/justenoughlibtexpdf.c
```diff
@@ -331,6 +331,7 @@ int pdf_metadata(lua_State *L) {
   texpdf_add_dict(p->info,
                texpdf_new_name(key),
                texpdf_new_string(value, len));
+  return 0;
```
ctrlcctrlv, line 334: looks fine.  normally the "right" solution would be to just make the function void, but most in the file are int.

Omikhleia: Well caught, actually! The integer value returned by these C wrappers is the number of Lua values pushed on the stack (= returned to the Lua calling context[^1]) so returning 0 is clearly the right/proper solution here.

[^1]: As per the definition of `lua_CFunction`: "To return values to Lua, a C function just pushes them onto the stack, in direct order (the first result is pushed first), and returns the number of results."

ctrlcctrlv:
@Pi-Cla I force-pushed your branch due to the commitlint failure but won't merge as I've never done that for this project and so even though Caleb was nice enough to make me able to I don't feel ready to :-) but can at least save him some time making easy fixes to user patches. Thanks for contributing.

src/justenoughlibtexpdf.c
ctrlcctrlv: figured it was something like that. read a rule once in a programming book. "Assume the programmer knows what [they're] doing unless given good reason to think otherwise." definitely applies here; no one who knows what they're doing would've repeated so many seemingly useless return's of 0 for no reason. :-)

ctrlcctrlv (approved):

alerque (approved):
Thanks for the contribution!

PR #1541, opened 2022-08-26
Rework simple and multilevel counters

### LINKED ISSUE #1536: Multi-level counter pitfalls (counters package)
(For the record, sharpening counters was evoked in #455)

When working on my book class (and a pending article class to go with it), I quickly hacked the multilevel counters. Since I am refactoring it for 0.14, it's perhaps time for me to report my findings / issues. (I can make a PR with my changes so far, just mentioning the main points below, some could be questionable.)

- `formatMultilevelCounter()` issues
  - It behaves quite badly if the (optional) level parameter is greater than the current level (robustness)
  - An option for skipping _leading_ zeroes would really be handy for classes that would accept, e.g. unnumbered chapters to have sections naturally starting at 1, **not** 0.1 (and so on).
  - BTW, I would question the "minlevel" option, what's the use case for it? Having a counter "1.2.3.4", is one really expecting a use case for formatting, say, "2.3"? ~~For the sake of simplicity over over-design, I'd suggest removing it.~~ Bah, could be kept anyway. Just weird maybe.

- `\set-multilevel-counter` missing
  - For single-counters, we have `\set-counter`, so it sounds logical having it here too.
  - The class may need to reset sectioning counters to some level (e.g. entering an appendix, resetting chapter level to 1, with alpha display, etc.)

- `\show-multilevel-counter`
   - It claims "optionally displaying it with the `<display>` format", but this is really confusing as:
     - These are multilevel counters where all levels may have different display format
     - The option currently acts on the current level, without regards for the requested level.
     - The option (as for the `\show-counter` too) does not only use that display for showing, it also _alters_ it (so this is still changed afterwards).

Omikhleia:
Closes #1536 - I suffered a bit on this one, so these are at least some proposals for review and scrutiny...

Things proposed here:
- Robutness checks added to counters:
  - `SU.cast()`, `SU.boolean()` checks, etc.
  - for multilevel counter, checks the passed level(s) vs. maximum (i.e. current) level.
- New options:
  - Add option `noleadingzeros=true` to multilevel counter formatting, to possibly skip leading zeros
- New commands:
  - `set-multilevel-counter`
- Proposed deprecations (the versions might have to be adapted)
  - `\increment-counter[set-to=...]`: it was not documented, and an increment that does a set is weird. We already have `\set-counter` for that, let's keep things simple, no?
  - `\show-counter[display=...]` and `\show-multilevel-counter[display=...]`: these are actually _altering_ the counter, this does not seem expected on a "show" command which shouldn't have side effects, and we can already change the display format in a much more understandable way with the "set" or "increment" commands. Again, let's keep simple and clear. A "show" just shows without side-effects.
- Potential breaks:
  - I tried to make sense of the `reset` undocumented option on `increment-multilevel-counter` but it was pretty broken (it could not be used from SIL-language, as it was lacking a cast) and what it did when the passed `level` is greater than the current level did not seem right.
  - In the same case (level > current level), what it did was dubious (e.g. 1.3 set at level five would yield 1.3.3.3.1... I can't understand the rationale for repeating the previous levels); I changed it to something that seems more predictive (1.3.0.0.1), **but** this is breaking some tests. Are these really correct, though?

Things not yet done, for the record:
- Package documentation for the mutilevel counters. Will do, if we reach consensus regarding the above.
- Add test cases. _Idem_

Another unaddressed point I'd like to raise here: it doesn't seem logical to me to export some functions at class level (e.g. `class:getCounter`, while some aren't (well, still are, but deprecated, e.g. `class:formatCounter`). There's some hard-to-understand asymmetry.


packages/counters/init.lua
```diff
@@ -42,9 +42,14 @@ function package.formatCounter (_, counter)
 end
 
 function package:formatMultilevelCounter (counter, options)
-  local maxlevel = options and options.level or #counter.value
-  local minlevel = options and options.minlevel or 1
+  local opts = options or {}
```
alerque, line 45:
```suggestion
  options = options or {}
```

... and rename `opts` → `options` throughout. Unless there is a reason to use a different variable throughout I'd rather just use the table passed rather than a copy with a similar name.

alerque (dismissed):
I think this all makes sense.

Re tests: I would suggest Busted based unit tests rather than full page layout regression tests since we're after the math here more than the resulting typesetting.

Re class scope: The exporting stuff is mostly to support v0.13 classes and packages that didn't have another way to access functions from other packages. Now in v0.14 we do and it's possible we could eventually deprecate those exports entirely. For now I would leave the ones we have but encourage not adding any more unless absolutely necessary. Packages can now access other packages methods directly via `self.class.packages["other-package"]:method`. We could possibly even make that more ergonomic, but for now lets keep an eye on what we need it for so we know what to do with the existing export system.

The one caveat is exporting functions make it easy for them to access the *class* as `self` rather than the package, which might be a good reason to keep it for certain senarios.

packages/counters/init.lua
Omikhleia: > Re tests: I would suggest Busted based unit tests rather than full page layout regression tests since we're after the math here more than the resulting typesetting.

That makes perfect sense. I'll work on it, then, and update the PR.

> Re class scope: (...)
Yep, points noted. For the record about my "asymmetry" concern, I now have code (after replacing the corresponding deprecated `SILE.xxx` previous calls) such as:
```
local n = self.class.packages.counters:formatMultilevelCounter(
   self.class:getMultilevelCounter(counter)
)
```
I would have _kind of_ expected, either:
```
local n = self.class:formatMultilevelCounter(
   self.class:getMultilevelCounter(counter)
)
```
Or
```
local n = self.class.packages.counters:formatMultilevelCounter(
   self.class.packages.counter:getMultilevelCounter("mycounter")
)
```
With _probably_ a preference for the second form - The first makes it sound, at least at first glance, that it could be harder for one to use another "counter" package (compatible at API level, but perhaps with additional options and methods, etc.) instead of the stock one.

BTW, this does raise another possible idea: to accept either a counter object _or_ a counter id in methods such as `formatMultilevelCounter`, so the above could become, without loosing any current functionality:

```
local n = self.class.packages.counters:formatMultilevelCounter("mycounter")
```
I'm not sure how such methods accepting different types for an argument are regarded here, though. (Developers' tastes sometimes differ on such an approach)

alerque: I'd like to move *away* from functions and methods that accept multiple types. They will be the hardest parts to substitute in Rust or Teal or whatever in the future. I know there are still a lot of use cases for this and some of them are even quite useful. The gray area here is where type casting comes into play. Any type enforcement we implement is going to have to account for the cases where we can cast to the right type. For example anything coming from SIL input is going to be a string, but depending on the command we cast it no numbers or lengths or whatever. So the long term rule will be "only accept type X or anything that can be cast to X". Lua doesn't handle this very gratefully, but Rust for example has features that can be used to automatically handle that kind of situational conversion. The difference is it is backwards from Lua: the function doesn't test for a string and then try to convert it to the right type, the string would have to have a trait that enabled it to be cast to something else.

I'm kind of torn on this. I actually typed up my fist answer as "I suggest adding a similar method that accepts an id, fetches the counter, then passes it to the current method such as `:formatMultilevelCounterId()`", but given all the existing precedent I wouldn't object if you thought just implementing it a test for string vs. counter object and used strings to fetch counters by ID.

Omikhleia:
Updated:
- Added a bit more robustness (e.g. `SU.required()` and check on type of counters)
- Added multilevel counters to package doc
- Added busted tests. (N.B. There was no existing package, AFAIK, with busted tests, so I hope I did it the right way)
- Fixed existing text expectations.

Omikhleia:
> Omikhleia dismissed alerque’s stale review via e691329

Hmm. Not sure what I did here, it was unintended at best^^ Maybe the rebasing?

alerque (dismissed):
Looks great to me!

alerque:
>     * Added busted tests. (N.B. There was no existing package, AFAIK, with busted tests, so I hope I did it the right way)

Yup that's exactly what I was thinking.

Also need to make sure that sort of layout works for 3rd party packages so external repos can be tested the same way without too much fuss. I'm guessing the easiest way to do this will be with `nix`. Then we should pretty easily be able to test a 3rd party module against an array of SILE versions (via tagged flakes) and also have busted on hand with whatever version(s) of Lua the 3rd party module supports.

alerque:
I rewrote the commits (via advanced `git rebase` and `git revise` ninja foo) to split them up a little — mostly for the sake of the change log. In general, `refactor:` should indicate a different way to code exactly the same function and I have those commits suppressed in the release notes. In this case the overhaul changed the expected output (for the better) and also added options and functions, so I used `fix:` and `feat:` so that those changes show up automatically in the release notes. I didn't change any code, just mucked with the commit sequence.

Issue #1283, opened 2021-11-12
open-double-page is (almost) broken


Omikhleia:
SILE 0.12

The "open-double-page" command from the "twoside" package doesn't work well.
- If the document starts with a chapter directly (so you are already on an odd page)... Then one gets blank initial pages.
- Likewise if the previous content within the book already ejected the page (e.g. because one needed to vfill something at the bottom properly), one gets extra blank page.

But having it inside the "twoside" package might also be a wrong separation of concerns: As implemented, the "blank" page that will get inserted if needed will likely get page numbers and headers. This is at the least controversial, and should be a (book) class decision - so perhaps to consider there https://github.com/sile-typesetter/sile/projects/2#card-42057180

_As an aside note_, in my in-progress book class I ended up not using it. The re-implementation inside that book class cannot just be reused as-is, though, as it relies on another feature (= mostly, an abstraction of the headers as we have it for the folios, with "headers / noheaders / noheaderthispage"). Commenting that part, it comes along the lines of:

```
SILE.registerCommand("omibook-double-page", function (_, _)
  SILE.typesetter:leaveHmode() -- Important, flushes output queue.
  if #SILE.typesetter.state.outputQueue ~= 0 then
    -- We are not at the top of a page, eject the current content.
    SILE.call("supereject")
  end
  SILE.typesetter:leaveHmode() -- Important again.
  if not SILE.documentState.documentClass:oddPage() then
    SILE.typesetter:typeset("")
    SILE.typesetter:leaveHmode()
    SILE.call("nofoliosthispage")
    -- SILE.call("noheaderthispage") -- Commented 
    SILE.call("supereject")
  end
  SILE.typesetter:leaveHmode() -- and again!
end, "Open a double page without header and folio")
```

When I see all the tasks listed in the "Redesign the book class" project, I am not sure where it will go.

alerque:
Duplicate of #454.

----

My current implementation is at:

https://github.com/sile-typesetter/casile/blob/6529e93163add0ec9c2ff87dee52b437595108b7/casile.lua#L213-L232

PR #1547, opened 2022-08-31
Twist out some odd-page oddities

### LINKED ISSUE #454: \open-double-page is overzealous
And forces a pagebreak even when it doesn't need to.

alerque:
Fixes #454

Upstreams `\open-spread` from CaSILE. Just replacing `\open-double-page` didn't seem like a good idea because it would be a breaking change for existing documents (because of the automatic suppression of headers and folios). This might be expected anyway, but it is different enough I decided to just provide it as an alternative.


alerque:
@Omikhleia Any input on whether suppressing headers and folios by default on `\open-double-page` would be expected or unexpected behavior? We could make it an option to turn off ... or we could just go with my current commits that leave it alone and add a new function `\open-spread` instead. I know my use cases always expect *blank* pages not *empty* pages with headers and page numbers when stuffing to get to an odd or even layout, but I didn't want to presuppose that applies to all/most use cases.

alerque:
What the blazing boondocks ... *none* of this works as expected. I started working up a stress test and neither the old `\open-double-page` nor the patched edition nor `\open-spead` work as expected in all situations. Bah humbug.

alerque:
Okay that was all just because I dropped an important line along the way. I think the test does everything I expect now out of both commands.

alerque:
Nope, not quite. But at least I worked out an MWE for the original issue!

alerque:
For real this time. And the test actually tests the already overfull frame test case that was causing too aggressive page outputs. Both old and new functions handle that case now.

Omikhleia:
> Any input on whether suppressing headers and folios by default on \open-double-page would be expected or unexpected behavior

AFAIK, I would say it depends - that's why I suggested in this [comment](https://github.com/sile-typesetter/sile/issues/1283#issue-1051933941) that a general-purpose command currently proposed (but undocumented) in the "twoside" package is perhaps not even achievable and that may such decisions are probably a matter of class-design. Hence my suggestion that the command, since it's currently undocumented, could be moved to the book class -- or perhaps, nowadays, to a dedicated package: it might not necessarily even be depending on being in twoside mode (i.e. assume different masters on odd and even pages).

An _example_ of design amongst other could be:
1. blank pages after a `\part` may be wanted without folios and header...
2. blank pages before a `\chapter` (unless it is the first of a part)
    - May have folios
    - May have headers _or not_...
       - And when the do, this could be either the header of the previous chapter (the one we just completed) or the current one (which triggered the blank page and has it facing) - but that also depends on what the right/left headers are intended to contain (i.e. what the class allows for them).

But _another design_ could also want those "blank" page to actually have content, e.g. some generic graphic layout to fill the place, or a place anyway where stock images could be deferred.

The problem (IMHO) is not really to provide a command that inserts such page breaks in a way that cannot possibly be fully generalized. It's rather to provide a way to check if such a break is necessary or not, i.e. whether we are at the top of an empty page or not (`\ifattop` doesn't seem - last time I checked - to really work properly for this), and where we would land. (The weird hbox trick in the current PR hints that this might not be that obvious)

packages/twoside/init.lua
```diff
@@ -65,28 +66,74 @@ function package:_init (options)
...
   end)
 
 end
 
 package.documentation = [[
 \begin{document}
 The \code{book} class described in chapter 4 sets up left and right mirrored page masters; the \autodoc:package{twoside} package is responsible for swapping between the two left and right frames, running headers and so on.
-It has no user-serviceable parts.
+It's main function in mirroring master frames does not provide any user-serviceable parts.
+It does supply a couple user facing commands for convenince.
```
Omikhleia, line 123: "convenience" (typo)

packages/twoside/init.lua
```diff
@@ -65,28 +66,74 @@ function package:_init (options)
...
+
+The \autodoc:command{\open-double-page} ejects whatever page is currently being processed, then checks if it landed on an odd page.
```
Omikhleia, line 125: Regarding the discussion whether to put this in a book class, but the lack of class documentation mechanism:
Why not have these commands in a standalone package, loaded by the book class. This way it could have documentation, and I guess be used even if twoside is not active?

packages/twoside/init.lua
```diff
@@ -27,18 +27,14 @@ local mirrorMaster = function (_, existing, new)
...
 local function switchPage (class)
-  if class:oddPage() then
-    tp = "even"
-    class:switchMaster(class.evenPageMaster)
-  else
-    tp = "odd"
-    class:switchMaster(class.oddPageMaster)
-  end
+  _odd = not class:oddPage()
+  local nextmaster = class[(_odd and "odd" or "even") .. "PageMaster"]
```
Omikhleia, line 36: `local nextmaster = _odd and class.oddPageMaster or class.evenPageMaster` could be more legible, or did I miss something?

packages/twoside/init.lua
```diff
@@ -65,28 +66,74 @@ function package:_init (options)
...
+      return (not double or spread_counter > 1) and
+             (odd and self.class:oddPage()) or
+             (not odd and not self.class:oddPage())
     end
+    spread_counter = 0
     SILE.typesetter:leaveHmode()
+    -- Output a box, then remove it otherwise we can't prove what page new content will land on
+    SILE.call("hbox"); SILE.typesetter:leaveHmode(); table.remove(SILE.typesetter.state.nodes)
+    if spread_counter == 1 and optionsMet() then
+      SILE.typesetter.state.outputQueue = {}
```
Omikhleia, line 100: Why do we need this? Can we investigate what's left in the outputQueue (vglues, perhaps) and document the rationale?

tests/bug-1280.sil
```diff
@@ -2,7 +2,7 @@
 \font[filename=.fonts/NotoSerifCJK-Regular.ttc]
 \nofolios
 \neverindent
-\script{
+\begin{script}
```
Omikhleia, line 5: BTW, wouldn't we use `lua` rather than `script`? (That was my understanding seeing `lua` being added, but not so quite sure about it after all.)

alerque: 99% honestly? Because I haven't updated vim-sile with syntax highlighting for it yet and since we aren't throwing any deprecation warnings yet I am not in a hurry to switch. The other 1% is because this is an old test file I was just updating the style for readability not refactoring the API surface area.

packages/twoside/init.lua
alerque: I've expanded the comment to explain. This is for an edge case where the page is busting at the seams already but `:leaveHmode()` won't trigger a page break because of widow penalties. Adding the hbox here *after* having left hmode forces a line to be added and hence the next tme we leave hmode *will* break the page. For this edge case the *only* thing in the output queue will be the extra line we just added.

packages/twoside/init.lua
alerque: Roger that. I was trying to do something else here but when it didn't work I forgot to go back and simplify again.

packages/twoside/init.lua
```diff
@@ -65,28 +66,74 @@ function package:_init (options)
...
+      return
+    end
+    repeat
+      if spread_counter > 0 then
+        SILE.call("hbox")
+        SILE.typesetter:leaveHmode()
+        if blank then
+          SILE.scratch.headers.skipthispage = true
```
Omikhleia, line 108: Hmm wait! Who's defining this? We don't have a "headers" package abstraction as we have a "folios" one (N.B. I did a quick one in my own bundle...). This comes from the book class, right? But doesn't it mean that any other book class _not even inheriting the standard book class_ has to respect some unspecified scratch thing if it wants to use this command? It does seem to me there's a serious separation of concerns issue ;-)

alerque: We already stuff the entire contents of R/L headers in a scratch variable, I'm not concerned about a scratch variable that suppresses the same without erasing it—at least not more than I think **all** the scratch variables should be refactored to be scoped to the class/package/typesetter/whatever. Yes anything wanting to muck with the existing book stuff would need to use the save variables. In this case some other header system could ignore this just fine, but then would have to implement their own way to suppress headers in spreads if they want that feature.

packages/twoside/init.lua
```diff
@@ -65,28 +66,74 @@ function package:_init (options)
...
+          SILE.scratch.counters.folio.off = 2
```
Omikhleia, line 109: Again breaking separation of concerns. We have `\nofolios`, `\nofoliothispage` etc. but this package hits directly into scratch variable. Maybe my soon-to come re-implementation of folios won't have this scratch variable, though be compatible at command level? How much internal knowledge of scratch things has to be know by class & packages designers?

packages/twoside/init.lua
alerque: Because I'm trying **not** to refactor anything that would be a breaking change here, I'm trying to fix a bug so I can publish a book with chapter headings on odd pages.

I 100% agree refactoring this elsewhere makes sense, but I don't have the bandwidth to figure out whether the class or another package makes more sense or how to do that without it being a breaking change. Since it was already in this file and the version I'm submitting has bugs fixed and gets things on the right pages (which the original does not) I'm going with it for now.

packages/twoside/init.lua
alerque: Good call on using the public facing folios command. I switched to that since it exists. When we have a book class with proper public API for handling headers we can switch to that too, but we don't right yet and it's out of scope for this PR.

packages/twoside/init.lua
alerque: I agree on the refactoring needed, and also have implemented my own header system at least 3 times without submitting any iteration to SILE because it's easy to work up what a project needs but a lot harder to generalize robustly.

PR #1573, opened 2022-09-14
fix(math): Fix insertion order of MathML children

OlivierNicole:
I noticed a strange behaviour when typesetting math tables with 11 rows or more:

```
\begin[class=plain]{document}
\script[src=packages/math]

\math[mode=display]{
  \table{
    0 & 0 & 0 \\
    1 & 1 & 1 \\
    2 & 2 & 2 \\
    3 & 3 & 3 \\
    4 & 4 & 4 \\
    5 & 5 & 5 \\
    6 & 6 & 6 \\
    7 & 7 & 7 \\
    8 & 8 & 8 \\
    9 & 9 & 9 \\
    10 & 10 & 10 \\
  }
}
```

Some of the rows are permuted non-deterministically, depending on the execution… it turns out that I make the mistake of relying on the iteration order of `pairs()` when compiling to MathML. This PR fixes it.

### DIFF
--- a/packages/math/texlike.lua
+++ b/packages/math/texlike.lua
@@ -124,10 +124,10 @@ local objType = {
   str = 2
 }
 
-local function inferArgTypes_aux (acc, typeRequired, body)
+local function inferArgTypes_aux (accumulator, typeRequired, body)
   if type(body) == "table" then
     if body.id == "argument" then
-      local ret = acc
+      local ret = accumulator
       table.insert(ret, body.index, typeRequired)
       return ret
     elseif body.id == "command" then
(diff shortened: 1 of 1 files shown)


alerque (dismissed):
This looks good to me, the bug fix is a pretty straightforward one. That being said it took me a while to sort out what was going on and why it mattered, so while I was there I refactored the code a little bit to:

* Use longer descriptive variable names would have put me on the right track sooner.
* Use `pl.utils.kpairs()` instead of `pairs()` with a custom check for number keys. It's a touch more robust and just saves re-inventing the wheel.
* Dropped an argument that was never used.
* Changed the argument order on the folder function to accept `(val, key, ...)` argument order. This is counter intuitive when many things return `key, val`, but since `key` may or may not exist many Lua tools take arguments in the opposite order. This doesn't matter *yet* but it would open the door to greatly simplifying this code later with `pl.tablex.map()` or `pl.tablex.foreach()` or `pl.tablex.pairmap()`. Other non Penlight tooling does the same thing with argument order for mapping functions, so I think this is a reasonable way to setup these sort of callback functions by default.

OlivierNicole:
Thank you. I agree with the changes, except maybe the function name `folder` which i fear may evocate filesystem folders.

alerque:
> Thank you. I agree with the changes, except maybe the function name `folder` which i fear may evocate filesystem folders.

Fair enough. I was a bit sketchy on what it even did when renaming it. Do you have a suggestion?

OlivierNicole:
What about something that describes the effect, like `compile_and_insert`?

alerque (approved):

PR #1582, opened 2022-11-02
s/--deps-only/--only-deps/

snan:


alerque (approved):
Thanks for the contribution.

It looks like `--deps-only` got [added as an alias in 3.4.0](https://github.com/luarocks/luarocks/blob/master/CHANGELOG.md#whats-new-in-luarocks-340) but I'm happy to use the more canonical variant in our docs.

alerque (changes requested):
The manual needs an update to match. I'll be happy to patch that onto this PR so don't worry about it.

snan:
I already have sent a separate PR for the manual:

https://github.com/sile-typesetter/sile/pull/1583

alerque (approved):

PR #1583, opened 2022-11-02
path to sile + same only-deps change from other PR

snan:


alerque (approved):
I dropped the unrelated luarocks change that belonged in the other PR and extended this one to cover another instance of the same thing later in the manual.

Good catches and thanks for the contributions.

PR #1584, opened 2022-11-02
Paper size present documentation

snan:
It's confusing to me that it lists a4, b5 etc followed by the word also then a list that... includes those aforementioned A/B/C ISO sizes.

### DIFF
--- a/documentation/c03-input.sil
+++ b/documentation/c03-input.sil
@@ -25,8 +25,8 @@ To manually change the paper size, an optional argument may be added to the docu
 \line
 \end{verbatim}
 
-SILE knows about the ISO standard A, B and C series paper sizes using names like a4 and b5.
-Here is a list of presets for traditional sizes:
+SILE knows about the ISO standard A, B and C series paper sizes using names like a4 and b5 as well as many other traditional sizes.
+Here is a full list of papersize preset names:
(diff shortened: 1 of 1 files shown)


documentation/c03-input.sil
```diff
@@ -25,8 +25,8 @@ To manually change the paper size, an optional argument may be added to the docu
 \line
 \end{verbatim}
 
-SILE knows about the ISO standard A, B and C series paper sizes using names a4, b5, etc.
-Also available are presets for some traditional sizes:
+SILE knows about the ISO standard A, B and C series paper sizes using names like a4 and b5.
+Here is a list of presets for traditional sizes:
```
alerque, line 29:
```suggestion
SILE knows about the ISO standard A, B and C series paper sizes using names like a4 and b5 as well as many other traditional sizes.
Here is a full list of papersize preset names:
```

alerque (changes requested):
Good catch, that was definitely not good wording. I would even take the edits a step farther. Does this suggestion sound clear to you?

documentation/c03-input.sil
snan: Yes, that's much better than what I suggested 👍🏻

PR #1585, opened 2022-11-02
Remove misleading text about latin hyphenation

snan:
It said that you turned off the latin hyphenation but the rendered text in the manual as downloadable from your homepage does have that latin text with some hyphens.

### DIFF
--- a/core/font.lua
+++ b/core/font.lua
@@ -19,8 +19,8 @@ SILE.registerCommand("font", function (options, content)
   if options.variant then SILE.settings:set("font.variant", options.variant) end
   if options.features then SILE.settings:set("font.features", options.features) end
   if options.direction then SILE.settings:set("font.direction", options.direction) end
-  if (options.language) then
-    if icu and icu.canonicalize_language then
+  if options.language then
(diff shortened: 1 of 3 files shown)


snan:
Sure, getting to the bottom of that bug is another option. I decided against it since it didn't serve the purposes of that particular section 💁🏻‍♀️

(And also because I didn't know how to fix the bug 🤷🏻‍♀️)

alerque:
Understood. I'll be looking into it.

alerque:
I agree with you the example is not serviced by adding the hyphenation topic into the mix. 

- First, I fixed the underlying bug. At some point we started normalizing language codes more aggressively and "und" was getting normalized to "en" because it wasn't a valid ICU canonical language code and the preceding context was "en".

- Next I've tweaked your commit to not only remove the description of it but also actually remove the hyphenation suppression. In the process it became apparent that the hyphenation that was being done was being done in English, while Latin actually has somewhat different rules. So I switched to the actual language so the output is actually *correct* as well as expected. Ironically now there is no *actual* hyphenation in the sample paragraph even though enabled because no break points were juicy enough to use.

Thanks for contributing towards making the manual more useful for everyone.

PR #1612, opened 2022-11-17
Don't count some passthrough tags twice in the AST

### LINKED ISSUE #1611: AST parser is duplicating nodes
Our last few versions have gone out with the version string on the cover page duplicated:

image

I've actually seen it before, but I always assumed the cause was too trivial to open an issue for a copy/paste duplication and that I'd just fix it. As it turns out the root cause is not trivial at all. Our new shiny SIL input parser is massaging the AST returned by the PEG parser and detecting some things twice. Notably some kinds of pass-through content is getting returned as both a string and a more detailed info node, but when the AST is cleaned up with end up with both versions. Hence anything in a `\lua{}` command is actually getting executed twice! This is almost certainly responsible for some weird stuff I've seen lately because running Lua code twice is not always idempotent (although I guess this would be a good way to stress test that it is!).

alerque:
Closes #1611

### DIFF
--- a/inputters/sil.lua
+++ b/inputters/sil.lua
@@ -170,17 +170,20 @@ local function massage_ast (tree, doc)
   end
   if tree.id == "document"
       or tree.id == "texlike_braced_stuff"
+      or tree.id == "passthrough_stuff"
       or tree.id == "passthrough_braced_stuff"
-    then return massage_ast(tree[1], doc) end
+      or tree.id == "passthrough_env_stuff"
+    then
+      return massage_ast(tree[1], doc)
+  end
   if tree.id == "texlike_text"
-      or tree.id == "passthrough_text"
-      or tree.id == "passthrough_env_text"
-    then return tree[1] end
+    or tree.id == "passthrough_text"
+    or tree.id == "passthrough_env_text"
+    then
+      return tree[1]
+  end
   for key, val in ipairs(tree) do
-    if val.id == "texlike_stuff"
-      or val.id == "passthrough_stuff"
-      or val.id == "passthrough_env_stuff"
-      then
+    if val.id == "texlike_stuff" then
       SU.splice(tree, key, key, massage_ast(val, doc))
     else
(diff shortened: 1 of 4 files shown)


Omikhleia:
Hmm this doesn't explain why the tree had two duplicated nodes (possibly a deeper root cause)

alerque:
The tree starts out with duplicates of *everything*, first the raw text capture then a breakdown of what the command, position, and inputs are. For most node types we flatten those into just one final AST entry, but I had some command syntax passthrough content both the string and info were getting spliced back in instead of descending into the node and flattening it.

Omikhleia:
Two small things:
- The fact `\script` works is that it does a `content[1]` rather than a `SU.contentToString(content)` as `\lua` and friends.
- The duplication may come from
  ```
    passthrough_stuff = C(Cg(
      V"passthrough_text" 
      + V"passthrough_debraced_stuff"
    )^0)
    ```
    (I don't get the double capture here, and the intent of that bit)

Just removing the extra "C", I had the expected output, but later math broke in the manual. Removing the math documentation package, it seems the manual compiled. That doesn't mean this is it at all (you merged/closed quickly, I haven't tested much furher, but there's code smell perhaps here worth investigating, no? Perhaps worth re-opening this issue for closer inspection?

alerque:
The use of `content[1]` was always a hack, being able to use `SU.contentToString()` seems more correct to me, the problem was we were iterating on a node type that has two different representations of the same data when we should have just been descending into it.

The `C(Cg())` capture is mostly to bring the tree into line with the way all the *other* captures work so the same parse tree to AST munging works across the board.

I do think the whole parser could be re-written from scratch with a much more efficient set of capture patterns that would almost or completely eliminate the need for the whole `munge_ast()` shinanigans, but I've been reluctant to do that because .... other priorities. The current system is inefficient and I agree it smells, but it's also remarkably fast for being so inefficient. I'd be happy to see a new issue opened to refactor the whole capture system so we go straight to a usable AST.

Issue #1614, opened 2022-11-18
Boustrophedon package issue at page break


Omikhleia:
Observed with SILE 0.14.4

I was working on the SILE manual with the aim of experimenting and proposing some improvements, when due to my page changes, I stumbled upon this issue. Extracting a scenario from the documentation:

```
\begin[papersize=a7]{document}
\neverindent
\nofolios
\use[module=packages.boustrophedon]
\font[size=20pt,family=Gentium Plus,language=grc]

Lorem ipsum dolor...

Good case
\skip[height=4cm]

\begin{boustrophedon}
ΧΑΙΡΕΔΕΜΟΤΟΔΕΣΕΜΑΠΑΤΕΡΕΣΤΕΣΕΘΑΝΟΝΤΟΣΑΝΦΙΧΑΡΕΣΑΓΑΘΟΝΠΑΙΔΑΟΛΟΦΘΡΟΜΕΝΟΣΦΑΙΔΙΜΟΣΕΠΟΙΕ
\end{boustrophedon}

After

Lorem ipsum dolor...

\end{document}
```

Observed:
![image](https://user-images.githubusercontent.com/18075640/202768143-08779ab6-1b45-4d8d-82cd-7250b400a374.png)

~~I did not investigate further, but~~ something is still active outside the boustrophedon environment after the page break occurred.

Obviously, removing or commenting out the vertical skip introduced here just to push the text at page boundary, the issue doesn't appear:
![image](https://user-images.githubusercontent.com/18075640/202768690-8328bee0-f031-4761-870b-15969a42e849.png)

FWIW, in the broken scenario above, adding more regular paragraphs at the end of the document: things are actually back to normal after a subsequent page break.

Omikhleia:
Issue closed in https://github.com/sile-typesetter/sile/milestone/40 (or is there a reason to keep it open?)

alerque:
Nope, just the PR link to this issue didn't explicitly mark it as closed and I missed it in triage.

PR #1615, opened 2022-11-18
fix(packages): Ensure a page switch does not break boustrophedon

Omikhleia:
One possible way to address #1614, by in-place modification of the vbox's output routine. It works for me, but beyond that issue I am also interested to know if that specific solution is sound[^1]

Important note: These kind of things will likely break if pushback[^2] occurs. I'm pretty sure the previous solution would have too, though!

[^1]: This is the kind of tricks I'm experimenting in a package for rotating by 180° the answers below a question (obviously additionally inverting the line order in that case, and the vglues are causing me troubles, but that's beyond the point here ^^).

[^2]: Because pusback - If I understand it correctly - occurs when the next frame has a different width (at least) and has to unpack the vbox's to resubmit the internal hbox's and nnode's to linebreaking. Is that correct?

### DIFF
--- a/packages/boustrophedon/init.lua
+++ b/packages/boustrophedon/init.lua
@@ -3,8 +3,6 @@ local base = require("packages.base")
 local package = pl.class(base)
 package._name = "boustrophedon"
 
-local _swap
-
 function package:_init (class)
   base._init(self, class)
   SILE.hyphenator.languages.grc = { patterns={} }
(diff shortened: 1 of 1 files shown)


alerque (approved):

alerque (approved):
As you've noted this isn't a 100% robust solution yet, but it does appear to be less brittle than before.

PR #1620, opened 2022-11-19
Better code blocks in SILE manual and package docs

Omikhleia:
This PR is a part of #1194 upon which I've been sitting too long - so I rebased all I had started, completed it, and "Voilà!".

I think it's helpful for package documentation writers and for casual readers of the Manual or package doc sources...

  - It entirely removes the `\line` command calls (from the SILE manual's macro) in existing package documentations
  - Instead, it proposes, via the **autodoc** package, an `autodoc:codeblock` environment and a raw handler going by the same name.
    - The first takes care of adding the lines around the code block, so the appearance is the same as now by default, but with less impediment for the writer. Old:
       ```
       \begin{verbatim}
       \line
       Some stuff possibly SILE-escaped
       \line
       \end{verbatim}
      ```
      Now becomes:
      ```
       \begin{autodoc:codeblock}
       Some stuff possibly SILE-escaped
       \end{autodoc:codeblock}
      ```
      It's sparingly used (where some command expansion was indeed required in the content), due to the following point.
    - The second allows avoiding all those escaped sequences. Most of the time, they are annoying to get right, and sometimes a few were even missed... So we also have, cool and easy-to-use:
       ```
       \some{stuff}% All being a true verbatim without needing the escapes
       \end{raw}
       ```
Regarding the Manual itself
  - I applied the same autodoc commands as much as possible in my limited time. Chapters c09, c11, c12 still use a lot the old method... And therefore still have several `\line` calls. That's for future documentation writers/reviewers to address, I'm afraid.
  - I've nevertheless redefined `\line` to that now provided in autodoc, so at least they have the same "look" everywhere. That line is now 100% of the text block, by the way. The previous 450pt looked dubious (stopping just at a small distance from the margin in the current manual), and wouldn't have worked with other paper width...

On the way I also added a few missing "examplefont" around examples, and removed a few `\line` in them (most already didn't hve them, and it's marked with a different font, which is sufficient IMHO).

Overall, I feel it makes the tooling for package documentation writer a bit more helpful and independent from the SILE manual (... but there would still be a lot of things needed to achieve that -- as per #1194, sure, `examplefont` and `code` are on the list....)


alerque (approved):
This looks great! It's work I had in mind to do and definitely an improvement.

PR #1621, opened 2022-11-19
Flip book class gutter margin to left of odd pages

alerque:
The default book class has had a layout that suggests a gutter margin, but the margin has been on the wrong side of the isle since its conception. Since the default language and layout is LTR and the default first page is odd and the book class uses the default frameset as the right master, it seems to me the wider margin should be on the left where a LTR book binding would traditionally be.


alerque:
@simoncozens Is there something I'm missing here? Why did you put the wider margin on the right of the right master back in the first commit for the book class 634a80f9f?

@Omikhleia Can you sanity check this too? This is a big change but for me it only affects the manual because I never use the book class for anything but always CaSILE's cabook class that has different frames. But every time I see the manual spreads they look backwards to me.

Omikhleia:
@alerque I don't use the regular book class much (i.e. except for quick tests, which are seldom spreading over many pages, I use my own book class)...

This being said, to be real safe in that discussion, let's rather think in terms of _inner margin_ (where the book pages would be bound together) and _outer margin_ (er... where one puts one's fingers to flip the page, so to say?).

So you seem to want (inner margin) > (outer margin) in this PR, right? That might perhaps be wrong.

In the usual "canons" of page construction (but there are several, depending on countries, periods of time...), the inner margin is _shorter_ than the outer margin. In other terms, on a European book (LTR languages) at least (I don't know the non-European usages), regarding the an _odd_ page (= on the "right"), the left (inner) margin is shorter than the right (outer) margin.

With that in mind, It **seems** to me the book class is actually correct. Just to be sure I don't mix might right and left again as I do too often, this is the (current) SILE manual:
image

That does seem correct to me (although I am not sure were the percentages used in the frame definitions come from).

For an extreme-case comparison, here is Vencentinus" layout:

image

(Vencentinus, in the 15th century, used that layout for a prayer book. It is assumed, if I read that correctly, that he wanted huge outer and bottom margins for marginalias and manual annotations. If so, this is fairly cool!)

See also https://github.com/sile-typesetter/sile/discussions/1251#discussioncomment-3522248 = I actually implemented several page layouts in my class (various "French canons"; the above Vencentinus; Honnecourt (a.k.a. Van de Graaf layout); layout by division methods (a generalization of the two former ones); last but not least, Tschichold's "canonical" layout.) The images there show an odd page (i.e. chapter-opening page). All inner margins are shorter than outer margins.

I hope it helps -- For a PDF, that should be correct.

For print, aye, it gets intricate... The question of an additional "gutter margin" _may_ be considered, i.e. an _additional_ margin added to the inner margins in order to compensate the part of the paper that would be used by the binding process. But AFAIK, this is not considered as part of the "page layout", rather as an extra trim area (i.e. if my book is in 6 x 9 inches, the margins would still be computed using one of the above method (which sum up, to simplify most Canons but the French, to using a division via the diagonal, or a circle in the case of Tschichold). The gutter would be added to that _afterwards_, if I get it right. I don't think -but I might be wrong- it's still used a lot. Most print-on-demand services (to speak of what I know) expect a trim area (equal on both sides), notably for full-bleed printing, but otherwise they compensate by themselves (theoretically...) the area needed for gluing the pages without asking.

So in other terms, it doesn't seem to me we have to account for gutters. If we want to follow the usual canons of page construction, all that matter is (inner margin) < (out margin) - with proportions depending on the actual canon. (The French one, or "canon des Ateliers") is weird, though I am obviously used to it... Now that I have seen Tschichold's, I tend to prefer it. But paper size also have to be accounted... Some layout are better in 2/3 ration (e.g. 6x9in in american novel book) than in 1/√2 (e.g. ISO formats), and reciprocally.

**In brief**: inner margins (where the binding is) < out margins (we humans have big fingers for page flipping or need space for pencil annotations).

alerque:
The paperback binding systems I'm used to publishing with eat a large amount of paper. Since the book does not open up flat a substantial bit of the inner margin is not accessible, and if I don't provide extra gutter margin there books would be unreadable or at least look like they got printed offset. I suppose different production methods might account for this other ways (automatically compensating at print time, or using other binding types that don't need the space, etc.)

Omikhleia:
Indeed. That's, I think, the core issue: the inside gutter ("binding offset" in the image below from the LaTeX geometry package), for pages to remain legible depends on many factors: the number of pages (the thicker is the book, the more folding space is needed), the kind of paper (its weight/thickness), and eventually the type of binding (gluing, stitched in quires, etc.). So it cannot be "generally" computed without knowing these.

image

Hmm, that would be an interesting challenge, to properly designed a frame set with appropriate (class?) options for flexibility... I just checked Bringhurst's _Elements_, but apart from a general discussion, he's not very helpful here (3rd edition). That is, while various proportions between paper size and the content text block are discussed (always leading to smaller inner margins than outer ones), the actual placement of a possible binding offset does not seem to be addressed...

His own book (on a non standard paper size...) has wider outer margins than inner ones - with respect for the main text block, that is: he uses most of the outer margin for notes and headings. E.g. (blurred as this is copyrighted material):

image

As I wrote, I am not sure where the default book class percentages come from.

The above image from the LaTeX geometry package is "suspiciously" good (it does show some proportions, e.g. the diagonal of the offset text block seems canonical), I'll try to see if I can derive how they made it...

Omikhleia:
What one can do with SILE and a bit of Lua...
image

As I said, I don't know where this `8.3%pw` inner margin in the book class comes from, but one thing seems sure: it is **too small** for book-sized pages such as A5 or 6x9in (US trade) in glued binding... And far too small, henceforth, for anything smaller (e.g. "pocket" books) Typical recommendations is a **minimum** of 0.5in for margins. Below 60 pages, it's "just" decent with 0.5in; between 61-150 pages, a minimum of 0.625in is recommended (so a gutter of 0.125in), and for 151-400 pages, it goes up to 1in (so a gutter of 0.5in).

In my own book, I empirically (at the time) tweaked these percentages and used an inner margin of 0.6in for 360 pages, and I confirm it's a bit too small, especially in the middle of the book (maybe 1in would have been too much, but something in-between would have looked better, very likely - empirically again, an extra 0.20in at least...).

This PR here results in:
image

The inner margin is a bit better, but I _really_ don't like the short outer margin. We _could_ attempt having better defaults, e.g. something such as one of the following constructions:
image

alerque:
Having better default frames is definitely something I'd consider for the book class, but I would also only consider it in a breaking major release. This (if I was right about the flip) would be a bug fix I could but in a minor release. Also it has the advantage of not breaking any of our test cases because the exact size and proportions of the frame don't change, only the absolute position.

Thanks for the extra research into this.

alerque:
Dropping this to a future breaking release. Since I was wrong about it just being backwards (at least for *some* production workflows) I think a proper frame overhaul should be the one to address this, just flipping it and making it wrong for the *other half* of people doesn't make much sense.

Omikhleia:
The PR as proposed here is thus invalid.

We could have a discussion on how to properly handle page layouts and binding offsets (e.g. in my own classes, I now have an `offset` class option applied to pre-composed layouts), but this is another topic for which a dedicated issue is preferable.

PR #1623, opened 2022-11-22
+§ Installing third-party packages

ctrlcctrlv:
Closes Omikhleia/markdown.sile#17.


README.md
```diff
@@ -271,6 +271,39 @@ Add to your repository as `.github/workflows/sile.yaml`.
...
+
+Third-party SILE packages rely on the `luarocks` package manager.
+
+To get your Lua version which is used for the execution of `sile`:
+
+```bash
+#!/bin/bash
+# This command finds SILE, then reads its shebang line, and calls whatever the
+# interpreter for SILE is to find its version.
+SILE=$(which sile) || >&2 echo "No sile in '$PATH'?" && \
```
alerque, line 284: Why not just parse `sile --version`? In fact to save the trouble of parsing the output to get just the Lua version just use the information from inside SILE itself with `sile -e 'print(SILE.lua_version);os.exit()' 2> /dev/null`.

README.md
```diff
@@ -271,6 +271,39 @@ Add to your repository as `.github/workflows/sile.yaml`.
 This work flow assumes your project has a source file `my-document.sil` and will leave behind a `my-document.pdf`.
 Note the comments in [the section about Docker](#docker) regarding version tags.
 
+## Installing third-party packages
+
+Third-party SILE packages rely on the `luarocks` package manager.
+
+To get your Lua version which is used for the execution of `sile`:
```
alerque, line 278: I think we should probably note that for most cases just using the LuaRocks default is probably okay, manually specifying the Lua version only applies when there is both more than one available and SILE was configured to use something other than the default.

README.md
```diff
@@ -271,6 +271,39 @@ Add to your repository as `.github/workflows/sile.yaml`.
...
+  [[ $___LUA_VERSION =~ ^5\.[1-4]$ ]] || \
+  (>&2 printf 'Cannot auto-detect Lua version!? ' && unset ___LUA_VERSION)
+>&2 printf 'Lua binary: %s\n' "$(which lua$___LUA_VERSION)"
+```
+
+To install the latest development version, for the given package, you may often use its provided “rockspec”:
+
+```bash
+# This example uses markdown.sile, a highly useful plugin that converts Markdown documents to SILE `.sil`.
+luarocks --lua-version $___LUA_VERSION install --server=https://luarocks.org/dev markdown.sile --local
```
alerque, line 298: Two things here (besides the above note about Lua versions). First the extra server argument is only necessary for things that are published to non-default manifests, and the need for that argument should probably be documented in the plugin docs more so than here. Once things are published to the default manifest this shouldn't be necessary. Second we should probably note the various options besides local for project, user, and system wide installations.

alerque (changes requested):

README.md
ctrlcctrlv: Because this can happen:

```
$ sile --version
SILE v0.14.5 (LuaJIT 2.1.0-beta3)
```

In that case we error so user can research the issue, ultimately realizing they need to install for lua5.2.

alerque: Hm, yes LuaJIT is always fun. The second cammand I sent though should give the right answer for LuaJIT too:

```console
$ sile -e 'print(SILE.lua_version);os.exit()' 2>/dev/null
5.1
```

At least for my system that's the right version to install modules for. I'm curious though why you said 5.2 above?

ctrlcctrlv: Hmm. Not sure , I just know LuaJIT supports some 5.2 APIs.

alerque: Yes it does, but I think the binary module API (except some of it's exclusive FFI stuff) is a match for 5.1. It has some language features added similar to 5.2, but is fundamentally forked from 5.1.

ctrlcctrlv: I see, you'd know more about it than I I'm sure :)

README.md
```diff
@@ -271,6 +271,36 @@ Add to your repository as `.github/workflows/sile.yaml`.
...
+Third-party SILE packages rely on the `luarocks` package manager.
+
+To install the latest version, for the given package, you may often use its provided “rockspec”:
+
+```bash
+# This example uses markdown.sile, a highly useful plugin that converts Markdown documents to SILE `.sil`.
+luarocks install markdown.sile --local
+```
+
+You may remove `--local`, in which case a global installation will be done (on most systems this requires `root`).
```
Omikhleia, line 285: > on most systems this requires `root`
Or adding `sudo`? (keeping in mind Linux users are not always devs and that terminal-friendly)

README.md
```diff
@@ -271,6 +271,36 @@ Add to your repository as `.github/workflows/sile.yaml`.
...
 Note the comments in [the section about Docker](#docker) regarding version tags.
 
+## Installing third-party packages
+
+Third-party SILE packages rely on the `luarocks` package manager.
+
+To install the latest version, for the given package, you may often use its provided “rockspec”:
+
+```bash
+# This example uses markdown.sile, a highly useful plugin that converts Markdown documents to SILE `.sil`.
```
Omikhleia, line 281: Not that I mind ;) but it doesn't converts to .sil (yet...). It allows reading and processing Markdown documents with SILE

README.md
```diff
@@ -271,6 +271,36 @@ Add to your repository as `.github/workflows/sile.yaml`.
...
+eval $(luarocks path)
+```
+
+### Finding Lua version in use for running SILE
+
+To get your Lua version which is used for the execution of `sile`:
+
+```bash
+#!/bin/bash
+___LUA_VERSION=$(sile -e 'print(SILE.lua_version);os.exit()' 2> /dev/null)
```
Omikhleia, line 299: Maybe we should consider eventually having a `sile --lua-version`, to hide the weirdness of that one-liner Lua script.

README.md
alerque: I don't mind hints reminding people to use their system package manager, but (with by distro developer hat on here) I'm against doing so in anything other than the default options and minimal hints. There are many ways to customize most install commands as a downstream project we don't want to wade into that too much. There are also many ways of running root. Some package managers handle achieving root themselves, others expect you to run them in a root shell, with `sudo`, with `doas`, or some other tooling. People should follow their distro usage guidelines for which way to achieve elevated permissions, we don't want to spec that.

README.md
alerque: I reworded this with that in mind.

README.md
alerque: Maybe, but given the only use case is for where SILE is *not* using the system default I'm afraid there might be a lot of pitfalls to trying to simplify. What exactly are we trying to solve for?

(I cleaned up the usage and explanation a bit more, I'm willing to consider exposing something more precise if there is a demand and notable use case for it.)

alerque (approved):

Issue #1339, opened 2022-03-05
Enhancing the indexer package


Omikhleia:
Main goal: I need multiple index support for a potential project.
Stretch goal: As #1194 progresses, it could be nice for the SILE manual to include an index of all mentioned settings, commands, key concepts, etc.

The **indexer** package is a nice start.

I however noted a few things to fix and improve, from the easiest to the more challenging. Let's track them here.
- [x] docs: No, `\printindex` does not "deposit (entries) in a list". That's vague and obscure. It actually renders the index.
- [x]  ~~docs: `index:item` is passed the indexed content as argument, and an option `pageno` (which is a list of comma-separated pages, so it could be a plural here). Could be nice to document.~~ or we have better hooks, so entries might be styled in a way (e.g. italic), and page numbers another (e.g. oldstyle numbers, and the index name shall be passed too, if different indexes are to be style differently ;)
- [x] fix: if nothing was indexed, `\printindex` without option aborts with a low-level error (`pairs()` being called on nil...). Let's play fair, it could just be a warning that the "main" index is empty, and let the document be processed to end.
- [x] fix: If index "myindex" does not exist, `\printindex[index=myindex]` aborts with that same low-level error. Let's play fair again, it could just be a clear error g that the "myindex" index is unknown (it could be a typo, whatever). If it exists, but is empty - this shouldn't happen in the current implementation - a warning would be ok.
- [x] feat: The book class should support the indexer by default, so we have one potentially class using it...

That's the simple bits. Call that "Stage 0" tasks.

Let's now go for real things....
- [x] feat: PDF link annotations (page linking to the index call), for the index to be useful in a PDF reader. Would be easy a first glance, but see next item below...
- [x] feat: Collecting subsequent pages into ranges, so one does not get "someentry 62, 63, 64, 65, 70, 71, 72" but rather "someentry 62-65, 70-72" (in which case linking could be on the first occurrence of a range).
- ~~feat: Explicit ranges. The raw entry might be "someentry 62, 65", but the author knows 62-65 would be more appropriate. (Several LaTeX packages use delimiters such as `|(` and `|)`, for the record).~~
[...]

PR #1632, opened 2022-11-27
Number formatting for many languages and sorting (collation) with ICU

### LINKED ISSUE #1630: Number formatting in foreign languages
Why does SILE have the 3 "hacks" `SU.formatNumber.en`,  `SU.formatNumber.tr`,  `SU.formatNumber.eo` (with methods such as `nth()`, `string()` etc.) - not covering many languages, thus - while it _also_ has ICU (via the "justenoughicu" wrapper to Lua) and could cover, with little efforts, a whole lot of languages with less _ad hoc_ coding?

Just playing 5mn with ICU...
```
> print(icu.format_number_ext(1984, "en-US", 0)) -- 1984
> print(icu.format_number_ext(1984, "en-US", 1)) -- 1,984
> print(icu.format_number_ext(1984, "en-US", 5)) -- one thousand nine hundred eighty-four
> print(icu.format_number_ext(1984, "en-US", 6)) -- 1,984th

> print(icu.format_number_ext(1984, "fr-FR", 0)) -- 1984
> print(icu.format_number_ext(1984, "fr-FR", 1)) -- 1 984
> print(icu.format_number_ext(1984, "fr-FR", 5)) -- mille neuf cent quatre-vingt-quatre
> print(icu.format_number_ext(1984, "fr-FR", 6)) -- 1 984e
```

Or for useful purposes, not doubts :grin:  :
```
> print(icu.format_number_ext(1984, "eo", 0)) = 1984
> print(icu.format_number_ext(1984, "eo", 1)) = 1 984
> print(icu.format_number_ext(1984, "eo", 5)) = mil naŭcent okdek kvar
> print(icu.format_number_ext(1984, "eo", 6)) = 1 984.
```

No, seriously:
```
>  print(icu.format_number_ext(1984, "sr-Cyrl", 5)) -- једна хиљаду деветсто осамдесет и четири
>  print(icu.format_number_ext(1984, "sr-Latn", 5)) -- jedna hiljadu devetsto osamdeset i četiri
```

Was there a good rationale **not** using ICU, and having those above-mentioned limited language hooks?
If not, should I make a PR?

EDIT: Of course eventually to unleash the power of the beast, we'd need using locales instead of the current use-my-country-code-as-a-language...

Omikhleia:
**Rationale**
- SILE should be able to correctly format numbers in appropriate number system and with appropriate format style for as many languages as possible, and not just a few selected ("en", "eo", "tr")
  - in "default" form (en: 1234 or ar: ١٢٣٤ or zh: 1234)
  - in "decimal" form, i.e. usually with thousands separator etc. (1,234 or ١٬٢٣٤ or 1,234)
  - in "ordinal" form, i.e. the "nth's" (1,234th or ١٬٢٣٤. or 第1,234)
  - as an extra in spelt-out form (one thousand two hundred etc., or some stuff I can't copy, or 一千二百三十四)
  - EDIT: (N.B.) I took the most neutral names I could thing of, for these... What typographic rules say, for a given language, may be different; e.g. some French typographers distinguish a _cardinal use_ (= "decimal" with separator, in that case a space) and an _ordinal use_ (as in a year date, a page number, etc. = "default" without separator). Terminology is sometimes ambiguous, it seems (because of course, there are also the actual ordinals: with spaces at thousands... and the raised "e" &c.)
- SILE should be able to correctly sort strings with appropriate rules according to the (main) language usages: **indexer** and packages of that vein may need it or their sorting entries will be garbage.

**First part** (number formatting) = Closes #1630 

Notes
- Deprecations are currently set to **0.14.6** for removal in 0.16.0
- For counters, it is _slightly_ breaking though - if we consider that what was in `tests/counters.sil` is a "reference" of sorts, though fully undocumented. I would tend to think it is acceptable (for the _main_ intended use of _counters_, i.e. sectioning and lists). It's unlikely to affect many users anyway; and the benefits are worth it: We get "default" (cardinal), "decimal" (with thousands separator, etc.), "ordinal" (our former "nth"), as well as "string" (spelt-out) support, _for most languages_ (i.e. as long as supported by ICU).

> En français, on écrit « en l'an 1984 », mais « dans 4 500 ans ».
> Probably English and others have similar rules, when to use 1984 or 1,984

N.B. I haven't considered _all_ format style options from ICU (e.g. "duration", "currency", etc.) because I felt they were unlikely to be needed. That could be added, if someone really wants - But I'm not convince it would have a real use in SILE (and in my tests at least, the ICU library is somewhat inconsistent in how it honors these).

**Second part** (string sorting) = Add language-dependent sorting with ICU collation options

 As noted, an imperative for indexes (= relates to #1339), etc.

```
> SILE.call("language", { main = "fr" })
> t = { "Jean100", "Jean2", "Alinoé", "Alinéa" }
> table.sort(t)
> print(table.concat(t, ", "))
```
--> Alinoé, Alinéa, Jean100, Jean2 = bad indexing order

```
> t = { "Jean100", "Jean2", "Alinoé", "Alinéa" }
> SU.collatedSort(t)
> print(table.concat(t, ", "))
```
--> Alinéa, Alinoé, Jean2, Jean100 = good indexing order, yay!

Supports options passed to `SU.collatedSort`, or also defined in `SU.collatedSort.xx` for language-specific override of the default values if need be.


alerque:
Cardinal vs. Ordinal is also very applicable in Turkish, and to a lesser degree English. It's probably a distinction we should make and supply both for rather than assuming one or the other given the most common use case.

Otherwise :100: to the analysis here.

core/utilities-numbers.lua
```diff
@@ -0,0 +1,189 @@
...
+        if options:match("^%l") then
+          case = "lower"
+        elseif options:match("^.%l") then
+          case = "title"
+        else
+          case = "upper"
+        end
+      end
+      if options:lower() == "nth" then
+        SU.deprecated("Format 'nth' in SU.formatNumber", "'ordinal' in SU.formatNumber", "0.14.6", "0.16.0")
```
alerque, line 134: I think my only quibble with this whole PR is the naming fiasco for ordinal/ordinal-string/nth/cardinal etc. As I understand it the current shim you have in here will work to shim old documents to the new format, but it will also *fail* when old documents get updated because of the re-use of "ordinal" to mean something different. One would have to update documents exactly at the version boundary this shim was removed, not before. I don't think that's ideal at all.

At the moment I'm inclined to keep the "nth"/"ordinal" style names. As I understand it ICU doesn't have an equivalent to our "nth" style at all right?

alerque (changes requested):

core/utilities-numbers.lua
alerque: I suppose one way to deal with this is introduce a document level parameter that specifies the *target* SILE version. That might be handy for reference, but it could also allow us to automate cases like this where it is ambiguous whether the old behavior through a shim or new behavior is intended.

alerque: Oh yuck. Thanks for setting the record straight.

Turkish has equivalents of both English forms, "1st" and "first" are "1." and "birinci", the period is not punctuation it is explicitly the cardinal marker and vocalized as "-inci".

What do you think about the backwards compatibility issue with the shim? Should I add a marker for documents that have been converted to target a specific API level or revert the scheme rename? Or some third option I'm missing?

alerque: Nevermind the trailing question. I overlooked the argument changing from a string to a table, so that will be enough to know whether a shim should be used or not without conflating new/old meaning of "ordinal".

alerque: > > As I understand it ICU doesn't have an equivalent to our "nth" style at all right?
> 
> It does: "ordinal" is the ICU name (`UNUM_ORDINAL`) for SILE's idiosyncratic English-sounding "nth". Both Eo and En had the correct ordinal _numbers_, resp. "1234-a" and "1234'th"[1](#user-content-fn-1-eb534ee3e1bd9c19d8f53e0201eb8748) here.

Oh, right. In that case the one it doesn't have is our "ordinal-string" combo right? It has "spellout" and "ordinal" but not a spelled out ordinal. Or am I still missing something? It seems to have an input parser for spelled out ordinals but I can't find a combination of output options to achieve it.

alerque:
I'm looking into dropping the bits of our implementations that are now duplicated. Somewhat amusingly our "string" (spellout) implementation in English handles bigger numbers than the ICU one which starts failing at 32 bits. It leaves off at quadrillions, ours goes on to handle quintillion, sextillion, septillion, and octillion!

Similarly for the Turkish implementation, ours is more complete.

The English apostrophe thing is probably wrong according to most style guides though.

core/utilities-numbers.lua
Omikhleia: Correct, it only has our "string" as `UNUM_SPELLOUT` (1000 = "one thousand"). It doesn't have spelt-out ordinals (but perhaps these are complex in the general case, or the Consortium just considered they would do it anyway... Too many language-specifics to address lol).

Also, by the way, note that ICU doesn't know anything about genders, plurals etc. (so for instances, 1 as ordinal is "1st" in English and "1e" in French - the latter only works actually for masculine singular.) Well, ICU can't be perfect I guess, for many reasons.

Omikhleia:
> (our implementation) is more complete.

Yep, or even better (with tunable settings, in the case of Esperanto). I'm ok with SILE having a mechanism for bypassing ICU when it can do better, I didn't intend to have them necessarily replace. The main point is to address other languages generically with decent defaults.[^1]

[^1]: That what we already do e.g. for node breakers, defaulting to ICU breakpoints but possibly overriding the logic as e.g. in `SILE.nodeMakers.fr`, `SILE.nodeMakers.ja`, etc. (to handle resp. punctuations, some JIS class things, etc.).

alerque (approved):

PR #1633, opened 2022-11-27
Small typo in Esperanto language section

Omikhleia:
Seen in passing.
@ctrlcctrlv Can you double check this is indeed a typo? Seems to me "chapter" is "ĉapitro" in Esperanto and that it was the intended example here, no?


ctrlcctrlv (approved):
Yes, you're right.

PR #1638, opened 2022-12-01
Cleanup inputter AST nesting expectations

### LINKED ISSUE #1637: Line no issue in AST
Using SILE 0.14.5 with my markdown inputter, I got a `This isn't a SILE document!` error, which I obviously didn't have before.

The reason is here (introduced via #1596)
https://github.com/sile-typesetter/sile/blob/1a86520d198d7709966e1fd1bfd8dfaceda86393/inputters/base.lua#L39

I was about to just shrug: "Ok, your inputter didn't add (lno, col, pos) information to the parsed tree, because it doesn't keep track of them... _Now_ they are mandatory, well, _regression aside_, just add something at zero and let it go..."

But the AST not being documented, there might be some code smell...
- More generally, I am not sure how well it plays with other inputters than SIL (e.g. XML)...
- There are places where `lno` may indeed be considered as optional...
   E.g.
  https://github.com/sile-typesetter/sile/blob/1a86520d198d7709966e1fd1bfd8dfaceda86393/core/sile.lua#L355
- Programmatic command creation in **inputfilter** assume a `line` and no `lno`
  https://github.com/sile-typesetter/sile/blob/1a86520d198d7709966e1fd1bfd8dfaceda86393/packages/inputfilter/init.lua#L30
  (I have used these tricks in so many places...)

Having a properly documented AST would be neat (any new inputter need that information...), but until then, we'd need fixing the expectation (`lno` or `line`) here...

BTW, is the offending `lno` check in the first quoted bit of code above really needed?

alerque:
Closes #1637

Please do sanity check me on this one @Omikhleia. This completely moves the cheese around but I think it makes more sense overall and likely fixes your use case of a different inputter. I moved the check for leading/trailing junk that was specific to SIL to the SIL inputter. I also reworked the check for a document to potentially wrap loose leaves which should make document fragments more viable with no top level tags. It won't necessarily sniff but it should parse if asked to explicitly. In the process the expecations about top level tags changed, so XML no longer has to add an extra level of nesting it just runs with whatever the top level is. Of course this changed all the unit test expecations, but the change seems way saner than the original.

Am I breaking it or fixing it?

### DIFF
--- a/inputters/base.lua
+++ b/inputters/base.lua
@@ -33,28 +33,18 @@ end
 
 function inputter:requireClass (tree)
   local root = SILE.documentState.documentClass == nil
-  local document
   if root then
-    for _, leaf in ipairs(tree) do
-      if leaf.lno then
-        if document then
-          SU.error("Document has more than one parent node!")
-        end
-        if leaf.command == "sile" or leaf.command == "document" then
-          document = leaf
-        end
-      end
-    end
-    if not document then
+    if tree.command ~= "sile" and tree.command ~= "document" then
       SU.error("This isn't a SILE document!")
     end
-    self:classInit(document.options or {})
+    self:classInit(tree.options or {})
     self:preamble()
   end
 end
 
 function inputter:process (doc)
-  local tree = self:parse(doc)
+  -- Input parsers can already return multiple ASTs, but so far we only process one
+  local tree = self:parse(doc)[1]
   self:requireClass(tree)
   return SILE.process(tree)
 end
--- a/inputters/sil.lua
+++ b/inputters/sil.lua
@@ -196,13 +196,31 @@ function inputter:rebuildParser ()
 end
 
 function inputter:parse (doc)
-  local tree = epnf.parsestring(self._parser, doc)[1]
-  if not tree then
+  local parsed = epnf.parsestring(self._parser, doc)[1]
+  if not parsed then
     return SU.error("Unable to parse input document to an AST tree")
   end
   resetCache()
-  local ast = massage_ast(tree, doc)
-  return ast
+  local top = massage_ast(parsed, doc)
+  local tree
+  for _, leaf in ipairs(top) do
+    -- Content without position tracking is likely comments or whitespace outside of a tag.
+    -- We're just fishing for one tag here.
+    if leaf.lno and leaf.command then
+      if tree then
+        SU.error("Document has more than one parent node that looks like a master document!")
+      end
+      tree = leaf
+    end
+  end
+  if not tree then
+    SU.error("This isn't a SILE document!")
+  end
+  local rootelement = tree.command
+  if rootelement ~= "sile" and rootelement ~= "document" then
+    tree = { tree, command = "document" }
+  end
+  return { tree }
 end
 
 return inputter
--- a/inputters/xml.lua
+++ b/inputters/xml.lua
@@ -78,8 +78,8 @@ function inputter.parse (_, doc)
   -- to supply handling far whatever that element that is in a specific format.
   -- Hence we wrap the actual DOM in an extra element of our own if and only if
   -- it doesn't look like a native SILE one already.
-  local root = tree.command
-  if root ~= "sile" and root ~= "document" then
+  local rootelement = tree.command
+  if rootelement ~= "sile" and rootelement ~= "document" then
     tree = { tree, command = "document" }
   end
   return { tree }
--- a/documentation/c02-gettingstarted.sil
+++ b/documentation/c02-gettingstarted.sil
@@ -382,7 +382,6 @@ But be warned that since GitHub rebuilds containers from scratch on every such i
 Pulling the prebuilt Docker images is recommended instead.
 
 With these ideas in mind, other CI systems should be easy to support as well.
-\end{document}
 
 \section{Installing third-party packages}
 
@@ -412,3 +411,4 @@ To get your Lua version which is used for the execution of SILE:
 $ export LUA_VERSION=$(sile -e 'print(SILE.lua_version);os.exit()' 2> /dev/null)
 $ luarocks install --lua-version $LUA_VERSION ...
 \end{raw}
+\end{document}
--- a/inputters/sil_spec.lua
+++ b/inputters/sil_spec.lua
@@ -9,13 +9,13 @@ describe("#SIL #inputter", function ()
   describe("should parse", function ()
 
     it("commands with content", function()
-      local t = inputter:parse([[\foo{bar}]])[1]
+      local t = inputter:parse([[\foo{bar}]])[1][1]
       assert.is.equal("foo", t.command)
       assert.is.equal("bar", t[1][1])
     end)
 
     it("commands without content", function()
-      local t = inputter:parse([[\foo{\foo bar}]])[1]
+      local t = inputter:parse([[\foo{\foo bar}]])[1][1]
       assert.is.equal("foo", t.command)
       assert.is.equal("foo", t[1][1].command)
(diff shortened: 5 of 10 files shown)


inputters/base.lua
```diff
@@ -33,24 +33,11 @@ end
...
-    end
-    if not document then
-      SU.error("This isn't a SILE document!")
-    end
-    self:classInit(document.options or {})
-    self:preamble()
+  if root and tree.command ~= "sile" and tree.command ~= "document" then
+    SU.error("This isn't a SILE document!")
   end
+  self:classInit(tree.options or {})
```
Omikhleia, line 39: I think this is wrong and SILE <= 0.14.4 was actually right here:
It make sense that we only need to init a class and check preambles if the document is "root", i.e. not inputted while there's already a class, i.e. typically from CLI.


 local root = SILE.documentState.documentClass == nil
  if root then
    if #tree ~= 1
      or (tree[1].command ~= "sile" and tree[1].command ~= "document") then
      SU.error("This isn't a SILE document!")
    end
    self:classInit(tree[1].options or {})
    self:preamble()
  end

inputters/sil.lua
```diff
@@ -201,8 +201,17 @@ function inputter:parse (doc)
...
+  local ast = {}
+  for _, leaf in ipairs(top) do
+    if leaf.lno and leaf.command then
+      table.insert(ast, leaf)
+    end
+  end
+  if #ast == 0 then
+    SU.error("This isn't a SILE document!")
+  end
+  return #ast == 1 and ast[1] or { command = "document", ast }
```
Omikhleia, line 214: Franly, I have no idea what this tries to do.

inputters/xml.lua
```diff
@@ -74,15 +74,7 @@ function inputter.parse (_, doc)
...
-  -- XML documents can have any root element, and it should be up to the class
-  -- to supply handling far whatever that element that is in a specific format.
-  -- Hence we wrap the actual DOM in an extra element of our own if and only if
-  -- it doesn't look like a native SILE one already.
-  local root = tree.command
-  if root ~= "sile" and root ~= "document" then
-    tree = { tree, command = "document" }
-  end
-  return { tree }
+  return tree
```
Omikhleia, line 77: I can understand why there was a "document" level wrapped around the "tree".
Say you have an XML file not containing a `<document>` or `<sile>` root, but its own (schema-specific) root tag.
How to process it with SILE?
I'd say, possibly[^1], with something such as:
```
sile -c myxmlclass myxmlfile.xml
```
Before, this was applied to the `<document>` pseudo-encapsulation.
If it is removed, it will apply to the top-level tag, but possibly doing weird things with its own attributes.

[^1]: I formerly used the `-I` preamble, and am still at loss finding an appropriate replacement.

inputters/sil.lua
alerque: Not as a root document no, but in the context of an imported fragment, I could see asking the parser to parse a fragment with no top level element, `foo \code{bar} baz` being a viable SIL fragment (not document, just valid syntax for a fragment to include into another one.

inputters/base.lua
alerque: Good catch, the nesting of `if root` is borked here, I combined two conditionals incorrectly.

inputters/xml.lua
alerque: In what scenario would an input document have something as a top level tag that we should reject entirely and not process as the top level AST element that brings with it the class instantialization?

Omikhleia: I don't get your question.
The previous code was not rejecting anything, it was adding a document level when the root tag isn't "sile" or "document". In other terms, given an XML document:
```
<tei>
  <!-- TEI dictionary stuff -->
 </tei>
 ```
Then SILE regarded it as wrapped in a `<document>`:
```
$ sile mydict.xml
SILE v0.14.4 (Lua 5.2)
<xml.xml> as xml

! Unknown command tei at xml.xml:1:1
```
And assuming "myclass" defines all the necessary tags, `sile -c myclass mydict.xml` worked.

The `<document>` encapsulation was required, for `inputter:requireClass()` to work as expected.

File: inputters/sil.lua (hunk 2 above)
Comment (Omikhleia) [reply inputters/sil.lua:214]: I don't follow the rationale fully for this. But admittedly, being sick this week, my brain might be malfunctioning.

File: inputters/xml.lua (hunk 3 above)
Comment (alerque) [reply inputters/xml.lua:77]: You're right, I lost track of the fact that classes don't provide the XML input parser in those cases and they may *want* to wrap the whole document.

File: inputters/sil.lua (hunk 2 above)
Comment (alerque) [reply inputters/sil.lua:214]: I still think this use case makes sense but it needs a test and it shouldn't be mashed in the other unrelated change. I'll back it out for now.

Comment (Omikhleia) [reply inputters/sil.lua:214]: Just re-reading myself, the above comment might sound ironic but that wasn't intended. I'm really sick, functioning at reduced brain level.

**Comment by alerque** [comment]
I marked all the reviews as resolved because all three got completely rehashed, but that's not to say the new rehash doesn't need scrutiny. Have at it fresh! Same two commits but totally different scope of changes.

File: inputters/xml.lua [hunk 4]
```diff
@@ -70,19 +70,19 @@ function inputter.appropriate (round, filename, doc)
...
   -- it doesn't look like a native SILE one already.
-  local root = tree.command
-  if root ~= "sile" and root ~= "document" then
-    tree = { tree, command = "document" }
+  local rootelement = ast.command
+  if rootelement ~= "sile" and rootelement ~= "document" then
+    ast = { ast, command = "document" }
   end
-  return { tree }
+  return ast
```
Omikhleia, line 85: So before there was one additional table level (`{ xxx }` vs. `xxx`) now.
Do we have any idea why?

inputters/base.lua
```diff
@@ -33,22 +33,11 @@ end
...
-          document = leaf
-        end
-      end
-    end
-    if not document then
+    if tree.command ~= "sile" and tree.command ~= "document" then
       SU.error("This isn't a SILE document!")
     end
-    self:classInit(document.options or {})
+    self:classInit(tree.options or {})
```
Omikhleia, line 40: So compared to 0.14.4, this is acting on `tree` now (vs. `tree[1]` before).

Does this mean I need to change:
https://github.com/Omikhleia/markdown.sile/blob/53851473b631af24d9f1a008ce524c08a13015f4/inputters/markdown.lua#L354

From (working in 0.14.4)
```
tree = { { tree, command = "document", options = { class = "markdown" } } }
```
To (maybe working on 0.14.x, x > 5)
```
tree = { tree, command = "document", options = { class = "markdown" } }
```
?

Omikhleia:
#1637 was also raising a possible discrepancy between the AST expectations (`lno`) and inputfilters (`line`). Any guidance on this aspect?

inputters/xml.lua
alerque: Last night I didn't have any idea why. Today I remembered it was prep work for a future where we could process multiple documents in one file. Also the Lua inputter is already capable of using this, the XML and SIL inputters just have the extra wrapper as future proofing and don't make use of it yet.

I should not have removed it in this PR.

inputters/base.lua
alerque: No, I'm reverting this. It was/is a mechanism for passing multiple documents in the future.

alerque:
> #1637 was also raising a possible discrepancy between the AST expectations (`lno`) and inputfilters (`line`). Any guidance on this aspect?

See the issue for my commentary on that side of things, and the new commit here for what I think is the fix for inputfilters related variable naming mismatches.

Issue #1631, opened 2022-11-26
Using qualified language names instead of mere language code


Omikhleia:
Most of the code (and the manual also states it) assumes `document.language` is an ISO 639 language code (e.g. `fr`, `en`...). There are a number of cases where this is not sufficient for actual typography.
- Hyphenation. E.g. Serbian is "sr", which currently has Cyrillic hyphenation pattern, but in written form the language is "digraphic", using the Cyrillic ("sr-Cyrl") or Latin ("sr-Latn") script. Likewise for Azeri, and a bunch of others.
- Internationalization (`fluent` and friends)
  - For the same reason as above, obviously
  - But also because various regional use of a base language might have other habits. E.g. "es" and "pt" _vs_ resp. "es-AR", "pt-BR", etc.
- Number formatting #1630 (and, slightly related, #1248 - for cases when either latin digits or a different native script are used)
- Smart typography. E.g. see [smartquotes.sile](https://github.com/Omikhleia/smartquotes.sile) (a dependency of markdown package): "fr-FR" and "fr-CH" would need being distinguished, but also "en-US" and "en-UK".
- ...

This also indirectly relates to #1367 and #1157.

alerque:
Yes, all true. In fact I thought I had an issue for tracking this but I don't see it. I suppose [this comment](https://github.com/sile-typesetter/sile/issues/1368#issuecomment-1100728951) is what I was thinking of.

The next step is probably to figure out if changing document.language to a region qualified language identifier by default is going to be a breaking change.

Omikhleia:
Breaking for the user? No, it shouldn't.
I've been using "en-GB" and "fr-CA" for the fun, with just one clever but ugly hack to the exiting code.
The question now is how far are we ready to go for refactoring the internal logic of several things to avoid the really ugly hack... I can push a draft branch by the week-end, annotated with comments, if you think that might help ascertaining the problem and finding a good solution.

alerque:
> Breaking for the user? No, it shouldn't.
> I've been using "en-GB" and "fr-CA" for the fun,

[...]

Issue #1368, opened 2022-04-16
You put what in my cannon(ical language)?


alerque:
Having introduced a CLDR database in #675, I started using it to verify language settings.

Then I spotted the `\font` function does it's own form of verification. I'll let it speak for itself though because I have no idea what it thinks it is doing:

```console
$ sile
SILE v0.12.4 (Lua 5.4)
> icu = require("justenoughicu")
> icu.canonicalize_language("jaberwalkie")
en
> icu.canonicalize_language("ukuaoeu")
ukuaoeu
```

c.f. #1367 for related brain teasers setting languages.

alerque:
Also we've short-circuited the distinction between languages and locales. To some extend we have scripts separated, but we cross-wire that with languages sometimes too.

Omikhleia:
`jaberwalkie` is not a valid (parsable) Locale format for ICU (... there are constraints on field lengths), so it fallbacks to "en".
`ukuaoeu` is valid. It perhaps doesn't exist, but it's canonical form. What was the expectation in this issue?

alerque:
Interesting. I guess the expectation is that we always either get a KNOWN valid language code or "und" in the event it doesn't exist.

Issue #276, opened 2016-02-22
Canonicalize language tags


brawer:
It might be a good idea to bring user-supplied language tags into a canonical form. For example, normalize casing (`fr-ca` → `fr-CA`); rewrite deprecated tags to their modern equivalents (`iw` → `he`, `i-klingon` → `tlh`); remove redundant subtags (`en-Latn` → `en`); and so on. The full logic is quite involved, but ICU has an implementation. Since SILE already uses ICU for other purposes, here’s a snippet that shows how this can be done by calling ICU.

After implementing this, a SILE input file could interchangeably write `\font[language=iw]`, `font[language=he]` or `font[language=he-IL]`. Internally, these equivalent variants would all get mapped to `he`, for example when choosing hyphenation patterns.

```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include "unicode/utypes.h"
#include "unicode/uenum.h"

char* canonicalizeLanguage(const char* lang) {
  char locale[200], minimized[200], result[200];
  UErrorCode error = 0;
  uloc_forLanguageTag(lang, locale, sizeof(locale), NULL, &error);
  if (!error) {
    uloc_minimizeSubtags(locale, minimized, sizeof(minimized), &error);
  }
  if (!error) {
    uloc_toLanguageTag(minimized, result, sizeof(result),
		       /* strict */ 1, &error);
  }
  if (!error) {
    return strdup(result);
  } else {
    return strdup("und");
  }
}

int main(int argc, const char** argv) {
  int i;
  for (i = 1; i < argc; ++i) {
    char* lang = canonicalizeLanguage(argv[i]);
    printf("%s --> %s\n", argv[i], lang);
    free(lang);
  }
  return 0;
}
```

PR #1641, opened 2022-12-02
Accept and resolve BCP47 language names (WIP)

Omikhleia:
Some "food for thought" regarding #1631 

I fancied to check what adding `\language[main=en-GB]`[^1] to the SILE manual would imply.
Regard this a (kind of working but ugly) draft attempt.

It raises some interesting questions for the future:
- Technical debt and/or code smell (apparently unused setting `font.script`, remainders of no-longer properly initialized `SILE.languageSupport.languages`, etc.)
- Architecture design and internal API consistency (too many things with too many side effects stuffed in language resource file, hard to manage state-oriented API of fluent, uncertain need for cldr in the current implementation, some language files return a table while most do not, etc.)

How to move forward certainly requires a longer and tougher discussion.

[^1] Or anything else you'd like. I played with "sr-Latn", which currently resolves to "sr" (= sr-Cyrl actually), no real interest obviously, except to see chapter titles in Cyrillic and hyphenations not working for the English :)

### DIFF
--- a/core/font.lua
+++ b/core/font.lua
@@ -1,4 +1,7 @@
-local icu = require("justenoughicu")
+-- BEGIN OMIKHLEIA HACKLANG
+-- No longer required here, see further below
+--   local icu = require("justenoughicu")
+-- END OMIKHLEIA HACKLANG
 
 local lastshaper
 
@@ -20,19 +23,36 @@ SILE.registerCommand("font", function (options, content)
   if options.features then SILE.settings:set("font.features", options.features) end
   if options.direction then SILE.settings:set("font.direction", options.direction) end
   if options.language then
-    if options.language ~= "und" and icu and icu.canonicalize_language then
-      local newlang = icu.canonicalize_language(options.language)
-      -- if newlang ~= options.language then
-        -- SU.warn("Language '"..options.language.."' not canonical, '"..newlang.."' will be used instead.")
-      -- end
-      options.language = newlang
-    end
+    -- BEGIN OMIKHLEIA HACKLANG
+    -- Commented out. Too soon and or at discrepancy with the "language" command anyway.
+    -- So let document.language is what the user wants, and loadLanguage take care of it
+    --
+    -- if options.language ~= "und" and icu and icu.canonicalize_language then
+    --   local newlang = icu.canonicalize_language(options.language)
+    --   -- if newlang ~= options.language then
+    --     -- SU.warn("Language '"..options.language.."' not canonical, '"..newlang.."' will be used instead.")
+    --   -- end
+    --   options.language = newlang
+    -- end
+    -- END OMIKHLEIA HACKLANG
     SILE.settings:set("document.language", options.language)
-    fluent:set_locale(options.language)
+    -- BEGIN OMIKHLEIA HACKLANG
+    -- Commented out. This is BAD design:
+    -- Fluent maintains a global state (locale) but we could be in a temporary environment
+    -- if the font command has contents, and we'd need this to remain scoped.
+    -- So we should act here!
+    --     fluent:set_locale(options.language)
+    -- END OMIKHLEIA HACKLANG
     SILE.languageSupport.loadLanguage(options.language)
   end
   if options.script then SILE.settings:set("font.script", options.script)
   elseif SILE.settings:get("document.language") then
+    -- BEGIN OMIKHLEIA HACKLANG
+    -- BAD STUFF WARNING: This SILE.languageSupport.languages[] is broken, (nearly always returning nil),
+    -- see comment in languages.lua
+    -- ON THE OTHER HAND, setting font.script doesn't seem to be used in any sensible way...
+    --  Possible code smell here.
+    -- END OMIKHLEIA HACKLANG
     local lang = SILE.languageSupport.languages[SILE.settings:get("document.language")]
     if lang and lang.defaultScript then
       SILE.settings:set("font.script", lang.defaultScript)
--- a/core/languages.lua
+++ b/core/languages.lua
@@ -1,5 +1,9 @@
 local loadkit = require("loadkit")
-local cldr = require("cldr")
+
+-- BEGIN OMIKHLEIA HACKLANG
+-- Disabled for now, see further below.
+--   local cldr = require("cldr")
+-- END OMIKHLEIA HACKLANG
 
 loadkit.register("ftl", function (file)
   local contents = assert(file:read("*a"))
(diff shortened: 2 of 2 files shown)


core/font.lua
```diff
@@ -20,19 +23,36 @@ SILE.registerCommand("font", function (options, content)
...
+    --   -- if newlang ~= options.language then
+    --     -- SU.warn("Language '"..options.language.."' not canonical, '"..newlang.."' will be used instead.")
+    --   -- end
+    --   options.language = newlang
+    -- end
+    -- END OMIKHLEIA HACKLANG
     SILE.settings:set("document.language", options.language)
-    fluent:set_locale(options.language)
+    -- BEGIN OMIKHLEIA HACKLANG
+    -- Commented out. This is BAD design:
```
alerque, line 40: As you note Fluent-lua does maintain state: the current locale—and also the messages loaded in each locale. Since SILE can and does flip back and forth between calling things for different locales rather than just setting one and sticking to it, the only way the Fluent API should be used from SILE at this point is to *always always* set the locale before all message operations. That is being done everywhere else and should be done here too. It should never be used to read or write a value without explicitly setting a locale.

core/languages.lua
```diff
@@ -9,30 +13,124 @@ end)
...
+    -- BEGIN OMIKHLEIA HACKLANG
+    -- Either done too soon or plain wrong (a BCP47 language can e.g. contain a script for instance, sr-Latn
+    -- and I don't see that in https://github.com/alerque/cldr-lua/blob/master/cldr/data/locales.lua
+    -- I don't really know why we would want that CLDR filtering here BTW...
+    -- language = cldr.locales[language] and language or "und"
+
+    -- The user may have set document.language to anything, let's ensure a canonical
+    -- BCP47 language...
+    if language ~= "und" then
+      language = icu.canonicalize_language(language)
```
alerque, line 63: Does this actually parse and match the closes canonical language or just validate the form (see [these comments](https://github.com/sile-typesetter/sile/issues/1368#issuecomment-1327941413))? If the latter I would propose we actually need the former, and the right place to add that functionality would actually be the [cldr-lua](https://github.com/alerque/cldr-lua) library not SILE.

core/languages.lua
```diff
@@ -1,5 +1,9 @@
 local loadkit = require("loadkit")
-local cldr = require("cldr")
+
+-- BEGIN OMIKHLEIA HACKLANG
+-- Disabled for now, see further below.
+--   local cldr = require("cldr")
```
alerque, line 5: See comment below, if ICU actually does what we need here that's fine, otherwise cldr-lua should be extended to provide the necessary functionality.

core/languages.lua
```diff
@@ -9,30 +13,124 @@ end)
...
+--    nil if not callback could process the language
+--    language, res if a callback could (returning the matched language pattern and the
+--    result of the callback)
+local function forLanguage(langbcp47, callback)
+  while langbcp47 do
+    local res = callback(langbcp47)
+    if res then
+      return res, langbcp47
+    end
+    langbcp47 = langbcp47:match("^(.+)-.*$") -- split at dash (-) and remove last part.
```
alerque, line 36: I think this matching/parsing bit needs to be in cldr-lua.

core/languages.lua
```diff
@@ -9,30 +13,124 @@ end)
...
-    local langresource = string.format("languages.%s", language)
-    local gotlang, lang = pcall(require, langresource)
-    if not gotlang then
-      SU.warn(("Unable to load language feature support (e.g. hyphenation rules) for %s: %s")
-        :format(language, lang:gsub(":.*", "")))
+
+    -- BEGIN OMIKHLEIA HACKLANG
+    -- We need to find language resources for this BCP47 identifier, from the less specific
+    -- to the more general.
+    local langresource, matchedlang = forLanguage(language, function (lang)
```
alerque, line 73: This whole specific -> general fallback parsing is something CLDR should take care of. The only catch is we need to track *TWO* sets of available resources (SILE Lua libraries and Fluent messages) and take a specific request and fall back to first available general case for *both* (either in sync or first available for each).

core/languages.lua
```diff
@@ -9,30 +13,124 @@ end)
...
+        require(string.format("i18n.%s", matchedi18n))
+      end
     end
-    if type(lang) == "table" and lang.init then
-      lang.init()
+
+    -- Most language resource files act by side effects, directly tweaking
+    -- SILE.nodeMarkers.xx, SILE.hyphenator.languages.xx, SU.formatNumber.xx (etc.? Doh)
+    -- BUT some don't do that exactly AND return a table with an init method...
+    -- Unclear API discrepancy, heh.
```
alerque, line 129: Yes, the ones that act by side effect predate my attempts to bring some modularity to the interfaces. We need to switch to a module system that can subclass other modules when desired. We've done it for document classes, packages, inputters, outputters, typesetters and more: but not yet languages. We need to.

core/font.lua
Omikhleia: (I don't like state-maintaining APIs, and think it could have been avoided.)

_Notwithstanding_, it doesn't have to be done _here_ (in `\font`), that's the point of this commenting out:
- `SILE.languageSupport.loadLanguage()` does it anyway _when it needs to load messages_.
- The `\fluent` command does it when invoked for _reading_ and rendering a message

alerque: A longer term proper usage here would be to instantiate and embed the correct Fluent messages class inside languageSupport when a language is loaded. The only "state" we're actually making use of is a toggle switch for which locale to access, the right way is to have each SILE language support module have it's own instance with the correlated locale.

core/languages.lua
Omikhleia: It does what is implemented in the SILE wrapper i.e. several things:
- Convert (possibly poor-formatted) BCP47 to correct Locale format.
  - This does _not_ perform any kind of validation. E.g. `ukuou` remains `ukuou` (it's _not_ forbidden by any standard, actually, to design "new language names"); `en-Cyrl` becomes `en_Cyrl` (well, English in Cyrillic script might not really exist, but likewise, why not).
  - It takes care of normalizing the format (capitals, etc.). E.g. `en-us` becomes `en_US`
  - Unrecognized elements are skipped
  - Note that it has actually some provision for scripts, variants, collations, etc. `fr-u-ks-level2-kn` becomes `fr@colnumeric=yes;colstrength=secondary`... I have _not_ taken advantage of this in the collation-enabled sorting, but we could possibly... (there's an interesting question then, but it's a slightly different topic).
- Then _minimizes_ it according to the "canonical" rules for subtags, e.g. (refer to ICU for more details, these are just examples)
  - `en_US` becomes `en` (that's the default implication), while `en_GB` remains itself.
  - Likewise `fr_FR` becomes `fr` (but obviously `fr_CA` etc. stay)
  - `sr_Cyrl` becomes `sr` (again, the default implication is Cyrillic script), while `sr_Latn` stays, etc.
  - Likewise `zh_Hans` = `zh` (the norm, again), and `zh-Hant` stays itself. 
- Convert back to BCP47
  - So we are back with a well formed minimized or "canonical" BCP47

In the end, to keep the same kind of examples:
`en-us` => `en` 
`en-gb` = `en-GB` 
`sr-Cyrl` => `sr`
`sr-Latn` => `sr-Latn`
`fr-u-ks-level2-kn` = `fr-u-kn-true-ks-level2` (yep, it also "normalizes" the optional ICU BCP47 extensions)

All of these are "valid".

What you are asking is something else, the _presumed_ validity of a known language... Well, it dosen't do that.
`ukuo-Kuou` passes (it would be the "Ukuo" language in "Kuou" script. I just invented it)
`en-Cyrl` passes, as stated (it would be the English language in Cyrillic script. Perhaps someone even invented it one day).

But do we have to really care about non-existing combinations?
- They wouldn't have any hyphenation rules anyway
- They wouldn't have any i18n translation keys... Erm, wait, until someone decides to provide them ;-)

Hopefully maybe...
`tlh-Piqd` passes... that's Klingon in the pIqaD script. Yep, it does exist "officially", being even registered in ISO 15924.

This being said, if you want to filter "invalid" languages in **cldr-lua**, you have huge tables to maintain.
- All scripts from ISO 15924 https://www.unicode.org/iso15924/iso15924-codes.html
- Lots of languages, regional variants, etc. https://icu4c-demos.unicode.org/icu-bin/locexp

Currently **cldr-lua** is aligned or CLDR 36 from 2019? We are in 2022, use CLDR 42 at least...

core/languages.lua
Omikhleia: I'd perhaps suggest splitting hyphenation, node breaking and number formatting as 3 separate resources.

core/languages.lua
Omikhleia: Wherever it eventually ends, the utility is also useful for other packages. E.g. I wish I had it for my "smartquotes" package.

core/languages.lua
Omikhleia: The discussion on ICU "canonicalize" also relates to https://github.com/sile-typesetter/sile/issues/276 (when it was implemented).

alerque: Thanks, that helps a lot!

> But do we have to really care about non-existing combinations?

No. Nor do we care about invalid languages. By the same token we don't care about **valid** languages if we don't have any support for them. What we want is to fall back to the closest thing *we support* in a given context, and that might not be the same for all contexts. A font might support a more specific locale than we have different hyphenation rules or localized strings for.

core/languages.lua
alerque: The ICU bits should probably get moved from the font code to some base language support module, and yes should be setup in a way that every package can use them. This bit which isn't really ICU still looks like it goes in CLDR to me and could also return information about what the segments *mean*.

core/languages.lua
alerque: Why not just methods on a class? If a language doesn't provide a method it would fall back to the parent class.

Omikhleia: Hyphenations at least could be handled differently: it's hard to reconstruct them from the TeX sources (or whatever other sources) and update them in the middle of the rest of the code. Ideally we should be able to script the conversion of the patterns to a Lua table (I guess that's what was done initially), crediting the original sources, etc., and keep our node breaking / number formatting / etc. apart. That'd also make maintenance easier.

Omikhleia: (But otherwise, class(es) and inheritance would be better indeed than the current logic.)

alerque: I think we're talking about different things. I was thinking of the API structure not the file layout. If hyphenation data comes from an external source I think it would be totally fine to keep the raw data in a dedicated file and require it up from wherever we put our code.

Omikhleia: I was embarrassed by the current file layout too, indeed. But if we are heading towards one class per file, we have the best of all worlds ;-)

alerque:
I think there are two important next steps for SILE.

One is having the tooling worked out to build Rust functions to both crates & Lua rocks (so SILE can still be used as a library and we have a two way bridge to implement anything however works best). That is kind of held up waiting on mlua v0.10 because it is going to make that effort much more ergonomic.

The other one is this. Handling everything with full BCP-47 locales internally and falling back gracefully to shorter ISO language codes where necessary has became what I see as the next big push I need to make. I don't know whether we can/should do this separately from the other item above, but I'm going to be looking into it.

Omikhleia:
> I think there are two important next steps for SILE.

I've started disentangling the hyphenation patterns in #2102 (and see also #2104 for what could be the next logical step here - I've kept those in a branch for now).

One implication is that the `language/xxx` files are much simpler (only some bootstrapping logic), some maybe this comment is then easier to address: https://github.com/sile-typesetter/sile/pull/1641#discussion_r1038222440

(So as to have a homogeneous implementation there, while currently some return tables, some affect globals, etc.)

alerque:
Thanks for the pointers on what end of the stick to grab first, I'll go start digging. (I'm also expecting a son to be born in the next few days so don't be surprised if I drop off the radar for a while.)

alerque:
Nevermind I see you already have it posted here, I just missed it.

core/font.lua
```diff
@@ -20,19 +23,36 @@ SILE.registerCommand("font", function (options, content)
...
+    --     fluent:set_locale(options.language)
+    -- END OMIKHLEIA HACKLANG
     SILE.languageSupport.loadLanguage(options.language)
   end
   if options.script then SILE.settings:set("font.script", options.script)
   elseif SILE.settings:get("document.language") then
+    -- BEGIN OMIKHLEIA HACKLANG
+    -- BAD STUFF WARNING: This SILE.languageSupport.languages[] is broken, (nearly always returning nil),
+    -- see comment in languages.lua
+    -- ON THE OTHER HAND, setting font.script doesn't seem to be used in any sensible way...
```
alerque, line 53: The *script* parameter isn't used for all languages or fonts, but some locale and font combinations do make use of it and the shaper then needs this info to correctly shape them. We're passing this on to HarfBuzz if it is set, that's all we need to do with it.

core/font.lua
alerque: This has been addressed since this PR and comments were created.

core/languages.lua
```diff
@@ -9,30 +13,124 @@ end)
...
+    --   - so does languages/kn.lua
+    --   - and also packages/complex-spaces
+    -- END OMIKHLEIA HACKLANG
   loadLanguage = function (language)
     language = language or SILE.settings:get("document.language")
-    language = cldr.locales[language] and language or "und"
+    -- BEGIN OMIKHLEIA HACKLANG
+    -- Either done too soon or plain wrong (a BCP47 language can e.g. contain a script for instance, sr-Latn
+    -- and I don't see that in https://github.com/alerque/cldr-lua/blob/master/cldr/data/locales.lua
+    -- I don't really know why we would want that CLDR filtering here BTW...
```
alerque, line 57: From memory (obviously I'll be testing this here) one of the reasons for this is because using `icu.canonicalize_language()` gets us the canonical language code, but it does not validate that what we got could even be recognized, and just passes through garbage if you git it garbage. If my recollection is correct here, we should be doing both. First trying to parse a canonical name, then checking if that is a known language in the database and hence we can reasonably be expected to know anything about it.

alerque (commented):
Review comments are kind of rambling, I'm just poking around. I rebased your commit as best I could on master. I think it's broken now because I didn't rewire up the bits that changed, just figured out what hunks to pull out of the merge conflict. I'll keep going on it trying to sort out the parts.

PR #1651, opened 2022-12-10
Refactor typesetters as independent module and fix grid bugs

### LINKED ISSUE #886: Manual switches typesetters midstream
While testing #885 I tried applying it to the manual just to see what, if anything, blew up. Interestingly setting it at the top of the document worked up until p38 (section 5.14) where it abruptly switches to a new typesetter and looses the settings. I don't know why, and it seems like whatever funky typesetting things that section might want to do it should revert to the original one rather than staying in the new typesetter until the end of the document.

### LINKED ISSUE #1430: Where is the first baseline on a new page (alignment, grid or not)?
The first line on a page has a baseline depending on its actual height.

In regular typesetting:
if a page starts with a line with, say, no ascenders or capitals, it will not line up with another page where those would be present.

In grid typestting:
I am not too sure how the grid typesetting is supposed to work, but it seems to be the case here too.

I wonder if this was intended/expected.

(Admittedly, it seldom occurs in normal languages with full sentences...)

**Scenario A**
```
\neverindent
\nofolios
\font[size=50pt]
Some ascenders or capitals here...

\vfill
\break

none ov em
\end{document}
```

**Scenario B** (same but with grid typesetting)

```
\script[src=packages/grid]
\neverindent
\nofolios
\font[size=50pt]
\grid[spacing=60pt]
Some ascenders or capitals here...

\vfill
\break

none ov em
\end{document}
```

**Observed result**

Scenario A:

image

Scenario B = Likewise in grid typesetting:

image

alerque:
Fixes #1430
Fixes #886


alerque:
This should probably be expanded to fix #886 before merging.

PR #1655, opened 2022-12-13
bibtex package: Fix \cite{key}

raphCode:
Really tiny code change to make the bibtex packe work with LaTeX-style citations.
Before, `\cite{abc}` errored with:
`Using code at: ./packages/bibtex/init.lua:98: attempt to concatenate a table value (field 'key')`

Unsure if this is the right approach to fix this, so I am open to further suggestions.

### DIFF
--- a/packages/bibtex/init.lua
+++ b/packages/bibtex/init.lua
@@ -90,7 +90,7 @@ function package:registerCommands ()
   end)
 
   self:registerCommand("cite", function (options, content)
-    if not options.key then options.key = content[1] end
+    if not options.key then options.key = SU.contentToString(content) end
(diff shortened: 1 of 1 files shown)


raphCode:
Also, minor UI issue:
Citing an invalid key results in:
`Unknown reference in citation abc at ...: in \cite[key="abc"]`
while the actual command used could also have been `\cite{abc}`

raphCode:
Would this be a better solution?
`options.key = SU.contentToString(content)`
Like in the next command, "reference"

alerque (approved):

alerque (approved):
Thanks for taking the time to contribute, this does look right to me.

Issue #1265, opened 2021-10-29
OpenType FeatureVariations have no effect


ctrlcctrlv:
This is a bit different than #1182, because this doesn't require instancing or anything—we should still be applying the OT FeatureVariations even without that.

I tried this patch to get SILE to start setting the opsz OT FeatureVariation for use in FRB American Cursive v2, but it doesn't work:
```diff
diff --git a/src/justenoughharfbuzz.c b/src/justenoughharfbuzz.c
index 5ac8d61b..fd559247 100644
--- a/src/justenoughharfbuzz.c
+++ b/src/justenoughharfbuzz.c
@@ -178,6 +178,11 @@ int shape (lua_State *L) {
     hb_face_t* hbFace = hb_face_create (blob, font_index);
     hbFont = hb_font_create (hbFace);
     unsigned int upem = hb_face_get_upem(hbFace);
+    hb_variation_t hbVar;
+    hbVar.value = point_size;
+    hbVar.tag = HB_TAG('o', 'p', 's', 'z');
+    printf("hbVar.value : %g\n", hbVar.value);
+    hb_font_set_variations(hbFont, &hbVar, 1);
     hb_font_set_scale(hbFont, upem, upem);
 
     if (can_use_ot_funcs(hbFace)) {

```

Even though my code there matches what's used in the Harfbuzz tests, I don't get my numbers like I expect:

**Chrome (OK)**
![image](https://user-images.githubusercontent.com/838783/139399773-3b3ba4ec-2e10-45af-b8e5-8a5dd78e7894.png)

**SILE (broken)**
![image](https://user-images.githubusercontent.com/838783/139399920-9ee55988-c784-4f6e-b305-4fe7b4e95159.png)

ctrlcctrlv:
I give up on this one, but the solution has to be simple. It's just eluding me, and I have no more time to look. But I really need this fixed for proofing purposes.

So…you know what's coming…drum roll please…(lol) :drum: :drum: 


**The Modular Font Editor K Foundation, Inc. has selected this issue (sile-typesetter/sile#1265) as requiring particular attention.**

It offers a bounty of **$350** for _either_ a PR merger which closes this bug _or_ a patch which fixes the issue described in the original post. @ctrlcctrlv will shepherd the PR if you don't want to commit to that.

If you see this before it appears on foundation.mfek.org, do not worry, it takes some time for that website to update.
```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

[...]

PR #1662, opened 2022-12-18
fix(shaper): Respect variations when shaping (#1265)

khaledhosny:
Use `hb_shape_full()` instead of fiddling with shape planes as `hb_shape_plan_create_cached()` ignores variations set on the font (one needs to use `hb_shape_plan_create_cached2()`), but `hb_shape_full()` does all that for us and there is no need to re-implement it.

Fixes https://github.com/sile-typesetter/sile/issues/1265

### DIFF
--- a/src/justenoughharfbuzz.c
+++ b/src/justenoughharfbuzz.c
@@ -152,8 +152,6 @@ int shape (lua_State *L) {
     if (strlen(shaper_list_string) > 0) {
       shaper_list = scan_shaper_list(shaper_list_string);
     }
-    hb_segment_properties_t segment_props;
-    hb_shape_plan_t *shape_plan;
 
     hb_direction_t direction;
     hb_feature_t* features;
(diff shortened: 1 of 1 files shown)


ctrlcctrlv (approved):
Wow, thanks Khaled. I was right that I'd never figure this one out on my own, I would not have ever thought to look for a `2` version of an HB function.

If you want the payout mentioned in #1265, please just email me ([copypaste@kittens.ph](mailto:copypaste@kittens.ph)) and we can arrange it off-tracker. If I were you I'd submit the `opsz` patch as a separate PR because that is what directly closes #1265, and I think it's fine to put your `opsz` patch in because OT FeatureVariations and interpolation are different enough that we can fix one without necessarily fixing the other.

alerque (approved):

alerque:
Thank you for the contribution (especially of expertise!).

Issue #1182, opened 2021-07-19
Variable font support


simoncozens:
Lack of variable font support is a problem. We know *why* there's a lack: putting variable fonts into a PDF requires instantiation, which we don't currently have the tools to do. The Harfbuzz subsetter isn't quite there yet. But: I have recently discovered that Skia is able to produce PDF files with text from variable fonts. We need to do at least one and probably all of the following:

* Check upstream libtexpdf and see if they have variable font support there yet or not.
* Work out how Skia does its instantiation/PDF generation.
* If it's easy to implement, maybe implement it and contribute upstream.
* If not, consider writing a Skia outputter backend.

There *are* [Skia Lua bindings](https://chromium.googlesource.com/skia/+/chrome/m49/site/user/special/lua.md) but never prepackaged and they look tricky to build and doubly tricky to do so in a portable way. So while writing the Skia outputter itself would be fairly straightforward, getting the bindings buildable would not be.

simoncozens:
OK, so I found out a few things:

* I found out how Skia does it. They just instantiate the glyf table using the gvar table and a location. In theory, not difficult, but...
* Adding this to libtexpdf is possible, but it would be a nightmare, because libtexpdf does all the table parsing by hand and the gvar table is a real pain to implement. (I know, I've done it once.)
* I'm kinda working on a variable font instantiate in Rust. That might actually be our best bet.

alerque:
Great! Keep the Rust library as a separate project and version it properly, but we can definitely make use of that. Unrelated and not a requirement to make use of your idea since we can call into a Rust library from Lua and/or C, but I've also been thinking about swapping out a few bits of the SILE core with Rust—in such a way that the user could still use Lua replacements to override anything they want but just to crank some speed out of the slower iteration points.

simoncozens:
I have also been thinking this way, especially with regard to the line breaking engine.

ctrlcctrlv:
```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

The Modular Font Editor K Foundation, Inc. has selected this issue as requiring particular attention.

It offers a bounty of $3,000 for a PR merger which closes this bug.

If you see this before it appears on foundation.mfek.org, do not worry, it takes some time for that website to update.
-----BEGIN PGP SIGNATURE-----

iHUEAREIAB0WIQSY8o92dHASn747BUziFU3Rocd7iwUCYmHi8QAKCRDiFU3Rocd7
[...]

PR #1666, opened 2022-12-23
Correct font-variants using opsz axis

alerque:
Follow up to #1662 with patch mentioned in #1265.

### DIFF
--- a/src/justenoughharfbuzz.c
+++ b/src/justenoughharfbuzz.c
@@ -178,6 +178,9 @@ int shape (lua_State *L) {
     unsigned int upem = hb_face_get_upem(hbFace);
     hb_font_set_scale(hbFont, upem, upem);
 
+    hb_variation_t opsz = { HB_TAG('o', 'p', 's', 'z'), point_size };
+    hb_font_set_variations(hbFont, &opsz, 1);
+
     if (can_use_ot_funcs(hbFace)) {
       hb_ot_font_set_funcs(hbFont);
     } else {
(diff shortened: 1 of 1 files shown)


ctrlcctrlv (approved):

simoncozens:
OK, so I've been pretty absent all year and I probably don't have much right to speak into this, but this seems like a strange special case: one specific axis instead of arbitrary axes. I know we're not currently able to do *much* with variable fonts until we work out instantiation and putting them into a PDF, so we're just making GSUB variations work, but I would feel more comfortable with a plan to pass arbitrary designspace locations from the user into the shaper rather than hacking one axis in to make one of Fred's fonts work...

alerque:
I think that desire is reasonable but also covered by other issues. Passing all axis data is definitely on the list but also basically equivalent of variable font support. In the mean time I don't at all mind dabbling in support like this even without the whole enchilada.

ctrlcctrlv:
@simoncozens `opsz` will always need to be special cased, as you cannot get it directly from the input given to the other axes, it comes from the font's size; most (free) applications, if they support `opsz`, don't even let you set the axis independently of the font's size.

ctrlcctrlv:
(So what I'm trying to say is even though I benefit the most from this patch, this is a sane way to do it, and when we add generic varfont axis setting then users can change the `opsz` independently of the font's size if they so choose.)

khaledhosny:
I’m working on #1182 and will send a PR later today, and I used a slightly different approach to this.

PR #1672, opened 2022-12-26
Variable fonts support

### LINKED ISSUE #1182: Variable font support

khaledhosny:
Fixes #1182

This uses HarfBuzz subsetter to instanciate the font, saves it to a temporary file and passes it to libtextpdf for embedding into PDF file.

Fonts with CFF2 table are not yet supported and will result in an error.


shapers/harfbuzz.lua
```diff
@@ -42,16 +42,23 @@ function shaper:shapeToken (text, options)
     SU.warn("Font family '"..options.family.."' not available, falling back to '"..face.family.."'")
   end
   usedfonts[face] = true
+  local point_size = ("%g"):format(SILE.measurement(options.size):tonumber())
+  -- If opsz axis is not explicitly set, set it to the point size.
+  local variations = options.variations
+  if not variations:find("opsz=") then
```
ctrlcctrlv: Isn't the syntax `opsz = ` valid?

khaledhosny: I suppose, and it seems `hb_variation_from_string()` accepts also `opsz 100` without equal sign. Too bad SILE keeps these as free form  strings instead of parsing them into tables.

khaledhosny: Anyway, I did some refactoring locally and this is not needed anymore.

src/justenoughharfbuzz.c
```diff
@@ -278,44 +257,110 @@ int shape (lua_State *L) {
...
+  unsigned int data_l = 0;
+  const char * data_s = NULL;
+#ifdef HAVE_HARFBUZZ_SUBSET
+
+  hb_font_t* font = get_hb_font(L, 1);
+  hb_face_t* face = hb_font_get_face(font);
+
+  if (hb_ot_var_has_data(face) &&
+      /* hb-subset does not support instanciating CFF2 table yet */
+      !has_table(face, HB_TAG('C','F','F','2'))) {
```
ctrlcctrlv, line 275: Good issue for me to possibly offer a grant for next :o)

src/justenoughharfbuzz.c
```diff
@@ -278,44 +257,110 @@ int shape (lua_State *L) {
...
+
+      hb_subset_input_set_flags(input,
+                                HB_SUBSET_FLAGS_RETAIN_GIDS |
+                                HB_SUBSET_FLAGS_NAME_LEGACY |
+                                HB_SUBSET_FLAGS_GLYPH_NAMES |
+                                HB_SUBSET_FLAGS_NO_PRUNE_UNICODE_RANGES);
+
+      /* Keep all glyphs */
+      glyphs = hb_subset_input_set(input, HB_SUBSET_SETS_GLYPH_INDEX);
+      hb_set_invert(glyphs);
```
ctrlcctrlv, line 296: Obviously not required for this, but one wonders if we couldn't get smaller PDF files by also subsetting fonts inside SILE at the end of document generation, now that we optionally include the subsetter.

I've had some issues before submitting documents to Unicode Consortium via email due to their size and used various hacks to get them down to size. I think that in a font specimen with many different instances all in use at once it'd become more important, but ultimately for this PR is premature optimization. :)

khaledhosny: libtexpdf do subset the fonts somehow, it clears up the outlines of unused glyphs but keeps the glyphs IDs (it is is probably a bit less efficient file size-wise than complete subsetting, but probably not by much). We can’t subset here because we share the instantiated file across all font invocations.

khaledhosny:
The test failures seem to be cause by outdated test expectations. How are these updated?

shapers/harfbuzz.lua
```diff
@@ -69,13 +66,21 @@ function shaper.getFace (opts)
...
+  -- Try instantiating the font, hb.instanciate() will return nil if it is not
+  -- a variable font or if instanciation failed.
+  face.tempfilename = face.filename
+  local data = hb.instanciate(face)
+  if data then
+    local tmp = os.tmpname();
+    local file = io.open(tmp, "wb")
+    file:write(data)
+    file:close()
+    face.tempfilename = tmp
```
khaledhosny, line 81: We are never delete these temporary files and they litter the tmp dir. They need to be delete after the PDF file is created, but I don’t know if SILE has facilities to run code at such stage. May be lua garbage collection can be utilized, but defining `__gc` on the font table does not seem to ever get called.

ctrlcctrlv: I think the best analog of what you're doing in the `libtexpdf` source code is `texpdf_close_images` which gets called by `texpdf_close_document`. There's already a `texpdf_close_fonts` but it doesn't delete temporary files like `texpdf_close_images`. `struct pdf_ximage` has a field called `tempfile` which when set causes `dpx_delete_temp_file` to get called on `pdf_ximage.filename`.

khaledhosny: Thanks for the suggestion, that was helpful. But since the files are not created by libtexpdf, I decided to delete them from the Lua side where they are created.

khaledhosny:
I think I’m done here, this should be ready for review/testing for real now.

ctrlcctrlv:
Got it. Do you expect color fonts will work? If so I'll give the FRBAC specimen build a spin and add the Cantarell TTF as the font used to show the labels.

khaledhosny:
Color fonts should work.

ctrlcctrlv:
I'm sorry to say that opsz variations don't seem to be working for me.

```
Dumping "dist/ttf/FRBAmericanPrint-900-GuidelinesDottedBlack.ttf" to "dist/ttf/FRBAmericanPrint-900-GuidelinesDottedBlack.ttx"...
Dumping 'GlyphOrder' table...
Dumping 'head' table...
Dumping 'hhea' table...
Dumping 'maxp' table...
Dumping 'OS/2' table...
Dumping 'hmtx' table...
Dumping 'cmap' table...
Dumping 'loca' table...
Dumping 'glyf' table...
Dumping 'name' table...
Dumping 'post' table...
Dumping 'COLR' table...
Dumping 'CPAL' table...
Dumping 'GDEF' table...
Dumping 'GPOS' table...
Dumping 'GSUB' table...
Dumping 'HVAR' table...
Dumping 'STAT' table...
Dumping 'XOLR' table...
Dumping 'fvar' table...
Dumping 'gvar' table...
```

Could it be a table check gone awry? This isn't a traditional variable font in that I only use FeatureVariations in it.

ctrlcctrlv:
Interesting, some of my fonts are causing crashes, but not all.

```
! Unexpected Lua error

stack traceback:
	/opt/share/sile/lua_modules/share/lua/5.4/pl/class.lua:20: in upvalue 'call_ctor'
	/opt/share/sile/lua_modules/share/lua/5.4/pl/class.lua:171: in field 'measurement'
	/opt/share/sile/core/font.lua:80: in function 'core.font._key'
	/opt/share/sile/core/font.lua:118: in function 'core.font.cache'
	/opt/share/sile/shapers/harfbuzz.lua:33: in function 'shapers.harfbuzz.shapeToken'
	/opt/share/sile/shapers/base.lua:94: in function 'shapers.base.createNnodes'
	/opt/share/sile/core/nodefactory.lua:250: in method 'shape'
	/opt/share/sile/typesetters/base.lua:302: in function 'typesetters.base.shapeAllNodes'
	/opt/share/sile/packages/bidi/init.lua:159: in function 'typesetters.base.boxUpNodes'
	/opt/share/sile/typesetters/base.lua:596: in function 'typesetters.base.leaveHmode'
	...	(skipping 210495 levels)
	/opt/share/sile/classes/plain.lua:98: in field '?'
	/opt/share/sile/core/sile.lua:391: in function 'core.sile.call'
	/opt/share/sile/core/sile.lua:236: in function 'core.sile.process'
	/opt/share/sile/core/font.lua:53: in field '?'
	/opt/share/sile/core/sile.lua:391: in function 'core.sile.call'
	/opt/share/sile/core/sile.lua:236: in function 'core.sile.process'
	(...tail calls...)
	/opt/share/sile/core/sile.lua:300: in function 'core.sile.processString'
	/opt/share/sile/core/sile.lua:332: in function 'core.sile.processFile'
	(...tail calls...)
	[C]: in function 'xpcall'
	/opt/bin/sile:125: in main chunk
	[C]: in ?

error summary:
	Processing at: /tmp/silespecimens/eachfontFRBAmericanCursive-1000-GuidelinesArrowsUltra.ttf.sil:20:1: in \bigskip
	Using code at: /opt/share/sile/lua_modules/share/lua/5.4/pl/class.lua:20: stack overflow

```

ctrlcctrlv:
Determined it's only the color ones that cause crashes.

ctrlcctrlv:
@khaledhosny I added a quick hack to my build script so I could see what happens when we don't crash.

```bash
for f in `find /tmp/silespecimens -iname '*.pdf'`; do
    PDFHEADERONLY='00000000: 2550 4446 2d31 2e35 0a25 e4f0 edf8 0a    %PDF-1.5.%.....'
    if [ "$(head -c 30 "$f" | xxd)" = "$PDFHEADERONLY" ]; then
        rm "$f"
    fi
done
```

It's possible I'm doing it wrong, but it's not working for me at all for some reason.

image

I expected that Cantarell to be at its heaviest weight.

Here's the .sil

```sile
\begin[papersize=9in x 3in]{document}
% Note: This file is intended to be compiled by scripts/each_font_specimen.sh which expects you to also have installed pdftk
\script[src=packages/complex-spaces]
\script[src=packages/color-fonts]
\script[src=packages/rebox]
\set[parameter=font.filename,value=/home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf,makedefault=true]
\set[parameter=font.size,value=20pt]
\frame[id=specimen,left=5%pw,right=95%pw,top=5%ph,bottom=top(footnotes)]
\frame[id=specimen_overlay,left=5%pw,right=95%pw,top=5%ph,bottom=top(footnotes)]
\frame[id=arrows,left=5%pw,right=95%pw,top=5.22%ph,bottom=top(footnotes)]
\set[parameter=shaper.spaceenlargementfactor,value=1]
\frame[id=fontname,left=5%pw,right=95%pw,top=1%ph,bottom=3% + height(folio)]
\typeset-into[frame=fontname]{
\noindent\font[size=5pt,variations=wght=1000]{FRBAmericanCursive-1000-DashedUltra.ttf}
}
\begin[filename=dist/ttf/FRBAmericanCursive-1000-DashedUltra.ttf]{font}
\par \font[size=48pt]{FRB American Cursive}
\bigskip
\par FRB American Cursive Dashed Ultra
\bigskip
\par Aa Bb Cc Dd Ee Ff Gg Hh Ii Jj Kk Ll Mm Nn Oo Pp Qq Rr Ss Tt Uu Vv Ww Xx Yy Zz
\par ÁáÀàẢảÃãẠạĂăẮắẰằẲẳẴẵẶặÂâẤấẦầẨẩẪẫẬậĐđÉéÈèẺẻẼẽẸẹÊêẾếỀềỂểỄễỆệÍíÌìỈỉĨĩỊịÓóÒòỎỏÕõỌọ
\par ÔôỐốỒồỔổỖỗỘộƠơỚớỜờỞởỠỡỢợÚúÙùỦủŨũỤụƯưỨứỪừỬửỮữỰựÝýỲỳỶỷỸỹỴỵ
\par abcdefghijklmnopqrstuvwxyz
\par 1234567890 !@#$\%&*() []{};':",./ \\ № ₿ ₱ ₹ †
\par The quick brown fox jumps over the lazy dog.
\par Ekvidu! Laŭ Zamenhof bongustas Ĉeĥuje manĝaĵo freŝa spica.
\par В чащах юга жил бы цитрус? Да, но фальшивый экземпляр!
\par Şaban Türk'tür. Bir şıpka torpido yakalar.
\end{font}
\end{document}
```

khaledhosny:
> I expected that Cantarell to be at its heaviest weight.

It works for me. Can you pull the last commit and run with `sile --debug=fonts --debug=versions` and show me the output?

khaledhosny:
> Determined it's only the color ones that cause crashes.

Please send me and fonts and documents that are failing.

ctrlcctrlv:
Cantarell.zip is in #1182.

Here's an example of a successful invocation.

```
Shapers enabled: ot, fallback
ICU support enabled

Fonts used:
dist/ttf/FRBAmericanCursive-400-Regular.ttf:0	Version 2.0;Modular Font Editor K font №1
/home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf:0	Version 0.303
/home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf:0	Version 0.303
dist/ttf/FRBAmericanCursive-400-Regular.ttf:0	Version 2.0;Modular Font Editor K font №1
SILE v0.14.6.r20-g605aed8-dirty (Lua 5.4)
</tmp/silespecimens/eachfontFRBAmericanCursive-400-Regular.ttf.sil> as sil

[fonts] Looking for ;5;400;;normal;;'wght=1000';LTR;/home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf
[fonts] Resolved font family  -> /home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf
[fonts] Looking for ;10;400;;normal;;;LTR;dist/ttf/FRBAmericanCursive-400-Regular.ttf
[fonts] Resolved font family  -> dist/ttf/FRBAmericanCursive-400-Regular.ttf
[fonts] Looking for ;48;400;;normal;;;LTR;dist/ttf/FRBAmericanCursive-400-Regular.ttf
[fonts] Resolved font family  -> dist/ttf/FRBAmericanCursive-400-Regular.ttf[1] 
[fonts] Looking for ;10;400;;normal;;;LTR;/home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf
[fonts] Resolved font family  -> /home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf
```

And here's a failed one:

```
99% 104:1=0s dist/ttf/FRBAmericanCursive-1000-Ultra.ttf                                                                                                       SILE v0.14.6.r20-g605aed8-dirty (Lua 5.4)
</tmp/silespecimens/eachfontFRBAmericanCursive-1000-GuidelinesArrowsDottedUltra.ttf.sil> as sil

[fonts] Looking for ;5;400;;normal;;'wght=1000';LTR;/home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf
[fonts] Resolved font family  -> /home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf
[fonts] Looking for ;10;400;;normal;;;LTR;dist/ttf/FRBAmericanCursive-1000-GuidelinesArrowsDottedUltra.ttf
[fonts] Resolved font family  -> dist/ttf/FRBAmericanCursive-1000-GuidelinesArrowsDottedUltra.ttf
[fonts] Looking for ;48;400;;normal;;;LTR;dist/ttf/FRBAmericanCursive-1000-GuidelinesArrowsDottedUltra.ttf
[fonts] Resolved font family  -> dist/ttf/FRBAmericanCursive-1000-GuidelinesArrowsDottedUltra.ttf
! Unexpected Lua error

stack traceback:
	/opt/share/sile/lua_modules/share/lua/5.4/pl/class.lua:20: in upvalue 'call_ctor'
	/opt/share/sile/lua_modules/share/lua/5.4/pl/class.lua:171: in field 'measurement'
	/opt/share/sile/core/font.lua:80: in function 'core.font._key'
	/opt/share/sile/core/font.lua:118: in function 'core.font.cache'
	/opt/share/sile/shapers/harfbuzz.lua:33: in function 'shapers.harfbuzz.shapeToken'
	/opt/share/sile/shapers/base.lua:94: in function 'shapers.harfbuzz.createNnodes'
	/opt/share/sile/core/nodefactory.lua:250: in method 'shape'
	/opt/share/sile/typesetters/base.lua:302: in function 'typesetters.base.shapeAllNodes'
	/opt/share/sile/packages/bidi/init.lua:159: in function 'typesetters.base.boxUpNodes'
	/opt/share/sile/typesetters/base.lua:596: in function 'typesetters.base.leaveHmode'
	...	(skipping 210495 levels)
	/opt/share/sile/classes/plain.lua:98: in field '?'
	/opt/share/sile/core/sile.lua:391: in function 'core.sile.call'
	/opt/share/sile/core/sile.lua:236: in function 'core.sile.process'
	/opt/share/sile/core/font.lua:53: in field '?'
	/opt/share/sile/core/sile.lua:391: in function 'core.sile.call'
	/opt/share/sile/core/sile.lua:236: in function 'core.sile.process'
	(...tail calls...)
	/opt/share/sile/core/sile.lua:300: in function 'core.sile.processString'
	/opt/share/sile/core/sile.lua:332: in function 'core.sile.processFile'
	(...tail calls...)
	[C]: in function 'xpcall'
	/opt/bin/sile:125: in main chunk
	[C]: in ?

error summary:
	Processing at: /tmp/silespecimens/eachfontFRBAmericanCursive-1000-GuidelinesArrowsDottedUltra.ttf.sil:21:1: in \bigskip
	Using code at: /opt/share/sile/lua_modules/share/lua/5.4/pl/class.lua:20: stack overflow
```

And here's the font that failed w/the document above: 
[FRBAC_EXAMPLE_FONT_OT_LICENSE_INVALID_DO_NOT_REDISTRIBUTE.tar.gz](https://github.com/sile-typesetter/sile/files/10313973/FRBAC_EXAMPLE_FONT_OT_LICENSE_INVALID_DO_NOT_REDISTRIBUTE.tar.gz)

ctrlcctrlv:
(Note that as mentioned I'm not seeing Cantarell at the weight I expected even when the invocation of `sile` does not crash.)

image

[eachfontFRBAmericanCursive-400-Regular.ttf.pdf](https://github.com/sile-typesetter/sile/files/10313983/eachfontFRBAmericanCursive-400-Regular.ttf.pdf)

khaledhosny:
> Cantarell.zip is in #1182.
> 
> Here's an example of a successful invocation.
> 
> ```
> Shapers enabled: ot, fallback
> ICU support enabled
> 
> Fonts used:
> dist/ttf/FRBAmericanCursive-400-Regular.ttf:0	Version 2.0;Modular Font Editor K font №1
> /home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf:0	Version 0.303
> /home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf:0	Version 0.303
> dist/ttf/FRBAmericanCursive-400-Regular.ttf:0	Version 2.0;Modular Font Editor K font №1
> SILE v0.14.6.r20-g605aed8-dirty (Lua 5.4)
> </tmp/silespecimens/eachfontFRBAmericanCursive-400-Regular.ttf.sil> as sil
> 
> [fonts] Looking for ;5;400;;normal;;'wght=1000';LTR;/home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf
> [fonts] Resolved font family  -> /home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf
> [fonts] Looking for ;10;400;;normal;;;LTR;dist/ttf/FRBAmericanCursive-400-Regular.ttf
> [fonts] Resolved font family  -> dist/ttf/FRBAmericanCursive-400-Regular.ttf
> [fonts] Looking for ;48;400;;normal;;;LTR;dist/ttf/FRBAmericanCursive-400-Regular.ttf
> [fonts] Resolved font family  -> dist/ttf/FRBAmericanCursive-400-Regular.ttf[1] 
> [fonts] Looking for ;10;400;;normal;;;LTR;/home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf
> [fonts] Resolved font family  -> /home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf
> ```

I was expecting to see the HarfBuzz version here, and some output about instanciation, but I see neither. You need harfbuzz-subset >= 6.0.0 for instanciation to happen.

> And here's a failed one:
> 
> ```
> 99% 104:1=0s dist/ttf/FRBAmericanCursive-1000-Ultra.ttf                                                                                                       SILE v0.14.6.r20-g605aed8-dirty (Lua 5.4)
> </tmp/silespecimens/eachfontFRBAmericanCursive-1000-GuidelinesArrowsDottedUltra.ttf.sil> as sil
> 
> [fonts] Looking for ;5;400;;normal;;'wght=1000';LTR;/home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf
> [fonts] Resolved font family  -> /home/fred/Workspace/cantarell-fonts/src/variable_ttf/Cantarell-VF.ttf
> [fonts] Looking for ;10;400;;normal;;;LTR;dist/ttf/FRBAmericanCursive-1000-GuidelinesArrowsDottedUltra.ttf
> [fonts] Resolved font family  -> dist/ttf/FRBAmericanCursive-1000-GuidelinesArrowsDottedUltra.ttf
> [fonts] Looking for ;48;400;;normal;;;LTR;dist/ttf/FRBAmericanCursive-1000-GuidelinesArrowsDottedUltra.ttf
> [fonts] Resolved font family  -> dist/ttf/FRBAmericanCursive-1000-GuidelinesArrowsDottedUltra.ttf
> ! Unexpected Lua error
> 
> stack traceback:
> 	/opt/share/sile/lua_modules/share/lua/5.4/pl/class.lua:20: in upvalue 'call_ctor'
> 	/opt/share/sile/lua_modules/share/lua/5.4/pl/class.lua:171: in field 'measurement'
> 	/opt/share/sile/core/font.lua:80: in function 'core.font._key'
> 	/opt/share/sile/core/font.lua:118: in function 'core.font.cache'
> 	/opt/share/sile/shapers/harfbuzz.lua:33: in function 'shapers.harfbuzz.shapeToken'
> 	/opt/share/sile/shapers/base.lua:94: in function 'shapers.harfbuzz.createNnodes'
> 	/opt/share/sile/core/nodefactory.lua:250: in method 'shape'
> 	/opt/share/sile/typesetters/base.lua:302: in function 'typesetters.base.shapeAllNodes'
> 	/opt/share/sile/packages/bidi/init.lua:159: in function 'typesetters.base.boxUpNodes'
> 	/opt/share/sile/typesetters/base.lua:596: in function 'typesetters.base.leaveHmode'
> 	...	(skipping 210495 levels)
> 	/opt/share/sile/classes/plain.lua:98: in field '?'
> 	/opt/share/sile/core/sile.lua:391: in function 'core.sile.call'
> 	/opt/share/sile/core/sile.lua:236: in function 'core.sile.process'
> 	/opt/share/sile/core/font.lua:53: in field '?'
> 	/opt/share/sile/core/sile.lua:391: in function 'core.sile.call'
> 	/opt/share/sile/core/sile.lua:236: in function 'core.sile.process'
> 	(...tail calls...)
> 	/opt/share/sile/core/sile.lua:300: in function 'core.sile.processString'
> 	/opt/share/sile/core/sile.lua:332: in function 'core.sile.processFile'
> 	(...tail calls...)
> 	[C]: in function 'xpcall'
> 	/opt/bin/sile:125: in main chunk
> 	[C]: in ?
> 
> error summary:
> 	Processing at: /tmp/silespecimens/eachfontFRBAmericanCursive-1000-GuidelinesArrowsDottedUltra.ttf.sil:21:1: in \bigskip
> 	Using code at: /opt/share/sile/lua_modules/share/lua/5.4/pl/class.lua:20: stack overflow
> ```
> 

I can reproduce this. Investigating, though since you don’t seem to have any instanciation going on, I’m suspecting it is an unrelated issue.

khaledhosny:
> I can reproduce this. Investigating, though since you don’t seem to have any instanciation going on, I’m suspecting it is an unrelated issue.

I can’t run using this font with SILE v0.14.5 either. Looks like a bug in the color-fonts package (I would have replaced this code with the better maintained HarfBuzz API (which now supports COLRv1, too), but it when I suggested using HarfBuzz API to read MATH table others didn’t like it).

khaledhosny:
If you add `--debug=color-fonts` you can see it is getting into some kind of infinite loop before exhausting lua stack.

khaledhosny:
The infinite loop is caused by `leaveHmode` in, but I don’t understand this code to debug further: 
https://github.com/sile-typesetter/sile/blob/771d87f24fa0f4599655fba23bcade15a7a5e7cb/core/font.lua#L130-L134

Also, conceptually, color support should be handled by the outputters not the shapers (I know, I wrote the initial `color-fonts` package, guilty as charged).

alerque:
An issue with looping after using color fonts sounds like the modified shaper object not getting relpaced with the original when done. I'll see if I can replicate that, but I ran into something like that recently with the fallback font stack so it sounds familiar.

alerque:
Just a couple release management notes:

1. My plan is to amend the configure setup just a little to *default* to compiling with subsetter support and erroring out it the dependency versions are not up to it. An option to disable it will be provided for distros that are substantially behind the curve, but for all future releases it will just be assumed that the target audience has all the recent goodies and others can specifiably configure a degraded feature experience.
1. I'll have to figure out how to bake in some warning message for when people try to render documents that use variable fonts but they have a SILE compiled with limited capabilities.
1. This is targeting v0.14.8 and the plan is to make that release a one-trick-pony with this as the showcase. The configure system changes in v0.14.6 turned out to be an unexpected fiasco for distro packagers and I want to get that cleaned up (and some other regressions fixed) with v0.14.7 before hitting distros with the new dependency issues this feature brings.

khaledhosny:
> 1. My plan is to amend the configure setup just a little to _default_ to compiling with subsetter support and erroring out it the dependency versions are not up to it. An option to disable it will be provided for distros that are substantially behind the curve

This is what I had originally, but since HarfBuzz 6.0.0 is too fresh (~2 weeks old), it felt a bit too much, but that is your call. Windows cmake build probably needs to be updated to build HarfBuzz 6.0.0 and the subset library as well.

> 2. I'll have to figure out how to bake in some warning message for when people try to render documents that use variable fonts but they have a SILE compiled with limited capabilities.

This is currently written as a debug message, but it can be turned into a warning if you prefer so.

khaledhosny:
We are also not testing any of this in the CI builds since they all use old HarfBuzz, so something might be needed to be done about this as well.

ctrlcctrlv:
> This is what I had originally, but since HarfBuzz 6.0.0 is too fresh (~2 weeks old), it felt a bit too much, but that is your call. Windows cmake build probably needs to be updated to build HarfBuzz 6.0.0 and the subset library as well.

Wow this might be my issue. I thought it had to be a typo and forced it to use 5.0. _Mea culpa._

khaledhosny:
> > This is what I had originally, but since HarfBuzz 6.0.0 is too fresh (~2 weeks old), it felt a bit too much, but that is your call. Windows cmake build probably needs to be updated to build HarfBuzz 6.0.0 and the subset library as well.
> 
> Wow this might be my issue. I thought it had to be a typo and forced it to use 5.0. _Mea culpa._
> 
> I applied this erroneous patch when trying to get it working...

Since `hb_subset_input_pin_axis_to_default()` and `hb_subset_input_pin_axis_location()` were made public on in 6.0.0, I’m surprised you got this to compile at all.

src/hb-utils.c
```diff
@@ -0,0 +1,138 @@
...
+    free(axes);
+    free(newCoords);
+  }
+#endif
+
+  hb_face_destroy(face);
+  hb_blob_destroy(blob);
+
+  lua_pushlightuserdata(L, font);
+  lua_setfield(L, index, "hbFont");
```
khaledhosny, line 133: We never call `hb_font_destroy()`, but since the fonts are cached and never garbage collected, this seems to not be an issue, this will be released when the program exits anyway. It should be easy to fix though (adding a method to be called from `SILE.font.finish()`.

ctrlcctrlv:
> > > This is what I had originally, but since HarfBuzz 6.0.0 is too fresh (~2 weeks old), it felt a bit too much, but that is your call. Windows cmake build probably needs to be updated to build HarfBuzz 6.0.0 and the subset library as well.
> > 
> > 
> > Wow this might be my issue. I thought it had to be a typo and forced it to use 5.0. _Mea culpa._
> > I applied this erroneous patch when trying to get it working...
> 
> Since `hb_subset_input_pin_axis_to_default()` and `hb_subset_input_pin_axis_location()` were made public only in 6.0.0, I’m surprised you got this to compile at all.

I think it still compiled because I do a lot of font devel (like you :smile:) so I had a self-built version of `harfbuzz` in `/opt/lib` with `experimental-apis` enabled. But yeah..._because of that experience_, having just rebuilt my HB like a month or two ago, I felt it was safe to force the 6.0.0 down not realizing it was intentional and there actually had been a 6.0.0 release. Really dumb of me, apologies.

khaledhosny:
I tried to get variable fonts working on the Windows CMake builds, but now it feels that these builds are unused and untested. The build passes, but the result is unusable. First issue, there is no `core/version.lua` but a `core/version.lua.in`, fixing this manually reveals that `lua-utf8` is also missing. I wonder if this is used at all?

alerque:
The CMake setup was only ever partially implemented and has since fallen out of date too. Don't worry about it. We'll decide how to handle Windows builds when somebody with expertise wants to tackle it (autotools + mingw, cmake, meson, whatever). If anything we just need to add a big warning so people don't expect it to work right now.

configure.ac
```diff
@@ -32,6 +32,33 @@ AM_CONDITIONAL([DEVELOPER], [test "x$enable_developer" = "xyes"])
...
 
+AC_ARG_WITH([harfbuzz],
+            AS_HELP_STRING([--without-harfbuzz], [Compile without HarfBuzz support for shaping backend]))
+AM_CONDITIONAL([HARFBUZZ], [test "x$with_harfbuzz" != "xno"])
+
+AM_COND_IF([HARFBUZZ], [
+    PKG_CHECK_MODULES(HARFBUZZ, [harfbuzz >= 2.7.4], [
+        PKG_CHECK_MODULES(FONTCONFIG, fontconfig, [
+            ], [
+                AC_MSG_FAILURE([Harfbuzz found but not fontconfig])
```
khaledhosny, line 43: Those two checks should be decoupled, the code that uses HarfBuzz and the code that uses FontConfig are not dependent and one should be able to build without FontConfig (e.g. on macOS). I’d make at least one font manager a requirement, so at least FontConfig or AppKit.

alerque:
I realize I'm breaking CI here, I'll get it back together. My plan is to setup a hybrid approach were some tests only run if we have support for them, then add `--without-harfbuzz-subset` to most of our test runners and use Nix to test the `--with-harfbuzz-subset` end of things as soon as [this PR clears](https://github.com/NixOS/nixpkgs/pull/207041) and makes it as far as [into unstable](https://nixpk.gs/pr-tracker.html?pr=207041).

configure.ac
alerque: Roger that, I was wondering while I did that refactoring...

tests/variable-axis.sil
```diff
@@ -0,0 +1,7 @@
+\begin[papersize=a6]{document}
+\neverindent
+\nofolios
+\font[family=Tourney,size=48pt]
+\font[variations="wght=150,wdth=120"]{Foo bar}\par
+\font[variations="wght=750,wdth=80"]{Foo bar}
```
alerque, line 6: This test is working for me as written here, but I was a bit surprised when working this up to find this did *not* work at all:

```suggestion
\font[weight=750,variations=wdth=80]{Foo bar}
```

When including *any* variations parameter value at all, the weight option gets ignored. It works fine to select any value from the weight axis right up until any other axis is specified, then it stops working an one must add the `wght` axis to the variations to get the effect back.

Is there some way we can map the axis that correspond to parameters we already have? I have to imagine this is going to affect more combinations too, such as when `style=Italic` is part of a slant axis.

alerque:
A note to self so I don't forget to deal with this: ever using `--disable-harfbuzz-subset` at configure time does not stop the code from trying if the host system's Harfbuzz has the subset library. Fixing that may have the same root cause as the `if` block that isn't working to skip the variations test.

Quite unexpectedly the overall metrics of a font testing the width axis are respected even though the actual subsetter doesn't get used! The result is a passing test that **should fail**. I guess harfbuzz is being asked for shaping metrics and doing all the right work, then when each chunk gets rendered it is rendering with the wrong font because it isn't instantiated right. I'm going to have to rethink how we test so we actually catch failures!

khaledhosny:
> A note to self so I don't forget to deal with this: ever using `--disable-harfbuzz-subset` at configure time does not stop the code from trying if the host system's Harfbuzz has the subset library. Fixing that may have the same root cause as the `if` block that isn't working to skip the variations test.
> 
> Quite unexpectedly the overall metrics of a font testing the width axis are respected even though the actual subsetter doesn't get used! The result is a passing test that **should fail**. I guess harfbuzz is being asked for shaping metrics and doing all the right work, then when each chunk gets rendered it is rendering with the wrong font because it isn't instantiated right. I'm going to have to rethink how we test so we actually catch failures!

We currently always apply variations during shaping (this what makes #1265 work even without the subsetter), but we can make applying variations conditional on the presence of the subsetter.

src/hb-utils.c
```diff
@@ -0,0 +1,138 @@
...
+
+  blob = hb_blob_create_from_file(filename);
+  face = hb_face_create(blob, face_index);
+  font = hb_font_create(face);
+  upem = hb_face_get_upem(face);
+  hb_font_set_scale(font, upem, upem);
+
+  hb_ot_font_set_funcs(font);
+
+#if HB_VERSION_ATLEAST(3, 3, 0)
```
khaledhosny, line 66: If we want to apply variations only when the subsetter is present, we can do:
```suggestion
#ifdef HAVE_HARFBUZZ_SUBSET
```

tests/variable-axis.sil
khaledhosny: The `weight` option should be respected if the font manager finds a named instance matching it (the same for other style settings we ask for in the font manager). If this is not working, then something is broken.

But generally, there might nit be a named instance matching the style and it will be ignored. What can we do it to treat `weigh`, `width` (or `stretch` if SILE supports any) and `style=italic`/`style=oblique` the same way we treat font size; set the corresponding variation axes to their values.

So, what is the list of styles we want to apply as variations?

src/hb-utils.c
```diff
@@ -91,14 +91,18 @@ hb_font_t* get_hb_font(lua_State *L, int index) {
...
+      switch (axes[i].tag) {
+        case HB_TAG('o', 'p', 's', 'z'):
+          lua_getfield(L, index, "pointsize");
+          if (lua_isnumber(L, -1)) { newCoords[i] = luaL_checknumber(L, -1); }
+          break;
+        case HB_TAG('w', 'g', 'h', 't'):
+          lua_getfield(L, index, "weight");
+          if (lua_isnumber(L, -1)) { newCoords[i] = luaL_checknumber(L, -1); }
+          break;
+        default: break;
```
khaledhosny, line 105: This code is now generalized and can be extended to set more axes.

src/hb-utils.c
khaledhosny: I have done a slightly different variation of this on 2f71ef4d90b14dc234d6c6c6efbc55ee3eac7332, but I’m not sure it is making any difference with the failed test.

alerque: Thanks. I think this is the direction I'd lake to take it. From a troubleshooting and support standpoint I think it will be clearer what to expect if we deliberately turn this on or off rather than just letting it try. Builders are going to start getting errors for HB<6 now anyway and need to explicitly disable features to continue. They will be told some things won't work at that point, and then I want to add an error when rendering documents that request access to those features. I think lumping all variations support behind the same feature flag is reasonable. I might reword the flag and help messages with that in mind though.

khaledhosny: I have done this part now.

core/sile.lua
```diff
@@ -112,10 +112,11 @@ end
 SILE.init = function ()
   -- Set by def
   if not SILE.backend then
-    if pcall(require, "justenoughharfbuzz") then
+    local status, result = pcall(require, "justenoughharfbuzz")
+    if status then
```
khaledhosny, line 116: Is this a typo or does libtexpdf backend really require justenoughharfbuzz module?

configure.ac
```diff
@@ -23,38 +23,59 @@ LT_PREREQ([2.2])
...
 
 AC_ARG_ENABLE([developer],
-            AS_HELP_STRING([--enable-developer], [Check for and enable tooling required only for SILE developers]))
+            AS_HELP_STRING([--enable-developer],
+                           [Check for and enable tooling required only for SILE developers]))
 AM_CONDITIONAL([DEVELOPER], [test "x$enable_developer" = "xyes"])
 
+AC_ARG_ENABLE([font-variations],
+            AS_HELP_STRING([--disable-font-variations],
+                           [Disable support for OpenType variations and variable fonts that requires very recent HarfBuzz]))
```
khaledhosny, line 37: Wording like `very recent` does not age well, we are probably better off spitting the required HarfBuzz version (and to avoid repeating that, we can move the required versions to variables at the top of the file, many projects do this and it helps grouping them in one place).

alerque: Good call.

core/sile.lua
alerque: By default we require libtexpdf & harfbuzz pairing if not otherwise specified, and this looks on purpose ... make usre we have a working default shaper before we assume the default backend. But the code does look fishy, I'll give it a review it a bit...

khaledhosny: At least the error message can be improved if this is on purpose. It talks about missing libtexpdf, but it is harfbuzz that is missing.

configure.ac
```diff
@@ -121,7 +121,7 @@ AM_COND_IF([DEPENDENCY_CHECKS], [
                                   [AC_MSG_FAILURE([--with-harfbuzz was given, but harfbuzz not found])]) ])
 
     AM_COND_IF([FONT_VARIATIONS],
-               [PKG_CHECK_MODULES(FONT_VARIATIONS, [harfbuzz-subset >= 6.0.0],
+               [PKG_CHECK_MODULES(HARFBUZZ_SUBSET, [harfbuzz-subset >= 6.0.0],
```
alerque, line 124: Ha sorry about this. I have it fixed locally in a more expansive rebase branch but forgot to stuff the fixup here for visibility.

khaledhosny: No problem. But I wonder why CI is still running and failing the variations test when they are not being built with variations support.

core/sile.lua
alerque: I believe the this bit of obtuse code has outlived its original purpose. I've simplified it and errors should make more sense now. We don't have a fallback for a different backend than the libtexpdf/harfbuzz combo right now. If/when we do configure flags should be able to change the default, not just a try/fail.

alerque:
I think we're getting close here. I still want an end-user facing error when trying to render a document that expects features they don't have, but I'll be looking at that now.

One for you @khaledhosny: we're failing `make busted` right now due to a change in open type font loading. I suspect the index in this line probably just changed:

https://github.com/sile-typesetter/sile/blob/master/spec/opentype_spec.lua#L19

Do you have any ideas on that?

khaledhosny:
> I think we're getting close here. I still want an end-user facing error when trying to render a document that expects features they don't have, but I'll be looking at that now.
> 
> One for you @khaledhosny: we're failing `make busted` right now due to a change in open type font loading. I suspect the index in this line probably just changed:
> 
> https://github.com/sile-typesetter/sile/blob/master/spec/opentype_spec.lua#L19
> 
> Do you have any ideas on that?
> 
> **Edit:** This is only failing for builds configured `--enable-font-variations`. With it disabled they work okay. The difference seems to be whether the Tourney font is available in FonConfig's search path. Whet it is, the `SILE.shaper:getFace({ family = "Libertinus Serif" })` request is returning a cache item Tourney! Something is cross-wired in the cache mechanism now.

I’m trying to debug this, but I’m on macOS and `make busted` assumes fontconfig is used but it dies not actually enforce this, so it my case macfonts is used and it does not find the fonts in `.fonts`. So I guess I have to disable compiling macfonts first.

ctrlcctrlv:
@khaledhosny Feel free to email me for a free Linux shell (via SSH) if you want one and it'd make debugging work easier. :-)

I already have SILE installed on most of my servers / VPS so I can easily make you an account if you want one.

ctrlcctrlv:
(Also I don't mind giving you root.)

khaledhosny:
> @khaledhosny Feel free to email me for a free Linux shell (via SSH) if you want one and it'd make debugging work easier. :-)
> 
> I already have SILE installed on most of my servers / VPS so I can easily make you an account if you want one.

Thanks, I found an easy way to force FontConfig for this test which is actually the only one that needed it.

khaledhosny:
> > I think we're getting close here. I still want an end-user facing error when trying to render a document that expects features they don't have, but I'll be looking at that now.
> > One for you @khaledhosny: we're failing `make busted` right now due to a change in open type font loading. I suspect the index in this line probably just changed:
> > https://github.com/sile-typesetter/sile/blob/master/spec/opentype_spec.lua#L19
> > Do you have any ideas on that?
> > **Edit:** This is only failing for builds configured `--enable-font-variations`. With it disabled they work okay. The difference seems to be whether the Tourney font is available in FonConfig's search path. Whet it is, the `SILE.shaper:getFace({ family = "Libertinus Serif" })` request is returning a cache item Tourney! Something is cross-wired in the cache mechanism now.
> 
> I’m trying to debug this, but I’m on macOS and `make busted` assumes fontconfig is used but it dies not actually enforce this, so it my case macfonts is used and it does not find the fonts in `.fonts`. So I guess I have to disable compiling macfonts first.

I don’t know what is going on, but it seems that calling `SILE.shaper:getFace()` directly is broken (adding a `print()` in `shapers/harfbuzz.lua:shaper.getFace()` shows that `opts` is an empty dict no matter what is passed to the function. Surprisingly, this works:
```diffdiff --git a/spec/opentype_spec.lua b/spec/opentype_spec.lua
index 7edf848b..3856bc49 100644
--- a/spec/opentype_spec.lua
+++ b/spec/opentype_spec.lua
@@ -13,7 +13,7 @@ if not fcf then return end

 describe("The OpenType loader/parser", function()
   local ot = require("core.opentype-parser")
-  local face = SILE.shaper:getFace({ family = "Libertinus Serif" })
+  local face = SILE.shaper.getFace({ family = "Libertinus Serif" })
   local font = ot.parseFont(face)

   it("should convert Microsoft-platform name strings to UTF8", function()
```

My Lua is a bit rusty to remember the exact difference between the two.

khaledhosny:
OK, so with colon `shaper` is passed as first argument, like `string.len("foo")` vs `"foo":len()`, so I think this test only worked so far by accident.

core/font.lua
```diff
@@ -106,7 +109,17 @@ local font = {
...
     if not options.hyphenchar then options.hyphenchar = SILE.settings:get("font.hyphenchar") end
+    if options.variations ~= "" and not SILE.features.font_variations then
+      SU.error([[This build of SILE was compiled with font variations support disabled,
+  likely due to not having the subsetter library included in HarfBuzz >= 6.
+  This document specifies font variations which cannot be correctly rendered.
+  Please rebuild SILE with the necessary library support. Alternatively to procede
+  anyway *incorrectly* render this document run:
+      sile -e 'SILE.features.font_variations = true' ....
+  Or modify the document to remove variations options from font commands.]])
+    end
```
khaledhosny, line 122: We already do a similar check in `shapers/harfbuzz.lua:shaper.getFace()` and it also checks for the face index (which can carry the index of a named instance). It also catches an other potential failures of instantiation (like CFF2) even which harfbuzz-subset is enabled.

shapers/harfbuzz.lua
```diff
@@ -63,19 +62,44 @@ function shaper:shapeToken (text, options)
...
+  local data = hb.instanciate(face)
+  if data then
+    local tmp = os.tmpname()
+    local file = io.open(tmp, "wb")
+    file:write(data)
+    file:close()
+    face.tempfilename = tmp
+    SU.debug("fonts", "Instanciated", _pretty_varitions(face), "as", face.tempfilename)
+  elseif (face.variations ~= "") or (bitshim.rshift(face.index, 16) ~= 0) then
+    SU.debug("fonts", "Failed to instanciate", _pretty_varitions(face))
```
khaledhosny, line 100: We can make this an error and move the message here.

configure.ac
```diff
@@ -90,40 +110,49 @@ AM_COND_IF([DEPENDENCY_CHECKS], [
...
-            AC_MSG_FAILURE([--with-harfbuzz was given, but test for harfbuzz failed])
-        fi
-    ])
+    AM_COND_IF([FONTCONFIG],
+               [PKG_CHECK_MODULES(FONTCONFIG, [fontconfig],
+                                  [],
+                                  [AC_MSG_FAILURE([Fontconfig package not found])])])
+
+    AM_COND_IF([HARFBUZZ],
+               [PKG_CHECK_MODULES(HARFBUZZ, [harfbuzz >= 2.7.4],
```
alerque, line 119: > We were not requiring any specific version, but it is unlikely that the
> code works with such old versions of HarfBuzz. Instead, require the
> newest version of HarfBuzz we can get away with, the version available
> in stable Debian/Ubuntu.

So, this is turning out to be a little bit more complicated. Stable Debian/Ubuntu sounded like a good minimum baseline so I skipped over this. It might be true 2.7.4 is the minimum for Debian, but unfortunately for Ubuntu we have LTS releases out there with 1.x releases still in support. Bionic is on 1.7.2 and Focal is on 2.6.4.

I am not sure I want to revisit this since the most *recent* supported LTS will clear this bar, but it does mean we will be dropping support for Bionic and Focal.

I think I'm okay with that. I'm also looking at what it will take to backport HarfBuzz 6 back to at least Jammy so we can have full support everywhere we support Ubuntu at all, but I haven't found the bottom of the rabbit hole yet.

alerque: No 6.x yet, but it looks like [this PPA](https://launchpad.net/~reviczky/+archive/ubuntu/harfbuzz/+packages) has at least 2.7.4 backported to all current LTS releases.

shapers/harfbuzz.lua
alerque: Will do.

khaledhosny: I just fixed it. face.variations can be nil, I just made sure it is always set.

khaledhosny:
BTW, I’m seeing:
```
Tests not enabled by configuration:
Use of uninitialized value $_ in print at ./tests/regressions.pl line 60.
⚠
```

I think this was meant to show disabled variation tests, but since they are enabled some perl code is not happy.

alerque:
As far as I understand the SILE support end of things and even test suite are all done here.

What remains on my end is...

* waiting for [this PR](https://nixpk.gs/pr-tracker.html?pr=207041) to progress until we can easily use it in CI to test the `--enable-font-variations` end of things in CI.
* sorting out how we're going to handle Ubuntu builds. Since I've been doing PPA builds that get quite a bit of usage I need to figure out if backporting a harfbuzz package is even possible without breaking the whole distro.
* getting everything currently queued up into a release.

khaledhosny:
> sorting out how we're going to handle Ubuntu builds. Since I've been doing PPA builds that get quite a bit of usage I need to figure out if backporting a harfbuzz package is even possible without breaking the whole distro.

Is this a hard requirement? I can try to figure a way to keep support for older versions of HarfBuzz if you really really want this.

alerque:
> Is this a hard requirement? I can try to figure a way to keep support for older versions of HarfBuzz if you really really want this.

No, I'm totally okay with saying people don't get the fancy new features unless they upgrade HB. I do want to not take away any platform support we currently have, but if there are backport packages readily available (which there seem to be) I'm okay with requiring those. In fact I should be able to add them to our PPA and make it pretty seamless. I don't want to sink a bunch of time into keeping legacy support unless somebody *really* needs it. I'd rather help move the whole ecosystem forward by setting up HB backports than write a bunch of code to maintain HB 1.x compatibility.

alerque (approved):
Sorry for the delay getting this merged and released. The Nix stuff I was waiting for happened in a couple days but my kids have been sick. For my youngest daughter with TSC that means more seizures and I haven't gotten much work in the past few weeks.

I think we're good to go here. It is easy to only run supported tests but I couldn't come up with a mechanism for only running tests not supported by other runs that I liked, so I just hard coded the HB6 specific tests for now. Eventually we'll probably run the whole test suite there anyway, but for now I just wanted to get that one running.

khaledhosny:
Thanks @alerque for getting this to the finish line.

Issue #210, opened 2015-11-19
Value of rskip inside quote environment carried over into first footnote on page


alerque:
Check out this three page spread:

![selection_101](https://cloud.githubusercontent.com/assets/173595/11263420/3cd176cc-8e93-11e5-9e68-d86c7ebfd885.png)

The first page is all good. So is the second really, but note that the page break between two and three happens in the middle of a quote environment. Now check out the indentation of the first footnote on the third page (footnote 80).

simoncozens:
Could you try adding `SILE.settings.reset()` into the footnote package here:

    local material = SILE.Commands["vbox"]({}, function()
      SILE.Commands["font"]({size = "9pt"}, function()
        SILE.settings.reset() -- this line
        SILE.call("noindent")
        SILE.typesetter:typeset(SILE.formatCounter(SILE.scratch.counters.footnote)..".")
        SILE.call("qquad")
        SILE.process(content)
      end)
    end
    )

Let me know if it breaks other stuff.

alerque:
Edit: this isn't actually a page break issue, I just realized it's happening when the `\footnote` marker is inside the quote.

(Trying your code now…)

alerque:
No change when adding `SILE.settings.reset()` to my footnote function.

alerque:
Here's some output with the reset in place. It's also from the middle of a page and note the only one affected is the one marked from inside the quote. The font size is wrong, but otherwise everything is the same as before.

![selection_103](https://cloud.githubusercontent.com/assets/173595/11263972/6a0e2766-8e98-11e5-83a8-aa1ab7b34efa.png)

alerque:
Hey that's better. Thanks!

PR #1700, opened 2023-01-27
Ensure parindent and other state gets handled before entering temporary settings wrapper

### LINKED ISSUE #1699: Paragraph have inconsistent indents after chapter openings
I'm new to SILE, but with lots of experience with similar systems. I'm typesetting a book, and am finding that paragraphs are indented inconsistently after chapter openings. It seems to have something to do with whether the first paragraph following the \chapter command starts with another environment — in my case, emphasis. (`sile --version` says `SILE v0.14.8 (Lua 5.4)`, running under a Docker container on macOS.)

I don't really know Lua, so haven't tried to determine a fix yet. But a minimal example is below, with my some admittedly naive workarounds. My guess is that `current.parindent` is not being properly reset.

```
\begin[class=book]{document}

%
% This test file is designed to illustrate a problem, where if the first
% paragraph after a \chapter command doesn't start with plain text, then
% the following paragraph's indentation is not correct.
%
% For each case, the line 'a b c' should not be indented, and the 'd' and
% 'e' lines should be indented.
%

%%%

\chapter{fail with inital \\em}

\em{a} b c

d

e

%%%

\chapter{success with no initial \\em}

x \em{a} b c

d

e

%%%

\chapter{success with prefixing \\par}

\par \em{a} b c

d

e

%%%

\chapter{success with adding \\indent}

\em{a} b c

\indent

d

e

%%%

\end{document}
```

alerque:
Closes #1699

### DIFF
--- a/core/font.lua
+++ b/core/font.lua
@@ -3,7 +3,10 @@ local icu = require("justenoughicu")
 local lastshaper
 
 SILE.registerCommand("font", function (options, content)
-  if SU.hasContent(content) then SILE.settings:pushState() end
+  if SU.hasContent(content) then
+    SILE.typesetter:initline()
+    SILE.settings:pushState()
+  end
   if options.filename then SILE.settings:set("font.filename", options.filename) end
   if options.family then
     SILE.settings:set("font.family", options.family)
--- a/tests/bug-1699.expected
+++ b/tests/bug-1699.expected
@@ -1,0 +1,18 @@
+Set paper size 	209.7637818	297.6377985
+Begin page
+Mx 	10.4882
+My 	22.3038
+Set font 	Gentium Plus;10;400;Italic;normal;;;LTR
+T	73 82 82 w=11.1670	(foo)
+Mx 	30.4882
+My 	34.3038
+Set font 	Gentium Plus;10;400;;normal;;;LTR
+T	69 68 85 w=13.6279	(bar)
+Mx 	10.4882
+My 	46.3038
+T	73 82 82 w=13.1689	(foo)
+Mx 	30.4882
+My 	58.3038
+T	69 68 85 w=13.6279	(bar)
+End page
+Finish
--- a/tests/bug-1699.sil
+++ b/tests/bug-1699.sil
@@ -1,0 +1,13 @@
+\begin[class=plain,papersize=a7]{document}
+\nofolios
+% Both "bar" entries should have parindent applied
+\noindent
+\em{foo}
+
+bar
+
+\noindent
+foo
+
+bar
+\end{document}


alerque:
This fix may not turn out to be quite as simple as shown, but the *failure cases* are quite interesting.

core/font.lua
```diff
@@ -3,7 +3,10 @@ local icu = require("justenoughicu")
 local lastshaper
 
 SILE.registerCommand("font", function (options, content)
-  if SU.hasContent(content) then SILE.settings:pushState() end
+  if SU.hasContent(content) then
+    SILE.typesetter:initline()
```
Omikhleia, line 7: It will only work for the `\font` command, but not generally for all cases where one can enter a temporary settings wrapper. It seems an ad hoc workaround for but one case only, I don't know but it feels kind of contrived. As I wrote in the issue, `\set[parameter=font.style, value=italic]{content}` would still fail. Admittedly, changing font this way is not very clever, but that would be true for any other temporary setting. E.g. the following fails similarly:

```
\section{fail}

\language[main=fr]{doh} b c

d

e

\section{expected}

a b c

d

e
```

image

In other terms, doing this in `\font` seems wrong to me. And cramming it into `SILE.settings:pushState()` (to get our back covered here, but also at every place where settings are explicitly pushed, or implicitly via a `SILE.settings:temporarily` wrapper), besides not really being doable, would introduce a severe separation of concerns issue...

The scoping issue is hard here. My gut feeling is that current.parindent is not a setting comparable to language or font changes, but a typesetter state.

alerque: All true, this is only a band-aid on what is probably the most common iteration of a bigger problem. It's whacking off the visible top of the iceberg without doing anything about the floating mass, and something else is going to float up after this.

That being said I think the band-aid is better than the unexpected results in this common case and I expect a it will be *forced* into a better resolution when we have a settings stack that can be managed in the scope of a typesetter.

alerque:
This whole mess is also related to #210 that regressed in #1216 and now tracked as #1758.

core/font.lua
Omikhleia: I'd prefer working towards an actual fix rather than inserting a _ad hoc_ workaround for one case only, which make it easy to forget about it and later face other concerns. I'm not sure what the actual fix would be -- i.e. not sure we even need a "settings stack" for typesetter-scoped parameters (The typesetter already has a stack with states, and this could be sufficient).

alerque: The typesetter has a state-stack, but typesetter related settings are not tracked in the "state" at all. That's part of our problem. A real fix for this is already tracked in other issues, namely #1327 and friends.

Omikhleia: > A real fix for this is already tracked in other issues, namely #1327

Exactly! So perhaps it's premature to add a "workaround" of sorts here for #1699 and turn to the real thing...
But we have to consider and better define what are states vs. settings. IMHO, not all settings called "typesetter.xxx" need to be typesetter _states_, and some "yyy.param" are actually typesetter states. -- `current.parindent` could _possibly_ be one. 

I'd like to suggest a path towards it at some point, but it likely requires thinking out-of-the-box i.e. attempting to forget the current implementation and re-think it from scratch (even if ending very similar eventually), so as to be sure asking the right questions.

Omikhleia: I have been attempting a "hack" to properly "scope" the parindent. It shows other things possibly done wrong (or wrong assumptions relying on side effects, e.g. in the book class -- I don't have the same issue in resilient.book at the same places)... And there are still a few failing non-regression tests that need analysis. I can try to clean that up and make a draft PR. It's probably not the _right_ way to go obviously but it does show something about proper scoping and may help understanding the issue at stakes.

Omikhleia: ... And it helps finding tests that were likely wrong, e.g. `tests/chapterverse.sil` seems affected too by #1699 (no chapter opening, but temporary closure in header sets current.parindent to 0, yet a parindent is applied -- and as in the issue, this is the scoping in a `\font`...)

image

Omikhleia: See #1760

Omikhleia: Additionally, regarding the initial proposal here, invoking `SILE.typesetter:initline()` outside the temporary environment will imply using the parindent from the upper context, which might be wrong if it is set to a font-relative unit, no?

alerque:
Abort! Abort! Abort!

I might grab the regression test from here, but #1760 has a working fix that moves just a little bit closer to where we actually want to end up.

PR #1702, opened 2023-01-28
Support migrating content in hbox building logic

### LINKED ISSUE #583: Discressionaries in \href aren't counted towards width

### LINKED ISSUE #1326: Add an API for typesetter takebacks
There are quite a few situations where doing something then immediately taking it back (measuring, seeing what else it would affect, stashing for later, etc.) is useful. There are a couple simple Lua idioms that work for this, but knowing what table to operate on when is a bit obtuse and I think the whole node queue handling could probably benefit from some more user friendly APIs.

c.f. discussion [here](https://github.com/sile-typesetter/sile/pull/1323#discussion_r815028639)

### LINKED ISSUE #1648: The question of footnotes in an hbox
There are plenty of case where one could want to put some content in an `\hbox{...}`
- For the purpose of framing it (as our old `\boxaround` did, or my own **framebox** package, etc.).
- For astute `\rebox`ing or to `\rotate` it...
- In the current implementation of `\pdf:link`, `\underline` and similar friends (... though something such as #1334 _might_ eventually change it)...

Still, footnotes in the content are "lost":

```
\begin[papersize=a9, class=book]{document}
\neverindent
\use[module=packages.rules]

Footnote\footnote{I am a normal footnote}

\underline{Footnote\footnote{I am a footnote in an hbox}}
\end{document}
```

image

In my own implementation of table building blocks (`\parbox` in module **ptable.sile**), I had the idea of moving the migrating material to the upper context. Something similar could be done too for footnotes, leading to:

image

**But** this is not as obvious at it seems (and my own implementation suffers from accounting that fact) - Besides hboxes re-wrapped in pseudo-hboxes, there is also the case of hboxes just built for measurement, and "stolen" from the typesetter afterwards with a dirty trick (and possibly re-pushed or not _later_ - This is actually what `\underline` currently does for instance...).

(La)TeX also has troubles, IIRC, with boxing content that might include a footnote. There's some interesting design challenge there.

Omikhleia:
Close #1648 

EDIT: Also Closes #1326 (see comment below)

EDIT: Also Closes #583 (see comment below)

**What it does**: provides a way for migrating content (a.k.a. footnote-like content) in a hbox constructed from content to be effective...

image

... That is, it now makes it to upper context, instead of just "disappearing"[^1]

[^1]: The example here is not made into a test because (a) underlines, strikethroughs and links shouldn't be hboxes eventually as per #1334; (b) reboxing for overprint etc. is not that impressive and realistic; (c) links and rotations require the libtexpdf backend currently... TBH, rotation is however probably the most interesting case in the existing set of standard packages.

**How it does it**:

- Moves the hbox building logic (from the plain class command) into a typesetter method, which IMHO is more sound.
- Extracts migrating nodes when building the hbox, so these can be pushed back outside the hbox
- Removes, as much as possible, the use of the "hbox stealing trick" in the code base (classes, packages, docs)

**Words of caution**

- I tried to be very cautious with the changes, but lots of things are not covered by existing tests... e.g. to cite but a few, the `\discretionary` command, the `\tate-chu-yoko` (while `\latin-in-tate`  from the same package has tests), etc.
- I tried to remain compatible / non-breaking, which implied a few decisions (warnings where I'd have used errors ideally, and a few hacks with in-code comments where I'd have coded it a bit differently if it wasn't _slightly_ modifying some assumptions, whether these are legit or not - probably the latter...).


alerque:
Testing this on one of my own projects blew up, bit it might be due to my ... heterodox use of `\discretionary`. I'll have to give it another shakedown, but in general what I saw of the diff here makes sense.

Omikhleia:
@alerque 
> Testing this on one of my own projects blew up, bit it might be due to my ... heterodox use of \discretionary. I'll have to give it another shakedown, but in general what I saw of the diff here makes sense.

I can perhaps help, if you have a MWE of the case that breaks?

Omikhleia:
#1326 called for an API for "typesetter takebacks" (a.k.a "hbox stealing trick") => This PR makes the point moot, as (most if not all) hbox stealing is removed from the code, and this was the main use case for it (building hboxes for mere measurements)

Omikhleia:
Things to consider before any merge
- remove hbox stealing logic from scalebox (who made in in the meantime)
- #583 (possibly)
- #1718 (possibly)

alerque (approved):

PR #1703, opened 2023-01-28
Update Cairo/Podofo hbox debug API to match libtexpdf

### LINKED ISSUE #1701: Inconsistent debugHbox signature in alternate outputters
In the **libtexpdf** ouputter:
`function outputter:debugHbox (hbox, scaledWidth)`

This is also how it is used the very few cases where it is invoked (e.g. **packages.math** and **base** typesetter).

But the **base** outputter mentions up to 4 arguments (incl. the self instance):
`function outputter.debugHbox (_, _, _, _) end`

The **podofo** and **cairo** outputters require a typesetter argument (thus matching the "wrong" base API):
`function outputter.debugHbox (_, typesetter, hbox, scaledWidth)`

They should likely rely on `outputter:getCursor()` instead. I am not sure any build of SILE uses them, but there's anyway an API discrepancy (which would only be noticed when debugging hboxes).

alerque:
Closes #1701

### DIFF
--- a/outputters/base.lua
+++ b/outputters/base.lua
@@ -32,7 +32,7 @@ function outputter.drawRule (_, _, _, _, _) end
 
 function outputter.debugFrame (_, _, _) end
 
-function outputter.debugHbox (_, _, _, _) end
+function outputter.debugHbox (_, _, _) end
 
 function outputter.getOutputFilename (_, ext)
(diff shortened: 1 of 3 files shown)


alerque:
Neither Cairo nor Podofo are known to work or tested at this point, but this is a good-faith attempt to sketch out what the current interface should look like in each.

PR #1704, opened 2023-02-01
Add Docker workaround to avoid user mapping reqirement

alerque:
This comes out of recent experiences over on CaSILE with new Git and Docker security measures. The effective user ID stuff is perfect for CaSILE that operates primarily on *repositories* not *files*, so writing as the directory owner almost always will make sense. I'm not so sure it is a match for SILE, hence the draft status. It might be better to actually parse the arguments passed to SILE to find the input file(s), then check their ownership. This would necessitate some extra logic to handle cases without an input file but with `-o` for output. Also we can do some debugging and see if we have any ENV information available now that we didn't have before, but this should get the process started before I forget how it works.


alerque:
Closing in favor of cherry picking the differences from here that look good into #2037. No wonder why I had that feeling of *déjà vu* when working in this...

PR #1709, opened 2023-02-04
Urlstyle hook in url package and code command moved to plain class

Omikhleia:
@ctrlcctrlv This the proposal I made while discussing your PR #1538

Since things didn't move there for 5 months or so, I am proposing it again as a clean PR.

```
\url{http://https://sile-typesetter.org/}

\define[command=urlstyle]{<{\process}>}
\url{http://https://sile-typesetter.org/}

\define[command=urlstyle]{<{\color[color=blue]{\underline{\process}}}>}
\url{https://sile-typesetter.org/}
```

image

IMHO, simpler and more general, in the same vein as `\foliostyle`. (If it makes it, I'll certainly use it in my classes to use my own style paradigm in an overridden hook).

_Of course_, the definition of `\verbatim:font` (and hence `\code`) is lame, but I don't think it necessarily _has_ to be addressed here. (It's just yet another technical debt item.)


Omikhleia:
N.B. This would also "tick" a checkbox in #1194 :
> Command \code is problematic and has to be reconsidered: It is defined in the url package, but re-defined in the Manual with a less-than-general definition (hard-coded 9pt, etc.). The situation is a bit messy; one may want URLs in regular font but keep the code-formatting option elsewhere, and it's hard to make right currently. (Redefining \url to temporarily kill the \code command is one possible way, but it's not that nice.)

alerque (dismissed):

alerque:
I'm not 100% satisfied with this as a final resting place for URL styling, but it is a step forward for sure. It gets `\code` moved where it belongs, gives a way for people to tweak what they want (mostly), and doesn't break anything. We can iterate on this.

alerque (approved):

PR #1710, opened 2023-02-04
feat(packages): New scalebox package for arbitrary box re-scaling

### LINKED ISSUE #1035: \reflectbox ?
I have macros for \LaTeX and \SILE. Is it possible to typeset the XeTeX and XeLaTeX logos, which has flipped &lsaquo;E&rsaquo;'s?

image

I suppose I need to make a custom font with a flipped E, Ǝ, if it doesn't exist in the font? Or...

XƎLaTeX / XƎTeX ?

### LINKED ISSUE #1178: How do I horizontally squish content?
Let's say I want to take a letter and make it 0.5 of its width—how?

I'd find this useful for Ruby. It's often done in Japan.

Omikhleia:
Closes #1178

EDIT: Also Closes #1035 (see comments below)

Box re-scaling via PDF transformation matrix:

`\scalebox[xratio=<number>, yratio=<number>]{<content>}`

_Rationale_

Both @ctrlcctrlv and @alerque in the above-mentioned issue said it should be proposed to the core distribution... So here it is, basic logic extracted from textsubsuper.sile and re-vamped as as standalone package.

_Caveats_
- It's only tested (see below) in LTR writing direction: Checking RTL and vertical writing directions is left as an exercise to readers who know how to use these.
- It's using the "hbox stealing trick": When/if #1702 makes it, it is left as an exercise to adapt the code so that migrating contents are honored.
- It is also left as an exercise where to include the package documentation in the manual.

image

The first series of lines just demonstrate the basic feature. The second series ensures line stretching/shrinking is correctly applied. The very last line is just for the show. 

Example code attached as .txt so it can be uploaded here:
[feat-scalebox.txt](https://github.com/sile-typesetter/sile/files/10608913/feat-scalebox.txt)

### DIFF
--- a/packages/scalebox/init.lua
+++ b/packages/scalebox/init.lua
@@ -1,0 +1,80 @@
+local base = require("packages.base")
+
+local package = pl.class(base)
+package._name = "scalebox"
+
+function package:registerCommands ()
+
+  self:registerCommand("scalebox", function(options, content)
+    if SILE.outputter._name ~= "libtexpdf" then
+      SU.warn("Output will not be scaled: \\scalebox only works with the libtexpdf backend")
+      return SILE.process(content)
+    end
+    SILE.outputter:_ensureInit()
+    local pdf = require("justenoughlibtexpdf")
+
+    local hbox = SILE.call("hbox", {}, content)
+    table.remove(SILE.typesetter.state.nodes) -- Remove the box from queue
+    local xratio, yratio = SU.cast("number", options.xratio) or 1, SU.cast("number", options.yratio) or 1
+    if xratio == 0 or yratio == 0 then
+      SU.error("Scaling ratio cannot be null")
+    end
+
+    local W = hbox.width * math.abs(xratio)
+    local H, D
+    if yratio > 0 then
+      H = hbox.height * yratio
+      D = hbox.depth * yratio
+    else
+      H = hbox.depth * -yratio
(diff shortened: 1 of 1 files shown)


Omikhleia:
Oops wrong file on first push... fixed and rebased. Sorry about it.

(But in case you look at the old commit hash... Yep, the rotate package is doesn't like `\hfill` in content. I was looking at it out of curiosity, but it wasn't a fix, just quick tests... And I am not going to fix it anyway.)

Omikhleia:
N.B. #1035 called for a `\reflecbox` (with also hints to achieve it via PDF matrix transform). This would just be a special case = `\scalebox[xratio=-1,yratio=1]`

packages/scalebox/init.lua
```diff
@@ -0,0 +1,81 @@
...
+      height = H,
+      depth = D,
+      outputYourself = function(node, typesetter, line)
+        local outputWidth = SU.rationWidth(node.width, node.width, line.ratio)
+        local X = typesetter.frame.state.cursorX
+        local Y = typesetter.frame.state.cursorY
+        local x0 = X:tonumber()
+        local y0 = -Y:tonumber()
+
+        typesetter.frame.state.cursorY = Y
```
Omikhleia, line 44: Re-reading my code, this line doesn't seem useful, perhaps a leftover.

alerque (approved):

PR #1712, opened 2023-02-09
Documentation fixes (no code changed):

jslabovitz:
- Spelling/grammar/syntax/usage:
  - Generally rephrase text to read better and avoid inessential details or noise/slang.
    - Avoid: of course, obviously, very, and (at start of sentence), so, etc.
  - Rework some sentences to have fewer semicolons. (An unfortunately common pattern...)
  - Add some commas between clauses, and remove some others, as needed (Oxford style).
  - Expand a few terms that may be otherwise unknown to a new user.
  - Correct spelling errors.
  - Correct capitalization errors.
    - According to majority style, only capitalize first word of section & subsection titles.
    - 'Lua' (not lua or LUA)
    - Chapter number references (eg, Chapter 4).
  - Try to be more consistent about using a colon at the end of a sentence that introduces a code block.
  - In `math`, use direct voice (one may -> you may).

- Typography/punctuation:
  - Use em-dashes consistently (replace --, en-dash, spaces before/after, etc.).
  - Use quotes consistently (generally double-quotes, not single).
  - Be consistent about punctuation within quotes, unless clearly indicating a single word, etc.
  - Try to put footnote references after punctuation, not before.
  - Use conventional enumerated items within text paragraph.
    - eg, instead of 1) red, 2) blue -- use (1) red, (2) blue
  - Remove excessive punctuation (commas, semicolons) from a few itemized lists.

- Formatting:
  - Try to be consistent with itemized lists -- use colon instead of dashes (definition lists would be better in many of these places!).
  - Avoid hyphenating proper names of software products.
  - Fix a few code blocks that are wrapping badly or otherwise poorly formatted.
  - Ensure inline code text is really in \code{} blocks.
  - Ensure mentions of classes is consistently in \code{} blocks (though probably should use autodoc:class?).
  - Ensure mentions of values (especially defaults) is consistently in \code{} blocks.
  - Use \autodoc:parameter{}, etc., in appropriate places (incomplete).
  - Add a few \noindent's in select places.
  - Move a couple of references to systems (eg, Fluent) into footnotes.
  - Correct & simplify SILE bumpy-road examples.
  - In `linespacing`, use \itemize instead of manual bulleted lists.

- Other:
  - Update/normalize references to products like InDesign, Pages, etc.
  - Update incorrect URLs.

### DIFF
--- a/documentation/c09-concepts.sil
+++ b/documentation/c09-concepts.sil
@@ -88,7 +88,7 @@ output queue, the other new vertical content can be processed. At any point you
 can force the current queue of horizontal content (the node queue) to be shaped
 into lines and added to the vertical output queue by calling the function
 \code{SILE.typesetter:leaveHmode()}. This is handy when for writing custom
-functions, but it is a fairly low level control. (Tt is unlikely to be
+functions, but it is a fairly low level control. (It is unlikely to be
 useful while writing a document.) A related but higher level command,
 \code{\\par}, is more frequently used when writing a document and embedded in
 the content. The \code{\\par} command first calls
--- a/documentation/c09-concepts.sil
+++ b/documentation/c09-concepts.sil
@@ -11,29 +11,29 @@ you are a class designer, and will be able to follow the details of how
 SILE implements these commands and features; we will also explain how to
 interact with these components at the Lua level.}
 
-\section{Boxes, Glue and Penalties}
+\section{Boxes, glue, and penalties}
 
-SILE’s job is, looking at it in very abstract terms, all about arranging
+SILE’s job, looking at it in very abstract terms, is all about arranging
 little boxes on a page. Some of those boxes have letters in them, and
 those letters are such-and-such a number of points wide and
 such-and-such a number of points high;
-some of the boxes are empty but are there just to take up space; when a
-horizontal row of boxes has been decided (i.e. when a line break is determined)
+some of the boxes are empty but are there just to take up space.
+When a horizontal row of boxes has been decided (i.e., when a line break is determined)
 then the whole row of boxes is put into another box and the vertical list
 of boxes are then arranged to form a page.
 
 Conceptually, then, SILE knows about a few different basic components:
-horizontal boxes (such as a letter); horizontal glue (the stretchable, shrinkable
+horizontal boxes (such as a letter); horizontal glue (the stretchable or shrinkable
 space between words); vertical boxes (a line of text); vertical glue (the space
 between lines and paragraphs); and penalties (information about where and when
 not to break lines and pages).\footnote{Additionally there are three more types of box
 that SILE cares about: N-nodes, unshaped nodes, and discretionaries.}
 
-The most immediately useful of these are horizontal and vertical glue. It
-is possible to explicitly add horizontal and vertical glue into SILE‘s processing
+The most immediately useful of these are horizontal and vertical glue.
+Horizontal and vertical glue can be explicitly added into SILE‘s processing
 stream using the \code{\\glue} and \code{\\skip} commands. These take a
-\code{width} and a \code{height} parameter respectively, both of which are
-glue dimensions. So, for instance, the \code{\\smallskip} command is
+\code{width} and a \code{height} parameter, respectively, both of which are
+glue dimensions. For instance, the \code{\\smallskip} command is
 the equivalent of \code{\\skip[height=3pt plus 1pt minus 1pt]}; \code{\\thinspace}
 is defined as being \code{\\glue[width=0.16667em]}.
 
@@ -41,7 +41,7 @@ Similarly, there is a \code{\\penalty} command for inserting penalty nodes;
 \code{\\break} is defined as \code{\\penalty[penalty=-10000]} and \code{\\nobreak}
 is \code{\\penalty[penalty=10000]}.
 
-You can also create horizontal and vertical boxes from within SILE. One obvious
+You can also create horizontal and vertical boxes from within SILE. One
 reason for doing so would be to explicitly avoid material being broken up by a
 page or line break; another reason for doing so would be that once you box some
 material up, you then know how wide or tall it is. The \code{\\hbox} and \code{\\vbox}
@@ -67,7 +67,7 @@ local glue  = SILE.nodefactory.glue({ width =  l })
 local vglue = SILE.nodefactory.vglue({ height = l })
 \end{raw}
 
-SILE’s typesetting is organised by the \code{SILE.typesetter} object; it
+SILE’s typesetting is organised by the \code{SILE.typesetter} object. It
 maintains two queues of material that it is still working on: the node queue and
 the output queue. Material in these queues is content that has been parsed but
 not yet rendered to the canvas and can still be manipulated. The node queue
@@ -78,28 +78,28 @@ that have not yet been broken up into lines. The output queue
 breaking happen when the typesetter moves between horizontal and vertical mode.
 
 As new content is parsed it is added to the node queue in as small chunks as
-possible—chunks that must remain together no matter where they end up on
-a line. For example this might include individual symbols, syllables, or objects such as
+possible. These chunks must remain together no matter where they end up on
+a line. This might include individual symbols, syllables, or objects such as
 images. As soon as new content which requires a vertical break is encountered,
 the node queue is processed to derive any missing shaping information about
 each node, then the sequence of node is broken up into  lines. Once all the
-"horizontal mode" nodes are broken into lines and those lines are added to the
+“horizontal mode” nodes are broken into lines and those lines are added to the
 output queue, the other new vertical content can be processed. At any point you
 can force the current queue of horizontal content (the node queue) to be shaped
 into lines and added to the vertical output queue by calling the function
-\code{SILE.typesetter:leaveHmode()}. This is a handy move for writing custom
-functions, but it is a fairly low level control (i.e. it is unlikely to be
-useful while writing a document). A related but higher level command,
+\code{SILE.typesetter:leaveHmode()}. This is handy when for writing custom
+functions, but it is a fairly low level control. (Tt is unlikely to be
+useful while writing a document.) A related but higher level command,
 \code{\\par}, is more frequently used when writing a document and embedded in
 the content. The \code{\\par} command first calls
 \code{SILE.typesetter:leaveHmode()}, then inserts a vertical skip according to
 the \autodoc:setting{document.parskip} setting, then goes on to reset a number of
-settings that are typically paragraph related such as hanging indents.
+settings that are typically paragraph-related such as hanging indents.
 
 When writing a custom command, if you want to manually add a vertical space to
-the output first ensure that the material in the current paragraph has been all
+the output, first ensure that the material in the current paragraph has been all
 properly boxed-up and moved onto the output queue by calling
-\code{SILE.typesetter:leaveHmode()}; then add your desired glue to the output
+\code{SILE.typesetter:leaveHmode()}, then add your desired glue to the output
 queue.
 This is exactly what the \code{\\skip} and similar commands do.
 
@@ -115,7 +115,7 @@ SILE.typesetter:pushVglue({ height = l })
 Adding boxes yourself is a little more complicated, because boxes need to
 know how to display themselves on the page. To facilitate this, they normally
 store a \code{value} and an \code{outputYourself} member function. For instance,
-the \code{image} package actually does something very simple; it adds a horizontal
+the \code{image} package does something very simple: it adds a horizontal
 box to the node queue which knows the width and height of the image, the source,
 and instructions to the output engine to display the image:
 
@@ -144,9 +144,9 @@ done with the \code{SILE.typesetter:pushPenalty(\{penalty = x\})} and
 S\lower[height=0.5ex]{I}L\kern[width=-.2em]\raise[height=0.6ex]{\font[features=+smcp]{e}}}}
 \define[command=SILEglue]{\font[family=Gentium Plus]{%
 S\lower[height=0.5ex]{I}L\glue[width=-.2em]\raise[height=0.6ex]{\font[features=+smcp]{e}}}}
-\code{\\kern}’s are a type of \code{\\glue}. The difference between the two is
-that, while a \code{\\glue} can be broken at the end of a line, a \code{\\kern}
-can’t be. Hearkening back to our \SILEkern example from the \em{Macros and
+\code{\\kern}’s are a type of \code{\\glue}, only different in that
+while a \code{\\glue} can be broken at the end of a line, a \code{\\kern}
+can’t. Hearkening back to our \SILEkern example from the \em{Macros and
 Commands} chapter, consider that example, repeated enough times to cause a
 linebreak, but with \code{\\glue}’s everywhere \code{\\kern}’s are used
 instead:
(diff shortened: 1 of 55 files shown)


documentation/c10-classdesign.sil
```diff
@@ -243,17 +241,17 @@ Once again this is done in an object-oriented way,
 with derived classes overriding their superclass’ methods where necessary.
 
 Some packages may provide functions that need to be run as part of the class output routines.
-One way packages can accomplish this is registering hook functions that get run at known locations in the provided classes.
+One way packages can accomplish this is by registering hook functions that get run at known locations in the provided classes.
 Class authors may also provide their own hook locations for packages, or run any set of registered hooks in their own outputs.
 The other way is to export functions that get added to the class itself and can be run manually from the class.
 
-For examples check out the \autodoc:package{tableofcontents} package for the hooks it sets, but also the \autodoc:command{\tocentry} command it registered that gets called manually in the \code{book} class.
+For examples check out the \code{tableofcontents} package for the hooks it sets, but also the \autodoc:command{\tocentry} command it registers that gets called manually in the \code{book} class.
```
Omikhleia, line 248: Why `\autodoc:package{tableofcontents} package` > `\code{tableofcontents}` here? (The idea being that package names are typeset as `\autodoc:package{}` rather than code, for possible cross-reference/indexing later)

documentation/c11-xmlproc.sil
```diff
@@ -6,29 +6,29 @@ We’ll be looking at the DocBook processor that ships with SILE.
...
+The class initalization for DocBoox isn’t too fancy; it just loads up a couple packages that will get used later.
 
 \begin{note}
-Much of the example code in this chapter is in SIL format using macros.
-The actual docbook class currently uses Lua functions to specify these commands.
+Much of the example code in this chapter is in SILE format using macros.
+The actual \code{docbook} class currently uses Lua functions to specify these commands.
 The functionality is the same, but the Lua syntax is more flexible and recommended for most use cases.
-The SILE define macros shown here can still be used in a preamble file if desired.
+The SILE \code{\\define} macros shown here can still be used in a preamble file if desired.
```
Omikhleia, line 19: Could be `\autodoc:command{\define}` I'd assume.

packages/date/init.lua
```diff
@@ -32,9 +32,9 @@ end
 
 package.documentation = [[
 \begin{document}
-The \autodoc:package{date} package provides the \autodoc:command{\date} command, which simply outputs the date using the system’s date function.
-You can customize the format by passing the \autodoc:parameter{format} parameter, following the formatting codes in the Lua manual.
-(\url{https://www.lua.org/pil/22.1.html})
+The \autodoc:package{date} package provides the \autodoc:command{\date} command, which simply outputs the current date using the system’s date function.
```
Omikhleia, line 35: Not necessarily the _current_ date, if I remember correctly (depending on options).

packages/frametricks/init.lua
```diff
@@ -260,24 +260,25 @@ We just issued a \autodoc:command{\breakframevertical} command at the start of t
...
 It should only be used at the top of a frame, as it reinitializes the typesetter object.
 
 Combining all of these commands, the \autodoc:command{\float} command breaks the current frame, creates a small frame to hold a floating object, flows text into the surrounding frame, and then, once text has descended past the floating object, moves the frame back into place again.
 It takes two optional parameters, \autodoc:parameter{bottomboundary=<dimension>} and/or \autodoc:parameter{rightboundary=<dimension>}, which open up additional space around the frame.
-At the start of this paragraph, I issued the command
 
-\begin[type=autodoc:codeblock]{raw}
-\float[bottomboundary=5pt]{\font[size=60pt]{C}}.
-\end{raw}
+% At the start of this paragraph, I issued the command
```
Omikhleia, line 273: Oh well noted, regarding commenting this out. I remember this was used in some older version of the manual. I wonder why it is no longer the case (on purpose, or just by mistake?)

documentation/c10-classdesign.sil
jslabovitz: Yes, I think this was my mistake. You’re right, it should be \autodoc:package{}.

packages/lists/init.lua
```diff
@@ -280,16 +280,18 @@ end
 package.documentation = [[
 \begin{document}
 \font:add-fallback[family=Symbola]% HACK Gentium Plus (SILE default font) lacks the circle bullet :(
-The \autodoc:package{lists} package provides enumerations and bullet lists (a.k.a. \em{itemization}\kern[width=0.1em]) which can be nested together.
+The \autodoc:package{lists} package provides enumerated and itemized (also known as \em{bulleted lists}) which can be nested together.
 
 \smallskip
-\em{Bullet lists.}
+\noindent
+\em{Itemized lists}
```
Omikhleia, line 287: Out of curiosity, is "itemized lists" more understandable that "bullet lists" for casual readers (not necessarily used to LaTeX etc.)? Being French, I can't say, but I had the feeling that "bullet lists" was more clear maybe. In French we'd likely use "listes à puces" - Google Translate translates it as "bulleted lists"; but again, I don't know what's best for English readers.

packages/lists/init.lua
```diff
@@ -370,21 +376,23 @@ The way they do is best illustrated by an example.
...
+\indent
 The package tries to ensure a paragraph is enforced before and after a list.
 In most cases, this implies paragraph skips to be inserted, with the usual \autodoc:setting{document.parskip} glue, whatever value it has at these points in the surrounding context of your document.
 Between list items, however, the paragraph skip is switched to the value of the \autodoc:setting{lists.parskip} setting.
 
 \smallskip
-
-\em{Other considerations.}
+\noindent
+\em{Other considerations}
```
Omikhleia, line 390: For the fun of it, an interesting case where French typographers would likely disagree. (Likely = depends on whom you ask, lol).

packages/lists/init.lua
jslabovitz: I'd say that for people who are not typesetters (or LaTeX nerds), the English term would be 'bulleted lists' (and not 'bullet lists'). But because the package itself uses 'itemize,' I figured that was a better primary term to use. Hence why I put 'bulleted lists' as the 'aka' clause as opposed to the other way around.

packages/math/init.lua
```diff
@@ -175,12 +176,12 @@ Here is a slightly more involved equation:
...
 Its main difference from the SILE syntax is that \code{\\mycommand\{arg1\}\{arg2\}\{arg3\}} is translated into MathML as \code{<mycommand> arg1 arg2 arg3 </mycommand>} whereas in normal SILE syntax, the XML equivalent would be \code{<mycommand>arg1</mycommand> arg2 arg3}.
 
-\code{\\sum}, \code{\\infty} and \code{\\pi} are only shorthands for the Unicode characters \math{\sum}, \math{\infty} and \math{\pi}.
+\code{\\sum}, \code{\\infty}, and \code{\\pi} are only shorthands for the Unicode characters \math{\sum}, \math{\infty} and \math{\pi}.
 If it’s more convenient, you can use these Unicode characters directly.
 The symbol shorthands are the same as in the TeX package \href[src=https://www.ctan.org/pkg/unicode-math]{\code{unicode-math}}.
 
-\code{\{formula\}} is a shorthand for \code{\\mrow\{formula\}}.
-Since parentheses — among other glyphs — stretch vertically to the size of their englobing \code{mrow}, this is useful to typeset parentheses of different sizes on the same line:
+\code{\\\{formula\}} is a shorthand for \code{\\mrow\{formula\}}.
```
Omikhleia, line 183: I don't think there's a `\{formula}` thing in code. (Never really understood that bit, though, perhaps it needs a more drastic rewording.

packages/frametricks/init.lua
jslabovitz: I tried it, and it just didn't work very well. It seemed a bad implementation of the `dropcaps` package. (Which is what it says in the docs, too.)

packages/lists/init.lua
jslabovitz: This actually brings up something I noticed: both the `lists` package and the `math` package have this sort of heading. I'd call it a sort of section, yet perhaps not one that should be numbered or in the TOC. As \subsubsection isn't being used in the docs, I wonder whether this could be defined (in `macros.sil`) to do the 'right' thing consistently?

packages/tableofcontents/init.lua
```diff
@@ -221,16 +221,16 @@ end
...
-At the end of each page the class will call a hook function (\code{moveTocNodes}) that collates the table of contents entries from that pages and logs which page they’re on.
+At the end of each page the class will call a hook function (\code{moveTocNodes}) that collates the table of contents entries from that pages and records which page they’re on.
 At the end of the document another hook function (\code{writeToc}) will write this data to a file.
 The next time the document is built, any use of the \autodoc:command{\tableofcontents} (typically near the beginning of a document) will be able to read that index data and output the TOC.
 Because the toc entry and page data is not available until after rendering the document,
 the TOC will not render until at least the second pass.
-If by chance rendering the TOC itself changes the document pagination (e.g. the TOC spans more than one page) it might be necessary to run SILE 3 times to get accurate page numbers shown in the TOC.
+If by chance rendering the TOC itself changes the document pagination (e.g., the TOC spans more than one page) it might be necessary to run SILE three times to get accurate page numbers shown in the TOC.
```
Omikhleia, line 233: While we are at it: "three" times > "several" times? (as soon as there are other packages doing the same type of things, or more than one ToC, etc. this is going to be hard to predict!)

jslabovitz: Ah yes, in `c01-whatis.sil` I changed 'three' to 'several.' Missed this one.

packages/twoside/init.lua
```diff
@@ -134,22 +134,22 @@ end
...
+If so, it ejects another page to assure content starts on an odd page.
 
 The \autodoc:command{\open-spread} is similar but a bit more tailored to use in book layouts.
-By default headers and folios will be suppressed automatically on any empty pages ejected, making them blank.
+By default, headers and folios will be suppressed automatically on any empty pages ejected, making them blank.
 It can also accept three parameters.
-The \autodoc:parameter{odd=false} parameter can be used to disable the opening page being odd, hence opening an even page spread.
-The \autodoc:parameter{double=false} parameter can be used to always output at least one empty  even page before the starting an odd page.
-The \autodoc:parameter{blank=false} parameter can be used to not suppress headers and folios on otherwise empty pages.
+The \autodoc:parameter{odd} parameter (default \code{false}) can be used to disable the opening page being odd, hence opening an even page spread.
```
Omikhleia, line 147: I think it's wrong here. The defaults for all these settings are true. Setting xxx=false changes the behavior as explained.

packages/lists/init.lua
Omikhleia: IMHO,we can't use `\sub(sub)section` (or any book sectioning level) as we have no idea where the package documentation would end up (without making assumptions on how the manual is organized). I'd call that a "sort of simple heading", and maybe it could be provided by autodoc.

packages/twoside/init.lua
Omikhleia: I mean, e.g. odd=true is the default (opening is always on odd page), so one has to explicitly add odd=false for opening on any page.

documentation/c11-xmlproc.sil
```diff
@@ -6,29 +6,29 @@ We’ll be looking at the DocBook processor that ships with SILE.
...
-Now we can start defining SILE commands to render XML elements.
-Most of these are fairly straightforward so we will not dwell on them too much.
-For instance, DocBook has tags like \code{<code>}, \code{<filename>}, \code{<guimenu>} which should all be rendered in a monospaced typewriter font.
-To make it easier to customize the class, we abstract out the font change into a single command:
+The class initalization for DocBoox isn’t too fancy; it just loads up a couple packages that will get used later.
 
 \begin{note}
-Much of the example code in this chapter is in SIL format using macros.
-The actual docbook class currently uses Lua functions to specify these commands.
+Much of the example code in this chapter is in SILE format using macros.
```
Omikhleia, line 16: SIL / SILE format: I think it was discussed somewhere (@alerque might certainly remember) on how to call the TeX-like input format, and SIL was the choice (to distinguish that input format from SILE as a software that can use other input formats, IIRC). Might not be very clear, though.

documentation/c12-tricks.sil
```diff
@@ -161,10 +161,10 @@ function diglot:endPage = ()
...
 \end{verbatim}
 
-Similarly for the end of the document, but in this case we will use the
-emergency \code{chuck} method; whereas \code{leaveHmode} means “call the
-page builder and see there’s enough material to build a page”, \code{chuck}
-means “you must get rid of everything on your queue now.” We add some infinitely
+At the end of the document, we will use the
+emergency \code{chuck} method. Where \code{leaveHmode} means “call the
+page builder and see there’s enough material to build a page,” \code{chuck}
+means “you must get rid of everything on your queue \em{now.”} We add some infinitely
```
Omikhleia, line 167: Not sure the quotation mark should be in italic. (Some typographers would perhaps frown: the quote end and start mark should be in the same style, most likely.)

documentation/c10-classdesign.sil
alerque: Touched up.

packages/date/init.lua
alerque: Touched up.

packages/frametricks/init.lua
alerque: Yes, one of the proof-of-concept usages for `\float` was for dropcaps, but we have a *much* more robust package for that now and using this as an example was always terrible. There are still use cases for this, but due to limitations they don't involve contiguous paragraphs flowing around the float.

packages/tableofcontents/init.lua
alerque: While theoretically possible, this was worded as 3 because it would be pretty tricky come up with an example that took 4. It would require the addition of a TOC to change the total page count so substantially that the raw size of the page numbers (e.g. number of digits) caused a third reflow of the document onto new pages even after the body of the ToC contents had been typeset with *close* page numbers once. If the number of pages taken up by your ToC content changes based on page numbers 999 overflowing to 1000 and now your TOC takes up 5 pages instead of 6, *AND* the overflow of 10000 to 1002 causes the ToC to take up 7 pages instead of 6 then you have bigger problems than number of SILE runs. This isn't "several", 3 is a magic number. 2 will probably work 80% of the time, but 3 will work cover the 50%-99.9999% of the time. If you have to go to four to get accurate results you should reconsider your life choices.

alerque: The entry in c01 is unrelated to the ToC (and other packages that save references for future runs). That has to do with Knuth's line and breaking algorithm. We can gloss over that with "several" even though it is a fixed number too because the user doesn't need to do anything different based on that knowledge. For rerunning SILE to get accurate ToC page numbers they do need to know that 3 is the sweet spot.

documentation/c11-xmlproc.sil
alerque: Yes, the file type and input language should be SIL not SILE. I touched this up. We probably have lots more places in need of an update since this was a change in early development.

documentation/c12-tricks.sil
alerque: Going to have to side with the French on this one ;-)

documentation/c09-concepts.sil
```diff
@@ -78,28 +78,28 @@ that have not yet been broken up into lines. The output queue
...
-"horizontal mode" nodes are broken into lines and those lines are added to the
+“horizontal mode” nodes are broken into lines and those lines are added to the
 output queue, the other new vertical content can be processed. At any point you
 can force the current queue of horizontal content (the node queue) to be shaped
 into lines and added to the vertical output queue by calling the function
-\code{SILE.typesetter:leaveHmode()}. This is a handy move for writing custom
-functions, but it is a fairly low level control (i.e. it is unlikely to be
-useful while writing a document). A related but higher level command,
+\code{SILE.typesetter:leaveHmode()}. This is handy when for writing custom
+functions, but it is a fairly low level control. (Tt is unlikely to be
```
Omikhleia, line 91: Small typo introduced here: "Tt"

packages/lists/init.lua
alerque: I'm leaving this comment unresolved because I do think there is room for improvement here, but I'm not waiting for it to be part of this PR. If anybody wants to pop open a new PR or issue go for it.

alerque (approved):

PR #1715, opened 2023-02-15
Overhaul SIL parser, write formal grammar

### LINKED ISSUE #1435: Formal specification of SILE grammar documentation
Is the formal specification of the SILE grammar documented?

This would help write other tools such as a Language Server, as per https://github.com/sile-typesetter/sile/issues/1406

alerque:
Closes #1435 
See #105


sil.abnf
```diff
@@ -0,0 +1,71 @@
...
+environment =/ %s"\begin" [ options ] "{" passthrough-command-id "}" passthrough-content %s"\end{" passthrough-command-id "}"
+
+; Nothing to see here.
+comment = "%" utf8-octets CRLF
+
+; Input strings that are not wrapped in some SILE command.
+text = *text-char
+
+; Input content wrapped in braces can be attatched to a command or just used to
+; create a settings wrapper.
```
Omikhleia, line 31: I'm not sure what a "settings wrapper" is.

The name might imply it does a temporary settings context, which is not true:

```
\begin{document}

Before {\set[parameter=font.size, value=30]Inside} Outside

\end{document}
```

image

sil.abnf
```diff
@@ -0,0 +1,71 @@
...
+braced_content = "{" content "}"
+
+; As with environments, can be raw passthrough or regular content depending on the command. With raw variants we require contents.
+command =  "\" command-id [ options ] [ braced_content ]
+command =/ "\" passthrough-command-id [ options ] braced-passthrough-content
+
+; Building blocks
+options = "[" parameter *( "," parameter ) "]"
```
Omikhleia, line 39: Current syntax (pairsep in parserbits.lua) also allows the semi-colon as separator

sil.abnf
alerque: I believe that was the intended use, the fact that it is not working (stopped working?) is a bug. We did fix the `\set` command so that it can be used as a wrapper itself which was broken for a long time, but it seems to me what you show is a regression (unless I understand something wrong).

```sile
Before \set[parameter=font.size, value=30]{Inside} Outside
```

sil.abnf
alerque: Yahoo, who's idea was it to have alternate syntax there? (I'll probably `git blame` and find out it was me). How bad would it be do you think to walk that back and only have one syntax? We just make a long tail of work for future tooling if the parser is full of alternative ways to do things.

Omikhleia: The SILE manual actually mentions it: "... multiple parameters are separated by commas or semicolons"
But on my own, I would never have the idea of using the semicolumn, and I would agree to only have one syntax eventually. Unless there was a good reason to have it too, we could deprecate it, as far as II am concerned, or keep it undocumented.

sil.abnf
Omikhleia: I can think to one usage where grouping with braces was probably _not_ expected to result in anything else than AST grouping without temporary settings: the **gutenberg** package, which uses `\alternative{{and}{&}}`.

But it seems to be some weird attempt at working around the "one argument" only in SILE commands, and actually manage more than one argument under the hood...

It _could_ be interested with that purpose in mind, but with also XML parity in mind, I then wonder how to make sense of it.. I mean, if `\command[options][content}` more or less equates to `<command options>content</command>`, I am not sure to understand how the above "alternative" construct would translate. An implicit tag? Huh.

sil.abnf
alerque: A little more thought early on and we might have had a way to pass a table or list as an option without quoting (e.g. if commas were used for one and semicolons for another, `\foo[string=foo; table=val1,val2,val3]`) but I suspect that ship has sailed. Unless we reassign the semicolon to be the *lesser* operator which would just rub me and probably most English speakers the wrong way, we're stuck with the quoting method `\foo[string=foo,list="val1,val2,val3"]`.

Having more that one possible separator just makes more work for users needing to know about both and escape them, more work for parsers, editor syntax grammars, etc.

I vote we un-document this pronto as deprecated and remove it in a future release. I might be able to work out a way to throw a deprecation warning from the LPEG grammar, or maybe not.

sil.abnf
alerque: Yes lets think about XML parity and how other formats might parse and generate a compatible AST. There is no sense in cramming some language feature in SIL markup that is hard/impossible in other input formats. I can see how it current gives deliberate meaning to `content[1]`, `content[2]`, etc. in the AST and I have no idea how to do that in XML.

sil.abnf
alerque: All these months later and I am still of the same mind. We should not have pointless alternate syntax variants: they just make more work all the way around. Hopefully there isn't too much of this in the wild. I wish we'd caught it a lot earlier and could have used the semicolon variant and been able to use commas for multiple values, but I'm pretty sure any attempt to backtrack that is going to make downstream users very unhappy. It just isn't a nice easy thing to shim or reformat.

I un-documented the alternative as a starting point. We'll leave it in the actual parser for a while but look into a way to throw a deprecation warning on it.

sil.abnf
alerque: I've played around with this a bit and it doesn't work the way I would have wanted it to, but since the thing I wast also makes compatibility with other inputters tricky, I think I'll forgo fixing/implementing what I expected here for now.

In the mean time I think we should make `\set{}` work as a settings wrapper even with no settings in the actual command.

alerque:
Surely plenty to iterate on here, but I think it's time for this to join the party in the develop branch.

PR #1730, opened 2023-03-10
fix(packages): Text conversion in bookmarks has spacing issues

Omikhleia:
Several small points:
 - Kerns were overlooked and ignored with a warning.
 - Kerns and glues were always converted to a space, but "small" enough kerns/glues should rather be ignored. E.g. using a small adjustement kern for pseudo-italic correction shouldn't result in a space, although the logic cannot be anything else but heuristic.
 - Resulting string was not trimmed, and could contain repeated internal spaces.

N.B. back-porting this from a bunch of pending fixes I have for my resilient.tableofcontents too.
(I really think the nodeToText function should be moved into the utilities, but this is a separate topic.)

### DIFF
--- a/packages/tableofcontents/init.lua
+++ b/packages/tableofcontents/init.lua
@@ -58,27 +58,34 @@ end
 -- (Similar to SU.contentToString(), but allows passing typeset
 -- objects to functions that need plain strings).
 local function _nodesToText (nodes)
-  local spc = SILE.measurement("0.8spc"):tonumber() -- approx. see below.
+  -- A real interword space width depends on several settings (depending on variable
+  -- spaces being enabled or not, etc.), and the computation below takes that into
+  -- account.
+  local iwspc = SILE.shaper:measureSpace(SILE.font.loadDefaults({}))
+  local iwspcmin = (iwspc.length - iwspc.shrink):tonumber()
+
   local string = ""
   for i = 1, #nodes do
     local node = nodes[i]
     if node.is_nnode or node.is_unshaped then
       string = string .. node:toText()
     elseif node.is_glue or node.is_kern then
-      -- What we want to avoid is "small" glues or kerns to be expanded as
-      -- full spaces. Comparing to a "decent" ratio of a space is fragile and
-      -- empirical: the content could contain font changes, so the comparison
-      -- is wrong in the general case. It's still better than nothing.
-      -- (That's what the debug text outputter does to, by the way).
-      if node.width:tonumber() > spc then
+      -- What we want to avoid is "small" glues or kerns to be expanded as full
+      -- spaces.
+      -- Comparing them to half of the smallest width of a possibly shrinkable
+      -- interword space is fairly fragile and empirical: the content could contain
+      -- font changes, so the comparison is wrong in the general case.
+      -- It's a simplistic approach. We cannot really be sure what a "space" meant
+      -- at the point where the kern or glue got absolutized.
+      if node.width:tonumber() > iwspcmin * 0.5 then
         string = string .. " "
       end
     elseif not (node.is_zerohbox or node.is_migrating) then
       -- Here, typically, the main case is an hbox.
-      -- Even if extracting its content could be possible in regular cases
-      -- (e.g. \raise), we cannot take a general decision, as it is a versatile
-      -- object (e.g. \rebox) and its outputYourself could moreover have been
-      -- redefine to do fancy things. Better warn and skip.
+      -- Even if extracting its content could be possible in some regular cases
+      -- we cannot take a general decision, as it is a versatile object  and its
+      -- outputYourself() method could moreover have been redefined to do fancy
+      -- things. Better warn and skip.
       SU.warn("Some content could not be converted to text: "..node)
     end
   end
--- a/packages/tableofcontents/init.lua
+++ b/packages/tableofcontents/init.lua
@@ -83,7 +83,7 @@ local function _nodesToText (nodes)
     end
   end
   -- Trim leading and trailing spaces, and simplify internal spaces.
-  return string:match("^%s*(.-)%s*$"):gsub("%s+", " ")
+  return pl.stringx.strip(string):gsub("%s%s+", " ")
 end
 
 if not SILE.scratch.pdf_destination_counter then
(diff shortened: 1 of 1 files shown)


packages/tableofcontents/init.lua
```diff
@@ -58,14 +58,19 @@ end
 -- (Similar to SU.contentToString(), but allows passing typeset
 -- objects to functions that need plain strings).
 local function _nodesToText (nodes)
+  local spc = SILE.measurement("0.8spc"):tonumber() -- approx. see below.
```
alerque, line 61: Any chance we could use `1 space - spaceshrinkfactor`? And/or if we are pulling numbers out of a hat I think I would have used the width of a thin space for this, not a regular space.

packages/tableofcontents/init.lua
```diff
@@ -77,7 +82,8 @@ local function _nodesToText (nodes)
       SU.warn("Some content could not be converted to text: "..node)
     end
   end
-  return string
+  -- Trim leading and trailing spaces, and simplify internal spaces.
+  return string:match("^%s*(.-)%s*$"):gsub("%s+", " ")
```
alerque, line 86:
```suggestion
  return string:match("^%s*(.-)%s*$"):gsub("%s+", " ")
  return pl.stringx.strip(string, "%s%):gsub("%s%s+", " ")
```

Probably six of one half dozen of the other, but Penlight's `strip()` uses something faster than `:match()` and the more targeted `:gsub()` that only matches multiples not replacing singles is also faster. It also doesn't do quite the same job so if the intention was to replace tabs or other spaces with spaces then go back to what you had. I don't care about this case since it will be run so few times, go with whatever code makes you happy. Just wanted to point out there are a few more ergonomic / efficient things available to you in Penlight for strings, tables, lists, etc.

alerque (dismissed):

alerque:
> (I really think the nodeToText function should be moved into the utilities, but this is a separate topic.)

 I think I've mentioned it before, but I'm definitely interested in some sort of tree walker that converts to text. Note the text output backend acts as something similar, but doesn't fit the use case so well.

packages/tableofcontents/init.lua
Omikhleia: We could. It's still empirical because cannot know easily what settings (incl. fonts) were active when the kerns or glues were absolutized. Of course, most of the time we'd be quite safe: a basic `\section{Header\kern[width=...] stuff}` will be ok (the kern is absolutized it the context of the header font). The edge cases are things such as `\section{Header\font[size=...]{stuff\kern[width=...] stuf}}` -- I don't expect these to often occur, but who knows... E.g. a small kern within a smaller superscript text, etc.

Omikhleia: Hmm. The real proper way to get a "minimal" space (depending on various settings... we are not _necessarily_ in a context where space shrink/stretch is active, esp. if used e.g. in section titles which might be ragged right...) would be:
```
local sp = SILE.shaper:measureSpace(SILE.font.loadDefaults({}))
local spc = (sp.length - sp.shrink):tonumber()
```
It then takes out the spaceshrinkfactor if the latter is active (= default settings)... But currently gets bigger than 0.8spc due to the 1.2 spaceenlargementfactor we have...

And whatever formula is used, we'd need to apply some margin, to avoid rounding effect (at best) or again when spaces are not variable... In other terms, when comparing "node.width:tonumber() > spc" to decide whether a kern/glue should give a space:
- The earlier implementation was equivalent to `> 0`, i.e. even a small glue or kern would result in a full space...
- The initial PR uses " > 0.8spc" = as the text outputter currently does (without explanation, though) https://github.com/sile-typesetter/sile/blob/c8a1467494b6cbcf3582c56598cb68f55c11df83/outputters/text.lua#L48
- We _could_ make it a bit more clever and use a space with possibly the shrink factor removed as above, but would still need a ratio of it to be safe, or when spaceshrinkfactor is not active... I'm afraid an empirical 0.x is always needed...

packages/tableofcontents/init.lua
Omikhleia: Interesting, I didn't know `pl.stringx.strip()`!

`pl.stringx.strip(string):gsub("%s%s+", " ")`  does the intended thing (IMHO)

There's no need for a second argument to it (and it does work this way, actually, as it expect characters, not a pattern, when specified). I can't say whether it's faster or not - it uses two string.find and a string.sub, FWIW - but it's certainly more readable.

Omikhleia:
> >   (I really think the nodeToText function should be moved into the utilities, but this is a separate topic.)
>
> I think I've mentioned it before, but I'm definitely interested in some sort of tree walker that converts to text. Note the text output backend acts as something similar, but doesn't fit the use case so well

This is a bit orthogonal to having the function (however implemented) in the utilities.

But ok to discuss it... Are you sure you know what the full picture implies?

"A sort of tree walker" is what this function does: the tree walker is just our `SILE.process`, it does its stuff and then we extract the text from the typesetter node queue afterwards :)

In 90% of the cases, it does what's expected; the only limitations I can think to are:
1. It assumes horizontal content (since it works on the typesetter node queue, before shaping, breaking and outputting occur...). This one is real hard to fix properly, if we really really want paragraphs in bookmarks, etc. ^^
2. It skips content wrapped into hboxes, with a warning. This one has "decent" solutions, at various costs. Here I can come with two implementations (naive or not-so-naive). I could clean them up in a separate issue, but note they are overkill for most cases... Unless we _really_ want, say, "Some \rebox[phantom=true]{hidden} content and \rotate[angle=5]{rotated} lines" to be extracted as  "Some content and rotated lines". Currently, indeed, we get "Some content and lines" with warnings on hboxes not being extracted... The naive solution (not using an outputter and not fully generic) yields "Some hidden content and rotated lines" (Oops, the "hidden" stuff appears here). The not-so-naive solution yields "Some content and rotated lines", yay, but at a greater cost and some other issues (cursor "guesses" in the pseudo-text outputter that then needs to be invoked here have to take into account TTB, RTL, etc. - and the existing one doesn't do it very well, IMHO, in such a case where the text will be reprocessed e.g. in bookmarks, etc.). It's all doable, but at a cost that might be too big for the purpose (unless one consider a phantom rebox in a section title not to be an edge case!)

packages/tableofcontents/init.lua
alerque: Yes I think that is better. I don't have a hard evidential use case for why, but at least in my head based on past experiences I think that feels like a more robust heuristic.

Omikhleia: Done

alerque (approved):

PR #1734, opened 2023-03-21
Fix a typo

qur2:
Hey, I was getting started with Sile and while reading the PDF, I spotted a typo. English is not my native language, but checking some dictionaries online suggests that the word `'ether'` shouldn't be used here. 
Happy to hear what you think!


Omikhleia:
I believe the one in the document is actually already covered in #1712 
(But not the one in the README, it would seem).

alerque (approved):
Thanks for taking the time to contribute. I believe you are correct. There are undoubtedly more typos around, don't be shy to correct them or clarify points that were not clear to you on a first-read.

Issue #1570, opened 2022-09-08
Processing XML with the deprecation of the -I cli option


Omikhleia:
Before 0.14, one could process an XML file using a `-I` preamble:
```
sile -I examples/dict-sd-fr-preamble.sil  examples/dict-aq-fr.xml
```
Where the preamble was a self-contained SILE-document (specifying a class with options, possibly tweaking some settings and hooks, etc.)

In 0.14, this still works currently but of course emits a deprecation warning:
```
! -I/--include() was deprecated in SILE v0.14.0. Please use -u/--use or -p/--preamble() instead.
```

However:
- The `-I` cannot just be replaced with just a `-u` obviously (it's not a module, and thus errors) 
- It cannot just be replaced with `-p` as the preamble then file is loaded without honoring the declared class (so settings won't be available, etc.)

I came up with:
- Either `sile examples/dict-sd-fr-preamble.sil -P examples/dict-aq-fr.xml`: works, but inappropriate as the output PDF has, quite logically, the name of the preamble, not of the processed XML file (and adding a `--output` is not very user-friendly).
- Or `sile -u classes.teibook -p examples/dict-sd-fr-preamble.sil  examples/dict-aq-fr.xml`: kind of works (after locally fixing #1569), but the command line is not very user-friendly either (esp. if there were also class options to specify)

Migrating some of these classes, I wonder what would be the best approach to provide something "as simple" as it used to be for such a use case.

alerque:
What kind of content is in your `dict-sd-fr-premble.sil` that would not belong in a dedicated class module? The solution I've used elsewhere was to write a class that inherited (to use your example` *classes.teibook* and adds the use case specific commands for the type of XML I was loading.

I was envisiouning preambles as actual content that would end up inside the primary document. Kind of like a fax cover sheet (yes my beard is greying). I always thought the method of stuffing macros is a preamble to handle a document type was weird.

Another approach I used on a non XML file and later for an XML variant of the same content was to write an inputter module that also specified the default class, and made the default class for that file input type be something that made sense for the content type. Again no preamble required.

I'm quite willing to evaluate other use cases, but if the preamble is just a bunch of tag handling commands it makes more sense to me to write it as a module: whether a class, package, or inputter.

Omikhleia:
> What kind of content is in your dict-sd-fr-premble.sil that would not belong in a dedicated class module

Things that are dependent from the input XML file, a multilingual dictionary, and cannot be fully known by the class:
[...]

PR #1738, opened 2023-03-24
Handle multiple input documents

alerque:
I expect the first document to set the class, and all others (in any input format) to be concatenated at the end with processing just continuing. Some inputters (such as the Lua one) also can return more than one document at once, not sure how that will play here. Not much different than just using a master document and including others into it, but I can think of some CLI workflows there this would be helpful.


Omikhleia:
Interesting idea.
Note that this would perhaps address #1570 if instead of...
```
sile -I preamble.sil document.xml
```
... One could just do a:
```
sile preamble.sil document.xml
```
But I'm unsure, then, regarding the newer preamble/postamble options and how they are supposed to play in this game.

alerque:
Probably need to land this (including a test of some kind) before v0.15.0 so we're more clear on the Rust based CLI/Lua interpreter is actually a drop in replacement.

alerque:
> But I'm unsure, then, regarding the newer preamble/postamble options and how they are supposed to play in this game.

The postamble wouldn't be significantly different than just concatenating a document on the end with more inputs. The preamble is fundamentally different because it is *not* the master document and hence does not set the class or class options. The master document is loaded for that (in the case of the PR, the first input document is assumed to be master, previously only one input could ever be named). Once the class and options from the master document are initialized, then the preambles are processed.
The `--include` option was always ambiguous and your `sile -I preamble.sil document.xml` usage (where I presume you are setting the class in the preamble is not actually a preamble at all, it's the master document and the rest should be a postamble, `sile -P document.xml preamble.sil` or something like that.

PR #1744, opened 2023-03-26
new typesetter instance shall not reset settings

### LINKED ISSUE #1708: New typesetter instantiation kills settings since 0.14.6 (regression)
Document "doh.sil"

``` sile
\begin[papersize=a7,class=book]{document}
\set[parameter=document.letterspaceglue, value=5pt]

Lorem ipsum

Lorem ipsum\footnote{test}

Lorem ipsum

\end{document}
```

SILE v0.14.5

```
alias SILE145='docker run -it --volume "$(pwd):/data" --user "$(id -u):$(id -g)" siletypesetter/sile:v0.14.5'
SILE145 doh.sil 
```

image

SILE v0.14.6 and upper (... up to 0.14.8 current)

``` sh
alias SILE146='docker run -it --volume "$(pwd):/data" --user "$(id -u):$(id -g)" siletypesetter/sile:v0.14.6'
SILE146 doh.sil 
```

image

That's kind of a **regression**! After the footnote (here used as an example), the letter spacing setting (here used as an example) is reset...

Let's add this snippet at the beginning of the document to find the culprit (running with `-t` to get the stack):

``` lua
\lua{
SILE.settings.declare = function (self, spec)
  -- BEGIN MOD
  if self.declarations[spec.parameter] then
    return SU.warn("Settings redeclaration ignore " .. spec.parameter)
  end
  -- END MOD
  self.declarations[spec.parameter] = spec
  self:set(spec.parameter, spec.default, true)
end
}
```

Ignoring re-declarations "fixes" the issue...

Where do these re-declarations come from? This is due to 6bb02d558e551898513cadc3225420f4bf49f423 - with the base typesetter being changed into a class, with the settings declaration moved to its `_init` method (https://github.com/sile-typesetter/sile/blame/v0.14.6/typesetters/base.lua)

Each time a typesetter is instantiated (= here in this example, for the footnote, but that would be true for headers, etc. or any environment that creates a new typesetter for some purpose), it re-declares the same settings, which are reset to their default value.

(It's kind of a colleague to https://github.com/sile-typesetter/sile/issues/1531 ... The shift to full classes, with settings being declared in them, leads to such things too there).

Omikhleia:
Closes #1708 (regression), also adding a test for it.

I am not fully happy with the fix, but it is the only easy way I found that doesn't introduce any additional layer of weird things. I do think it's acceptable **in the current state of things** (for fixing a regression), but the fact that it affected the (folio) on the bidi test somehow shows that you were onto something at #1327, @alerque -- at one point we may indeed need to revisit settings and states with a finer-grain approach.


typesetters/base.lua
```diff
@@ -1,3 +1,81 @@
+--- SILE typesetter (default/base) class.
```
alerque, line 1: How about putting these back in the `class:declareSettings()` but redefining the `SILE.settings:declare()` function to check for existing settings of the same name and warning on collisions and *NOT* doing anything, just skipping the duplicate declaration?

(Note they were in `class:_init()` which is wrong, they should be in `class:declareSettings()` as a future-proofing against scope refactors...)

alerque (changes requested):
Proposed a possible alternate way to avoid the regression that moves us closer to saner scopes...

typesetters/base.lua
alerque: I'll take another look, but my gut reaction is to say that *correctness* is far more important than *efficiency* for this issue. We're about to try to untangle a giant web of cross wired setting scopes and I'm much more concerned about getting them mapped logically than how many typesetters are instanciated. If there is a speed problem (say with `ptable`) then you could instantiate one typesetter, then clone it for each new cell with similar properties.

core/settings.lua
```diff
@@ -111,6 +111,10 @@ function settings:declare (spec)
   if spec.name then
     SU.deprecated("'name' argument of SILE.settings:declare", "'parameter' argument of SILE.settings:declare", "0.10.10", "0.11.0")
   end
+  if self.declarations[spec.parameter] then
+    SU.warn("Refusing to declare previously declared setting: " .. spec.parameter)
```
Omikhleia, line 115: That was my own workaround for now... https://github.com/Omikhleia/resilient.sile/blob/main/resilient/hacks.lua#L19-L21

But isn't a warning going to occur _a lot_ in "normal" situations? (I went for a mere debug log.)

alerque (approved):

typesetters/base.lua
alerque: On second look I'm still of the same mind. I am deeply concerned about the speed of SILE (it should be 100× faster, multi threaded, and support resuming from the last change, not just little optimizations here and there) but the real way to fix that is to rewrite some bits of it in Rust. More than 95% of our time is spent iterating over node structures over and over in a way that Rust could do several orders of magnitude faster.

But I can't even think about making that happen until we have same separation of concerns. Until the typesetter/frame/class/package/pagebuilder/shaper/outputter interfaces have consistent inputs and outputs and don't rely on side effects (like global variables being changed when loading modules rather than on initializing them) I can't start swapping out pieces for higher performance.

alerque: Ya not *executing* code in any modules when they load is going to be super important, everything needs to be return as a table or function, no side effects of `require()` itself...

I should probably setup a CI check for that.

PR #1749, opened 2023-03-28
Speed up CI runs on PRs by aborting obsolete runners

alerque:
- style(build): Cleanup formatting in line with other projects
- ci(actions): Cancel previous long running jobs on new commits


alerque:
Needs to also match the trigger type so PR test merge and branch runs don't cancel each other.

PR #1751, opened 2023-03-28
Get un-numbered chapters back in the ToC

### LINKED ISSUE #1707: Unnumbered chapters do not appear in TOC (regression)
I'm at a loss to explain exactly why, but chapters marked with `numbering=false` do not appear in the table of contents. (Sections and subsections appear as expected.) This seems to have something to do with the `book:chapter:post` hook inserting `\par`. I found that if I redefined that command to do nothing, the unnumbered chapter *did* appear in the TOC. Mystifying!

I also found that the `...:post` hooks for sections and subsections inserts a space, which I think shouldn't happen if the headers are unnumbered as well. Similarly, redefining the hooks to null commands seems to fix the problem.

```
\begin[class=book,papersize="a5"]{document}

% uncommenting the following makes the chapter appear in the TOC, and also avoids extra spaces in sections/subsections
% \define[command="book:chapter:post"]{}
% \define[command="book:section:post"]{}
% \define[command="book:subsection:post"]{}

\ftl{tableofcontents-title=Contents}
\tableofcontents

\chapter[numbering=false]{A chapter}

foo bar zot

\section[numbering=false]{A section}

foo bar zot

\section[numbering=false]{A subsection}

foo bar zot

\end{document}
```

alerque:
Fixes #1707


alerque:
@Omikhleia This *works* but I have no idea why.

The same applies to sections too: if the hook functions don't output some sort of node the caller functions don't end up writing an info node.

Omikhleia:
> @Omikhleia This _works_ but I have no idea why.
> 
> The same applies to sections too: if the hook functions don't output some sort of node the caller functions don't end up writing an info node.

The reason why some "special" boxes (such as the info node) disappear when immediately followed by a line break seems very similar to https://github.com/sile-typesetter/sile/issues/1495 - There's something dubious at play...

But in this very case, why are the pre/post hooks called when the header is unnumbered? (E.g. in a chapter, I was assuming the `\par` was here so that the numbering would be a different line that the title, but why would be still do it if the chapter is unnumbered?

alerque:
> But in this very case, why are the pre/post hooks called when the header is unnumbered?

Because disabling numbering has to do with counters and the hook has to do with page layout. It's quite likely that people have already worked around the spacing they want to use for numbered vs. unnumbered chapters and there isn't a good reason to fix the logic yet.

A good reason is coming up when we add `\volume`, `\part`, etc. in (or can be extracted from) #597.

Omikhleia:
> A good reason is coming up when we add \volume, \part, etc. in (or can be extracted from) https://github.com/sile-typesetter/sile/pull/597.

Well, honestly, I can't really judge here. After all, I took quite different directions very soon in my **resilient.book** (and for more than 1 year now, I have working parts, captioned environments, custom ToC, etc. all with customization and styling possible nearly everywhere at several levels -- and I hope to soon focus appendices and indexes, having no idea what "volumes" are, heh) -- So the directions might take the standard default book class (and, IMHO, _really_, its all too complex hooks) is a bit beyond my reach.

It's just bad to have regressions, so these need at least to be fixes...
But beyond that point, I can't really comment on how it should evolve :cat: Very honestly again, beyond using it for me tests and helping here and there if need be, I'm obviously not really wanting to involve too much time into the default book class. As it stands (with its "hard-coded" noindents, for instance), it's no even usable for French. So... I'm afraid I'll just pass here.

alerque:
I agree the *book* class is terrible. I've been using *cabook* that has all this stuff overhauled for years and years now.

This PR is to fix the regression, not fix the book class. I want to see that happen to so new users have a better starting point, but that will be a breaking release.

Omikhleia:
>  I've been using cabook that has all this stuff overhauled for years and years now.

So we are both using our own book class for real projects :) It tells something. At some point, we may need to re-assess what the default book class in SILE ought to be, but that might be premature for now.

alerque:
That's a good point, the default book class doesn't have to be the most tricked out one. I was (and still am!) a `koma` user, but it's probably just as well that isn't the LaTeX default.

On the other hand the default class should be cover all the basics well, be robust, look pretty good, and get ease people into customization. I don't think it does any of those really.

PR #1752, opened 2023-03-29
Discarded glues at the top of a frame shall not account in glue adjustement

### LINKED ISSUE #1647: \par creates space in the wrong location
In this document:
```
\begin[papersize=a9]{document}
\set[parameter=document.parskip, value=30pt]
\par  % troublemaker
aaa
\vfill
bottom text
\par
\break
\end{document}
```
(papersize a9 is just to create small screenshots)

The first `\par` creates a space on the bottom of the page, lifting the "bottom text" from the bottom of the content frame:
image

Adding a `\hbox` at the beginning of the document or removing the first `\par` yields the correct result with the bottom text aligned to the bottom:
image

Is that the intended behavior?

Omikhleia:
Closes #1647 

It turned out to be a bit trickier than I thought and needs scrutiny...

Regarding impacted tests (besides a dedicated test for that case)

- bibtex.sil (using parskip) was actually revealing the same issue.
  
  image

  The extra space is no longer showing and the page is maximally stretched.

- bug-252 is impacted on page 3, the small space below now disappears. Seems better.

  image

- bug-301 is much harder to judge. It doesn't seem worse, but it's hard to tell.


alerque (approved):

alerque:
I tried this on one of my book project where I was having some troubles and it seems to be a solid improvement. I don't *like* the way settings and variables are scoped, but at least this code makes it *explicit* what is being tracked where and what subsystems need to interface with others where. I think this will make a settings rescope involving typesetters and frames easier in the future.

Is there any reason we can't bump this up to v0.14.9?

I'll look into the `bug-301` test now too.

alerque:
301 has always been a hard bug to replicate. It looks like the change in it is for the better in line with the other affected tests you described. It is still showing something that looks like the same bug, but it actually is different. The unfilled frames are the way they are because of the non-stretchable skips with specific metrics used to trigger the bug.

Omikhleia:
> Is there any reason we can't bump this up to v0.14.9?

Nope, just added it. It was 2AM or so when I completed it, I must have been too tired to notice I had the milestone on the issue but not on the related PR ^^

PR #1753, opened 2023-03-29
docs(manual): Reword introductory sections in a more neutral way

Omikhleia:
Following discussion #1729:
A proposal to rephrase some of the sentences in the manual's introduction, to avoid unwanted comparisons and judgement of values.

I'm not that good at that game - feel free to comment/improve on the matter, @jslabovitz and others!

### DIFF
--- a/documentation/c01-whatis.sil
+++ b/documentation/c01-whatis.sil
@@ -5,121 +5,81 @@
 % earlier without messing up folios in the TOC.
 \set-counter[id=folio,value=1,display=arabic]
 
-SILE is a typesetting system. Its job is to produce beautiful printed
-documents from raw content. The best way to understand what SILE is and what it
-does is to compare it to other systems which you may have heard of.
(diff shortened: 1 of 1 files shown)


jslabovitz:
All these fixes look good to me! They seem to remove a lot of the sticky points about trying to be better than TeX.

alerque (approved):

Omikhleia:
> I've been slowly redoing the wrapping anywhere 

Noted. I tend to wrap at 80 chars naturally (call that age and or small laptop screen, lol), but it makes sense yes!

alerque:
I prefer reading at about an 80 character wrap too, although for code projects I've finally started to agree with Linux Torvalds that 120 is a better target for max-line-width. In the mean time there are some really nice editor settings or plugins depending on you editor that will let you soft-wrap lines at your preferred length without having to hard-wrap them in the saved version.

PR #1754, opened 2023-03-31
fix discretionary handling mess

### LINKED ISSUE #1528: Discretionaries used for measurements incorrectly added to node queue
While working in #1527 on fixing #355 I ran into some crazy pants stuff in regard to hyphenation and discretionaries. I papered over the issue but trying to fix multiple bugs at the same time was turning into a Yak-shave session so I'm breaking out the other bug for solving separately.

As an MWE, one can apply this patch:

```diff
diff --git a/core/hyphenator-liang.lua b/core/hyphenator-liang.lua
index e095816a..46492e78 100644
--- a/core/hyphenator-liang.lua
+++ b/core/hyphenator-liang.lua
@@ -126,8 +126,10 @@ local hyphenateNode = function (node)
           segments[j+1] = luautf8.gsub(nextSegment, "^['’]", "")
         end
         if leadingApostrophe then
-          local replacement = SILE.shaper:createNnodes(leadingApostrophe, node.options)
-          local newNode = SILE.nodefactory.discretionary({ prebreak = hyphen, replacement = replacement })
+          local hyphen = SILE.shaper:createNnodes("_", node.options)
+          local postbreak = SILE.shaper:createNnodes("&", node.options)
+          local replacement = SILE.shaper:createNnodes("^", node.options)
+          local newNode = SILE.nodefactory.discretionary({ prebreak = hyphen, postbreak = postbreak, replacement = replacement })
           newNode.parent = node
           table.insert(newnodes, newNode)
         end
@@ -139,7 +141,10 @@ local hyphenateNode = function (node)
         end
       end
       if j < #segments and not nextLeadingApostrophe then
-        local newNode = SILE.nodefactory.discretionary({ prebreak = hyphen })
+        local hyphen = SILE.shaper:createNnodes("=", node.options)
+        local postbreak = SILE.shaper:createNnodes("#", node.options)
+        local replacement = SILE.shaper:createNnodes("%", node.options)
+        local newNode = SILE.nodefactory.discretionary({ prebreak = hyphen, postbreak = postbreak, replacement = replacement })
         newNode.parent = node
         table.insert(newnodes, newNode)
       end
diff --git a/core/nodefactory.lua b/core/nodefactory.lua
index 478eda76..89e9af7f 100644
--- a/core/nodefactory.lua
+++ b/core/nodefactory.lua
@@ -275,11 +275,6 @@ function nodefactory.discretionary:toText ()
 end

 function nodefactory.discretionary:outputYourself (typesetter, line)
-  -- TODO this is an indication of a deeper bug. If we were asked to output
-  -- discretionaries but the parent nnode is not hyphenated then we're
-  -- outputting things that were only used for measuring "what if" scenarios in
-  -- break point calculations. These nodes shouldn't exist at this point.
-  if self.parent and not self.parent.hyphenated then return end
   if self.used then
     local i = 1
     while (line.nodes[i].is_glue and line.nodes[i].value == "lskip")
diff --git a/packages/tableofcontents/init.lua b/packages/tableofcontents/init.lua
```

Then have a look at the output of `./sile tests/bug-355-tr-hyphenation.sil`. I expect to see a mess of debug characters added in various places, but the mess one gets is not the mess I expect!

image

Omikhleia:
Closes #1528

I'll edit the comment here later to add some insights and explanations, but pushing here what I have so far.

The tests that breaks are those I expected to see failing... The output seems ok to me, but it's our Turkish tests using indeed complex discretionaries, so @alerque another pair of experienced eyes would be good ^^


core/nodefactory.lua
```diff
@@ -213,12 +213,24 @@ function nodefactory.nnode:absolute ()
...
+    -- I guess this was done to ensure proper kerning (vs. outputting each of
+    -- the nodes separately).
     if not self.parent.used then
       self.parent:outputYourself(typesetter, line)
     end
     self.parent.used = true
   else
+    -- It's possible not to have a parent, e.g. N(word) without hyphenation points.
+    -- Either that, or we have a hyphenated parent but are in the case we are
+    -- outputting one of the elements e.g. N(out)D(-) [line break] N(put).
```
Omikhleia, line 233: > `N(out)D(-) [line break] N(put)` 

Actually this would be `N(out)D(-)` (followed by some G(margin) and other zerohbox'es) on one line, and `D(-)N(put)` (followed by some G(margin) and other zerohbox'es) on the subsequent line. In case it matters to be that precise about it...

core/nodefactory.lua
```diff
@@ -275,14 +287,27 @@ function nodefactory.discretionary:toText ()
...
+  -- Example: consider the word "out-put-ter".
+  -- The node queue contains N(out)D(-)N(put)D(-)N(ter) all pointing to the same
+  -- parent N(output), and here we hit D(-)
+
+  -- Non-hyphenated parent: when N(out) was hit, we went for outputting
+  -- the whole parent, so all other elements must now be skipped.
   if self.parent and not self.parent.hyphenated then return end
+
+  -- It's possible not to have a parent (e.g. on a discretionary directly
+  -- added in the queue and not coming from the hyphenator logic).
```
Omikhleia, line 301: > a discretionary directly added...

Typically what shows up in test "bug-1528-discretionary" with using a `\discretionary` call. Only discretionary nodes from the Liang line breaking code have parents. (Again, if that matters for understanding?)

core/nodefactory.lua
```diff
@@ -275,14 +287,27 @@ function nodefactory.discretionary:toText ()
...
+  -- Eiher that, or we have a hyphenate parent.
   if self.used then
+    -- This this the actual hyphenation point.
```
Omikhleia, line 304: > This this
Read "This is". It was late at night. Some of my comments might have misspellings, but I hope the code has none :smile_cat:

typesetters/base.lua
```diff
@@ -728,42 +728,53 @@ function typesetter:breakpointsToLines (breakpoints)
...
       else
         linestart = point.position + 1
       end
+
+      -- Then only we can add some extra margin glue...
+      local mrg = self:getMargins()
+      self:addrlskip(slice, mrg, point.left, point.right)
+
+      -- And compute the line...
+      local ratio = self:computeLineRatio(point.width, slice)
```
Omikhleia, line 745: Erm. Could we do it earlier, before adding the margin skips (and related zerohbox'es etc.), since one of the things it does is actually to skip those? I was tempted initially (it simplifies its code), but it would then break compatibility. E.g. I used the method here, to recompute a full line:

https://github.com/Omikhleia/ptable.sile/blob/9b9ec3bbb5539acb8938fb094a50915facfb4f25/packages/parbox/init.lua#L161

core/hyphenator-liang.lua
```diff
@@ -113,23 +113,19 @@ local hyphenateNode = function (node)
   if #segments > 1 then
     local hyphen = SILE.shaper:createNnodes(SILE.settings:get("font.hyphenchar"), node.options)
     local newnodes = {}
-    local leadingApostrophe, nextLeadingApostrophe
     for j, segment in ipairs(segments) do
+      local leadingApostrophe = nil
```
alerque:
```suggestion
      local leadingApostrophe
```

The purpose here is scope not assignment, I think this style says that more clearly.

core/hyphenator-liang.lua
```diff
@@ -113,23 +113,19 @@ local hyphenateNode = function (node)
...
-        leadingApostrophe, nextLeadingApostrophe = nextLeadingApostrophe, nextSegment and luautf8.match(nextSegment, "^['’]")
-        if j < #segments then
-          segments[j+1] = luautf8.gsub(nextSegment, "^['’]", "")
-        end
-        if leadingApostrophe then
-          local replacement = SILE.shaper:createNnodes(leadingApostrophe, node.options)
-          local newNode = SILE.nodefactory.discretionary({ prebreak = hyphen, replacement = replacement })
-          newNode.parent = node
-          table.insert(newnodes, newNode)
+        local hasNextApostrophe = j < #segments and luautf8.match(segments[j+1], "^['’]")
```
alerque: As a general rule variables or methods named `has...` would be booleans, lets name this based on the content it may or may not have, not just whether it has it or not.

alerque (dismissed):
The old book project I just reran with this looks great, including perfect fixes for several cases I knew about that came out wrong in the print edition.

typesetters/base.lua
alerque: I also have several places where I'm calling that directly, possible for similar off-label use case.

Also I'm not 100% sure it would be *correct* to do it earlier before margins are added. For example I have a project where I'm using a hybrid justifications scheme using a small but *stretchable and shrinkable* right margen. It's not like a full ragged scenario where inter-word spaces are always the same size and the right margin is infinitely flexible, but it allows only relatively minor stretching inside the line that is offset with a little bit more generous stretch *at* the margin. This keeps hyphenation to a minimum even with chunky words in abundance (Turkish can be a bit of a mouth-full) which is especially desirable for the context while still allowing it to hyphenate when it really needs to without bad artifacts like overflows or half-full lines.

In that case not including the margins in the *ratio* of correction being applied to stretches would be a mistake.

alerque:
I've been squinting at the changes in `tests/bug-355-tr-hyphenation.sil` and the ones on the ones on the second page are definitely good fixes. The ones on the first page I'm not so sure about, but neither can I suggest which way is correct. I'll give the test outputs another gander when I've had more sleep.

core/hyphenator-liang.lua
Omikhleia: Agreed!

core/hyphenator-liang.lua
Omikhleia: Agreed, and your `nextApostrophe` is better indeed.

alerque:
I reviewed the 355 test again with sleep and coffee. I improved it a little and documented what was being tested and what the expected results are. I'm convinced it is doing the right thing it was originally meant to test and that the changes in metrics with line ratio calculation differences are all for the better. The remaining abundance of overflow lines is entirely deliberate and expected. I'll open a new issue for possible ways to improve those expectations.

alerque (approved):

typesetters/base.lua
Omikhleia: Agreed. So let's keep the current behavior and API for now. I do think at some point we may wish to revisit the typesetter (and the entanglement with frames, pagebuilder and insertions, notably...), but until then, fixing issues without breaking anything else is the way to go.

Issue #1718, opened 2023-02-23
Paragraph indent applied twice, inside and outside hboxes at paragraph start


Omikhleia:
SILE 0.14.8 (also tested on 0.12.5)

Initially found by  @nphyx 

**MWE**

```
\begin{document}
\use[module=packages.rules]
\use[module=packages.rotate]

\noindent{}Lorem...

Lorem (paragraph indented)

\underline{ipsum}

\rotate[angle=5]{dolor}

\boxaround{toto}% deprecated command, but just to show it there too

\end{document}
```

**Expected**

The content underlined / rotated / boxed should not include the paragraph indent inside it.

**Observed**

![image](https://user-images.githubusercontent.com/18075640/220862597-4cdaae57-efb2-4168-982b-047a47ec9b01.png)

Omikhleia:
Besides the issue as reported by the user, it turns out I had separately hit the same core problem for paragraphs with a _negative_ parindent (and lskip in effect, so to have the whole paragraph indented on all lines but not the first), and I resorted to a kind of hack to compute the width of an initial content (for use in my new footnotes).

https://github.com/Omikhleia/resilient.sile/blob/9cd95ba206b594eea38717cb32e418bad79d19fb/packages/resilient/styles/init.lua#L411-L421

Omikhleia:
Note that with the state introduced in proposal #1702, it seems it becomes a no-brainer fix actually: the typesetter should not init a new line within the context of an hbox creation, so:

```
function typesetter:initline ()
  if self.state.hmodeOnly then return end -- ADDED HERE
  if (#self.state.nodes == 0) then
    self.state.nodes[#self.state.nodes+1] = SILE.nodefactory.zerohbox()
    SILE.documentState.documentClass.newPar(self)
  end
end
```

Omikhleia:
And it also seems, in my own case, that I can get rid of my "hack" with negative parindent (or even avoid the negative parindent wholly, now, with a well placed kern instead, since the box gets exactly where it ought to be)

![image](https://user-images.githubusercontent.com/18075640/221021920-92b82c26-188a-4e99-947c-05cf7adc6703.png)

alerque:
It seems #1770 is a duplicate of this issue and is fixed by the proposed solution. I've opened a PR that includes a test for this case.

PR #1760, opened 2023-04-06
Hack for parindent scoping

### LINKED ISSUE #1699: Paragraph have inconsistent indents after chapter openings

Omikhleia:
Closes #1699

My take at another approach than #1700, trying to go further and "disentangle" `current.parindent` scoping.

This is neverthless a very conservative approach, changing `current.parindent` into a typesetter state rather than a setting, but still "hidden" behind a setting, just to identify the other impacts without breaking anything else (= it could even be used as  compatibility shim, in a way). N.B. The code also includes the suggested fix for #1718 as I wanted to check it too below.

Failing tests:
- tests/bug-192.sil = Language urd with a foreign font, test starting with an empty hbox... I'm at lost understanding the old issue = expectations to check!
- tests/bug-262.sil = Doesn't seem to show any visual difference (pixel-wise comparison)... So seems good
- tests/bug-39.sil = Seems better to me (I can't explain the parindent in the first line of the running header in the old result)
- tests/chapterverse.sil = Also seems better (old test result also showing a parindent in the first line of header, likely unexpected).
- tests/disappearing-skip.sil = Difference seems logical, but this test is very dubious: It says "Where did my skip just go?" -- Which ones? There's a missing parindent indeed, due to #1699 exactly... now solved, but did the text refer to vertical skips rather, and why?

Dedicated check for the issue:
image

```
\begin[class=book]{document}
\nofolios

\section{Raw text}

a b c

d

e

\section{Inital font change}

\em{a} b c

d

e

\section{Initial hbox}

\hbox{}a b c

d

e

\section{Initial language change}

\language[main=fr]{a} b c

d

e

\end{document}
```


Omikhleia:
I'm (almost) probably away next week - so let's take the time to chew this thing and digest it, to consider the next step, after some vacations, some good sleep, and enough coffee or tea ;)

alerque (dismissed):
This seams like a good transition step towards settings *actually* being scoped to our effective modules.

alerque:
* [x] `tests/192.sil` — good, change reflects removal of hbox hack no longer needed and was never applicable to RTL being enabled by BiDi anyway
* [ ] `tests/262.sil`
* [ ] `tests/39.sil`
* [ ] `tests/chapterverse.sil`
* [ ] `tests/disappearing-skip.sil`

alerque (dismissed):

PR #1766, opened 2023-04-08
Fix subsequent line disappearing when a break occurs on a blank line

### LINKED ISSUE #479: Line disappears
```
\begin[class=book]{document}
        \set[parameter=document.parskip,value=12pt]
        \set[parameter=document.parindent,value=0pt]
        \font[family=DejaVu Sans,size=10pt]{

\skip[height=0pt]\glue[width=0em]This is a second line block.\glue[width=0mm plus 100%fw]\break
% \glue[width=0em]\glue[width=0mm plus 100%fw]\break
\glue[width=0em]Blank lines are permitted internally, but they must begin with a "|".\glue[width=0mm plus 100%fw]\break

fooofoo

}\end{document}
```

If you uncomment the commented line, the line saying "Blank lines are permitted internally" disappears from the output PDF.

Omikhleia:
Closes #479 

Conservative approach, skipping empty lines and lines containing only discardable nodes (incl. glues).

Reverting the bit of #1751 that is no longer needed now that the root cause is fixed.


alerque (approved):

alerque:
Commit 0279556 from #1511 was also to address a similar bug. Is there any chance that needs to be rewound as well?

Omikhleia:
> Commit [0279556](https://github.com/sile-typesetter/sile/commit/0279556fa271f5117f04b520bec392d195c0d3db) from #1511 was also to address a similar bug. Is there any chance that needs to be rewound as well?

Hmm. Pretty good question, actually! See comment https://github.com/sile-typesetter/sile/issues/1495#issuecomment-1201911082 (and subsequent exchanges): that `seenHbox` check was already under the radar since last summer... When realizing recently that the line disappearance issue in #479 was somehow related to the same piece of code, it hinted towards understanding the actual issue... It did ring a bell, so to say.

This being said, the changes you and @matteodelabre made for the mbox at the time in https://github.com/sile-typesetter/sile/commit/0279556fa271f5117f04b520bec392d195c0d3db still seems good to me, upon another review now. Before, the math mbox was typed as a "special" node, which sounded fairly dubious anyway, and did not invoke its parent init (so it wasn't even a "is_box" despite apparent inheritance)... 

Hence, it doesn't seem to me we need to change it again -- As I stated at the time, I'm not fully sure regarding the math features, but what I wrote in https://github.com/sile-typesetter/sile/issues/1495#issuecomment-1202880394 comes to a new light now:

> the line-breaking test above might be a bit wrong, as are already a lot of subtle things in the typesetter (...). On the other hand, it also reveals some other things are possibly a bit wrong too, or (at best) have unexplained weird choices (e.g. the mbox, the info node, the insertion nodes).

This very issue here shows that the `seenHbox` logic indeed _really_ proves to be dubious and wrong (and not just "a bit wrong", lol)

Yet, the (then) obscure mbox creation logic was _likely_ wrong _too_ at the time -- Or so I'd say, until shown otherwise: In other terms, we likely had two issues, unrelated but entwined ("a train may hide another"). Commenting every obscure part of the typesetter is important to me, so we know in the future what it does and, most of all, why! (Hint: the pagebuilder would need some love too...)

alerque:
Thanks. That's about the same conclusion I came to as well but the confirmation is very helpful.

Great work ferreting out this problem BTW.

PR #1768, opened 2023-04-09
docs(manual): Add section on measurements and lengths

### LINKED ISSUE #1267: Clarifying weird units, please...

Omikhleia:
Close #1267 (well, the last unaddressed checkbox)

In Chapter 9 (The Nitty Gritty):
- Add "Measurements and lengths" new section (moving content from below and extending it)
- Slightly refactor "Boxes, glue, and penalties" section as appropriate.
- Move "Kerns" (sub)section closer to the above (sounds more logical, as of introducing concept progressively)
- Split "The typesetter" section apart below these

In all the above, add a bit more of autodoc's magic here and there.

### DIFF
--- a/documentation/c09-concepts.sil
+++ b/documentation/c09-concepts.sil
@@ -37,7 +37,7 @@ casting a length to a measurement returns just the base measurement and discards
 casting a measurement to a length sets its stretch and shrink properties to zero.
 
 \begin[type=autodoc:codeblock]{raw}
-local l3 = SILE.length(SILE.measurement("10pt")) -- 10pt, without strech and shrink
+local l3 = SILE.length(SILE.measurement("10pt")) -- 10pt, without stretch and shrink
 local m3 = SILE.measurement(SILE.length("10pt plus 2pt minus 1pt")) -- 10pt
 \end{raw}
 
--- a/documentation/c09-concepts.sil
+++ b/documentation/c09-concepts.sil
@@ -2,153 +2,106 @@
 \include[src=documentation/macros.sil]
 \chapter{The Nitty Gritty}
 
-\noindent{}We are finally at the bottom of our delve into SILE’s commands
-and settings. Here are the basic building blocks out of which all of the
-other operations in SILE are created.
+We are finally at the bottom of our delve into SILE’s commands and settings.
+Here are the basic building blocks out of which all of the other operations in SILE are created.
 
-\note{At this point, it is expected that
-you are a class designer, and will be able to follow the details of how
-SILE implements these commands and features; we will also explain how to
-interact with these components at the Lua level.}
+\note{At this point, it is expected that you are a class or package designer, and will be able to follow the details of how SILE implements these commands and features;
+we will also explain how to interact with these components at the Lua level.}
+
+\section{Measurements and lengths}
+
+Before dabbling into more advanced topics, let’s introduce “measurements” and “lengths” in SILE,
+the two available Lua constructs for representing dimensions.
+
+Measurements are specified in terms of \code{SILE.measurement} objects.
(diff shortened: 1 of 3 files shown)


documentation/c09-concepts.sil
```diff
@@ -2,71 +2,152 @@
...
+\end{raw}
+
+Both of these are used for various purposes.
+In many cases, they are nearly interchangeable.
+Casting from one to the other is straightforward:
+casting a length to a measurement returns just the base measurement and discards the stretch and shrink properties;
+casting a measurement to a length sets its stretch and shrink properties to zero.
+
+\begin[type=autodoc:codeblock]{raw}
+local l3 = SILE.length(SILE.measurement("10pt")) -- 10pt, without strech and shrink
```
Omikhleia, line 40: "without strech ..." --> "without stretch'. It seems I cannot get this word right, despite knowing I do the same mistake over and over :)

Omikhleia: Fixed in a separate commit: a few other occurrence (not mine only, pfew).

alerque (changes requested):

alerque (approved):

PR #1771, opened 2023-04-10
Bring back the drop caps

### LINKED ISSUE #1718: Paragraph indent applied twice, inside and outside hboxes at paragraph start

### LINKED ISSUE #1770: Dropcaps package broke
As mentioned in [this comment](https://github.com/sile-typesetter/sile/issues/1764#issuecomment-1500968177) the dropcaps package is broken on *master*. A usable test case is the website example, which blows up.

alerque:
Closes #1770
Closes #1718

Or will when the time comes. As of opening the PR I don't know what the fix is, only what the problem bisects to. So here is a test case to use when fixing...


alerque:
I'm going ahead with a merge on this, there is just touch of duplication with #1760, but this part *at least* seems to be sound, so merging this will help evaluate the other bits on their own merits.

PR #1776, opened 2023-04-12
Drop internal package manager

alerque:
BREAKING CHANGE: The internal package manager that installed stuff to
the system from inside SILE was deprecated back in v0.13.2. It is now
completely removed. External 3rd party packages are fully supported
using LuaRocks and are much more robust.


alerque:
Marking as draft to keep it out of my visual radar until we're ready to drop the v0.14.x work and commit to a breaking release being the next one. I think we might have another round or two of touchups before we take that dive.

Issue #1465, opened 2022-06-29
Suggestion: a "demo" class and associated document


dontlaugh:
Creating a class can be challenging. I think the manual is pretty good at explaining the low-level mechanisms. And the built in "book" and "letter" classes are instructive.

Perhaps there is room for a "kitchen sink" kind of class, and an associated example document, that marches through some features that SILE is good at.

Some features to demonstrate in a kitchen sink class.

* Section numbering
* Folio/page number manipulation
* Running header manipulation
* Font changes
* Different kinds of line breaks and page breaks
* Widening/narrowing margins (e.g. for leading article abstract)

_I'll add to this list if more comes to mind_

----

Demonstrating these features in a "single page, single column" report style layout would be my vote. It should yield simpler Lua, if I understand correctly.

dontlaugh:
Relates to #455

baubleb:
A mininum working example boilerplate class would be nice to have so we can just build on that.

I tried to take an existing class, strip it down to bare basics (or so I thought) and it still won't work.

Omikhleia:
The default book class is instructive, and a pretty short read (about 250 LoC). I recommend starting with it -- It's what I did back in sept. 2021 when I started writing my own classes, and though I eventually departed a lot from it (abstracting my headers, styles, sectioning logic etc. in separate packages, so kind of an overhaul in the end), it was a neat example to get the work started. Of course, it's not perfect and has its subtleties, but nothing blocking. Still, it could be interesting to see where others get stuck.

alerque:
I minimum working class is not very informative either:

```lua
local plain = require("classes.plain")

local myclass = pl.class(plain)
myclass._name = "myclass"

return myclass
```

That's it. You have a working bare bones class. The next step is to start adding or overriding class functions to do what you want. The mostly likely bit you'll want to start with is an `_init()` function that doesn't just run itself, it runs the init from the parent class. That bit is non obvious but found in every existing class we have:

```lua
function myclass:_init (options)
  -- your stuff here
  plain._init(self, options)
  -- your stuff here
  return self
  end
```

This is spelled out in the manual by the way including code bits to setup a whole class that does ... something just for kicks. I'm not sure what else we should provide honestly.

Omikhleia:
Re-reading that manual section and @alerque 's note above, it would be doable, I think, to slightly refactor it and review the progression order. Maybe I can it a try at some point?

[...]

PR #1778, opened 2023-04-13
Rewriting of the manual's chapter 10 (designing packages and classes)

Omikhleia:
Back from a few days off away from the internet, here is a proposal to address (a part) of #1465

The very first commit adds `self:registerRawHandler()` to the package API, similar to what was done for commands in 4875972c, for the rationale explained in its commit log.

The subsequent commits is a rewrite of the manual's "c10" chapter:

- Change from "Designing Basic Class Files" to "Designing Packages & Classes"
- Change introductory text:
   - Explain the difference between a class and a package (a well-known concept for (La)TeX users, but let's target a general audience)
   - Introduce the Penlight OOP framework briefly. We didn't mention it earlier, in the subsequent sections. Here, users will see `pl.class()` constructs, so it's more than time...
- Add a first section covering packages first = it's much simpler, yet it allows to cover a lot of things.
  - Introduce the bare minimum "boilerplate" -- mostly expanded from @alerque 's comment on #1465
  - Move "Defining commands" from the class section here:
    - Move the "link" command from the class section here too, but remove the XML flavor at this point. It's too soon and makes things more complex that they ought to be. (Trying to keep focused on one thing at a time!)
    - Add a "blockquote" example to use the other things seen in the preceding chapters (dimensions, glue, leaveHmode, temporary settings, etc.)
  - Add "Defining settings", extending the blockquote example
  - Add "Defining raw handlers" -- These were never mentioned it seems.
  - Add "Loading other packages" -- Quite obvious.
  - Tease class hooks, so the picture is quite complete
- Change the section covering classes
  - Introduce the bare minimum "boilerplate" -- mostly expanded from @alerque 's comment on #1465
  - Brief subsection on things (now) identical with packages
  - Add "Defining class options" -- Quite obvious
  - Refactor the initial content as "Changing the default page layout"
  - Refactor "Output routines" as "Modifying class output routines"
  - Split "Interacting with class hooks" apart from the previous
     - This is somewhat distinct, as tackling with class-packages interaction
     - All the end is rewritten -- It was still mentioning the 0.12.x way of doing things in a `return`...
 
One thing I have voluntarily removed is all the discussion on "Exports".
I might be wrong, but it should eventually disappear (We can discuss that in a dedicated issue, though.)

### DIFF
--- a/documentation/c10-classdesign.sil
+++ b/documentation/c10-classdesign.sil
@@ -1,20 +1,301 @@
 \begin{document}
-\chapter{Designing Basic Class Files}
+\chapter{Designing Packages & Classes}
 
-Now we know how to define a frame layout for a single page, let’s try
-to define one for an entire document.
+This chapter describes how to implement your own add-on packages and classes in the Lua programming language, for you to extend the way that the SILE system operates, define new commands and page layouts, or indeed do anything that is possible to do in Lua.
 
-Document classes are Lua files, and live somewhere in the \code{classes/}
-subdirectory of either where your input file is located, your current working
-directory, or your SILE path (typically \code{/usr/local/share/sile}). We’re
-going to create a simple class file which merely changes the size of the
+The default formatting in SILE documents is usually determined by the class used by that document.
+This default look can be changed, and more functionalities can be added by means of a package.
+Sometimes it’s hard to make a decision when it comes to choose whether to write a package or a class, and the difference may seem subtle.
+The basic rule is that if your file contains commands that control the look of the \em{logical structure} of a given type of document, then it’s a class.
+Otherwise, if your file adds features that are independent of the document type, then it’s rather a package.\footnote{Obviously there’s nothing new here for seasoned (La)TeX users, but there’s no harm either stating it for a more general audience.}
+
+SILE relies on the Penlight Object-Oriented Programming (OOP) framework.
+Many components are therefore implemented as Penlight classes (here, in the usual OOP sense).
+Their use below is straightforward and is expected to be covered by examples, but you might also want to read more about it before you start.\footnote{See \href{https://lunarmodules.github.io/Penlight/libraries/pl.class.html}}
+
+\section{Designing a package}
+
+Packages live somewhere in the \code{packages/} subdirectory of either where your input file is located, your current working directory, or your SILE path (typically \code{/usr/local/share/sile}).
+
+\subsection{Implementing a bare package}
+
+A minimum working package inherits from the \autodoc:package{base} package.
+While it is possible to inherit from another existing package, let’s ignore this advanced use case in this primer.\footnote{Programmers will recognize the delegation over inheritance paradigm here.
+If you intend to develop a complete family of packages sharing several common methods, then of course you might be interested in first implementing all of these in a parent package, that your other packages will inherit.}
+
+We need to declare the name of our new package, override the package’s initialization method (that is, its class constructor) and possibly other methods as well, set a documentation string, and finally return our new package.
+
+While its presence is not mandatory, the documentation string usually comes in the form of an embedded SIL document, explaining the purpose of the package and possibly illustrating some of its features.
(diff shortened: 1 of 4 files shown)


documentation/c10-classdesign.sil
```diff
@@ -1,20 +1,301 @@
...
+    local text = content[1]
+    SILE.typesetter:typeset(text)
+  end)
+end
+\end{raw}
+
+\subsection{Loading other packages}
+
+Above, when introducing the \code{_init} method, we left a few placeholder comments.
+Let’s you want to ensure the \autodoc:package{color} package is also loaded, so that the custom \autodoc:command[check=false]{\link} command you implemented can safely invoke it in a \code{SILE.call}.
```
Omikhleia, line 197: "let's **assume**..." maybe? Can't remember what I was writing here :)

documentation/c10-classdesign.sil
```diff
@@ -1,20 +1,301 @@
...
+
+\begin[type=autodoc:codeblock]{raw}
+  self:registerCommand("blockquote", function (_, content)
+    SILE.call("smallskip")
+    SILE.settings:temporarily(function ()
+      local indent = SILE.measurement("2em"):absolute()
+      local lskip = SILE.settings:get("document.lskip") or SILE.nodefactory.glue()
+      local rskip = SILE.settings:get("document.rskip") or SILE.nodefactory.glue()
+      SILE.settings:set("document.lskip",
+        SILE.nodefactory.glue(lskip.width + indent))
```
Omikhleia, line 132: Erm. It works in Lua 5.4 (tested separately), but `lskip.width` is a length, `indent` a measurement. Mixing apples and oranges again?

documentation/c10-classdesign.sil
```diff
@@ -1,20 +1,301 @@
...
+
+\subsection{Defining raw handlers}
+
+“Raw handlers” allow packages to register new handlers (or callbacks) for use with the \autodoc:environment{raw} environment, which content is read as-is by SILE, without being interpretated.
+This is intended for advanced use cases where you may want to provide a way for users to embed arbitrary content (likely in another syntax), and you will provide the complete parsing and handling for it.\footnote{%
+This may be used to implement a “clever” verbatim environment.
+It is also used, for instance, by the \strong{markdown.sile} 3rd-party collection to embed Markdown or Djot content directly in a (SIL or XML) document.}
+
+You can define your own raw handlers at the Lua level.
+Overloading the \code{registerRawHandlers} package method; and within it, use the \code{self:registerRawHandler} function.
```
Omikhleia, line 178: "Overloading" > "Overload", likely.

documentation/c10-classdesign.sil
```diff
@@ -1,20 +1,301 @@
...
+end
+\end{raw}
+
+The \code{self:loadPackage} methods takes as argument a package name, and optionally packages options (as a table).
+
+\subsection{Registering class hooks}
+
+Some packages may provide additional functions that need to be automatically called at various points in the output routine of the document class.
+But let’s return to that topic later, when describing how to set up you own custom class.
+For now, we can conclude our primer on packages, as you should already have all the tools design great packages.
```
Omikhleia, line 213: "all the tools design" > "all the tools to design" likely ^^ Why can't I spell things right, doh.

alerque (approved):
I see a few things that could be tweaked but I'm too sick to actually do them now and this is a huge and necessary step in the right direction. Merging sooner rather than later ... almost wish we had a HEAD version of the manual being posted on our website so newcomers would be less likely to land on unhelpful docs.

Omikhleia:
> I see a few things that could be tweaked but I'm too sick to actually do them now and this is a huge and necessary step in the right direction. Merging sooner rather than later

Thanks! I'll go on working on other documentation-related issues progressively---I'd like to reduce the backlog for some of them which seem doable... And a some point maybe we could consider a full re-reading of the manual. There's no urge, though, take care of yourself!

PR #1781, opened 2023-04-14
docs(manual,packages): Make packages docs independent from the SILE manual

Omikhleia:
Several check-boxes from #1194
 - Package **autodoc** now provides `autodoc:example` and `autodoc:note`
 - Other packages no longer use `\examplefont` and `\note` from SILE's manual macros in their documentation -- and should now have their documentation tooling independent from the manual.
 - The SILE manual itself also uses the new autodoc-provided environments.
     - The changes also include a more thorough use of the other autodoc commands we already had.
     - Some sentences and examples are slightly rewritten (notably, the <note> nested macro obviously needed an update, and the XML code was wrong anyhow... I tested the proposed one (simplified) separately.

Regarding the notes, their design was quite problematic:
  - The 450pt horizontal rules in which they were wrapped were (still) going past the edge frame...
  - There are a few places where we had two notes followed by code examples (also wrapped between rules) - That made a lot of no-so-distinct rules on the page!

I opted for a slightly different visual approach -- I wanted something simple, that would not take me too much time and would be more satisfying...

image


Omikhleia:
Set to Draft for now (still working on improvements)

alerque (approved):

Issue #1338, opened 2022-03-05
docbook support is far from complete...


Omikhleia:
SILE 0.12.2

As a follow-up to #1336, still wanting to process the same markdown file, I went another path... Since we have a **dockbook** class, let's start a new journey!

```
pandoc -s -t docbook markdown.md -o markdown.db
```

Let's start a new journey, with the provided preamble, 

`sile -I docbook.sil markdown.db`

```
! Unknown command literal at markdown.db:17:5: in \literal
```

Erm? Oh this might be a long journey across weird oceans. Ok, let's not bother yet and replace "literal" tags by "emphasis" in the docbook file, and keep things rolling...

```
! Unknown command blockquote at markdown.db:40:3: in \blockquote
```

Ok, let's remove the blockquote manually, and keep things rolling...

```
! Unknown command docbook-section-3-title at markdown.db: in \docbook-section-3-title near 104:5: in \sectiontable: 0x55cb82609f30
```

Ok let's add a dumb `docbook-section-3-title` command to that preamble, and keep things rolling...

```
! Unknown command footnote at markdown.db:163:62: in \footnote
```

Ok let's load the footnote package in that preamble. That's plain wrong (the class might not have support at all for footnotes I guess), but that should skip the issue anyway. Roll, baby. roll.

```
! Unknown command colspec at markdown.db:177:11: in \colspectable: 0x55eb834b7f30
```

Ok, I had the feeling that tables would be an issue... Let's keep rolling by removing them for now in the docbook file...

```
! Unknown command variablelist at markdown.db:184:7: in \variablelisttable: 0x5608c6a07030
```

Doh, let's kill the variablelist for now...

```
! Unknown command literallayout at markdown.db:193:1: in \literallayout
```

Let's kill it too and see what comes next. Ah my bad at last, I need to have that JPEG file. Let's provide it and re-run...

```
! Unknown command textobject at markdown.db:203:11: in \textobject
```

Darn. Removing it too in the dockbook file, I want to see the end of it.

```
! Unknown command superscript at markdown.db:211:79: in \superscript
```

What? By Jove. Ok, let's have an emphasis there, and go on, my merry mariner, on that troubled sea in your frail ship.

Wow. "Land, land, land in sight!". Erm. Sorry. "Output, output, output generated!"

Yay! Yippee-ki-yay. Let's look at it and wonder over the wonders of the new-found lands.

Uh. Ah. Am I authorized to think that It's not really the promised land?

[...]

PR #1789, opened 2023-04-22
feat: DocBook class overhaul

Omikhleia:
WIP = Runny code for #1338 ^^

This ugly baby "decently" processes most of the native Markdown chapter from `markdown.sile`, converted to DocBook with Pandoc, i.e. it should cover a decent subset of DocBook as generated via Pandoc).

It processed less than decently a big technical documentation in DocBook, but at least it doesn't crash.

### DIFF
--- a/classes/docbook.lua
+++ b/classes/docbook.lua
@@ -8,6 +8,116 @@ SILE.scratch.docbook = {
   seccount = {}
 }
 
+-- BEGIN UTILITIES
+
+-- TAKEN FROM RESILIENT XUTILS (v2.1-draft)
+local function getFileExtension (fname)
+  return fname:match("[^/]+$"):match("[^.]+$")
+end
+
+-- String trimming
+local trimLeft = function (str)
+  return str:gsub("^%s*", "")
+end
+local trimRight = function (str)
+  return str:gsub("%s*$", "")
+end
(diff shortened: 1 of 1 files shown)


Omikhleia:
Due to inactivity, and as part of backlog cleaning, I would have been tempted to close it, but the topic was interesting, and the reasons why it didn't occur in this case are quite clear.

But I'd like to also question then:
 - [ ] The removal of the non-working and unmaintained docbook class from the core distribution
 - [ ] The replacement of all mention to it and all **fake** and unmaintained examples on "XML processing" in the User Manual
 - [ ] The eventual death of the **xmltricks** package (_it's a shame_)
 - [ ] New approaches to XML processing (#2111)

Omikhleia:
> The removal of the non-working and unmaintained docbook class from the core distribution

Another solution would be for this package/class mess to be extracted to a separated (3rd-party) module (distributed on luarocks etc.), while being under the "umbrella" of the SILE organization. (The main idea here implies that at least some people are interested in fostering it further, without impediments).

Yet, I wonder if it is worth the effort, as DocBook > Pandoc > Markdown/Djot > SILE > PDF is still a possible workflow with better results so far (Or So I Claim).

PR #1792, opened 2023-04-29
Automated italic correction heuristics

### LINKED ISSUE #428: Correct for advance direction changes when switching to/from italics

Omikhleia:
Closes #428 

Well, at least a working solution, ready for discussion.

It's slightly breaking, changing a warning in hbox construction, which we added in 0.14.9, to the better solution with a typesetter state (I could not find a way to avoid it) - I'd ideally like to have it (in that form or another) in 0.15.0, as it fixes many problems in my new book project, where manually adding kerns in the input is impracticable...[^1] but let's say it's at best a v0.x.0 target for now!

[^1]: The part that touches the harfbuzz Lua wrapper is the most critical to me -- I could run my project with a custom typesetter containing the proposed changes (and I actually already do that for other non-standard typesetter changes).


typesetters/base.lua
```diff
@@ -877,33 +1008,23 @@ function typesetter:makeHbox (content)
...
       migratingNodes[#migratingNodes+1] = node
     elseif node.is_unshaped then
-      local shape = node:shape()
-      for _, attr in ipairs(shape) do
-        recentContribution[#recentContribution+1] = attr
-        h = attr.height > h and attr.height or h
-        d = attr.depth > d and attr.depth or d
-        l = l + attr:lineContribution():absolute()
-      end
+      SU.error("DOH")
```
Omikhleia, line 1027: Doh indeed, a leftover!

typesetters/base.lua
```diff
@@ -315,18 +322,147 @@ function typesetter:breakIntoLines (nodelist, breakWidth)
...
+    for i = #nodelist, 1, -1 do
+      local n = nodelist[i]
+      if n.is_nnode then
+        local items = n.nodes[#n.nodes].value.items
+        last = items[#items]
+        break
+      end
+      if n.is_kern and n.subtype == "punctspace" then
+        -- Some languages such as French insert a special space around
+        -- punctuations. In those case, we should not need italic correction.
```
Omikhleia, line 339: > In those case, we should not need italic correction.

@sile-typesetter/lang-fr Note that is is debatable, and I couldn't find anything about it in typography resources I can access to... I hesitated between:
- No italic correction (the current choice here)
- Italic correction always applied, but it looked really exaggerate with most fonts I tried... and I mean even reasonable choices (e.g. EB Garamond).
- Adding an additional setting so as to define a ratio to apply to the italic correction (0..1) for maximum flexibility... But it sounded far-fetched for that small niche language... and not many words in French end with an _f_... Er. Ok, _turf_ does, for instance. It's not even really French. And it doesn't look that bad anyway...)

Omikhleia: To better explain:
The current implementation cancels italic correction in the cases I marked in red below.
image

Seems to be the french punctuation spacing does not need them indeed.

src/justenoughharfbuzz.c
```diff
@@ -254,6 +256,12 @@ int shape (lua_State *L) {
       lua_pushstring(L, "depth");
       lua_pushnumber(L, -tHeight - height);
       lua_settable(L, -3);
+      lua_pushstring(L, "x_bearing");
+      lua_pushnumber(L, x_bearing);
+      lua_settable(L, -3);
+      lua_pushstring(L, "glyphWidth");
+      lua_pushnumber(L, glyphWidth);
+      lua_settable(L, -3);
```
khaledhosny, line 264: You might as well push the whole glyph extents and be done with it. Other code might need axis to some of it.

typesetters/base.lua
```diff
@@ -315,18 +322,147 @@ function typesetter:breakIntoLines (nodelist, breakWidth)
...
+        if isItalicLike(prec) and not isItalicLike(current) then
+          local precShape = getLastShape(precShapedNodes)
+          local curShape = getFirstShape(shapedNodes)
+          itCorrOffset = fromItalicCorrection(precShape, curShape)
+        elseif not isItalicLike(prec) and isItalicLike(current) then
+          local precShape = getLastShape(precShapedNodes)
+          local curShape = getFirstShape(shapedNodes)
+          itCorrOffset = toItalicCorrection(precShape, curShape)
+        end
+        if itCorrOffset then
```
Omikhleia, line 441: Maybe `and itCorrOffset ~= 0`  --- Kind of useless to insert a zero-width kern here.

alerque: Useless indeed ... as long as we're not also breaking line breaking. Are we supposed to be inhibiting breaks at these points, and if so are we still?

Omikhleia: In the case of the french punctuation spaces, they are non-breakable, so it's fine whatever we do.

But in the general case, your are guessing right... There's a problem indeed :(

image

(I tweaked the correction kern output routine so that is shows a red box)

Ah. A break does not occur any longer at the regular space glue. (If it did, we could still have a small justification issue due to the correction kern in some extreme cases, but that could be _quite_ acceptable... But see below).

It's fixable:

image

(Now, red = kern correction, blue = glue correction)

The implication is that depending on the presence of a (space) glue, the italic correction should be a glue rather than a space. (And also, end-of-line glues are kinda skipped in the line ratio, so it wouldn't affect justification it seems).

Omikhleia: Fixed.

typesetters/base.lua
```diff
@@ -315,18 +322,160 @@ function typesetter:breakIntoLines (nodelist, breakWidth)
...
+
+local function isItalicLike(nnode)
+  -- We could do...
+  --  return nnode and string.lower(nnode.options.style) == "italic"
+  -- But it's probably more robust to use the italic angle, so that
+  -- thin italic, oblique or slanted fonts etc. may work too.
+  local ot = require("core.opentype-parser")
+  local face = SILE.font.cache(nnode.options, SILE.shaper.getFace)
+  local font = ot.parseFont(face)
+  return font.post.italicAngle ~= 0
```
Omikhleia, line 422: Reading https://learn.microsoft.com/en-us/typography/opentype/spec/post#post-table-and-opentype-font-variations:
> Note: The italicAngle value is not adjusted by variation data since this corresponds to the 'slnt' variation axis that can be used to define a font’s variation space. Appropriate post.italicAngle values for a variation instance can be derived from the 'slnt' user coordinates that are used to select a particular variation instance. See the discussion of the 'slnt' axis in the [OpenType Design-Variation Axis Tag Registry](https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxisreg) for details on the relationship between italicAngle and the 'slnt' axis.

I followed the link, but it's all beyond my knowledge.
Would we possibly need to make this italic detection more robust, and how?

alerque: I think we will need to make this more robust, but I also don't think there is any harm in starting with this.

The two situations that don't seem to be covered are:

1. Incorrectly built fonts that *deserve* a value in this data and were not given one.
2. Variable fonts.

Rolling with this should catch *correctly built* static font instances, which is not a bad chunk to bite off first.

src/justenoughharfbuzz.c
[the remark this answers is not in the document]
alerque: Yes, this will need to be preceded by `lua_checkstack(L, 4)` because you're pushing 4 more values on the stack and you need to make sure the VM is ready to allocate enough addresses. The existing check can just be bumped by 4.

I would hold off rebasing this, since it isn't a breaking change (is it?) we might even get it in a point release. Once we're sure I can probably retarget it pretty easily having done the same operation a few times recently.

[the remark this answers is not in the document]
Omikhleia: It might be _very slightly_ breaking on hbox construction logic (but so slightly that I'm unsure any existing user might be affected, unless they did something weird), hence my initial suggestion initially for having it in 0.15 rather than a 0.14.x.

FWIW, for my next official[^2] published book made with SILE[^3], soon to be announced[^1], I'll use a special SILE build with this feature (amongst other things) --> So there's no urge for rebasing etc. -- This is just a heads up / reminder as I noticed #1793 might need an additional check here indeed, not to forget in due time. (But don't hesitate letting me know if a rebase is needed or whatever.)

[^1]: Reviews and proofreading are ongoing, and we still have a chapter missing -- but the plan so far is to get the book ready in Nov. 2023... We are late by all standards, and Nov. 2023 is a best attempt, I'll be fairly busy already that month -- Hence also my (relative) silence lately ;-)

[^2]: What I mean by "official" here = ISBN-registered + (French law applies) legal deposit completed. I've the first, still need to finalize the second ('cause it needs the final page count to be fully completed, and it thus depends whether a missing chapter makes it or not).

[^3]: It's not even a secret, the (current, partially reviewed at this point) book sources are available in one of my repos ;)

alerque:
Neither quite the same problem space nor the same implementation, but just for cross reference:

https://github.com/typst/typst/pull/2334

alerque (approved):

PR #1794, opened 2023-04-30
Revamping the manual's chapter 5

### LINKED ISSUE #1351: Small batch of documentation remarks
A bunch of things I noted at various times and didn't report yet. The first items are mostly documentation-related, the final ones are more on the boundary of bugs/refactors.

_Notes "in passing" on a few package documentation_

chordmode package
- [x] Setting `chormode.lineheight` is undocumented.
  - Its effect (with an absolute default "4mm") is dubious at best. (I haven't checked it, but the rationale for a fixed height not based on font size or baseline skip would have to be explained...)

color-fonts package
- [x] Documentation mentions the COLR table, the code also references CPAL: maybe worth mentioning for informed readers? It's pretty technical in both cases, though.

converters package
- [x] `converters:check` is not documented. (At quick glance, it does perform a file conversion, but all the user will get is (maybe) warnings: not sure why there ought to be such a command exposed to SILE-language level.)

inputfilter package
- Documentation mentions `class:loadPackage()`. Confusing: it doesn't seem necessary (in many other cases of packages using it, the package is just loaded with `SILE.require`) i.e. doesn't seem this package HAS to be class-loaded (as opposed, say, to footonotes etc. where the class SHALL use some of the provided methods). -- _Removing the checkbox here, this was written at the time the new class/package API wasn't fully the same as of now (May 2023), and it does seem decent enough_

features package
- [x] "As mentioned in Chapter 3": Can't we make package documentations independent from the SILE manual and self-explanatory?

footnotes package
- [x] Refers to the "book" class, not generic enough (other classes could use it as well, doesn't need to be a book).
- See also below.

lorem package
- [x] Has an undocumented "counter" boolean option (which might not work well from SIL documents - it's not cast to a boolean...)

twoside package
- [x] Documentation should ideally be self-consistent and shouldn't refer to a "chapter 4",
- [x] Documentation shouldn't ideally refer to the book class as we know it (the use case for twoside is more general that this very class)
- ~~"It has no user-serviceable parts." = Wrong currently, what about `\open-double-page`? But see further below though~~ _This was fixed earlier when new commands were added._

xmltricks
- "In chapter 9, we’re going to use SILE to typeset existing XML documents."... (But this package is maybe ill-named anyway, it doesn't have to be XML-related). -- _Removing the checkbox here, this package is quite bad and needs to be more globally re-discussed anyway_

_Other important notes_ 

- [x] Wrong details about outputInsertions --> Fixed during #1392
  - Manual (`c10-classdesign.sil`) mentions outputInsertions() having to be called at the end of each page.
  - Initial comment in "footnotes" package also mentions it.
  - The footnotes package just re-exports that of the insertions package...
  - ... Which is not actually exported at all by the latter :) (and so obviously, such a function is never to be found in our classes' endPage)
 
 - Regarding `\open-double-page` in the twoside package _N.B. Removing the checkbox here, other commands where added later and the design decision is still bad IMHO but anyway, it's not blocking..._
   - IMHO, it doesn't belong here. The "twoside" package ought to be focused on handling right and left masters correctly (and it already has more serious issues to deal with, #683) 
   - At "best", it should be moved to the "book" class. _Rationale_: The way to do such things properly is actually a class-design decision, and a generic all-purpose command doesn't address it correctly. Some books do not honor the old rule of opening chapters on odd pages. Most books do, but do not repeat the running header of the _previous_ chapter on the blank left page thus inserted (and there are good reasons for that...). Some books don't even have a folio on such page. Or they do for chapters, but not for parts, etc. So there's some sort of separation of concerns. Rather let the class provide its own implementation(s) of the expected sectioning-related logic.

_Other minor things_

 - [x] "There is no shortcut for boldface, because boldface isn’t good typographic practice and so we don’t want to make it easy for you to make bad books." (regarding `\font`). Erm, but:
   - Contradicted by the fact of having `\strong` (via the plain class, IIRC). That command is incidentally discussed in #1048 
   - Named font (width and) weight are mentioned in #285 
   - Whatever, it is quite untrue that  "boldface isn’t good typographic practice". As for underlining etc. it depends on the use case. It's pretty bad in novel-type books to abuse heavy weight, but there are Humanities where you cannot avoid it.

Omikhleia:
I think we can improve this chapter, fix some incorrect or unclear statement, remove/merge redundant details, and overall propose a better documentation structure.

Moreover, while at it, closes #1351


documentation/c05-packages.sil
```diff
@@ -1,33 +1,81 @@
...
+By default SILE will look for packages in a variety of directories:
+
+\begin{enumerate}
+\item{The directory where your input source file is located.}
+\item{The current working directory.}
+\item{The environment variable \code{SILE_PATH}, if defined.}
+\item{The default Lua search path.}
+\item{The directories where SILE is installed on the system, \code{/usr/lib/sile} and
+\code{/usr/local/lib/sile}.}
+\end{enumerate}
```
Omikhleia, line 29: As can be seen in the previous version, we were mentioning the search logic **3** times with different details...
The exact order is not clear to me and would have to be checked.

documentation/c05-packages.sil
```diff
@@ -1,33 +1,81 @@
...
-do anything that is possible to do in Lua.
-
-As mentioned above, loading a package is done through the \code{\\script} command,
-which runs Lua code. By convention packages live in the \code{packages/} subdirectory
-of either your input file’s location, your current working directory, or SILE’s
-installation directory. For instance, we’ll soon be talking about the
-\code{grid} package, which normally can be found as
-\code{/usr/local/lib/sile/packages/grid/init.lua}. To load this, we’d say:
+SILE comes with a number of standard packages which provide additional functionality.
+In fact, the actual “core” of SILE’s functionality is small and extensible, with most of the interesting features being provided by add-on packages.
```
alerque, line 5:
```suggestion
In fact, the actual “core” of SILE’s functionality is small and extensible, with most of the interesting features being provided by add-on packages.
SILE ships with the core libraries plus a small collection of packages covering some common needs; more can be added from 3rd party sources.
```

alerque (changes requested):
Haven't read through this all yet but what I did looks for the better. Just a few ideas here as I saw them.

documentation/c05-packages.sil
Omikhleia: According to the above command, there's some significant differences between my setups (Ubuntu PPA, Docker images, etc.). There's also `lua-libraries` and sometimes `lua_modules` subfolders in the mix, and the `sile` prefix too (e.g. even in the local path)...

Perhaps we should indeed be less detailed, e.g.

```
\item{The directory where your input source file is located.}
\item{The current working directory.}
\item{The environment variable \code{SILE_PATH}, if defined.}
\item{Various directories depending on where and how SILE is installed on your system.}
\item{The default Lua search path.}
```

N.B. The default Lua path _seems_ to come last.

alerque: That working sounds fine.

And yes the default Lua (as defined by how the system Lua was compiled/configured) does come last. All the SILE specific bits whether project specific or related to vendored libraries or SILE itself are prepended to it.

And yes there is a huge a mount of variations here. At least 5 factors are involved each adding (or not) entirely different batches of paths. It took a long time and trial and error to work out something that worked robustly across platforms.

alerque: What am I even talking about. The default Lua search path *needs* to be second to last as currently document. The SILE installation path should be last. Otherwise 3rd party packages installed via LuaRocks have no chance to *override* anything in the default SILE installation. Of course that runs the risk of users breaking stuff, but I think it gives more flexibility to the end user to know what is right for their project too.

If your installation doesn't reflect that then it's a bug.

Omikhleia: Indeed!
Done = simplified in a way that should be sufficient (removing explicit mention of a path in chap. 5 and 10).

alerque (approved):

PR #1805, opened 2023-05-18
Try to clarify that SILE doesn't require a custom input language

alerque:
This PR is because I've been a bit frustrated with chatter abut Typst and folks comparing it to SILE without having any idea what they are talking about. Don't get me wrong Typst has done some things right and other things well. Of course it also does some things poorly or not at all. Sometimes it just makes significantly different design choices and that's fine (e.g. sandboxing everything so documents are "safe" and cannot read the network or write to files at render time vs. SILE that just lets you have a full featured programming language ecosystem and treats documents as unsafe programs).

On the other hand it feels bad to be put down for things that just aren't true. Our documentation is a contributing factor to that confusion.

One of the most egregious issues brought up several times is that (supposedly) Typist uses a generic input language "based on Markdown" while SILE "requires everything to be written in a custom input language". In fact the exact opposite is true. Typst is the one that invented a Turing complete custom input language, while SILE is the one uses generic off the shelf components like XML for input and Lua for programability. And SILE is the one that can actually process Markdown. The Typst language has *similarities* to Markdown that it was derived from, but it's not interchangeable at all. SILE is. The SIL input syntax is there as an easier to type XML alternative, but certainly not required in any way.

&lt;/end rant&gt;

### DIFF
--- a/documentation/c03-input.sil
+++ b/documentation/c03-input.sil
@@ -1,7 +1,29 @@
 \begin{document}
-\chapter{SILE’s Input Language}
+\chapter{SILE’s Input}
 
-Let’s now go back and reconsider the first SILE file we saw:
+First, we need to clear some air.
+In our own earlier documentation we gave the impression the primary way to feed content into SILE was a propriatry language.
+This was \em{never} really true since XML was always fully supported as an alternative.
+It is even less true now that 3rd party plugins can add their own input formats.
+
+Hence this chanpter has been renamed.
+The original chapter title was "SILE’s Input Language", as if there was only one.
+The truth is there \em{is} an input syntax we call "SIL", but even that is perhaps best thought of as a structured data systax rather than a unique language.
+The input strings \code{\\em\{foo\}} in SIL input syntax is 100\% equivalent to \code{<em>foo</em>} in XML input syntax.
+The SIL input syntax is provided as an easier to type alternative than XML which can be a bit verbose and tedious to work with by hand.
+On the other hand if you're handling data written by some other program, XML might be a much better solution.
+
+3rd party packages can also add their own input formats.
+The stipulation is that an inputter parses some content and returns an AST with content tags and options recognized by SILE.
+Markdown, Djot, and others have packages available that provide inputters, allowing them to be first class input candidates.
+Many others like Dockbook and TEI are just XML with a special tag schema.
+SILE can be extended to cover any tag schema.
+
+With that in mind, lets go back and reconsider the first SILE file we saw that was in SIL input syntax.
+
+\section{The SIL flavor}
+
+In the Getting Started chapter we looked at this document:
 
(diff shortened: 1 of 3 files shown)


documentation/c03-input.sil
```diff
@@ -1,7 +1,27 @@
...
+
+Hence this chanpter has been renamed.
+The original chapter title was "SILE’s Input Language", as if there was only one.
+The truth is there is a SIL input syntax, but even that is perhaps best thought of as a structured data systax, not a language.
+The input strings \code{\\em\{foo\}} in SIL input syntax is 100\% equivalent to \code{<em>foo</em>} in XML input syntax.
+The SIL input syntax is provided as an easier to type alternative than XML which can be a bit verbose and tedious to work with by hand.
+On the other hand if you're handling data written by some other program, XML might be a much better solution.
+
+3rd party packages can also add their own input formats.
+The requrement is that they parse some input and spit out an AST with content tags and options recognized by SILE.
```
Omikhleia, line 17: typo "requirement"

Omikhleia:
@alerque I wholly agree, and I have no rant against Typst (I haven't tried it yet, actually) -- but I also came to the same conclusion as you: the SIL language is perhaps made too prominent; many people do not read past that point and assume SIL to be the only choice.

E.g. the "Are We Digital Yet" website lists SILE has having a "custom language", which I tried discussing notably here: https://github.com/tajpulo/arewedigitaltypesettingyet.com/issues/3 -- One of the appeal of SILE for me is that it's made in such a way that it can virtually process any language with an appropriate "inputter", and that everything being replaceable, one can add such inputters at will (and this is even made simpler now since 0.13 and 0.14 where you revisited the object model etc.)

So SILE comes with SIL "TeX-like" and SIL "XML-flavored" _by default_, but it's quite language agnostic by default - After all, indeed, we have existing solutions for Markdown, Djot, even raw Pandoc AST JSON.... And subsets of big XML schemas: DocBook (far from complete but it shows some way), TEI[^1] etc.

[^1]: I've back pain issues currently due to a bad fall, so I haven't been very prolific lately -- Can't stay too long sitting and typing  -- but one of my latest projects deals with another subset of TEI (different from the dictionary subset). Slowly but surely, I'll made it public when I can return to it decently ^^ ;)

alerque:
Just to be clear, my rant was not against Typst itself or the developer community, it was much more against bystanders commenting on typesetters from the sidelines without any clear understanding of what they are talking about. Maybe I should just stay off Hackernews, Lobste.rs, et al.

Omikhleia:
> Maybe I should just stay off Hackernews, Lobste.rs, et al.

Heh :p ... Maybe I should stay away from some Typst-related documents too:
https://github.com/sile-typesetter/sile/discussions/1729#discussioncomment-5695994

*Grins*

PR #1808, opened 2023-05-19
Add Cantonese and Standard Chinese Translations

Pi-Cla:
I also touched Japanese because I don't think there should be spacing between words in that language.


alerque:
Nice work, and thanks for the contribution. I tweaked your commit messages (you mentioned the JA TOC title but what you actually changed was the chapter headers). I also had to update the expected output of one test to reflect that change, but no big deal. 

As for what goes in the `language/*.lua` files, that is highly variable depending on the language. Those are support files that tweak various functionality to accommodate the language. Many languages in there have hyphenation patterns. Some have other special line and frame breaking rules. Some have special numbering functions. Some tweak other settings or typesetter internals.

For Cantonese and Chinese it will take somebody with the expertise to know what *should* happen to properly typeset the languages, then we can figure how to accomplish whatever those things are. Feel free to open issues about any special handling those languages require.

alerque (approved):

PR #1815, opened 2023-05-29
docs: inputters and outputters

### LINKED ISSUE #493: API documentation
I mentioned this over breakfast after your FOSDEM presentation, but unfortunately I haven't had much time to look at sile since then:

I'm interested in being able to use sile for typesetting inside other programs and some of the recent work on lua bridging to other languages (e.g. sol2) makes this even more attractive.  Unfortunately, there seems to be a lack of documentation for the Lua code.  Given the goal of allowing arbitrary parts of sile to be replaced, this is unfortunate.  For example, for on-screen rendering I would like to be able to provide sile with direct access to in-memory data structures describing markup (rather than having to convert to XML and send that to the lua code) and then retrieve the glyph positions within a set of frames defined in code, but then use my own code for displaying the glyphs.  In the absence of documentation, and without the desire to read *all* of the sile code, it's not clear where to start with such an endeavour.

Omikhleia:
Closes #493

- [x] Inputters
- [x] Ouputters
- [x] Remove unused `developers.sil` chapter
- [x] Fix chapter file numbers


documentation/fig-input-to-output.png
alerque, line 1: Is it possible to replace the raster asset with something vector like SVG or PDF?

Omikhleia: pdf it should be then, since it contains text. the quick graph could be made nicer too. not sure when i can go on with this pr though, got a finger problem (typing this very message with one hand, very slowly lol). feel free to move on id urgent, otherwise it will have to wait i am fully operational ^^

alerque: Geçmiş olsun, and don't stress over this.

Omikhleia: Done (with branch rebased and also other typos fixed in the initial text)
The PDF is obtained with `dot documentation/fig-input-to-output.dot -Tpdf > documentation/fig-input-to-output.pdf`.

(My finger strain issue is annoying... Just starting to recover from it very slowly, but I still must be cautious when typing, so I am slowly resuming my work here and there)

alerque: Lets add a rule for regenerating the PDF to the Makefile even if we already have the pre-generated asset tracked so we don't have to go hunting for this comment if we ever need to change it.

documentation/c11-inputoutput.sil
```diff
@@ -0,0 +1,197 @@
...
+inputter.order = 2
+
+function inputter.appropriate (round, filename, _)
+  -- We will later change it.
+  return false
+end
+
+function inputter:parse (doc)
+  -- We will later change it.
+  return tree
```
alerque, line 58: My work with stricter languages like Rust is poisoning my brain. Would it be helpful to make even these incomplete examples valid code?

```suggestion
  local tree = {}
  -- Later we will work on parsing the input document into an AST tree
  return tree
```

documentation/fig-input-to-output.pdf
alerque, line 1: WE actually generate the manual in CI and require extra tooling beyond what SILE and even developer made needs to regenerate the manual already, I might look into adding `dot` to that list. We might make more use of this elsewhere in the manual too....

alerque (approved):

documentation/fig-input-to-output.pdf
Omikhleia: Of course, if Graphviz dot is available in the build environment, SILE might consider using embedders.sile to do the conversion. :pig_nose:

alerque: Yes we could do that do, although I kind of don't want to get into the circular dependency game where we use downstream addons while building/testing the upstream typesetter, so at least for now I'm just as happy to generate the PDF file manually in our existing build system.

...which I just did, but ran into some funny business with LuaJIT when building the manual. Looking into it now.

documentation/fig-input-to-output.png
alerque: I did this, still working on CI...

alerque:
So I've bumped into something weird locally. Using your pregenerated PDF figure I can build the manual fine, but using my Graphiz generated one I cannot. The libtexpdf image thing throws an error about it being an invalid image because it doesn't have a valid bounding box. Looking at the PDF it looked fine to me. Diffing them shows we're not generating quite the same thing:

20230818_05h19m58s_grim

Not entirely different, but everything is shifted around just a bit. I would have expected identical outputs visually and possibly even reproducibly byte for byte the same PDF. Also why mine doesn't have a valid bounding box is a mystery.

Gota go put the kids to bed, this message is partly just a note to my future self, partly a hope that somebody recognizes the issue and knows how to make Graphiz behave off the top of their heads...

alerque:
The `dot` generated PDF by @Omikhleia is PDF 1.5 and has a `/Type /Page` as the top level object, the one I generated has `/Type /ObjStm` in the same location.

I assume we probably have wildly different versions of Graphiz. Mine is 8.1.0, how about you Didier?

With that out of the way we probably have a bug/deficiency in libtexpdf to work with in that it doesn't understand new PDFs very well. Also I need to figure out how to output an older flavor...

Omikhleia:
> I assume we probably have wildly different versions of Graphviz. Mine is 8.1.0, how about you Didier?

Might have been a 2.48, if I generated it on my main host, doh. Can't remember where I did it, but checking the PDF with `pdfinfo`, the "culprit" might actually be the version of the cairo lib  = 1.16 in my file, but what you describe seems to correspond to 1.17.x with object stream compression: https://gitlab.freedesktop.org/cairo/cairo/-/commit/bd514f6b08c1b31a75948fd99c147319e5aa649f

alerque:
Cairo does turn out to be the culprit (mine being 1.17.8) and there doesn't seem to be a way to make it use an older encoding either. I worked around it for now using Ghostscript to rewrite the PDF targeting PDF 1.5, but that's a hack on top of a hack. The real issue is our libtexpdf driver's bounding box functions not deciphering the content stream. I presume that's something that's fixed in newer versions, but backporting it is non trivial.

I also copy-edited the new chapter just a little and tweaked the graph as well. All look sane to you?

Omikhleia:
Thanks for the edits @alerque, looks good :+1:

PR #1821, opened 2023-06-24
Numbered equations

leorosa:
With this PR, a new 'numbered' mode is added in math, in addition to 'display' and 'text'. Equations in this mode have the same appearance of the ones in the 'display' mode, except that a counter is added at the right end of the line.


packages/math/init.lua
```diff
@@ -16,6 +16,8 @@ function package:_init ()
       return value * SILE.settings:get("math.font.size") / 18
     end
   })
+  self:loadPackage("counters")
+  SILE.scratch.counters.equation = { value = 0, display = "arabic" }
```
Omikhleia, line 20: Maybe we should avoid tapping into "scratch" variables (I know we do it in a lot of places, but it's kind of annoying IMHO to expose such internals at package boundaries).
Wouldn't `SILE.call("set-counter", { id="equation" })` just work?

Omikhleia: (Or even do nothing, increment-counter will create the counter if needed)

leorosa: Thank you for your insights. In fact, in my first attempt, I did not initialize the counter (didn't even load the counters package), and everything worked fine. But I thought that having equation set to 0 would be better, for whatever reason.

packages/math/typesetter.lua
```diff
@@ -129,6 +129,8 @@ end
 local function handleMath (_, mbox, mode)
   if mode == 'display' then
     mbox.mode = b.mathMode.display
+  elseif mode == 'numbered' then
```
Omikhleia, line 132: instead of a new mode ("numbered"), I wonder if it wouldn't be more general to have options on the "display" mode.
E.g.
```
\math[mode=display]{...}% unnumbered
\math[mode=display, numbered=true]{...}% numbered with counter 'equation'
\math[mode=display, numbered=true, counter=xxx]{...}% numbered with (user-specified) counter id 'xxx'
\math[mode=display, numbered=true, number=N]{...}% numbered with (user-specified) value N'
```

Or would it be an overkill?

packages/math/typesetter.lua
```diff
@@ -149,6 +151,18 @@ local function handleMath (_, mbox, mode)
...
+  elseif mode == "numbered" then
+    SILE.typesetter:endline()
+    SILE.typesetter:pushExplicitVglue(SILE.settings:get("math.displayskip"))
+    SILE.typesetter:pushGlue(SILE.nodefactory.hfillglue())
+    SILE.typesetter:pushHorizontal(mbox)
+    SILE.typesetter:pushGlue(SILE.nodefactory.hfillglue())
+    SILE.call("increment-counter", { id="equation" })
+    SILE.typesetter:typeset("(")
+    SILE.call("show-counter", { id="equation" })
+    SILE.typesetter:typeset(")")
```
Omikhleia, line 163: maybe we should move the formatting in a hook command, so that users or 3rd-party package could override it?
(If going with options as suggested in my other command, those could also be passed through to the hook)
This could offer a customization point for styling differently, rather than hard-code the parentheses in that low level function. (eg. add color or font style; subscribe to label refs, etc.)

packages/math/typesetter.lua
leorosa: Using options is more general, and the examples given are appealing. In fact, I had some trouble in naming a new mode, since in it the equations are the display mode, with numbers.
However, I also like the idea of not having to type a lot of options to get numbered equations. LaTeX also do something like this, with the `displaymath` and `equation` environments.
We can have several options, and provide a script, e.g. `\equation{...}`, that calls `\math[mode=display, numbered=true, counter=equation]{...}`. What do you think?

packages/math/typesetter.lua
leorosa: I agree with that. My initial intention was just to provide a quick mean to get what I almost always need, and not dealing with the plethora of variations in equations numbering. Letting anyone be free to achieve what they want (a numbering like "[chapter.equation]' comes to my mind) would be best. I'll see what I can do, but I'll probably need some help ;)

packages/math/typesetter.lua
Omikhleia: > What do you think?

As long as we the have the core / low-level features, personally I don't have a strong opinion regarding "shortcut" convenience commands.

In my recent projects I use the SIL language less and less, favoring Markdown or Djot whenever I can (using [markdown.sile](https://github.com/Omikhleia/markdown.sile))

In Markdown (resp. Djot) there's just `$...$` and `$$...$$` (resp. the Djot corresponding syntax) for inline and display math, but one _could_ use Pandoc spans (resp. Djot direct attributes). E.g. in the Djot support, I could add something such as:

```
$$`formula`{ numbered=true }
```

Or even a pseudo-class:
```
$$`formula`{ .equation }
```

This could perhaps be interesting, hence my previous suggestion for more flexibility, but to be honest I don't type much maths usually so I am not sure what people might expect :)

packages/math/init.lua
```diff
@@ -353,6 +353,16 @@ Finally, here is a little secret. This notation:
 
 \noindent In other words, the notation using \code{&} and \code{\\\\} is only a syntactic sugar for a two-dimensional array constructed with braces.
 
+\paragraph{Equations numbering}
+Equations can be numbered in display mode adding the options \code{numbering} and \code{counter}.
```
Omikhleia, line 361: `\autodoc:parameter{numbering]` etc.
And possibly `\autodoc:parameter{counter=<id>}` a bit below.
(This gets syntax coloring if enable, and one day I hope we'll subscribe to a revamped indexer too)

packages/math/init.lua
```diff
@@ -67,16 +68,28 @@ function package:registerCommands ()
...
+    local counter = (options and options.numbered) and "equation"
+    counter = (options and options.counter) and options.counter or counter
     xpcall(function()
       mbox = self:ConvertMathML(self:compileToMathML({}, self:convertTexlike(content)))
     end, function(err) print(err); print(debug.traceback()) end)
-    self:handleMath(mbox, mode)
+    self:handleMath(mbox, mode, counter)
+  end)
+
+  self:registerCommand("math-counterstyle", function (options, _)
```
Omikhleia, line 85: maybe `math:counterstyle` (most hooks use colons, but SILE might not be very consistent in naming, just asking)

packages/math/init.lua
```diff
@@ -67,16 +68,28 @@ function package:registerCommands ()
     xpcall(function()
       mbox = self:ConvertMathML(content, mbox)
     end, function(err) print(err); print(debug.traceback()) end)
-    self:handleMath(mbox, mode)
+    self:handleMath(mbox, mode, nil)
```
Omikhleia, line 71: I don't use the MathML syntax much (vs. TeXlike math syntax), but display math equations could be numbered there too, no?

packages/math/init.lua
```diff
@@ -67,16 +68,28 @@ function package:registerCommands ()
...
   end)
 
   self:registerCommand("math", function (options, content)
     local mode = (options and options.mode) and options.mode or "text"
     local mbox
+    local counter = (options and options.numbered) and "equation"
```
Omikhleia, line 77: `options.numbered` -> `SU.boolean(options.numbered, false)`

(Coerce to boolean, defaults to false, errors if not a boolean.)

packages/math/init.lua
```diff
@@ -67,16 +68,28 @@ function package:registerCommands ()
...
+    self:handleMath(mbox, mode, counter)
+  end)
+
+  self:registerCommand("math-counterstyle", function (options, _)
+    SILE.typesetter:typeset("(")
+    SILE.call("show-counter", { id=options.id })
+    SILE.typesetter:typeset(")")
+  end)
+
+  self:registerCommand("equation", function (_, content)
```
Omikhleia, line 91: I might be wrong (untested) but i don't think this will work, due to the fact that this won't trigger the right parser (regular AST vs. raw)

packages/math/typesetter.lua
```diff
@@ -144,9 +144,15 @@ local function handleMath (_, mbox, mode)
...
-      SILE.typesetter:pushHorizontal(mbox)
-    end)
+    SILE.typesetter:pushGlue(SILE.nodefactory.hfillglue())
+    SILE.typesetter:pushHorizontal(mbox)
+    SILE.typesetter:pushGlue(SILE.nodefactory.hfillglue())
+    if counter then
+      SILE.call("increment-counter", { id=counter } )
+      SILE.call("math-counterstyle", { id=counter } )
+    else
+      SILE.call("hbox")
```
Omikhleia, line 154: I suppose you did that to cancel the fact that regular glues are ignore at an end of line?
If so, i think this empty hbox trick can be avoid, if the glue is made explicit (i.e. on line 149 above, use `pushExplicitGlue`)

packages/math/init.lua
Omikhleia: Also I don't think you need to check "option" (IIRC, it is set to `{}` if no options are passed), so this might suffice:
```
local counter = SU.boolean(options.numbered, false) and "equation"
```

alerque:
Sorry I haven't been able to jump in with PR review on this yet. I do appreciate the contribution and am pretty confident it will be worth merging. That being said there are a few unresolved comments from @Omikhleia that are worth resolving, and then there will be some commit message cleanup to happen before this is mergable. I am going to bump this one minor patch release cycle because it's holding up a patch release for stuff that's been in master for a while and I kind of need to be available in production.

packages/math/typesetter.lua
Omikhleia: Checking how LaTeX behaves in justified/raggedleft/raggedright/center - additionally nested in a block (quote)[^1] 

image

[^1]: problematic with our current implementation of ragged environments (not nesting, not keeping parindent, so basically behaving differently, in a questionable way)

We should design a test scenario and check how we compare...

alerque:
Still a bunch of unhandled review comments here that look like they make sense to me. Bumping again...

packages/math/init.lua
leorosa: Thanks for the sugestion. In this regard, I'll also change `\equation` and `\math`, below, to use `autodoc:command{..}`.

packages/math/init.lua
leorosa: No problem. If most hooks use colons, it makes sense to do this change.

packages/math/init.lua
leorosa: I have never used it, but yes, they could be also numbered (I need to test it).

packages/math/typesetter.lua
leorosa: Thanks for the `pushExplicitGlue` suggestion -- I didn't knew that!

As a (really) side note, since you brought the ragged environments topic: when you suggested to use options to number equations in the display environment, I also wondered: what if we had an option to place the number on the left, or on the right side? Playing a bit, I saw that, surprisingly (to me), documents with `direction=RTL` already place the numbering on the left side (although the equation may be bugged -- eq. 1 is not centered, and parenthesis in equation 2 are wrong)

math-RTL

Since left numbered equations are not my use case, I opted to not add any option in this regard.

packages/math/init.lua
Omikhleia: Becase you are lucky with "a+b+c", the text is quite innocuous :smile:

Try, and see it failing:
```
\equation{e^{i\pi} = -1}
```

Ouch! When parsing the input `.sil` file, SILE invokes the SIL inputter, which implements the SIL language grammar and builds an abstract syntax tree (AST) for processing. The content of the `\math` command, however is not part of the "usual" language syntax and needs a different grammar. It is, after all, (a subset of) the TeX math syntax, with commands with multiple arguments, special use of brackets, etc.)

The inputter reserves a few special keywords, such as `\script`, `\lua` etc. -- And, you'll now guess it, `\math` too, to tell the parser that the content here should not interpreted as SIL, but just extracted and parsed as a string that will later be parsed by another dedicated inputter (or, in the case of math a pseudo-inputter, not exposed as one, but that's an implementation detail).

Now, we cannot reserve, obviously, too many keywords -- First we have to know them before parsing a file, so it cannot really be dynamic; moreover those can't be overriden or redefined...

Now instead of:
```
self:registerCommand("equation", function (_, content)
    SILE.call("math", {mode="display", counter="equation"}, content)
end)
```

Try this:
```
self:registerRawHandler("equation", function (_, content)
  SILE.call("math", {mode="display", counter="equation"}, content)
end)
```

And in you SIL document, use: `\raw[type=equation]{e^{i\pi} = -1}`
... And it works. As you may also guess, `\raw` is another of the reserved keyworks.

The implication is that SIL cannot have an `\equation` command as you wanted -- Unless the maintainers decide to add it to the list of reserved keywords (_but see below_). It has to go via the `\raw` trick.

That's exactly what a few other packages and inputters do -- With my own markdown.sile, for instance, I use it too, so one can have:
```
\begin[type=markdown]{raw}
Now I am in _[Markdown]{.underline}_\
And $e^{i\pi}$ is a formula.
\end{raw}
```

There's a brief mention of raw handlers in the manual, but it is not really explained. Perhaps something could be added (reusing some of the explanations above is ok, but I'd appreciate if other hands can write a better explanation.

_Unless the maintainers decide to add it to the list of reserved keywords:_ The problem I see there is that we cannot reserve keywords in-definitively for any new use case that pops up. Today `\math`, tomorrow `\equation`, then why not `\formula` later, and so on... And eventually, SIL loses flexible command redefinitions...
Besides, math could have been a 3rd-party package, it does not have to be (strictly speaking) a core feature, or extra ties to the SIL language.
We already have enough reserved keywords, in my opinion. (All things equal, XML has only _one_ similar concept, `<![CDATA[...]]>`.)

But that might be _my_ opinion only.
Anyhow, I hope the above explains the topic well enough.

packages/math/init.lua
Omikhleia: BTW, I haven't tested, but with math being a reserved keyword (see other command), I think that `\autodoc:command` might have troubles with it (or not do behaves as intended)

packages/math/init.lua
leorosa: Thank you for the explanation. Some concepts are new to me. In summary, it's not worth creating a (reserved) `\equation` just for the sake of not having to type `\math[...]`. Hence, I removed this command in the revised PR.

Omikhleia: I opened #1854 to track the documentation aspect. Resolving the conversation here.

leorosa:
So, I have made a few chances in this PR.
(As I am not very used to github,) Do I need to mark the respective conversations to 'resolved'?
Moreover, there are errors in some checks ('commitlint', 'luacheck', 'test'). How can I resolve them?

Omikhleia:
@leorosa 
> So, I have made a few chances in this PR.

Unless I am not fully awake (which is possible!), I don't see the changes here. Are you sure you pushed them.

> Do I need to mark the respective conversations to 'resolved'?

A good practice (that we don't always follow, perhaps) is for the person who made a remark to mark it as resolved after having reviewed the applied changes.

packages/math/init.lua
alerque: There are two levels of hackyness I'd like to mention, more for documentation purposes than because it changes anything here.

1. It is possible to pass strings to commands as option values instead of as content, e.g. `\equation[input="e^{i\pi} = -1"]` and fetch it out of `options.input`. This is just moving the cheese though and introduces a whole different set of parsing limitations because of the quoting and other splitting that might go on. I suggest only using this for very special purpose strings like single characters or alternate content where it makes sense to *not* have it marked as content.
2. It might be possible to pass functions some raw text version of their input content as long as we could guarantee that the content *also* parsed correctly into an AST tree. Again this is unlikely to be a good route to go for math syntax, but I could imagine some scenarios where it could be useful for a function to have both access to the parsed AST *and* the text-only pre-parsed version of said text. Note this would not get around our current limitation on raw inputs where one either has to have a balanced number of braces *or* never have the `\end{foo}` sequence corresponding to the tag.

alerque:
Live @Omikhleia I'd be happy to give this some more review, but I also don't see any new content. There was a merge commit pushed in the last 2 months but it seems to be all content I expect from a merge, not anything releated to this PR. Are you sure there are not more commits you have locally or even uncommitted stuff locally that needs to get pushed here?

leorosa:
Another try.
(I think that I've pushed these changes to "my master" a week ago, if that makes sense.)

alerque:
That's better, now there is something new to review. We'll have to fix the commit messages but don't worry about that yet, we can fix that after everything else is reviewed and ready.

packages/math/init.lua
```diff
@@ -344,6 +359,18 @@ Finally, here is a little secret. This notation:
 
 \noindent In other words, the notation using \code{&} and \code{\\\\} is only a syntactic sugar for a two-dimensional array constructed with braces.
 
+\paragraph{Numbered equations}
+Equations can be numbered in display mode adding the options \autodoc:parameter{numbered} and \autodoc:parameter{counter}.
+When \autodoc:parameter{numbered=true}, the equations are numbered using a default "equation" counter.
```
Omikhleia, line 364: nitpicking: "example" --> “example” with curly typographical quote marks.

packages/math/init.lua
```diff
@@ -344,6 +359,18 @@ Finally, here is a little secret. This notation:
...
+A different counter may be set by using the option \autodoc:parameter{counter=id}, and this setting will also enable numbering. 
+The default numbering format is "(xx)", but this style may be overriden by defining a custom \autodoc:command{\math:numberingstyle} function, e.g.
```
Omikhleia, line 366: "(xx)" --> I would suggest \autodoc:example{(xx)}` so that it is typeset in the example font (Cormorant Infant currently) as other examples of "rendered content". We could avoid the quote marks here, then: no need for double marking! (But if we want to keep them, then use typographical quote marks.

packages/math/init.lua
```diff
@@ -64,21 +65,35 @@ function package:registerCommands ()
...
+  self:registerCommand("math:numberingstyle", function (options, _)
+    SILE.typesetter:typeset("(")
+    SILE.call("show-counter", { id=options.id })
+    SILE.typesetter:typeset(")")
+  end)
+
+
+
+
+
```
Omikhleia, line 96: nit: Lots of extra blank lines inserted!

packages/math/init.lua
```diff
@@ -344,6 +359,18 @@ Finally, here is a little secret. This notation:
...
 \noindent In other words, the notation using \code{&} and \code{\\\\} is only a syntactic sugar for a two-dimensional array constructed with braces.
 
+\paragraph{Numbered equations}
+Equations can be numbered in display mode adding the options \autodoc:parameter{numbered} and \autodoc:parameter{counter}.
+When \autodoc:parameter{numbered=true}, the equations are numbered using a default "equation" counter.
+A different counter may be set by using the option \autodoc:parameter{counter=id}, and this setting will also enable numbering. 
+The default numbering format is "(xx)", but this style may be overriden by defining a custom \autodoc:command{\math:numberingstyle} function, e.g.
+
+\begin[type=autodoc:codeblock]{raw}
+\define[command=math:numberingstyle]{(\show-counter[id="equation"])}
```
Omikhleia, line 369: I am not fond of the example using a macro, because macros can't use options, so here this example would always use the "equation" counter regardless of what the user passed to the `\math` command. I'd suggest removing the example, and just expanding the description as follows (my changes in bold):

... this style may be overrid**d**en by defining a custom `\autodoc:command{\math:numberingstyle}` **command**. **The counter `id` is passed as parameter to this hook.**

(By the way, _overridden_ takes two -d-, I think)

Omikhleia:
Looks quite good to me now - I marked most of my previous remarks as resolved and just opened a few new ones, mostly documentation-oriented.
I just clone your branch locally and will try to perform some tests soon.

packages/math/init.lua
```diff
@@ -344,6 +359,18 @@ Finally, here is a little secret. This notation:
 
 \noindent In other words, the notation using \code{&} and \code{\\\\} is only a syntactic sugar for a two-dimensional array constructed with braces.
 
+\paragraph{Numbered equations}
+Equations can be numbered in display mode adding the options \autodoc:parameter{numbered} and \autodoc:parameter{counter}.
+When \autodoc:parameter{numbered=true}, the equations are numbered using a default "equation" counter.
+A different counter may be set by using the option \autodoc:parameter{counter=id}, and this setting will also enable numbering.
```
Omikhleia, line 365: Suggestion: `\autodoc:parameter{counter=id}` --> `\autodoc:parameter{counter=<id>}`
The angle brackets are replaced with nice brackets and the content is shown in a different color when `autodoc.autodoc.highlighting` is enabled. (And italic would possibly be used, although maybe the Hack Italic font is not installed in regular builds, IIRC)

packages/math/typesetter.lua
Omikhleia: So let's craft a regression test similar to my above LaTeX tests. That test eventually should be committed in `tests/math-display-numbered.sil` and has it's `-b debug` result also committed as `tests/math-display-numbered.expected`

[math-display-numbered.zip](https://github.com/sile-typesetter/sile/files/12837360/math-display-numbered.zip)

Before any change:

image

After your changes as of c4db864 

image

**It breaks expectations with both numbered and unnumbered display math.** That tells the importance of having (non-)regressing tests for features. :smile_cat: 

What I would expect, _at least_, is:

image

Which I achieved hacking your proposal as follows:
(N.B. it could possibly be simplified, it basically re-implements or rather adapts a raggedright environment, respecting possible nesting)[^1]

```
  if mode == "display" then
    SILE.typesetter:endline()
    SILE.typesetter:pushExplicitVglue(SILE.settings:get("math.displayskip"))
    SILE.settings:temporarily(function ()
      local lskip = SILE.settings:get("document.lskip") or SILE.nodefactory.glue()
      local rskip = SILE.settings:get("document.rskip") or SILE.nodefactory.glue()
      SILE.settings:set("document.lskip", SILE.nodefactory.hfillglue(lskip.width.length))
      SILE.settings:set("document.rskip", SILE.nodefactory.glue(rskip.width.length))
      SILE.settings:set("document.parindent", SILE.nodefactory.glue())
      SILE.settings:set("current.parindent", SILE.nodefactory.glue())
      SILE.settings:set("typesetter.parfillskip", SILE.nodefactory.glue())
      SILE.settings:set("document.spaceskip", SILE.length("1spc"))

      SILE.typesetter:pushHorizontal(mbox)
      SILE.typesetter:pushExplicitGlue(SILE.nodefactory.hfillglue())

      if counter then
        SILE.call("increment-counter", { id=counter } )
        SILE.call("math:numberingstyle", { id=counter } )
      end
      SILE.typesetter:endline()
    end)
    SILE.typesetter:pushExplicitVglue(SILE.settings:get("math.displayskip"))
 ```

_But the devil is in the details_, and as you noted it probably doesn't work well in RTL...

Also, what happens if the equation is long and takes up most of the line so the fill and the number cannot fit? This would also call from a dedicated test.

[^1]: The current implementation of `\raggedright` in SILE 0.14 is subpar, and I might suggest at some point to go for the re-implementation I made in my silex.sile module. Whatever, here for the purpose of display math, it's probably safer _anyway_ not to depend from it.

Omikhleia:
Closing this PR (inactive for 7+ months) so as to move forward, see #2020 for an updated/curated version.

PR #1829, opened 2023-07-02
Updated Fedora section in README.md

hegjon:
Still three packages left, but many have been added to Fedora. Only builds for Fedora 38 and later due to `harfbuzz-devel >= 6.0.0`

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,8 +72,8 @@ A VCS package is also available as [sile-git][aur-sile-git] to build from the la
 #### Fedora
 
 A [COPR][copr] repository is available for Fedora users with packages of SILE
-and all the necessary dependencies including fonts.
-Fedora 36 and Fedora 37 are supported.
+and all the necessary dependencies.
+Fedora 38 and later are supported.
 There is work in progress to get the packages added to the official Fedora repository.
 
 ```console


alerque:
The actual minimum HarfBuzz version is 2.7.4, so it should be possible to support pretty old Fedora systems. The only catch is that one optional feature we added does require HB6, specifically variable font support. For that we need the subsetter that was added in 6. You can build SILE without variable font support using `./configure --disable-font-variations`, at which point it will only check for HarfBuzz 2.7.4.

That being said I don't know how much point their is in supporting old versions. We support some old Ubuntu LTS releases because many environments like CI runners get stuck in them, but I don't have a feel for how many folks might be stuck in old Fedora releases. Personally I'm pretty comfortable telling folks that if they want to use shinny new toys with the best new tech features they need to be on the latest stable release of their OS. Maybe that's why I like Arch Linux: only the latest packaged version of anything is supported! But at the same time I understand not everybody works in a rolling-release world and I'm not trying to force that on people with SILE—we've made some accommodation for supporting pretty old platforms.

hegjon:
Thanks for having the option to use older versions of harfbuzz. I used this to build before I myself upgraded to Fedora 38:

```
export EXTRA_FLAGS="$(pkg-config --atleast-version 6.0.0 harfbuzz || echo --disable-font-variations)"
%configure $EXTRA_FLAGS \
  --disable-static \
  --with-system-luarocks
```

Supporting older Fedora versions is not a big issue, since we are releasing every 6 month and each version is only supported for 13 months. RHEL and their derivatives is another story.

alerque:
Given that response, I'll assume leaving 37 out of the party is the plan and merge this. If that changes docs can be updated to match at any time. If there is anything I (as a non Fedora user) can do to facilitate full adoption let me know.

hegjon:
> Given that response, I'll assume leaving 37 out of the party is the plan and merge this. If that changes docs can be updated to match at any time. If there is anything I (as a non Fedora user) can do to facilitate full adoption let me know.

Yes, as for Fedora 38; getting SILE into Fedora have taken longer than I expected, I assume when it is finally added to Fedora, then most users will have upgraded to Fedora 38.

For adoption; I think it would be great to publish an article to Fedora Magazine [1] after all the packages have been added to the standard official Fedora repository.
I can help write the article, I was thinking about something like this article [1] I wrote a while ago, the article was published after I added the Electrum bitcoin wallet to Fedora.

[1] https://fedoramagazine.org/
[2] https://fedoramagazine.org/bitcoin-fedora-electrum/

alerque:
If you draft up an article somewhere I can help contribute to it some when the time comes, just ping me with info on where you are drafting.

Incidentally I started a little side project a few weeks back called polytype ([repository](https://github.com/alerque/polytype/) | [site](https://alerque.github.io/polytype/hello-world/)) that might be useful in the future for introductory posts. It isn't even SILE specific, it's just a Rosetta stone for how to do various things in assorted typesetting engines with the intention of both making it easier for people to learn one engine based on their knowledge of another as well as highlight strengths and weaknesses to help pick one in the first place. It isn't ready for prime time yet but contributions are welcome.

PR #1835, opened 2023-07-12
Make masterFilename what it says on the tin

### LINKED ISSUE #1833: SILE 0.14.10 cannot properly include a file with same base name as the master file.
SILE 0.14.10 cannot properly `\include` a file with same base name as the master file.
It used to work in 0.14.9  and earlier.

MWE with a markdown file (my own initial full case was using an XML file and a much longer SIL wrapper) follows.

Main SIL file `myfile.sil`:

```
\begin{document}
\use[module=packages.markdown]
\include[src=myfile.md]
\end{document}
```

Included markdown file `myfile.md`:

```
_Hello world_
```

Expected: The markdown file (or resp. XML etc.) is included with appropriate inputter.

Observed: The inputter is not triggered and the file is included as SIL.
That's even the case if we enforce `format=markdown` on the `\include`

Analysis:
The "detection" bypass is here -- indeed ignoring file extension.

https://github.com/sile-typesetter/sile/blob/83d142305846800bb5121b13f0a626e3f205dba3/core/sile.lua#L294-L295

Introduced by #1738

Workaround:
Well don't use files with same base names but different extensions in your project... That's confusing though!

alerque:
Closes #1833.

As far as SILE itself is concerned this shouldn't be noticed, but that's not to say documents or other project code that generates files and assumes that `masterFilename` is a basename not a file name won't be affected. I had in mind to put this in the next breaking release just to be safe, but given the current situation I'm not sure adding more hacks to work around this silly problem in the short term knowing we want to fix it properly anyway makes sense either.

This will break CaSILE but I'll fix that. I don't know if any other 3rd party extensions will be affected.

@Omikhleia does this change make sense to you?

### DIFF
--- a/classes/base.lua
+++ b/classes/base.lua
@@ -575,9 +575,9 @@ function class:finish ()
   end
   SILE.typesetter:runHooks("pageend") -- normally run by the typesetter
   self:endPage()
-    if SILE.typesetter then
-      assert(SILE.typesetter:isQueueEmpty(), "queues not empty")
-    end
+  if SILE.typesetter then
+    assert(SILE.typesetter:isQueueEmpty(), "queues not empty")
+  end
   SILE.outputter:finish()
   self:runHooks("finish")
 end
--- a/core/makedeps.lua
+++ b/core/makedeps.lua
@@ -25,7 +25,7 @@ local makeDeps = {
   write = function (self)
     self:add_modules()
     if type(self.filename) ~= "string" then
-      self.filename = pl.path.splitext(SILE.masterFilename) .. ".d"
+      self.filename = pl.path.splitext(SILE.input.filenames[1]) .. ".d"
     end
     local depfile, err = io.open(self.filename, "w")
     if not depfile then return SU.error(err) end
--- a/core/makedeps.lua
+++ b/core/makedeps.lua
@@ -25,7 +25,7 @@ local makeDeps = {
   write = function (self)
     self:add_modules()
     if type(self.filename) ~= "string" then
-      self.filename = SILE.masterFilename .. ".d"
+      self.filename = pl.path.splitext(SILE.masterFilename) .. ".d"
     end
     local depfile, err = io.open(self.filename, "w")
     if not depfile then return SU.error(err) end
--- a/core/sile.lua
+++ b/core/sile.lua
@@ -291,7 +291,7 @@ function SILE.processString (doc, format, filename, options)
   -- a specific inputter to use, use it at the exclusion of all content type
   -- detection
   local inputter
-  if filename and pl.path.normcase(pl.path.normpath(filename)) == pl.path.normcase(SILE.masterFilename) and SILE.inputter then
+  if filename and pl.path.normcase(pl.path.normpath(filename)) == pl.path.normcase(SILE.input.filenames[1]) and SILE.inputter then
     inputter = SILE.inputter
   else
     format = format or detectFormat(doc, filename)
@@ -301,7 +301,7 @@ function SILE.processString (doc, format, filename, options)
     inputter = SILE.inputters[format](options)
     -- If we did content detection *and* this is the master file, save the
     -- inputter for posterity and postambles
-    if filename and pl.path.normcase(filename) == SILE.masterFilename then
+    if filename and pl.path.normcase(filename) == pl.path.normcase(SILE.input.filenames[1]) then
       SILE.inputter = inputter
     end
   end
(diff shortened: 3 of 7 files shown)


Omikhleia:
> This will break CaSILE but I'll fix that. I don't know if any other 3rd party extensions will be affected.
> 
> @Omikhleia does this change make sense to you?

Somehow makes sense, though breaking an internal compatibility might be best in 0.15 rather than 0.14.x... (Especially for something that should occur that often)

This will break too on my side of things. Hoping this quick check caught them all:
```
find *.sile -name "*.lua" | xargs -d\\n grep masterFilename
embedders.sile/packages/embedders/init.lua:  local dir = pl.path.join(pl.path.dirname(SILE.masterFilename), "embedded")
embedders.sile/packages/embedders/init.lua:  local source = pl.path.basename(SILE.masterFilename)
labelrefs.sile/packages/labelrefs/init.lua:  local tocfile, err = io.open(SILE.masterFilename .. '.ref', "w")
labelrefs.sile/packages/labelrefs/init.lua:  local reffile,_ = io.open(SILE.masterFilename .. '.ref')
printoptions.sile/packages/printoptions/init.lua:  local dir = pl.path.join(pl.path.dirname(SILE.masterFilename), "converted")
resilient.sile/packages/resilient/styles/init.lua:  local fname = SILE.masterFilename .. '-styles.yml'
resilient.sile/packages/resilient/styles/init.lua:  local fname = SILE.masterFilename .. '-styles.yml'
resilient.sile/packages/resilient/tableofcontents/init.lua:  local tocfile, err = io.open(SILE.masterFilename .. '.toc', "w")
resilient.sile/packages/resilient/tableofcontents/init.lua:  local tocfile, _ = io.open(SILE.masterFilename .. '.toc')
```
Those using the directory where is located the master file should be safe, so I assume the list of impacts is a bit smaller:
embedders.sile/packages/embedders/init.lua:  local source = pl.path.basename(SILE.masterFilename)
labelrefs.sile/packages/labelrefs/init.lua:  local tocfile, err = io.open(SILE.masterFilename .. '.ref', "w")
labelrefs.sile/packages/labelrefs/init.lua:  local reffile,_ = io.open(SILE.masterFilename .. '.ref')
resilient.sile/packages/resilient/styles/init.lua:  local fname = SILE.masterFilename .. '-styles.yml'
resilient.sile/packages/resilient/styles/init.lua:  local fname = SILE.masterFilename .. '-styles.yml'
resilient.sile/packages/resilient/tableofcontents/init.lua:  local tocfile, err = io.open(SILE.masterFilename .. '.toc', "w")
resilient.sile/packages/resilient/tableofcontents/init.lua:  local tocfile, _ = io.open(SILE.masterFilename .. '.toc')
```
... And adding `pl.path.splitext` there should be sufficient and compatible with earlier versions decently enough (well, that is unless the master file name contains several dots, but that's really an edge case).

alerque:
What if we left `SILE.masterFilename` as a basename for now with a deprecation notice until v0.15.0, but switched SILE's own internal use from the public API to a new undocumented private version (say `SILE._masterFilename`) that is actually the filename? Would that sit better with you? That would fix our internal bug temporarily without breaking the external API for now. Then later we just swap them out and drop the legacy one.

Omikhleia:
> Then later we just swap them out and drop the legacy one.

That extra work later for 3rd party package maintainers.
Maybe we can avoid the need to swap
- SILE.masterFile*N*ame = name with extension, and we use that where needed for now on
- SILE.masterBaseName = for convenience name without extension (and also to save a few cumbersome calls here and there to `pl.path.splitext`)

Later, 0.15 or even beyond, we remove the old SILE.masterFilename without extension, so it gets nil.

In the meantime, 3rd party packages can start replacing `SILE.masterFilename` by `SILE.masterBaseName or SILE.masterFilename` and there's no urge to remove that shim until it has reached all users having updated everything (SILE, main 3rd party packages, their possible deps, etc.)

Or something similar?

alerque:
Okay I'll leave `masterFilename` alone for now so anything 3rd party using it won't trip up, but I don't like *any* of the suggestions for working around it. As such I've just gone to using `input.filenames[1]` internally for everything we need it for. Probably we can go back to providing something sensible in a ready-made variable at some point, but I want to cut down on the number of things a user has to input by hand to use SILE as a library that the CLI was just guessing at. That means less magic assumptions but also not asking for things we don't even need.

Where and how we name intermediary files needs a bigger overhaul anyway. It really should be possible to use a cache dir somewhere separate from sources OR outputs, and right now that isn't really possible.

PR #1844, opened 2023-08-17
Use true semver parser for dependency/deprecation handling"

### LINKED ISSUE #1840: SU.deprecated issue after we reached 0.14.10
SILE 0.14.10 no longer warns/errors on some deprecations:
```
SU.deprecated("\\quote", "\\pullquote", "0.14.5", "0.16.0") -- prints nothing now!
```

Due to performing a mere string comparison.

alerque:
Closes #1840


alerque:
Of course this had to highlight the fact that our internal version string is *not* actually fully semver compatible. Bah.

PR #1846, opened 2023-08-19
Cleanup display of code blocks in the manual

### LINKED ISSUE #1782: Weird line spacing in SILE manual 0.14.9
It seems we have introduced a small regression somewhere, see image below.

... Or the issue was present, but not apparent? Commenting this line, the issue "disappears":

https://github.com/sile-typesetter/sile/blob/5c9805b8bc9ccf2e1dfdf0971818664333c75165/packages/linespacing/init.lua#L97-L99

Maybe another edge-case of multiple instantiation / partial instantiation? (#1531) 

image

alerque:
Somewhere along the line after we introduced rules above and below code blocks everything got all out of whack and the rules were not evenly positioned—even occluding the lines of code in some cases. In addition to cleaning up the rule alignment this adds some subtle color treatment and a bit more leading in an attempt to make the mess less visually assailing. It could use more work but I at least think this is a step in the right direction.

Unrelated but fixed while messing around with manual visuals: fixes #1782.


alerque:
Пл зыымж зпмв *кфщл* ын лов
... er, wrong keyboard layout let me start over.

It looks like *part* of the issue with alignment going haywire might be related to #1782.

Issue #427, opened 2016-12-26
Rotate frame broken


alerque:
The rotate package is supposed to add the ability to rotate frames. The example PDF as checked into the relo looks fine, but when I compile the same thing I get no visible folio frame.

It testing in my own document it does seem to be doing something but the axis is not in the center and the units are not degrees.

alerque:
Strangely this doesn't happen on all my machines. Even one that's just a few commits behind and seems to have no changes is this department works, while one I just compiled from HEAD goes splat

In the version that _works_, it only works for the first frame. After that other rotated frames have some bizarre relative coordinate system and debug-frames shows them it a different place than the content they render.

alerque:
See Gitter chat [starting here](https://gitter.im/simoncozens/sile?at=5880859b300f220a66077fa8) for an example of what I'm (partially successfully)  with rotated frames and [ending here](https://gitter.im/simoncozens/sile?at=5880ad02300f220a66085e22) with at example of what happens when it failes.

PR #1853, opened 2023-08-24
Another approach to full bleed printing, paper/sheet size, and pdf refactoring

### LINKED ISSUE #902: Setup pdf package to work with other backends
Loading the *pdf* package inhibits being able to use the text or debug backends. For example the manual cannot be generated using `sile -d debug`, which would actually be useful. Somehow the package should be refactored so that it can just skip anything it can't do without the libtexpdf backend. Maybe throw warnings when such functions are called, but don't error on load.

### LINKED ISSUE #916: Refactor rotate package to move backend specific code into backend
> This will fix it, by analogy with other things in pdf.lua:
> 
> ```diff
> diff --git a/packages/rotate.lua b/packages/rotate.lua
> index 3c430db7..f504ecb0 100644
> --- a/packages/rotate.lua
> +++ b/packages/rotate.lua
> @@ -46,6 +46,7 @@ local outputRotatedHbox = function (self, typesetter, line)
> 
>    local horigin = (typesetter.frame.state.cursorX + origbox.width.length / 2):tonumber()
>    local vorigin = -(typesetter.frame.state.cursorY + origbox.height / 2):tonumber()
> +  SILE.outputters.libtexpdf._init()
>    pdf:gsave()
>    pdf.setmatrix(1, 0, 0, 1, horigin, vorigin)
>    pdf.setmatrix(math.cos(x), math.sin(x), -math.sin(x), math.cos(x), 0, 0)
> ```
> 
> But calling private methods on the outputter is a bit of a bad smell. Perhaps a tidier fix would be to make packages/pdf.lua more of a wrapper around low-level libtexpdf functions, and move the gsave/grestore/setmatrix stuff into there.

_Originally posted by @simoncozens in https://github.com/sile-typesetter/sile/issues/913#issuecomment-655315591_

This does smell. The other aspect of this to consider is that (per #902) the rotate package introduces a bunch of stuff that _only_ works with one backend. This makes it hard to test and kind of makes a mockery of our plans to support multiple backends.

The PDF specific functions should definitely move to a backend-specific package and call them using some sort of wrapper so that other backends could theoretically implement them as well.

### LINKED ISSUE #1285: cropmarks package is broken

### LINKED ISSUE #1439: A thought on full bleed printing, trim areas and cropmarks

Omikhleia:
Freshly backported from https://github.com/Omikhleia/silex.sile/pull/2 (you can see some new visuals there), proposed for inclusion upstream here[^2]

This is a new take at #1439, reusing some of the ideas from my earlier attempt #1470 -- There are a few "slightly" breaking changes (notably on the cropmarks package), although nothing worse that in that previous attempt.

The main differences with the previous attempt are that the frames are not impacted in this solution, and the frames coordinates are de-correlated from those used by the outputter.
- It's simpler and safer
- After all, bleed and sheet size are output concepts[^4]

Another (minor) differences is that I went for different base class options (which sound much better, I think, and less intrusive in the existing code base).

**Achievements**:
- Closes #1439 
- Moves the low-level libtexpdf calls to the outputter
  - As a bonus, thus: running a document with `-b debug` (or another outputter) does work now. :+1: 

**Other details and fixes**:
- Closes #1285 - new working implementation of the **cropmarks** package (:warning: commands and behavior changed)
- ~~Closes~~ Starts addressing #902 - **pdf** package now delegates to the outputter[^1]
  - Also refactored the **url** package so it doesn't have to check which outputter is used
- Closes #916 - **rotate** package now delegates to the outputter
- Updates the **scalebox** package to also delegate the lower-level pdf calls to the outputter
- Fixes a bug in the typesetter when `-d hboxes` is used, the debug hbox could be misplaced
- Fixes a bug in the **pdf** packages, the metadata and bookmarks do not need to be wrapped in a hbox (avoiding the box prevents side-effects with paragraph indents and spread / odd page opening logic...)

**Other noteworthy comments**:
- #427 is not addressed here, but I annotated the code with comments[^3]

[^1]: Except for the seldom used `\pdf:literal` command, I have annotated the rationale in the code.
[^2]: Is it a _reasonable request_ to ask for it to be considered for **0.15**? Note by the way the deprecation warning in the **cropmarks** package currently assuming it. I'd be happy if 0.15 could also bring some cool typography features (such as this one and the automated italic correction).
[^3]: I have some understanding _why_ the issue occurs, but this is far beyond my current reach to address it.
[^4]: In PDF, we may eventually want to consider filling the CropBox, TrimBox, BleedBox etc. properties (but the current libtexpdf wrappers only sets the MediaBox). In other outputters, the concept might not be meaningful and be ignored. In debug outputter, the frame coordinates remain unchanged, relative to the page, which is best IMHO.


Omikhleia:
Erm:
> Fixes a bug in the pdf packages, the metadata and bookmarks do not need to be wrapped in a hbox (avoiding the box prevents side-effects with paragraph indents and spread / odd page opening logic...)

Perhaps a bad move, esp. for bookmarks. I'll revert that bit.

Omikhleia:
Hey @jodros the conflicts here are due to the introduction of the landscape option. But with two or three page size related options in this proposal, I'm unsure how to interpret `landscape`, now... Your opinion on how to solve it would be welcome.

(Let's admit it, the most important option for me is `bleed` -- because when going to use POD services such as Lulu.com, a 0.25in bleed is ideally requested.)

Omikhleia:
BTW, I did mention #1892 that it would break here. I asked there for a discussion that never came. Thats's ok, but how do we move forward? I've no idea :)

jodros:
> Hey @jodros the conflicts here are due to the introduction of the landscape option. But with two or three page size related options in this proposal, I'm unsure how to interpret `landscape`, now... Your opinion on how to solve it would be welcome.

Hello, I took a brief look at your code, mostly the base class, so I'd like to know whether there are others conflicting parts besides the `class:setOptions`?

Omikhleia:
> Hello, I took a brief look at your code, mostly the base class, so I'd like to know whether there are others conflicting parts besides the `class:setOptions`?

Well besides the coding, the question is that the PR proposes `sheetsize` and `bleed` in addition to `papersize` -- the sheet size defaults to the paper size plus possibly the bleed, and can be set to a specified size again augmented by the bleed if needed. For instance,
- `papersize=6in x 9in, bleed=0.25in` is what I'd use for a US trade book sent to Lulu POD
- `papersize=6in x 9in, papersize=a4` is what I'd use (likely with cropmarks) for a pre-print on my home printer.

With a single `landscape=true` option now, I wonder how it should be interpreted. Would it apply to the paper size, or both the paper size and sheet size? Consider for instance `papersize=a6, landscape=true, sheetsize=a4`... There would be no need for the sheet size to be in landscape.

jodros:
Well, I'm not sure whether I understood the problem, but here is what I tested right now:

> Consider for instance `papersize=a6, landscape=true, sheetsize=a4`... There would be no need for the sheet size to be in landscape.

In your example indeed there would be no need for that, as we can see: 

изображение

But what if the user just want to have  this document with  an `a5` paper size as well? We'd have the following:

изображение

But if we add the landscape option to sheet size

`SILE.documentState.sheetSize = SILE.papersize(size, self.options.landscape)` 

We'd have:

изображение

Omikhleia:
> Well, I'm not sure whether I understood the problem, but here is what I tested right now

Can you possibly post your diff? I could try and check.

jodros:
> > Well, I'm not sure whether I understood the problem, but here is what I tested right now
> 
> Can you possibly post your diff? I could try and check.

```diff
diff --git a/classes/base.lua b/classes/base.lua
index d73b5ac6..d013a579 100644
--- a/classes/base.lua
+++ b/classes/base.lua
@@ -85,9 +85,8 @@ function class:setOptions (options)
   options.landscape = nil
   self.options.papersize = options.papersize or "a4"
   options.papersize = nil
-  options.papersize = options.papersize or "a4"
   options.bleed = options.bleed or "0"
-  
+
   for option, value in pairs(options) do
     self.options[option] = value
   end
@@ -152,7 +151,8 @@ function class:declareOptions ()
   self:declareOption("sheetsize", function (_, size)
     if size then
       self.sheetsize = size
-      SILE.documentState.sheetSize = SILE.papersize(size)
+      -- SILE.documentState.sheetSize = SILE.papersize(size)
+      SILE.documentState.sheetSize = SILE.papersize(size, self.options.landscape)
     end
     return self.sheetsize
   end)

```

packages/cropmarks/init.lua
```diff
@@ -63,52 +57,36 @@ end
 
 function package:registerCommands ()
 
-  self:registerCommand("crop:header", function (_, _)
-    local info = SILE.input.filenames[1] .. " - " .. self.class:date("%x %X") .. " -  " .. outcounter
+  self:registerCommand("cropmarks:header", function (_, _)
+    local info = SILE.masterFilename
```
Omikhleia, line 61: `SILE.masterFilename` > `SILE.input.filenames[1]`

alerque:
I just took a stab at ⓐ resolving the merge conflicts to integrate the landscape option and ⓑ cleaning up a few tidbits I noticed in review. How are we looking?

alerque:
This does not actually close #902, but it does *start* down that road. I'm working on finishing that up in another branch.

Omikhleia:
> This does not actually close #902, but it does _start_ down that road. I'm working on finishing that up in another branch.

It should have, i had rotate working in my "draft" HTML outputter with no problem. If I can help, let me know, perhaps indeed I didn't address the debug backend?

alerque (approved):
This all looks like a good start. There is some more refactoring I want to do to add hooks to the outputters so packages can do more, but also move more backend specific logic out of packages. But in messing with it I think it makes sense to just start with this and move on rather than trying to refactor these commits.

Issue #1870, opened 2023-09-12
Make sure Rust Lua understands system Lua paths


alerque:
Even with shipping our own Lua interpreter we'll want to make sure it plays nice with distro Lua package search paths, with and without `--with-system-luarocks`.

PR #1871, opened 2023-09-12
fix(tooling): Fix Nix buld for rustier sile

### LINKED ISSUE #1865: Fix broken Nix Flake on develop branch
Per discussion [starting in this review](https://github.com/sile-typesetter/sile/pull/1762#pullrequestreview-1446399382) the Nix Flake hasn't gotten patched up for the Rust based build in the develop branch yet.

cc @doronbehar I'm a little over my head. Do we need crane here?

doronbehar:
Fixes #1865

Building sort of works now, only running `sile` fails with:

```
Error: runtime error: cannot open ./core/pathsetup.lua: No such file or directory
stack traceback:
        [C]: in ?
        [C]: in function 'dofile'
        [string "src/lib.rs:25:10"]:1: in main chunk
```


pkg.nix
```diff
@@ -0,0 +1,173 @@
...
+, lua
+, libiconv
+, darwin
+, makeFontsConf
+, gentium
+, runCommand
+}:
+
+let
+  luaEnv = lua.withPackages(ps: with ps; [
```
doronbehar, line 29: I always wondered, what is the lua version you feel most comfortable with? Nixpkgs' `lua` is `lua5.2`.

pkg.nix
```diff
@@ -0,0 +1,173 @@
...
+    icu
+    fontconfig
+    libiconv
+  ] ++ lib.optionals stdenv.isDarwin [
+    darwin.apple_sdk.frameworks.AppKit
+  ];
+
+  configureFlags = [
+    "--with-system-luarocks"
+    # TODO: Explain this flag
```
doronbehar, line 102: Note this.

pkg.nix
```diff
@@ -0,0 +1,173 @@
...
+    "PDFINFO=false"
+    #"--with-manual" In Nixpkgs we add this flag, here its not important enough
```
doronbehar, line 104: BTW we can create a few variants of the build using some Nix conditionals, and add varied packages to `flake.nix`'s `packages` attribute set.

alerque:
Thanks!

The runtime issue is #1870. It works run from the source directory but paths aren't worked out to run from the system—which is something I just discovered trying to spin up packaging for the develop branch. To date I'd been testing only from my devel environment. Working on that now.

pkg.nix
```diff
@@ -0,0 +1,173 @@
...
+  ];
+
+  postPatch = ''
+    patchShebangs build-aux/*.sh
+  '' + lib.optionalString stdenv.isDarwin ''
+    sed -i -e 's|@import AppKit;|#import <AppKit/AppKit.h>|' src/macfonts.m
```
doronbehar, line 110: Not necessarily related to this change, but I consider such `sed` hooks rather ugly, and usually they are a result of upstream developers who don't make efforts for their distributors. This is not the case with sile, as development here is very much involved with distribution, so I was wondering whether in the `develop` branch you would like to make the effort to fix this issue, or at least explain better why is it needed only for Nix builds. Note it is rarely needed in Nixpkgs from my experience.

pkg.nix
```diff
@@ -0,0 +1,173 @@
...
+  '';
+
+  NIX_LDFLAGS = lib.optionalString stdenv.isDarwin "-framework AppKit";
```
doronbehar, line 113: Same here, why is this needed? Can it be fixed via changes to `Makefile.am` or other files?

pkg.nix
```diff
@@ -0,0 +1,173 @@
...
+    fontDirectories = [
+      gentium
+    ];
+  };
+
+  enableParallelBuilding = true;
+
+  preBuild = lib.optionalString stdenv.cc.isClang ''
+    substituteInPlace libtexpdf/dpxutil.c \
+      --replace "ASSERT(ht && ht->table && iter);" "ASSERT(ht && iter);"
```
doronbehar, line 125: This as well can be fixed IMO and shouldn't be hacked like that only in Nix' build.

pkg.nix
```diff
@@ -0,0 +1,173 @@
...
+  '';
+
+  # remove forbidden references to $TMPDIR
+  preFixup = lib.optionalString stdenv.isLinux ''
```
doronbehar, line 129: This one on the other hand a rather common thing you see in Nixpkgs, but we should verify it is still needed.

doronbehar:
I will also note I get a lot of failure CI emails.

pkg.nix
alerque: SILE is 100% compatible with all versions of Lua *newer* than 5.0 (currently 5.1–5.4). Up through v0.14.6, we were recommending the distro default `lua` as the path of least resistance. For Ubuntu LTS and some BSDs that has been 5.2, for almost everybody else that is 5.4. Starting with [v0.14.7](https://github.com/sile-typesetter/sile/releases/tag/v0.14.7) we also have 100% feature parity on LuaJIT. Previously we supported it but debug features were limited; now it is no sweat off anybody's back to use. Since LuaJIT is handily 2× faster we now recommend it everywhere that happens to have distro support for it. At some point it will be the build time default, for now `./configure --with-luajit` is needed to switch, including relevant to this PR making that the VM we build into the Rust binary.

pkg.nix
alerque: That would be cool, not so much for the manual or anything but for the Lua version. While we don't much care upstream and recommend LuaJIT for the speed, no matter what we default to it is almost inevitable that some people will want to use something else in order to play nice with some 3rd party Lua package they integrate with.

pkg.nix
alerque: Sure, I'll look into it. I do know the only other please we frequently build for macOS is on Homebrew which works properly with the current setup, hence my being reluctant to hack on it.

pkg.nix
doronbehar: I added 3 more nix packages (see `nix search .`) of builds of sile with different lua variants.

doronbehar: Currently Nixpkgs' default `lua` which is 5.2, is used for the default build - by default I mean `defaultPackage = sile;`, and this is the build you get when you (will) run `nix run github:sile-typesetter/sile`.

pkg.nix
doronbehar: Quoting myself:

> This as well can be fixed IMO and shouldn't be hacked like that only in Nix' build.

BTW by changing `stdenv` to `clangStdenv`, you can pretty much test building `sile` with clang.

pkg.nix
doronbehar: Commit [feat(tooling): Add nix builds with various lua versions](https://github.com/sile-typesetter/sile/pull/1871/commits/c28cd9410d817d4a8c6e432076e2f8b5cdae80a7) resolves this conversation.

pkg.nix
alerque: https://github.com/sile-typesetter/sile/issues/1879

pkg.nix
alerque: https://github.com/sile-typesetter/sile/issues/1879

pkg.nix
alerque: Wow, that could actually be useful to run in CI. I'll make a note of that.

alerque: https://github.com/sile-typesetter/sile/issues/1879
and
https://github.com/sile-typesetter/sile/issues/1880

pkg.nix
```diff
@@ -99,15 +98,20 @@ in stdenv.mkDerivation (finalAttrs: {
 
   configureFlags = [
     "--with-system-luarocks"
-    # TODO: Explain this flag
+    # The automake check target uses `pdfinfo` to confirm the output of a test
+    # run. This flake build eschews that test because it is run from the source
+    # directory but the binary is already built with system paths, so it can't
+    # be checked under Nix until after install. By specifying some binary here
+    # we skip the configure time test for `pdfinfo`, by using `false` we make
+    # sure that check fails if run (which we don't).
```
doronbehar, line 106: But in Nixpkgs we do run this check? If we do, then perhaps that would be worth mentioning as another difference between the two expressions.

alerque: The derivation does run an equivalent check, but not straight out of the Makefile (which isn't available after installation when we actually want to test it). Instead we have a copy of the shell code from the target that we run in `passthru.tests.test` and run after everything is in the expected location. In the shell code there we have hard coded `pdfinfo` instead of the upstream `$(PDFINFO)` set by autoconf.

alerque: I added that nugget of info to the comment and also a few other cleanups. It seems to work now (with the exception of Lua 5.4 where `lua-zlib` is marked as broken. The LuaJIT, Lua5_2, and Lua5_3 options all run correctly.

alerque (approved):

alerque:
Just in case anybody is following along and has no idea what is happening here, when this merges you'll be able to run the development version of SILE complete with Rusty CLI just by installing Nix on your system then running:

```console
$ nix run github:sile-typesetter/sile/develop
```

LuaJIT is now the default, but you can run anything you want, e.g.:

```console
$ nix run github:sile-typesetter/sile/develop#sile-lua5_3
```

Add `--` followed by any arguments you want to pass to SILE.

alerque:
Many many thanks @doronbehar.

pkg.nix
alerque: How would I confirm if this is needed? Would the build fail with a warning about it if there was a problem? Because just dropping this block entirely seems to allow the Flake to continue running just fine. See #1941.

PR #1883, opened 2023-10-07
feat(tooling): Add nix packages building sile with clang

doronbehar:
Here's the idea I was mentioning to you regarding that `stdenv.cc.isClang` conditional `preBuild`. I haven't tested the build yet of `sile-clang`, but this is the idea.


build-aux/pkg.nix
```diff
@@ -139,6 +139,7 @@ in stdenv.mkDerivation (finalAttrs: {
     touch source/build-aux/rust_boilerplate.mk
   '';
 
+  # TODO: Remove this, and test whether it is needed according to nix package sile-clang
```
alerque, line 142: I tried the clang build (which worked out of the gate) both with Lua 5.2 and LuaJIT and both seemed to work both with and without this `preBuild` patch. Do you have any background on why it was added in the first place?

doronbehar: Someone else wrote the commit, but it all happened in https://github.com/NixOS/nixpkgs/pull/77738 according to `git blame` on Nixpkgs.

flake.nix
```diff
@@ -96,6 +98,11 @@
         sile-luajit = sile.override {
           lua = pkgs.luajit;
         };
+        sile-clang = sile.override {
+          lua = pkgs.luajit;
+          # Use the same clang version as Nixpkgs' rust clang stdenv
```
doronbehar, line 103: Excellent!

PR #1892, opened 2023-10-13
Landscape option added to base class

jodros:
I missed an option for a landscape mode so here it is.


classes/base.lua
```diff
@@ -139,16 +162,16 @@ function class.declareSettings (_)
   })
 end
 
-function class:loadPackage (packname, options)
+function class:loadPackage(packname, options)
```
Omikhleia, line 165: Greetings @jodros 
This PR touches several lines unrelated to your change (adding or removing spaces, changing indentation....) -- I think this should be avoided (or done separately when wanted, but I don't think it's the case here), notably;
- For `git blame` purpose when investigating issue
- For other PRs to be minimally impacted (e.g. my own #1853 touches upon the same areas) and avoid useless conflicts

core/papersize.lua
```diff
@@ -63,21 +63,26 @@ local papersize = {
...
-        size = string.lower(size:gsub("[-%s]+", ""))
-        if self[size] then
+  __call = function(self, size, orientation)
+    local _, _, x, y = string.find(size, "(.+)%s+x%s+(.+)")
+    if x and y then
+      return { SILE.measurement(x):tonumber(), SILE.measurement(y):tonumber() }
+    else
+      size = string.lower(size:gsub("[-%s]+", ""))
+      if self[size] then
+        if orientation == "landscape" then
```
Omikhleia, line 73: Could be good to check for "portrait" and error if orientation is not valid.

tests/landscape.sil
```diff
@@ -0,0 +1,8 @@
+\document[orientation=landscape,papersize=a5]{
+%\use[module=packages.frametricks]
```
Omikhleia, line 2: Commented commands not needed

tests/landscape.sil
```diff
@@ -0,0 +1,8 @@
...
+%\showframe[id=all]
+
+\use[module=packages.lorem]
+\lorem[words=1000]
+
+}
```
Omikhleia, line 8: Why not `\begin ... \end{document}` (more usual way to way to wrap a document!)

By the way, for tests, when sure it does the right thing, you need to run `sile tests/landscape.sil -b debug` to generate a `.debug` text file, rename it as `.expected` and commit it too. The test checks (and the CI) use this for regression testing.

Omikhleia:
Hello, that's something I considered too.

If it get accepted, it would need to be documented in the manual too. Not that we already have a PR on similar areas (#1853), so things will have to be done orderly.

Two small remarks in passing
- In general, I believe it's best, when possible, to first open an enhancement issue before a PR (to get discussions before heading toward an implementation)
- I see you merged master in your branch. Personally I would tend to recommend _rebasing_ rather _merging_ branches. It makes it easier when we change the target branch (e.g. to `develop` or other)

core/papersize.lua
Omikhleia: (Or possibly change it to a boolean `landscape=true/false`, are there any other orientation than landscape/portrait?)

classes/base.lua
Omikhleia: I also think the commit log is longer than expected, mentioning line numbers etc. - not so future-proof.

jodros:
Hello @Omikhleia , thanks for your reviewing and I appreciate your tips.

>  This PR touches several lines unrelated to your change (adding or removing spaces, changing indentation....) -- I think this should be avoided (or done separately when wanted, but I don't think it's the case here), notably;

I have a habit to always use  automatic formatting, so some of these indentation changes were made so. I'll pay attention to this detail next times.

> In general, I believe it's best, when possible, to first open an enhancement issue before a PR (to get discussions before heading toward an implementation)

Right,  I'll remember that.

jodros:
Hello @Omikhleia , thanks for your reviewing and I appreciate your tips.

>  This PR touches several lines unrelated to your change (adding or removing spaces, changing indentation....) -- I think this should be avoided (or done separately when wanted, but I don't think it's the case here), notably;

I have a habit to always use  automatic formatting, so some of these indentation changes were made so. I'll pay attention to this detail next times.

> In general, I believe it's best, when possible, to first open an enhancement issue before a PR (to get discussions before heading toward an implementation)

Right,  I'll remember that.

classes/base.lua
Omikhleia: Still applies see ex. below

classes/base.lua
```diff
@@ -139,16 +157,16 @@ function class.declareSettings (_)
...
   local pack = require(("packages.%s"):format(packname))
   if type(pack) == "table" and pack.type == "package" then -- new package
     self.packages[pack._name] = pack(options)
-  else -- legacy package
+  else                                                     -- legacy package
```
Omikhleia, line 164: Ex. of change introduced in untouched code.

classes/base.lua
```diff
@@ -77,21 +84,28 @@ function class:_post_init ()
   SILE.scratch.half_initialized_class = nil
 end
 
-function class:setOptions (options)
+function class:setOptions(options)
   options = options or {}
+  options.orientation = options.orientation or "portrait"
+
+  if not self.options.orientation then
+    self.options.orientation = options.orientation
```
Omikhleia, line 85: Should not be necessary (you added it to `options` on line 89, and lines 97-99 therefore propagate it to the class instance already, methink.

classes/base.lua
```diff
@@ -341,22 +361,23 @@ function class:registerCommands ()
...
+      elseif options.src then
+        return SILE.processFile(options.src, "lua", packopts)
+      elseif options.require then
+        local module = SU.required(options, "require", "lua")
+        return require(module)
+      else
+        SU.error("\\lua function requires inline content or a src file path or a require module name")
+      end
+    end,
+    "Run Lua code. The code may be supplied either inline, using require=... for a Lua module, or using src=... for a file path")
```
Omikhleia, line 373: Another big example of untouched code reformatted without relation to this PR, etc.

tests/landscape.sil
Omikhleia: Ok.
N.B. Good practice = Let the reviewer resolve the conversation after they have a chance to check.

Omikhleia:
This will conflict with #1853 which introduces a `sheetsize` in addition to the `papersize` -- and in the grand scheme of things, possibly, both could be distinct with respect to using landscape or not. We'd need to discuss it.

tests/landscape.sil
alerque: For updating expectations, you can just run `make tests/landscape.expected` once you are getting the results you actually do expect.

classes/base.lua
jodros: As I explained in the first commit, while testing this feature, sometimes it suddenly stopped to work and that is the only way I got to solve the problem.

> After some time I found out why: since there is no order in a loop using
> pairs(), sometimes the loop in class:setOptions() was getting orientation
> as the last element, so when SILE.papersize was called, the value of
> options.orientation was indeed nil, so the solution should be just
> set the orientation value to class.options before papersize.

But indeed, it looks a bit redundant and could be more straight...

alerque (changes requested):
As already noted, the unrelated style changes need to be removed from this PR. I can barely make out what is going on, much less give it a good review. I'm actually quite happy to accept code style fixes, but only on their own not as part of other changes. Also many of these changes go *against* our usual coding styles and won't fly anyway.

Adding a landscape option is fine but we do need it to be clean.

alerque:
Also have we thought about whether an enum option like this (two named options) is best or whether an orientation option that could allow more rotations (e.g. 4) would be more robust?

classes/base.lua
alerque: On the commits & merge vs rebase issue, for a small change like this I don't care too much since we can squash the whole thing to one `feat:` commit, but yes for larger change sets with any sort of incremental work done or large enough to merit separate testing and documentation commits having cleaner history helps PR review.

core/papersize.lua
alerque: What about 180/270° rotations? I could see those being more likely to be used per-page than per-document, but it would be nice to know if there is any precedent out there for any other orientations. If not we either need to validate the two enum options or make it a boolean as mentioned.

classes/base.lua
alerque: That suggests there is some sort of race condition here, probably the order in which options are handled. Papersize and orientation might be handled in the opposite order (table `pairs()` iteration is non-deterministic).

core/papersize.lua
jodros: These other rotations are interesting but given its use being per-page, they could be implemented as commands instead of being passed as options to the class.

Omikhleia: > What about 180/270° rotations?

That's possibly something different. 
-There are the paper dimensions (e.g. one does not really need rotating anything for making slides with SILE, but just inverted width x height).
- Rotating page/frames (possibly per-page) is something perhaps different, whether done in a similar way as https://github.com/PietroCarrara/sine did for their booklets, or in other ways (frame specification, outputter logic, etc.)

alerque: I think this is correct, other rotations are a kind of different thing.

Here is my current thought. We can go ahead and implement a landscape/portrait toggle. I think this should probably be a boolean where papersize is the portrait orientation and `landscape=true` just flip-flops the X and Y dimensions. Nothing else would change about direction/etc. If in the future we have a more generic rotation/orientation thing we can figure out what the interplay is, and probably the handling order will be papersize, adjust papersize based on landscape toggle, then rotate or otherwise orientate things.

classes/base.lua
```diff
@@ -101,6 +102,15 @@ function class:declareOptions ()
     end
     return self._name
   end)
+  self:declareOption("orientation", function(_, orientation)
```
alerque, line 105: Given other discussions and the suggestion that there are no other possible orientations of interest relative to the way this feature interacts with the papersize, I think we should make this option take a boolean value, `landscape=true` rather than `orientation=landscape`.

core/papersize.lua
```diff
@@ -63,21 +63,26 @@ local papersize = {
...
+    end
+    SU.error(string.format([[Unable to parse papersize '%s'.
   Custom sizes may be entered with 'papersize=<measurement> x <measurement>'.
   Predefined paper sizes include: %s]],
-  size, table.concat(pl.tablex.keys(papersize), ", ")))
-    end
-  })
+      size, table.concat(pl.tablex.keys(papersize), ", ")))
+  end
+})
```
alerque, line 86: More irrelevant formatting changes?

tests/landscape.sil
```diff
@@ -0,0 +1,4 @@
+\begin[orientation=landscape,papersize=a5]{document}
+\use[module=packages.lorem]
+\lorem[words=1000]
```
alerque, line 3: Lets not use lorem for this if we don't need to since we don't need a lot of content to prove that the papersize got changed, the first lie of debug output is enough. Everything else is noise in the expectation file that can change and make huge diffs to review when unrelated factors change. I think the MWE for this test can be a completely empty document, the debug output shows the canvas measurements and they reflect the landscape orientation.

alerque (changes requested):

core/papersize.lua
jodros: Sorry, I didn't know how relevant this kind of change is, I'll pay attention to this detail in my next PRs.

alerque: The change itself may be fine, like I said I'm fine with style improvements. They just need to be in isolated commits (and PRs) that can be checked by using a diff with ignore-whitespace enabled so they are easy to check—not mixed in with actual code refactoring, fixing, or features.

core/papersize.lua
```diff
@@ -63,13 +63,16 @@ local papersize = {
...
 setmetatable(papersize, {
-    __call = function (self, size)
+    __call = function (self, size, landscape)
       local _, _, x, y = string.find(size, "(.+)%s+x%s+(.+)")
       if x and y then
         return { SILE.measurement(x):tonumber(), SILE.measurement(y):tonumber() }
       else
         size = string.lower(size:gsub("[-%s]+", ""))
         if self[size] then
+          if SU.boolean(landscape) then
```
alerque, line 73:
```suggestion
          if landscape then
```

Lets get this cast to a boolean when we *set* the class option, thereafter just assume it is one.

classes/base.lua
```diff
@@ -79,6 +79,7 @@ end
 
 function class:setOptions (options)
   options = options or {}
+  self.options.landscape = options.landscape or false
```
alerque, line 82:
```suggestion
  self.options.landscape = SU.boolean(options.landscape, false)
```

Lets use our utility function for this. This instance isn't a huge deal but it's easier to reason about when we do it the same way each time.

classes/base.lua
```diff
@@ -101,10 +102,17 @@ function class:declareOptions ()
     end
     return self._name
   end)
+  self:declareOption("landscape", function(_, landscape)
+    if landscape then
+      self.landscape = landscape
+    end
+    SILE.documentState.landscape = self.landscape
```
alerque, line 109:
```suggestion
    if landscape then
      self.landscape = SU.boolean(landscape, false)
      SILE.documentState.landscape = self.landscape
    end
```

Lets again use our boolean helper here. More importantly this option handler is both a setter and a getter, so the "set" action being done touching the document state should only happen in the if condition that runs when this is used as a setter, i.e. when called with a value.

alerque (changes requested):
Thanks for working with us to get this cleaned up. It's a simple option but it's worth implementing right so it is robust for future users.

Just a few nitpicks at this point mostly about boolean handling but there is one other logic issue I see.

classes/base.lua
jodros: Right, all these suggestions were committed.

alerque (approved):

alerque:
@Omikhleia I'm pretty sure we can make a small adjustment and make this play nice with your PR for full bleed printing, but thinking about that does bring up a concern here. Are we going to have a problem with race conditions and the order (or Lua's non-deterministic order) of handling options?

Omikhleia:
> @Omikhleia I'm pretty sure we can make a small adjustment and make this play nice with your PR for full bleed printing, but thinking about that does bring up a concern here. Are we going to have a problem with race conditions and the order (or Lua's non-deterministic order) of handling options?

My remark https://github.com/sile-typesetter/sile/pull/1892#discussion_r1358765078 got closed, and the current implementation uses some workaround do avoid it. It does have code smell (or adds up to the existing one, regarding how class options are implemented). What else can I say?

classes/base.lua
alerque: @jodros Please don't mark review comments as resolved if you didn't start them. Leave it to the people that started the thread or a maintainer to determine whether they are resolved.

alerque:
I just rewrote this and even changed the way the existing papersize class option is handled. This should avoid the race condition, but also reduces the code smell rather than adding too it. Reduces not eliminates. Again we all know the crop marks package here needs replacing entirely. I did fix the landscape option so it can be set on custom papersizes not just presets, added a unit test for it, etc.

@jodros You might look over the contents of my commits for future reference. Some of the changes (like not handling the papersize option twice!) are fixes to bugs that existed before your PR, but others are fixes for things you didn't handle or not setting random variables you don't use, etc. No worries, just some pointers for the next go-round...

PR #1895, opened 2023-10-20
Add configure time option to embed all resources in a single binary

### LINKED ISSUE #1868: Look into embeding Lua resources in a single binary
The Rust based CLI is a nice step, but a really useful step farther might be to optionally have a build mode that embeds all the default Lua resources into the binary (with and without system LuaRocks of course).

alerque:
Closes #1868

This collects all the resource files we usually install into `$datadir` and adds a configure option `--enable-embeded-resources` that builds them all into the finished binary. Then it injects a special loader function into the Lua VM we already have embeded that allows Lua's `require()` function to find modules that are embeded. Of course it does still look at a collection of file system paths first so users can override anything as usual.

One caveat of the override is that system paths *override* anything found embeded. The embeded loader only runs if all the other package loaders have failed. One can (and probably would) combine this new option with `--without-system-luarocks` to also embed all the Lua dependencies at build time, but any versions found on the system would be used in preference to the ones that are embedded which is a bit weird for our dependencies but makes perfect sense for adding addons. I haven't figured out what makes sense to do about that.

- [x] Not yet covered is fixing the FTL file loader so it can find the embeded resources as well.
- [ ] Also there is one SVG file in the packages directory. We're embedding it but I haven't looked into how it is loaded.

This should go a long way towards our long standing problem on Windows: we should be able to just cross compile a single Rust binary with everything embeded and call it a day. It should also mean we have a way to bring Docker images down from a few hundred MB to tens or even single digits. Distro packages may still want to use the normal method. Currently the new embedding is off by default.

I haven't done performance testing on the difference yet.


alerque:
We could provide a loader for SVGs and anything else, but in the case of that one file the only place it is used is in the package docs built into the manual, so I don't think handling that is a prerequisite to merging this.

FTL embedded resources work now.

alerque:
Disabled by default, but now works both ways. Add `--enable-embeded-resources` to get a 16MB binary with all SILE's stuff (Lua sources, C modules, FTL resources, Lua Rocks, etc.) embedded vs. a 3MB binary that expects to find said-same resources in all the same places the `sile-lua` script did/does. There doesn't seem to be any significant performance trade-offs either way, just system/distribution preferences.

Issue #475, opened 2017-06-29
How to center vertically?


kstenerud:
Is there a way to center a block of text vertically on the page in sile?

simoncozens:
Hi. I'm really sorry I'm behind on responding to issues. Thankfully you got me at a good time and this is an easy one.

Long answer: Yes, this is possible. Here's how I would think about this. You want to put some infinitely stretchy vertical glue at the start of your block of text, then your text, then another infinitely stretchy vertical glue at the bottom. Then the block of text will be balanced perfectly between two large spaces.

Now you need to know that SILE always puts infinitely stretchy glue at the bottom of a page anyway, so that material on pages which aren't full don't get stretched. So we don't need to add that ourselves. You also need to know that when SILE outputs a page, it normally discards any initial glue from the page so that every bit of material starts at the same place at the top. So we need to put some dummy material in there before the first piece of glue to stop SILE from stealing it. The usual bit of dummy material is an empty `\hbox`. So, empty `\hbox`, then a `\vfill`, then your text, then `\eject`.

Short answer:
```TeX
\hbox
\vfill
{ Stuff here will be centered on the page }
% We don't need a \vfill here because SILE provides one.
\eject
```

kstenerud:
Perfect! Thanks!

PR #1898, opened 2023-10-21
Command \center now has an option to center its content vertically

jodros:
I found this command was lacking an option to do the centering vertically (like asked here #475).


jodros:
I'm unsure whether I should write the following after the settings:temporarily block

```lua
if SU.boolean(options.vertical) and then
      SILE.call("eject")
end
```

classes/plain.lua
```diff
@@ -310,10 +310,14 @@ function class:registerCommands ()
...
   end)
 
-  self:registerCommand("center", function (_, content)
+  self:registerCommand("center", function (options, content)
     if #SILE.typesetter.state.nodes ~= 0 then
       SU.warn("\\center environment started after other nodes in a paragraph, may not center as expected")
     end
+    if SU.boolean(options.vertical) then
+      SILE.call("hbox")
+      SILE.call("vfill")
```
Omikhleia, line 319: A vfill won't push the content middle of the page, _unless_ if it is a final page automatically vfill'ed (such as the final page of your example) or explicitly `\eject`'ed. Then another vfill is currently automatically added at the end of the page, resulting in centering.

Personally:
- I don't think that horizontal centering and vertical centering should be the same command
- I don' think the vfill added on final pages or via eject is actually good :)

alerque (changes requested):
I think @Omikhleia is right here, writing direction (linewise) centering and advance direction (framewise) are very different operations and don't belong in the same command. In fact any sort of advance direction centering (absolute to page?, relative to frame), fill available remaining space?) is so complicated and the interactions with line spacing, start-of-frame and end-of-frame queues, and so forth is so complicated that this should go in a dedicated package and even if all possible modes are not implemented at first, some thought should be given to the command names(s) so that the rest of the scenarios can be implemented without stepping all over existing documents.

I don think that some centering handlers are sufficiently basic enough and used as building blocks for other commands that this is something the core distribution should provide, so adding a package in this repo is probably warranted.

Omikhleia:

Issue #420, opened 2016-12-20
Floating frames eat insertions for lunch


alerque:
According to by experiments with drop caps (see #394) the current best way to handle them is using the `\float` command to break up the frames. It's touchy and requires some care to come out right, but it can work.

That is until I introduced the concept of footnotes. An that point my pages started looking like they took a trip through a meat grinder. Here's a sample.

![screenshot from 2016-12-20 10-49-13](https://cloud.githubusercontent.com/assets/173595/21342261/b9dbf400-c699-11e6-8fda-ca068c9a8b27.png)

If I eliminate the drop-cap, the footnote frame behaves just fine:

![screenshot from 2016-12-20 10-52-12](https://cloud.githubusercontent.com/assets/173595/21342332/17b264d8-c69a-11e6-9a76-5dcbaaba3fb6.png)

This is using a pretty simple version of outputting a drop cap. A flag `SILE.scratch.startOfChapterContent` is getting set to `true` at the very end of the chapter header code, then the drop cap is inserted like this:

```Lua
SILE.scratch.startOfChapterContent = false

local push = SILE.typesetter.pushHorizontal
SILE.typesetter.pushHorizontal = function (self, node)
  if SILE.scratch.startOfChapterContent then
    SILE.call("output-chaptercap")
  end
  return push(self, node)
end

SILE.registerCommand("output-chaptercap", function (options, content)
  SILE.scratch.startOfChapterContent = false
  SILE.call("float", { bottomboundary = "1.3ex", rightboundary = "2spc" }, function ()
    SILE.call("font", { family = "Symbola", size = "3em" } , { "✛" })
  end)
end)
```
Obviously something isn't wired up right here. Whatever happens to frames when introducing a float doesn't get connected up to the insertion frame that can steal space.

alerque:
I added an MWE for this is #421.

alerque:
It looks like the `\float` command in renaming the frames by appending _ to the current id. This breaks the insertion things which knows which frame to steal from by name. Either the naming needs to happen the other direction (rename the original and keep the new name going forward) or there needs to be some sort of callback that specifies the insertion handling that could be moved to the new frame object.

alerque:
I made several attempts to fix this. The most promising attempt was to rename the old frame and keep the id with the new frame when creating a break. However this didn't actually work.

[...]

Issue #248, opened 2015-12-09
Ability to move from two columns to one column and back again


simoncozens:
This is hard, but it's really key. No release until this is possible.

simoncozens:
Problems are:

* What happens with footnotes? (Maybe we just punt and say your footnote frame must be separate.)
* Moving from one to two works fine - see below code - but how to go back again, balancing the columns along the way?

```TeX
\begin[papersize=a5]{document}
\script[src=packages/frametricks]
\script[src=packages/lorem]
\lorem[words=50]

\breakframevertical
\makecolumns[columns=3]
\raggedright{\lorem[words=400]}
\showframe[id=all]
\end{document}
```

simoncozens:
Linking to bug #118, which blocks this one.

PR #1901, opened 2023-10-22
Fix on command makecolumns

### LINKED ISSUE #1900: Alignment problem when using \makecolumns
The right edge of the frames which are bond to the content one doesn't shift as the number of columns increases.

notaligned

So I did the following changes:

```lua
  self:registerCommand("makecolumns", function (options, _)
    local aligned = {}
    options.columns = options.columns or 2
    for key in pairs(SILE.frames) do
      -- this must be checked before makecolumns() call, since the function changes the right() value of content
      if SILE.getFrame(key):right() == SILE.getFrame("content"):right() and key ~= "content" then
         table.insert(aligned, key)
      end
    end

    makecolumns(options)

    for _, frame in ipairs(aligned) do
      SILE.getFrame(frame):constrain("right", SILE.getFrame("content_col"..options.columns-1):right())
    end
  end, "Split the current frame into multiple columns")
```
And now they do:

aligned

---

The footnote frame is also overwritten if its top and the bottom of content frame are linked, I know this has more to do with the frameset definition than with the makecolumns method, but I think it worth mentioning...

footnote

```sil
\script[src=packages/lorem]
\script[src=packages/footnotes]
\makecolumns
\lorem[words=100]
\showframe[id=all]
\footnote{\lorem[words=20]}
\lorem[words=500]
\end{document}
```

But when the bottom of content is set to a constant value, e.g. 80%ph like this, it works well:

footnote2

jodros:
Fixes #1900

### DIFF
--- a/tests/makecolumns.sil
+++ b/tests/makecolumns.sil
@@ -1,0 +1,5 @@
+\begin[papersize=a5]{document}
+\use[module=packages.frametricks]
+\makecolumns[columns=4]
+\showframe[id=all]
+\end{document}
--- a/packages/frametricks/init.lua
+++ b/packages/frametricks/init.lua
@@ -176,16 +176,28 @@ function package:registerCommands ()
   self:registerCommand("makecolumns", function (options, _)
     -- Set a default value for column count
     options.columns = options.columns or 2
-    local aligned = {}
-    for key in pairs(SILE.frames) do
-      -- this must be checked before makecolumns() call, since the function changes the right() value of content
-      if SILE.getFrame(key):right() == SILE.getFrame("content"):right() and key ~= "content" then
-         table.insert(aligned, key)
+    local current_frame = SILE.typesetter.frame
+    local original_constraints = {}
+    -- Collect existing constraints that may need updating after makecolumns() changes them
+    for frameid in pairs(SILE.frames) do
+      if frameid ~= current_frame.id then
+        local frame = SILE.getFrame(frameid)
+        for method in pairs(frame.constraints) do
+          -- TODO: Remove the assumption about direction when makecolumns() takes into account frame advance direction
+          if method == "right" then
+            if frame[method](frame) == current_frame[method](current_frame) then
+              table.insert(original_constraints, { frame = frame, method = method })
+            end
+          end
+        end
       end
     end
     makecolumns(options)
-    for _, frame in ipairs(aligned) do
-      SILE.getFrame(frame):constrain("right", SILE.getFrame("content_col"..options.columns-1):right())
+    for _, info in ipairs(original_constraints) do
+      local frame, method = info.frame, info.method
+      local final_column_id = ("%s_col%d"):format(current_frame.id, options.columns-1)
+      local final_comumn_frame = SILE.getFrame(final_column_id)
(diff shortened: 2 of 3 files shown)


alerque:
c.f. #420

alerque (commented):
How does this fare when *reducing* the number of columns (as noted is a problem in #248) and/or subsequently changing the value after doing it once? Fixing those cases might not be blockers for this PR but we should at least track whether they are fixed or not.

jodros:
I began to play with \makecolumns just after reading the issue #248, so I have interest in finding a solution for it as well...

> c.f. #420

You mention have done a "dirty hack", well, this description:

> adding ALL the broken frame id's to the insertion steal from list and iterating over ones that exist on a given page

looks like my own approach, why would it be a "dirty hack"? How should a clean fix look like?

alerque:
Sorry I seem to have inadvertently edited *your comment* instead of *reply to selection*! I restored your comment and here is my reply:

> why would it be a "dirty hack"? How should a clean fix look like?

It doesn't take into account things like the possibility of splitting frames (say to add a float) after starting columns, or ending columns after they are started. You don't want insertions stealing space from anything but the actually *adjacent* frames. Ones that were adjacent at one time but later got split or whatever shouldn't be eligible for theft.

alerque:
LOL, so I tried to fix this to work in any direction, but it turns out the column splitter assumes the direction anyway, and this isn't the right output either:

20231027_14h39m11s_grim

Good times. I do think the code I added will be useful later when `makeoptions()` takes into account the frame advance direction, but for now I'll backtrack a little bit to assuming "right" again.

alerque:
@jodros How does my last commit here look to you? It turns out it isn't too useful since it still only works in the `:right()` direction, but I hopefully this will make it easier not harder to eventually expand the column tooling to automatically generate columns in any advance direction.

tests/makecolumns.sil
```diff
@@ -0,0 +1,5 @@
+\begin[papersize=a5]{document}
+\script[src=packages/frametricks]
```
alerque, line 2: @jodros I'll fix this test, but just for your own reference in the future, the non-deprecated syntax for this is:

```suggestion
\script[src=packages/frametricks]
\use[module=packages.frametricks]
```

alerque (changes requested):

alerque (dismissed):

alerque:
Please don't merge from master into feature branches unless there is a real reason to do so, such as the feature being worked on being dependent on something else that landed. It just makes history tracking / git blame / etc. much harder to work with.

Omikhleia:
> Please don't merge from master into feature branches unless there is a real reason to do so, such as the feature being worked on being dependent on something else that landed. It just makes history tracking / git blame / etc. much harder to work with.

Rebase is _always_ a good friend ;)

jodros:
> Please don't merge from master into feature branches unless there is a real reason to do so, such as the feature being worked on being dependent on something else that landed. It just makes history tracking / git blame / etc. much harder to work with.

Sorry, I didn't know it was a problem, I just reverted the merging right now.

alerque (approved):

PR #1904, opened 2023-10-30
Merge develop into master (commit to next release being breaking)

alerque:
This is so I can start keeping an eye on CI status for the branch linked from the milestone...


alerque:
The big Lua restyle from #1931 needs to be run in conjunction to this. If I run it now merges from any future v0.14.x releases into the development branch will become very hard. Right now they are easy, so we need to avoid it until we commit to the v0.15.0 release being next.

Something like this:

```console
$ g co develop
$ stylua --respect-ignores -g '*.lua' -g '*.lua.in' -g '*.rockspec.in' .busted .luacov .luacheckrc build-aux/config.ld .
$ g au
$ g cm "style: Reformat Lua with stylua"
$ g co master
$ stylua --respect-ignores -g '*.lua' -g '*.lua.in' -g '*.rockspec.in' .busted .luacov .luacheckrc .
$ g au
$ g cm "style: Reformat Lua with stylua"
$ g merge develop -s theirs
```

alerque:
Actually scratch that. We can do fancy merge foo now to get both branches reformatted and still have easy future merges. Sometimes you *can* have cake and eat it too.

PR #1908, opened 2023-11-04
Rehash how package idempotency vs. usability works

### LINKED ISSUE #1531: Multiple package initialization case?
The same package can be (re)initialized multiple times (i.e. re-instantiated in `class:loadPackage`), e.g. infonodes gets initialized 2 or 3 times in the manual, notably leading to `Attempted to set the same function hook twice` warnings very early when processing the manual).

[Here](https://github.com/sile-typesetter/sile/blob/01be7155ab07057a51a80e82c07665cbfe268e95/classes/base.lua#L144), it seems to me that each time a package loads another one on which it is dependent, with `self.class:loadPackage()`, then a new instance is created.

_Originally posted by @Omikhleia in https://github.com/sile-typesetter/sile/issues/1366#issuecomment-1200265306_

alerque:
Closes #1531

Lets keep conceptual discussion in the issue, code specific review okay can be here.

### DIFF
--- a/classes/base.lua
+++ b/classes/base.lua
@@ -150,13 +150,17 @@ function class.declareSettings (_)
   })
 end
 
-function class:loadPackage (packname, options)
+function class:loadPackage (packname, options, reload)
   local pack = require(("packages.%s"):format(packname))
   if type(pack) == "table" and pack.type == "package" then -- new package
-    self.packages[pack._name] = pack(options)
+    self.packages[pack._name] = pack(options, reload)
   else -- legacy package
     self:initPackage(pack, options)
   end
+end
+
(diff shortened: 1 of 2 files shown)


alerque:
Lint error is fixed in *develop* separately from this PR.

I think I'm going to go ahead and merge. This seems to be along the lines of what @Omikhleia was reporting / expecting in the original issue. I've started relaying on this in some of my downstream projects and all has gone well enough that it's getting frustrating to patch this into my SILE develop builds to run those projects. If later feedback before release suggests implementation changes before 0.15.0 we can do that, but I'm pretty convinced we'll be going the route of having both possibilities available.

PR #1909, opened 2023-11-11
Showframe should have an option to apply for every page just like \background.

jodros:
I missed an option for this and here is a possible solution.

### DIFF
--- a/packages/frametricks/init.lua
+++ b/packages/frametricks/init.lua
@@ -153,12 +153,28 @@ function package:registerCommands ()
 
   self:registerCommand("showframe", function (options, _)
     local id = options.id or SILE.typesetter.frame.id
-    if id == "all" then
-      for _, frame in pairs(SILE.frames) do
-        SILE.outputter:debugFrame(frame)
+    options.allpages = options.allpages or true
+
+    local show = function()
+      if id == "all" then
+        for _, frame in pairs(SILE.frames) do
+          SILE.outputter:debugFrame(frame)
+        end
+      else
+        SILE.outputter:debugFrame(SILE.getFrame(id))
+      end
+    end
+
+    if options.allpages and options.allpages ~= "false" then
+      show()
+      local oldNewPage = SILE.documentState.documentClass.newPage
+      SILE.documentState.documentClass.newPage = function (self_)
+        local page = oldNewPage(self_)
+        show()
+        return page
       end
     else
-      SILE.outputter:debugFrame(SILE.getFrame(id))
+      show()
     end
   end)
 
--- a/tests/showframes.lua
+++ b/tests/showframes.lua
@@ -1,0 +1,13 @@
+return function()
+    local plain = require "classes.plain"
+    local class = plain()
+    SILE.documentState.documentClass = class
+
+    class:loadPackage("frametricks")
+
+    SILE.call("showframe",{ id = "all"})
+    for i = 1, 50 do
+        SILE.call("supereject")
+    end
+
+end
--- a/tests/showframes.expected
+++ b/tests/showframes.expected
@@ -1,0 +1,136 @@
+Set paper size 	595.275597	841.8897729
+Begin page
+Mx 	295.2916
+My 	780.6177
+Set font 	Gentium Plus;10;400;;normal;;;LTR
+T	20 w=4.6924	(1)
+New page
+My 	780.5689
+T	21 w=4.6924	(2)
+New page
+T	22 w=4.6924	(3)
+New page
+My 	780.6177
+T	23 w=4.6924	(4)
+New page
+My 	780.5103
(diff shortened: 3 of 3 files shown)


alerque:
I've never felt the need for this because if I'm debugging an entire document then `sile -d frames doc.sil` makes more sense to me than editing the doc to embed debug info. Only localized fiddling makes sense to me.

Not saying this can't go in, just thinking out loud about ergonomics...

jodros:
> sile -d frames doc.sil

What a shame, I had no idea of this command 🤦. I'm aware of a debugging facility but never tried it so far... Well, anyway, thanks for telling me that, I'll take a look at the debugging part of documentation...

packages/frametricks/init.lua
```diff
@@ -153,12 +153,28 @@ function package:registerCommands ()
 
   self:registerCommand("showframe", function (options, _)
     local id = options.id or SILE.typesetter.frame.id
-    if id == "all" then
-      for _, frame in pairs(SILE.frames) do
-        SILE.outputter:debugFrame(frame)
+    options.allpages = options.allpages or true
```
alerque, line 156: This should default to false and use SU.boolean() to do so.

packages/frametricks/init.lua
```diff
@@ -153,12 +153,28 @@ function package:registerCommands ()
...
+
+    local show = function()
+      if id == "all" then
```
alerque, line 159: Even local functions should not use variables outside of their scope with a documented reason for needing to. This is called a side effect and has lots of downsides, not to mention will make porting harder down the road. Just add an id field to the function and pass it through explicitly.

packages/frametricks/init.lua
```diff
@@ -153,12 +153,28 @@ function package:registerCommands ()
...
+        for _, frame in pairs(SILE.frames) do
+          SILE.outputter:debugFrame(frame)
+        end
+      else
+        SILE.outputter:debugFrame(SILE.getFrame(id))
+      end
+    end
+
+    if options.allpages and options.allpages ~= "false" then
```
alerque, line 168: See above, but since this should already be cast to a boolean at this point the logic is much simpler: `if not options.allpages then`.

packages/frametricks/init.lua
```diff
@@ -153,12 +153,28 @@ function package:registerCommands ()
...
+      show()
+      local oldNewPage = SILE.documentState.documentClass.newPage
```
alerque, line 170: No, don't do this. I know we do funky things like this elsewhere but we're trying to move *away* from this, not towards it. Set a new page hook instead and just leave the `class:newPage()` function alone.

tests/showframes.lua
```diff
@@ -0,0 +1,13 @@
+return function()
+    local plain = require "classes.plain"
+    local class = plain()
+    SILE.documentState.documentClass = class
+
+    class:loadPackage("frametricks")
+
+    SILE.call("showframe",{ id = "all"})
+    for i = 1, 50 do
```
alerque, line 9: When writing regression test, we want as little output as possible to prove the case. In this case just one supereject and a tiny bit of content should do the trick, no need for 50 pages. Unless you expect to stop functioning after passing 42 for ... galaxy reasons.

alerque (changes requested):
Besides some of the code comments, this option needs documentation in the package docs.

packages/frametricks/init.lua
jodros: Sorry, I don't know how to play with hooks for now. I read this part of the documentation and searched for uses of it in this repository, but still can't manage to set hooks around there... Can you give me at least an example?

packages/frametricks/init.lua
```diff
@@ -153,12 +153,28 @@ function package:registerCommands ()
...
+      SILE.documentState.documentClass.newPage = function (self_)
+        local page = oldNewPage(self_)
+        show()
+        return page
       end
```
alerque, line 175:
```suggestion
      self.class:registerHook("newpage", show)
```

packages/frametricks/init.lua
alerque: Sure, happy to help. I think in this case this whole section of code can be reduced as in [this code suggestion comment](https://github.com/sile-typesetter/sile/pull/1909/files#r1390349519).

packages/frametricks/init.lua
jodros: Thanks for the tip. I got the following error when testing:

`Using code at: ./core/frame.lua:261: Passed a table, expected a string

So when I do ` self.class:registerHook("newpage", show(id)` the error isn't raised, but it still gives the same output: just the first page is debugged...

jodros:
Sorry for the waste of time, after using the `-d frames` options I realized how unnecessary and silly this PR is...

Issue #1866, opened 2023-09-12
Clarify the Markdown scenarios


alerque:
Per #1610 closing a bunch of Markdown related issues I think we have some documentation updates to do. I'd like to

- [ ] Look into adopting `markdown.sile` into this GitHub org as semi-official.
- [ ] Prominently document approaches to handling Markdown based document workflows in the manual and README.
- [ ] Also document an approach to handling minimal Markdown strings without a full inputter / class if we have such a thing available.

alerque:
@Omikhleia Just a heads up one of the big hindrances for me in adopting / recommending `markdown.sile` is the current hard dependency on `silex.sile`. Given the latter's path handling, it ends up taking over core SILE stuff. This may be fine for some folks and some projects, but it definitely isn't appropriate as a default way for people to bolt on Markdown input support.

Given that I've started using the `develop` branch in production and we've dropped the old Markdown support there, I've had to patch it back into SILE just for rendering a few simple strings.

I think we need to reconsider bringing Markdown parsing back into SILE's core without introducing *too many* new dependencies. I realize the libraries we ripped out were a good riddance, but I'd like to get `markdown.sile`'s robustness a lot easier to reach for without dragging along a too many other things.

Omikhleia:
> I think we need to reconsider bringing Markdown parsing back into SILE's core  without introducing too _many new_ dependencies

As for dependencies: without tables, language support and smart quotes, proper super sub- and superscript, I am not sure how it can be achieved with SILE's (current) core without losing functionality... I'm certainly willing to discuss it, but it's not an obvious topic. Perhaps we could have an open discussion on our Gitter (even privately if needed first), at least to consider possible strategies? Also note that markdown.sile goes beyond "mere" Markdown, with Djot etc.

> ... the current hard dependency on silex.sile

Some things from silex.sile could be sorted out and ported here upstream. I had to move "fast" to be able to make book(s), but there would be a way for some features.
On the same vein as I proposed #1853, we could work on proposing other backports.
But beside my limited time, my main concern is that I unsure how to build the new develop branch, and most of all when you'd want 0.15 out. I'd need some perspectives and an idea roadmap to organize my time here.

> I've had to patch (the old Markdown support) back into SILE just for rendering a few simple strings.

_Of course_, I don't think this is a good move -- I'm sure we can do better than reviving that old _broken/limited_ solution.

Omikhleia:
[...]

Issue #1048, opened 2020-08-30
Does SILE (intend to) support nesting `\em`?


jmaibaum:
I see that SILE defines the `\em` command as follows:

https://github.com/sile-typesetter/sile/blob/f371bffab2db6d71ea3607ea34eba12a3eadc6b3/classes/plain.lua#L118

Which doesn't allow for LaTeX-style nesting of emphasis’ like:

```latex
\emph{Emphasize a \emph{word} in emphasised text.}
```

which would typeset *word* upright (again) to make it stand out from the already emphasised surrounding text. To sum up, SILE’s `\em` is more akin to LaTeX’s `\textit`, which inconditionally typesets italic text.

Is there already a SILE command to support nested emphasising? If not, would it be acceptable to change `\em` in order to support it? Or should there be another new command for this use case?

ctrlcctrlv:
I think in this case it makes more sense to rewrite the command. Not all of the time would you want nesting to disable a command, but in this case it does make sense to me.

simoncozens:
Yes, a command which flips the italic-ness sounds sensible to me. I'll have a play with it.

alerque:
Given that the typical correct way to emphasis text inside emphasis is to reverse the style, changing the current function makes perfect sense. Since we only have to look at out current typesetters font stack and extract the style and toggle that value, this should be pretty straight forward.

I would also do the same thing for `\strong`, although that is somewhat less obviously the right thing to do. For people that want specifically a style or weight the `\font` command can set it directly, or we could even have `\i` and `\b` commands that specifically set the value instead of toggling it. That would more easily allow mapping to and from HTML, LaTeX, and some other source formats that make a style vs. semantic distinction in tagging data.

ctrlcctrlv:
I would actually expect nested strong tags to increase the heaviness of the chosen font. So, one strong tag would equal bold and two strong tags would equal heavy. And if the font has an ultra, that would be three. This is how it works in HTML

alerque:
[...]

PR #1913, opened 2023-11-19
Em-nesting, alignment-nesting, greek arithmetic numbering

Omikhleia:
Proposed features from [silex.sile](https://github.com/Omikhleia/silex.sile) - precisely, the subset from **silex.fixes**

Motivated by #1866 

- Nesting-aware `\em` (#1048) - **breaking**
- Nesting-aware (i.e. honoring the margins) `center`, `raggedright` and `raggedleft` + the latter two no longer cancelling paragraph indent[^1] - **breaking**
   - The "blockquote" example in the manual is fixed to also properly honor alignment, and the command is provided in the plain class
- Greek alphabetical numbering special case (similar to "alpha").
- Small robustness fix/refactor for inputter options

[^1]: This is a bold move, but the two things are somewhat orthogonal here, and parindent-cancellation is doable by other means.

Each topic is a separate commit, with detailed commit logs/

Still to do if we agree on the solution:

 - [ ] Check failing tests
 - [ ] Add new tests
 - [ ] Check the manual (obvious change done in the above commits, but maybe we can improve other places)


Omikhleia:
> This is a bold move, but the two things are somewhat orthogonal here, and parindent-cancellation is doable by other means

But i'm not strongly attached to it either. It makes things simpler when one would nest various alignments, but that's kind of an edge case anyway.

alerque:
A cursory glance suggests I'll be up for most of this. I targeted it for v0.15.0 since I'd like to resolve especially the inputter changes by then as well as bathing as many breaking changes as we can without making migration *too* hard. If we cherry-pick some and not other bits that's fine too, but lets try to sort out yes now/no never/maybe later answers on breaking changes sooner rather than later.

alerque:
I cherry picked the obvious wins of the inputter touchup and Greek numbering (with a couple new tests) to *develop* directly, that should leave just the intertwined margin/nesting stuff for this PR.

classes/plain.lua
```diff
@@ -260,20 +248,110 @@ function class:registerCommands ()
...
+  self:registerCommand("justified", function (_, content)
+    SILE.settings:temporarily(function ()
+      local lskip = SILE.settings:get("document.lskip") or SILE.nodefactory.glue()
+      local rskip = SILE.settings:get("document.rskip") or SILE.nodefactory.glue()
+      SILE.settings:set("document.lskip", SILE.nodefactory.glue(lskip.width.length))
+      SILE.settings:set("document.rskip", SILE.nodefactory.glue(rskip.width.length))
+      SILE.settings:set("document.spaceskip", nil)
+      -- HACK. This knows too much about parfillskip defaults...
+      -- (Which must be big, but smaller than infinity. Doh!)
+      SILE.settings:set("typesetter.parfillskip", SILE.nodefactory.glue("0pt plus 10000pt"))
```
Omikhleia, line 304: I overlooked the possibility to reset a setting to its default value, which should do the trick here.

Omikhleia: Done.

classes/plain.lua
```diff
@@ -260,20 +248,111 @@ function class:registerCommands ()
...
+    end)
+  end, "Typeset its contents in a centered block (keeping margins).")
+
   self:registerCommand("raggedright", function (_, content)
-    SILE.call("ragged", { right = true }, content)
-  end)
+    SILE.settings:temporarily(function ()
+      local lskip = SILE.settings:get("document.lskip") or SILE.nodefactory.glue()
+      local rskip = SILE.settings:get("document.rskip") or SILE.nodefactory.glue()
+      SILE.settings:set("document.lskip", SILE.nodefactory.glue(lskip.width.length))
```
Omikhleia, line 273: So the question here (and on raggedleft) likewise is whether we should cancel parindent.
The old behavior was to cancel it, now we do not -- and it's a breaking change.

None of these behaviors is fully satisfying...
In ragged environments, one still needs to be able to see where paragraphs start... There are _usually_ two possibilities here:
- Either keeping a paragraph indent
- Or using a non-null paragraph skips

As I see it, there could be two approaches here:
- Either we let the user independently cancel the parindent if they do not want it.
- Or we check if there's a (fixed part) parskip and take decision upon it. But it might be too clever?
  ```lua
      -- If paragraphs are separated by a skip, the usual habit is to cancel the
      -- paragraph indentation.
      local pskip = SILE.settings:get("document.parskip") or SILE.nodefactory.glue()
      if pskip.height:tonumber() > 0 then
        SILE.settings:set("document.parindent", SILE.nodefactory.glue())
        SILE.settings:set("current.parindent", SILE.nodefactory.glue())
      end
   ```

alerque: I don't like the "too clever" approach, given how subjective the right answer is it will be wrong for same use cases.

I also don't like automatically forcing one style on folks just because they switched an unrelated setting.

My suggestion would be to go ahead with the breaking change and start not touching the parindent setting just because ragged was turned on. We can throw a warning about how to enable it manually for folks whose documents may re-flow.

alerque:
I think we'll move ahead with the current breaking change that does *not* cancel `parindent`, but we'll add a shim to the *retrograde* package that restores the old behavior. This should be relatively easy to shim this way, and since the actual value of `parindent` is also changing in this release I think it makes sense to allow the breakage but also support old documents with the same mechanism.

Obviously the retrograde shim will not be expected to also support nested use cases the way the new code does, just cover the very basic case of a single level usage and simulate the old behavior. This will probably mean either stashing a private setting, having some private hook, or registering a wrapper/replacement function.

classes/plain.lua
alerque: I reviewed all the tests and found two types of changes: fixes to bugs we didn't even know existed because the previous resetting behavior was making them and expected changes based on new behavior. I updated some of them using `packages.retrograde` to keep the original test purpose intact while others that were incidental I just updated the new expectations (after verifying that the retrograde package restored the original expectations).

Issue #1337, opened 2022-03-05
Consider whether LDoc could be leveraged for developer docs


alerque:
On a related note to documentation overhauls in #1194 it might be interesting to look at whether [LDoc](https://github.com/lunarmodules/LDoc) can be leveraged. I would guess this would be more useful for the kind of documentation a package developer might need that a SILE user, but it's possible that better documentation of our own internals could be a benefit.

Two possible outcomes might be a set of developer API docs separate from the user manual using HTML output or writing some sort of wrapper that leverages LDoc to extract all the documentation information then uses SILE itself to typeset this as part of the manual.

alerque:
We're also going to have Rust API docs. It might be worth looking at [tealr](https://github.com/lenscas/tealr) too.

PR #1914, opened 2023-11-21
Refactor AST utilities

### LINKED ISSUE #1196: Standard way for extracting a node from the AST tree

Omikhleia:
Proposal, motivated by #1866 
- Move existing AST utilities to `SU.ast`
  - _Rationale:_  While the SILE AST is just a table with some structure (tied to `SILE.process` etc.), having some convenience function to manipulate is good -- and a must-have for inputter and package designers. The existing utilities were however scattered (some at the root of `SU`, one as an inputter method, another in the inputfilter package, etc.)...
- Refactored as a separate utility file.
  - _Rationale:_  We have lots of different/unrelated stuff in `SU`, a better organization makes maintenance of similar functions easier and is the path for curating those utilities progressively. Later, we might at a glance easily check these if we ever want to change the AST (add new constructs, make it more object-oriented, etc.)
  - Adding in-code documentation: follows VSCode Lua extension syntax (I tried LDoc initially but could not make it work with VSCode... So there's nothing definitive here, I just thought something would be better than nothing.)
- Proposed additions (taken from my **silex.ast** layer)
   - `SU.ast.extractFromTree` - complements `SU.ast.findInTree`
      - Closes #1196 
   - `SU.ast.createCommand` - as in inputfilter, but in inputters etc. this is convenient (and we might not even have loaded any package yet, there, by nature)
   - `SU.ast.createStructuredCommand` - convenience when passing an array of nodes (avoiding useless nesting)
   - `SU.ast.trimSubContent` and `SU.ast.processAsStructure` - fairly useful esp. in XML parsing context, when the schema defines structured elements where text/spaces aren't expected (unless due do indentation, but to be ignored in the ouput).
- Other changes (also from **silex.ast**)
  - `SU.ast.subContent` is simplified and no longer includes a weird `id=stuff` entry - discussed in #1834 (items **B1** and **B2**)


alerque:
> Adding in-code documentation: follows VSCode Lua extension syntax (I tried LDoc initially but could not make it work with VSCode... So there's nothing definitive here, I just thought something would be better than nothing.)

Something is definitely better than nothing. We'll likely be going the LDoc route eventually (c.f. #1337) but it will be much easier to adapt something than nothing!

core/utilities-ast.lua
```diff
@@ -0,0 +1,213 @@
...
+    end
+  end
+end
+
+--- Find and extract (remove) a command node in a SILE AST tree,
+--- looking only at the first level.
+---@param tree      table       AST tree
+---@param command   string      command name
+---@return          table|nil   AST command node
+function ast.extractFromTree (tree, command)
```
alerque, line 24: I don't think "extract" is the right verb here, at least to me that implies getting a node out in isolation, but also I would have assumed it *left* the original tree alone. This removes it from the tree, so "slice" or perhaps "extricate" might be a better name that suggests the original is being removed.

alerque (commented):
This will need a `BREAKING CHANGE` log message, but I can take care of that. Overall I think it's looking good. Just doing some tinkering on my end to see what I think of the details.

core/utilities-ast.lua
Omikhleia: maybe simpy `removeFromTree` (and the `@return` becomes "Removed AST command node") ?

Omikhleia:
By the way @alerque I was about to ask whether we should actually split utilities in subfolders (rather than `-xxx` files), and I just see you felt alike :) :+1: 

Would that be ok if we go on splitting the utilities  in subfolders? Currently we still have in the main files
- Things pertaining to casting
- Things pertaining to error/warning etc.
- Things pertaining to general utilities (map, max, etc.)
- Things pertaining to unicode and characters
- A bunch of other things likely...

Perhaps starting to split them would allow us to properly group them by "nature" and document them progressively?

Omikhleia:
> Perhaps starting to split them would allow us to properly group them by "nature" and document them progressively?

That could be another issue addressed later, obviously, just asking here as it sounds quite logical.

alerque:
> Would that be ok if we go on splitting the utilities in subfolders?

Yes. Smaller files/modules with tighter scopes are almost always better for Lua in my opinion.

alerque (approved):

alerque:
P.S. I'm also seriously scratching my head why we have "core" and "utilities". It seems to me the way our source organization is sorting out almost everything in "core" is just utilities. Basically anything that is just a function as opposed to a class than gets instantiated and maintains some sort of state as SILE runs.

Omikhleia:
> P.S. I'm also seriously scratching my head why we have "core" and "utilities". It seems to me the way our source organization is sorting out almost everything in "core" is just utilities. Basically anything that is just a function as opposed to a class than gets instantiated and maintains some sort of state as SILE runs.

The line is blurry, indeed.
A few things stand apart, though and hint towards different natures:
- line breaker, page builder, hyphenator, fonts = core algorithm (some are classes or alike, actually): pretty specific to the field of digital typesetting
- color, units, nodes, frame... = low level common concepts and data structures
- cli, repl... = their name says it all, running on command line
- _most_ current "utilities" are things you one could work without
    - Many just avoid lots of code and/or make things uniform
    - Several could find their place in other pieces software actually (error handling with stack, string encoding or collation, etc.)

So there could be some different levels of concern and design here ;-)

PR #1926, opened 2023-12-09
fix(packages): Adjust dropcap logic for letters with a depth

### LINKED ISSUE #1925: Dropcap with descenders introduces uneven lines
Ran with docker `siletypesetter/sile:v0.14.13`:

```
\use[module=packages.lorem]
\use[module=packages.dropcaps]
\dropcap[join=true, lines=4]{J}\lorem[words=50]
\end{document}
```

image

Same issue with a Q.

No font specified, so using default font Gentium Plus -- which happens to have capitals with descenders. Might occur with other fonts, though I wouldn't have noticed used most of the time fancy fonts such as Zalmann Caps ^^ i.e. fonts without descenders.

And it seems indeed that the depth being overlooked or not being taken into account somewhere:

image

Omikhleia:
Closes #1925 

Here is a proposal.

**1.** Default behavior
Honors the specified hanged lines, size the characters accordingly (now taking into account the depth).

image

It works well with EB Garamond Initials (center).
It does what it has to do for Gentium Plus (left): Q, J and even the O are slightly smaller so as to fit within the constraints.
It's a bit sad for Gingerbread Initials (right), where the Q is small but could have decently fit within the margin of a reasonable baseline...

But that's the most robust way.

**2.** With new option `strict=false`
It allows adding extra lines when needed (with a reasonable tolerance) so that the characters are all scaled identically.

image

It does what it has to do for Gentium Plus: Q, J and O are now the same sizes as the others + Q and J require an extra line, but not O as its small depth is reasonable).
Sadly, due to its descenders, EB Garamond Initial now always takes more lines. Here 4, so it looks decent due do its design, but this may not be what one wants...
On the bright side, Gingerbread Initials are now neat (the small depth of the Q is reasonable and doesn't need an extra line).

**3.**  With new options `strict=false` and `depthadjust=true`

The most fragile method, using an "X" as reference to check if characters hanging below the baseline.

image

It works as (2) for Gentium Plus (= extra hanged lines when needed), as (1) for EB Garamond Initials (= 3 hanged lines as expected) and as (2) for Gingerbread Initial (perfect Q).

My favorite, though relying on some ad hoc heuristics -- sooner or later some people will find other pathological fonts.

Any opinion(s) on the approach?


packages/dropcaps/init.lua
```diff
@@ -31,32 +43,68 @@ function package:registerCommands ()
...
+    -- Typeset the dropcap with its final shape, but don't output it yet.
     local hbox = shapeHbox(options, content)
 
+    -- Compensation for regular extra depth.
+    local compensationHeight = depthAdjustment * options.size / curSize
+
+    if not strict then
+      -- Some fonts have descenders on letters such as Q, J, etc.
+      -- In that case we may need extra lines to the dropcap.
+      local hasExtraDepth = hbox.depth:tonumber() > compensationHeight
```
Omikhleia, line 91: This condition is wrong. I need to check here.

Omikhleia:
Jena Gotisch sucks in all cases (in my opinion)

image

But it helped finding a bug, so heh...

packages/dropcaps/init.lua
Omikhleia: Fixed. I have no idea what passed through my mind, trying to compensate for something that was already taken into account...

packages/dropcaps/init.lua
```diff
@@ -9,6 +9,18 @@ function package:_init ()
...
+local function getToleranceDepth ()
+  -- In non-strict mode, we allow using more lines to fit the dropcap.
+  -- However we cannot just check if the extra depth of the dropcap is above 0.
+  --  - our depth adjustment is but a best attempt.
+  --  - Some characters may have a small depth of their own, such as the "O" in Gentium Plus.
+  -- We must just ensure they stay within reasonable bounds with respect to the baseline,
+  -- so as not to flow over the next lines.
+  -- With well-formed text font, a good compromise is around 0.3bs (LaTeX does this too for defining its \strut).
+  -- This could be computed more precisely (e.g. computing the ratio between ascender and descender font metrics).
+  return SILE.measurement("0.3bs"):tonumber()
```
alerque, line 21: What do you think about making the *relative* measurement a setting (so a user could tweak it to play nice with a weird font without struggling to reach this private function and only using this function (or probably just its calling location) to absolutize the measurement returned by the setting?

alerque (changes requested):
This looks pretty sane, just one idea about making a hard coded value a user setting. If that doesn't sound like an improvement to you I'm quite open to hearing so and just putting this through.

packages/dropcaps/init.lua
Omikhleia: I dunno, actually.
Another way would be to use the font metrics to compute the real value, and we actually have the "fontmetrics" C code based on harfbuzz and currently used by the linespascing package -- but I don't understand why we have it (i.e. what is its status besides our "justenoughharbuzz"?).
I did a quick test, assuming EB Garamond for the main text, it gives a 0.36(..) bs ratio for it, so "quite "close"
Perhaps we should go for the real things here? I.e. compute the real ascender / descender ratio (assuming font designers set those rights in their metrics) to avoid the hardcoded 0.3bs? We'd still need to round it a bit down, though, for safety (it it's _exactly_ the ratio, then we'll still have lines pretty closed, unless document.lineskip ticks in, but this is fragile)

Omikhleia:
In example (1), the J seems a bit to high (its top reaches the top of the /l/ of lorem, but the A does not show that, nor the J of solution (2) -- I may have some math wrong here. Gonna check.

packages/dropcaps/init.lua
alerque: The "why we have it" for any of the `justenough*` libraries is simply glue. Those are all just thin wrappers around external binary libraries giving access to the functions we want to reach from Lua. That's it. Really they should all go away in favor of using full binding libraries, but since full binding sets don't always exist in Lua Simon built out "just enough" of a binding set to do the work we wanted to do.

For now we'll keep maintaining however many bindings we actually need in SILE, no more no less. It isn't a lot of work to keep around or even add new ones as needed.

So yes if you want to use font metrics for this we can do so, and if we need more functions we can add them.

This will be even easier down the road from the Rust side of things where there are more full libraries available with bindings. For example we won't need just enough HarfBuzz because we'll have 100% access to the HB API out of the box.

While using the volues Font designers encoded is the *right* thing to do, there will be lots and lots of cases where the values are invalid. Special decorative one-off projects are even more likely to be broken than the rest. Having some way to set this manually even if we computer a theoretically correct value might still be worth the effort.

Omikhleia:
> In example (1), the J seems a bit to high (its top reaches the top of the /l/ of lorem, but the A does not show that, nor the J of solution (2) -- I may have some math wrong here. Gonna check.

Computation fixed, options simplified (cases 2 and 3 from the initial discussion merged in one case) and depth adjustment character(s) made configurable rather than hard coded. Also added a specific test for these things.

Omikhleia:
Animation showing proposed behavior:

anim-dropcaps-mode

("strict=true" vs. "strict=false" and default (computed) adjustments)

alerque:
Looking like a good improvement to me. Are we ready to drop the draft status on this?

alerque (approved):

PR #1929, opened 2023-12-11
Change module load error to include suggestions of how to install 3rd party modules

### LINKED ISSUE #1882: Add suggestion for installing packages to error output
When a document uses some iteration of `\use` and the require fails, we currently throw a generic Lua error that at least shows the paths that were tried, but is otherwise unhelpful to new users. I think we could probably take that chance to suggest what it might look like to use `luarocks install`, perhaps suggesting `--local --tree ./lua_modules` as a way to have dependencies local to a project, and also suggesting `--lua 5.1` or whatever the matching Lua version is for the way SILE is currently being run. If we don't include a lengthy help message at least a few small tips and a link to a wiki page or something might help make it easier to download some sample document and get up to speed with it if it requires special packages.

The hardest part might be mapping modules to rock names, but maybe we can figure out a way to tag them in the manifest such that a search consistently returns the correct rock.

alerque:
Closes #1882

For some reason I'm getting doubled output right now so I'll sleep on this...

### DIFF
--- a/core/sile.lua
+++ b/core/sile.lua
@@ -146,10 +146,43 @@ SILE.init = function ()
   runEvals(SILE.input.evaluates, "evaluate")
 end
 
+local function suggest_luarocks (module)
+  local guessed_module_name = module:gsub(".*%.", "") .. ".sile"
+  return ([[
+
+    If the expected module is a 3rd party extension you may need to install it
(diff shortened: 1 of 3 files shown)


alerque:
Doubled error output has been going on for a while and somehow escaped my notice. What a mess. This now cleans that up too.

PR #1932, opened 2023-12-13
Change default parindent

### LINKED ISSUE #1722: Adopt a saner default parindent replacing the fixed 20pt
Along the same lines as #1371 ... towards saner defaults.

The default paragraph indent (`document.parindent`) is currently a fixed 20pt. 

I already mentioned it in PRs and discussion, but let's track it officially: According to all typography sources, this doesn't seem to be sane value. It shouldn't be fixed, but ought to be text-size dependent.

Let's quote Bringhurst for instance:

> The most common paragraph indent is one em. Another standard value is _one_ lead. If your text is set 11/13, the indent would then be either 11 pt (one em) or 13 pt (one lead). One en (half an em) is the practical minimum. Where the line is long and margins are ample, an indent of 1 ½ or 2 ems may look more luxurious than one em, but paragraph indents larger than three ems are generally counterproductive

Or Lacroux (my approximate translation from French)

> In “ordinary” or balanced compositions (font size, letter spacing and line spacing all well suited for justification, use
    of a "reasonable" font family), the traditional indent of one em-space seems appropriate to me. (...)
    However, in certain "more or less disproportionate" compositions (examples: text block too small for proper justification),
    it makes sense to increase the value of the paragraph indent up to one em-space and a half, or even two em-spaces. (...)
    The paragraph should not be petty: one will never attribute to the indentation less than one em-space.

They don't have to all agree on the details (e.g. Bringhurst considers ½ em as an acceptable minimum, Lacroux visibly rejects anything smaller than 1em). We just have to note that:
- **1em** is the usual "common" value
- one _lead_ (= 1bs ?) is possible (but not that common? I'm not so sure, see below!)
- 2em is the maximum, and 1 ½ a kind of compromise.

It's a matter of taste (personally, I find 1 ¼ em nicer... and realize it's close to the baseline skip) -- but anyhow SILE should have a decent sane default, i.e. **1em** for general use -- Certainly not a fixed size in points!

alerque:
Closes #1722


alerque:
I couldn't think of a good way to throw a warning on this without being obnoxious to new users and those who have already updated their docs without inventing some new system like a CLI flag to output breaking change hints in the current release.

alerque:
So the remaining test cases that are failing here are unexpected side effects to say the least. The `retrograde` package *not* giving them a pass was actually a pretty good warning sign.

The differences are not obviously wrong for the things intended to be tested, but still the side effects should be reviewed. All of them seem to involve rotation: either frames or tate or whatever.

PR #1933, opened 2023-12-13
Change default space width

### LINKED ISSUE #1371: Questioning SILE's default interword spacing
So the default behavior in SILE is to use, for interword spacing:
- `shaper.spaceenlargementfactor` = 1.2
- `shaper.spacestretchfactor` = 1/2
- `shaper.spaceshrinkfactor` = 1/3

The stretch and shrink are somewhat consistent with what TeX and Co do.

Where does the default 1.2 enlargement factor comes from, however? Is there any rationale (read: typography literature) for that value?

alerque:
Closes #1371

### DIFF
--- a/packages/retrograde/init.lua
+++ b/packages/retrograde/init.lua
@@ -12,6 +12,9 @@ end
 
 -- Default settings that have gone out of fashion
 package.default_settings = {
+   ["0.15.0"] = {
+      ["shaper.spaceenlargementfactor"] = 1.2,
+   },
 }
 
 function package:_init (options)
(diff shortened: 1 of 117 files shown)


alerque:
The `retrograde` package is going to be a big help with tests this release cycle. Obviously a lot of these test expectations could be updated because the exact metrics don't affect the thing being tested, but I don't want to do that review separately for each major breaking change we're making or considering separately. This way we can make sure the changes only affect settings we know about, then eventually work backwards to drop this from tests after manual review on a case by case basis.

Issue #872, opened 2020-05-25
Build Fedora packages


alerque:
Using a SILE official copy repository?

hegjon:
Dependency lua-cldr is also added for package review https://bugzilla.redhat.com/show_bug.cgi?id=2142653

alerque:
This is great! Let me know if you need anything from me. It's been a long time since I worked with RPM packaging (and it was PLD Linux long before Fedora existed) but if you get stuck let me know. You might also be able to take some hints from the recently added OpenSUSE packaging since it is RPM based on they have all the Lua dependencies built now. The distro specific packaging macros are bound to be different but you can probably still take some clues from them.

hegjon:
Thanks, I have looked at the OpenSUSE packaging, they are using luarocks RPM macros that is not included in Fedora. So my packages are mostly dropping files into /usr/share/lua/5.4/

hegjon:
I have made and added the following depenecies for package review: 

 * lua-cassowary: https://bugzilla.redhat.com/show_bug.cgi?id=2142399
 * lua-cosmo: https://bugzilla.redhat.com/show_bug.cgi?id=2142671
 * lua-cldr: https://bugzilla.redhat.com/show_bug.cgi?id=2142653
 * lua-epnf: https://bugzilla.redhat.com/show_bug.cgi?id=2142786

I have started on lua-linenoise, but I am not sure how to package it, since Fedora does not want bundled libraries and it is built against a forked version of linenoise.

alerque:
I think you'll need somebody with Fedora expertise to answer that one. As a general rule Arch Linux doesn't like bundling libraries either and when possible we don't — but Arch also takes a pragmatism-above-principles approach and when there isn't another way to get the job done we do what works. That means there are *some* bundled libraries in packages.

In the case of forked libraries and where the fork is (as in the case of Lua linenoise) the long stading defacto library to use for the ecosystem I think there should be a way to do it.

hegjon:
I started building the dependencies at COPR while I am waiting for peer review approval https://copr.fedorainfracloud.org/coprs/jonny/SILE/packages/

You can use COPR for CI if you have a static build/check, then you get a large coverage of RPM distros/arches.

alerque:
[...]

PR #1936, opened 2023-12-13
Document availability of official Fedora packages

alerque:
Pending review process actually completing and #872 being closed this should update the docs to reflect the official packaging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,10 +70,13 @@ If you install LuaRocks for use with SILE via `pacman`, use the `lua51-*` varian
 
 #### Fedora
 
-A [COPR][copr] repository is available for Fedora users with packages of SILE
-and all the necessary dependencies.
-Fedora 38 and later are supported.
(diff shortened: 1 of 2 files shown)


alerque:
Of course @hegjon I'll want a fact check on this after it's official so the actual supported release targets are correct. I'm hoping this documentation can land in the manual before the next release, but of course as with anything downstream it happens when it happens.

hegjon:
I think there are no show stoppers for me to include it for Fedora 38 and later, except that I have upgraded all my machines to Fedora 39, so it would be nice if someone else could test it on Fedora 38.

Fedora 37 was EOL Tuesday, 5 December 2023 [1], so I would remove Fedora 36 and Fedora 37 from the documentation.

 [1] https://lists.fedoraproject.org/archives/list/announce@lists.fedoraproject.org/thread/BOMYDF5U3FVAQSOGM6NZEZCTRBMQQJ56/

PR #1937, opened 2023-12-14
Add retrograde package to ease version transitions by restoring old defaults

### LINKED ISSUE #1934: Come up with a way to ease people into new settings
Some of the changes we're scheduling for v0.15.0 are going to be pretty disruptive. The basic way the typesetter works is unchanged, but the default font metrics are all different (spaces in #1371, indents in #1722, leading in #1118).

Throwing CLI warnings for these seems like a noise proposition and we can't easily detect when the user expectation will be the old settings.

Lets come up with an alternate way that bridges the gap between change logs and people actually getting back to rendering.

It occurred to be one option might be a shim package that could reset the defaults to v0.14.x wvalues. This could be simple use command away either in old documents or from the CLI, for example:

```console
$ sile -u package.defaults.014x <INPUTS>
```

This might be easier than expecting people to know what the original settings were or implement a bunch of individual work arounds based on reading the release notes.

alerque:
Closes #1934

### DIFF
--- a/documentation/c05-packages.sil
+++ b/documentation/c05-packages.sil
@@ -274,6 +274,10 @@ This section introduces packages that could not fit in another category.
 \status:low
 \package-documentation{ifattop}
 
+\subsection{retrograde}
+\status:high
+\package-documentation{retrograde}
+
 \section{Frames and page layouts}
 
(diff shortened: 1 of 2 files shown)


alerque:
Any thoughts on the ergonomics of this @Omikhleia?

PR #1941, opened 2023-12-19
Touchup Mac builds and Nix flake

### LINKED ISSUE #1879: Fixup appkit detection in configure.ac to not need patching later
> Not necessarily related to this change, but I consider such `sed` hooks rather ugly, and usually they are a result of upstream developers who don't make efforts for their distributors. This is not the case with sile, as development here is very much involved with distribution, so I was wondering whether in the `develop` branch you would like to make the effort to fix this issue, or at least explain better why is it needed only for Nix builds. Note it is rarely needed in Nixpkgs from my experience.

_Originally posted by @doronbehar in https://github.com/sile-typesetter/sile/pull/1871#discussion_r1322927935_
            
Also: https://github.com/sile-typesetter/sile/pull/1871#discussion_r1322928621

alerque:
Closes #1879

### DIFF
--- a/build-aux/pkg.nix
+++ b/build-aux/pkg.nix
@@ -118,8 +118,6 @@ in stdenv.mkDerivation (finalAttrs: {
 
   postPatch = ''
     patchShebangs build-aux/*.sh build-aux/git-version-gen
-  '' + lib.optionalString stdenv.isDarwin ''
-    sed -i -e 's|@import AppKit;|#import <AppKit/AppKit.h>|' src/macfonts.m
   '';
 
   NIX_LDFLAGS = lib.optionalString stdenv.isDarwin "-framework AppKit";
--- a/build-aux/pkg.nix
+++ b/build-aux/pkg.nix
@@ -138,15 +138,6 @@ in stdenv.mkDerivation (finalAttrs: {
     touch source/build-aux/rust_boilerplate.mk
   '';
 
-  # remove forbidden references to $TMPDIR
-  preFixup = lib.optionalString stdenv.isLinux ''
-    for f in "$out"/bin/*; do
-      if isELF "$f"; then
-        patchelf --shrink-rpath --allowed-rpath-prefixes "$NIX_STORE" "$f"
-      fi
-    done
-  '';
-
   passthru = {
     # So it will be easier to inspect this environment, in comparison to others
     inherit luaEnv;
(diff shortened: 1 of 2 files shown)


build-aux/pkg.nix
```diff
@@ -118,8 +118,6 @@ in stdenv.mkDerivation (finalAttrs: {
 
   postPatch = ''
     patchShebangs build-aux/*.sh build-aux/git-version-gen
-  '' + lib.optionalString stdenv.isDarwin ''
-    sed -i -e 's|@import AppKit;|#import <AppKit/AppKit.h>|' src/macfonts.m
   '';
 
   NIX_LDFLAGS = lib.optionalString stdenv.isDarwin "-framework AppKit";
```
doronbehar, line 123: Could it be this can be removed as well?

alerque: I don't know. Nor do I know how to trigger a Darwin build to find out.

It seems like if this had to get added at all, nothing has changed that would make it irrelevant now. The sed expression removal works because we just fixed it to be that way upstream. On the other hand Homebrew doesn't seem to need this LDFLAG (or perhaps it has it by default?).

doronbehar: > I don't know. Nor do I know how to trigger a Darwin build to find out.
> 
> It seems like if this had to get added at all, nothing has changed that would make it irrelevant now. The sed expression removal works because we just fixed it to be that way upstream. On the other hand Homebrew doesn't seem to need this LDFLAG (or perhaps it has it by default?).

Hmm I wonder whether this is still relevant. According to `git blame` those additions were done long ago:

https://github.com/nixos/nixpkgs/commit/4a8b2c29420ef11af6d9abad8d8b6710eb397a23

In general, if you want to setup a Darwin & Nix CI, I recently heard of this interesting project:

https://github.com/MatthewCroughan/NixThePlanet/#macos 

You can start by testing the build on that VM by yourself. Plugging that machine into a GitHub CI would be a nice Nix miniproject, but let's leave that aside for now.

Another option would be to use Nixpkgs' CI to test that - i.e open a Nixpkgs draft PR, and see how the Darwin builds get along. This also takes time though, as our CI runner has many Darwin requests to process.

You could also commit the Linux related removals, and only deal with these `NIX_LDFLAGS` later.

alerque: Thank for the options. I'll look into them, some now some later.

For the moment I think we'll move ahead with the bits I'm sure about, then when v0.15.0 actually drops and we update the official Nix pkgs repo with the stable release we can test whether than is needed or not. If not we can then remove it from the Flake too.

doronbehar:
Sounds good to me. Great!

PR #1950, opened 2023-12-29
Balancing in command makecolums set as an option

jodros:
The user should have the option of disabling the balancing when calling `makecolumns`. For example, this just happened to me right now, using the default `makecolumns`:

image

But I wanted their bottom to be the same as the original frame, which I got after setting balancing to false:

image

### DIFF
--- a/packages/frametricks/init.lua
+++ b/packages/frametricks/init.lua
@@ -42,6 +42,7 @@ end
 local makecolumns = function (options)
   local cFrame = SILE.typesetter.frame
   local cols = options.columns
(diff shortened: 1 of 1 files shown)


alerque (approved):

alerque:
Since this is a single-commit change I can easily fix this in merging, but do note that this project uses [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) which allows us to automatically generate change logs as well as make other management tasks easier when interacting with history.

PR #1959, opened 2024-01-20
feat(typesetters): Support for speaker change introduced by em-dash

### LINKED ISSUE #1958: Support emdash-introduced speaker changes in dialogues
(From https://github.com/Omikhleia/markdown.sile/issues/105)

In French and in Turkish (at least), it's common do have speaker changes in dialogues introduced by an em-dash:

_— Lorem ipsum dolor sit amet, consectetur adipiscing elit, dit il.
— C’est sûr !_

For comparison, English does not use such dashes normally.

_“Lorem ipsum dolor sit amet, consectetur adipiscing elit,” he said.
“Sure!”_

(English may use em-dashes, but at the end of a quote. Anyway, that's not the point here).

The tricky thing in such dialogues is that the space following the em-dash ought to be fixed (= not stretched or shrinked by line justification) so that subsequent dialogue lines start identically.

The workaround is to manually enforce it, by using explicit kerns in the input

```
—\kern[...]Lorem .... 

—\kern[...]Sure!
```

It's cumbersome, annoying when copying/pasting text.... and we'd be back to discussing a command for inserting a fixed inter-word space rather than craft a kern specifically.

This enhancement proposal is to automatically handle those case.

I'll make a PR with an implementation, for discussion.

Omikhleia:
Closes #1958 

Some conventions denote speaker changes in dialogues by starting a paragraph with an em-dash.
In these cases, the space after the em-dash shall be fixed by default.

This is a _naive_ proposal: intercept those from the input at typesetter level.

**However**, this implementation breaks test-865 which is a pullquote test.
The pullquote author stands on its own paragraph line and is introduced by an em-dash: _— Joe Smith_

If done this "naive" way, what it does show is a bit more general: the kerning (likely) may be slightly affected:

image

This is not illogical (our fixed kern, introduced to replace the space, causes shaping to occur on each part around it, rather than the whole text. Whether this is acceptable is another question...

If we wanted it addressed _after_ shaping, it probably gets a bit more complex, because we are then far past the point where we are sure to be at the start of a paragraph. What we could perhaps do is to only _mark_ the unshaped text in some way at the typesetter level (as here, but without doing the replacement), and handle that later (after shaping, perhaps even in unicode node breakers where spaces are replaced with glues -- EDIT: which implies propagating the "mark" from the unshaped node to the shaped one(s) ^^).


Omikhleia:
The **second** commit (refactor) does what's suggested above = marking the unshaped node, but doing the space replacement after shaping. No regression, and probably better?

The cost of the added complexity of this kind of trick might perhaps be acceptable.

core/nodefactory.lua
```diff
@@ -261,6 +261,18 @@ end
 
 function nodefactory.unshaped:shape ()
   local node =  SILE.shaper:createNnodes(self.text, self.options)
+  if self.markedForSpeakerChange then
```
Omikhleia, line 264: If going this route (space handling after shaping), rather than using a mark flag and then a dumb nil check in almost everything we shape, subclassing the node would certainly be better :)

Omikhleia: Done. "Avoid messing with standard node classes" should be a motto.

typesetters/base.lua
```diff
@@ -330,6 +337,31 @@ function typesetter:endline ()
...
+  local node = self._base.shape(self)
+  local spc = node[2]
+  if spc and spc.is_glue then
+    -- Switch the variable space glue to a fixed kern
+    node[2] = SILE.nodefactory.kern({ width = spc.width.length })
+    node[2].parent = self.parent
+  else
+    -- Should not occur:
+    -- How could it possibly be shaped differently?
+    SU.warn("Speaker change logic met an unexpected case, this might be a bug.")
```
alerque, line 360: If this is "Should not occur" error then `SU.error("...", true)` would be more appropriate to actually throw a bug error.

typesetters/base.lua
```diff
@@ -145,6 +145,13 @@ function typesetter.declareSettings(_)
     default = false,
     help = "When true, a warning is issued when a soft hyphen is encountered"
   })
+
+  SILE.settings:declare({
+    parameter = "typesetter.dialogue",
```
alerque, line 150: Something about this setting name doesn't quite sit right with me—possible just the lack of scope. I don't have a great suggestion though. What about `typesetter.em-dash-demarcates-dialogue` or something else similarly verbose that gives some more clues about what is being changed? just "dialogue" sounds more to me like an environment name where any number of settings could have been adjusted than a single setting.

typesetters/base.lua
```diff
@@ -145,6 +145,13 @@ function typesetter.declareSettings(_)
...
+    type = "boolean",
+    default = true,
```
alerque, line 152: Just to limit possible unforeseen consequences, I'm thinking we should default this to `false`, then explicitly enable it to true in the French and Turkish language modules (and possible a more languages that may have taken any queues from French).

alerque (changes requested):
A couple relatively minor thoughts...

typesetters/base.lua
Omikhleia: My rationale for a warning is that it won't really break (but it will just leave a stretchable glue): so we should perhaps not block users here if the edge-case is ever met for any reason -- it's not worse than what they'd be getting with the feature disabled. It can be changed to an error if deemed preferable (and if the case occurs, we'll need to be on the front-line for fixing it)

typesetters/base.lua
Omikhleia: I am not happy with `typesetter.dialogue` either :)
Erm, perhaps `typesetter.emdashSpeakerChange` simply?

typesetters/base.lua
Omikhleia: As for unforeseen consequences, they'd _only_ occur for an emdash + spacing at the start of a paragraph. I don't see any good reason one would want a stretchable space there, even if the language usage don't rely on this for dialogue speaker changes. Rather the contrary, e.g. if we get a "loose" list input in English
```
— Step 1 stuff

— Step 2 more stuff
```

Could be controversial, but even in those case I'd prefer seeing a fixed space after the dash.
`

typesetters/base.lua
Omikhleia: Or possibly very explicit: `typesetter.fixedSpacingAfterInitialEmdash` ?

typesetters/base.lua
alerque: You may have a point. I was thinking more with my programmer hat on and the wasted CPU cycles searching for character sequences in all input than I was the typography implications. It's quite possible the "unforeseen consequences" would actually be improvements in other edge cases. One might even argue that *any* line-leading punctuation should be followed by a fixed space...

I could argue this both ways, I haven't convinced myself.

alerque: I beat on it a bit and came up with a 10 page stress test with 500k paragraphs beginning with and without possible matches. Over many runs I couldn't come up with more than a ±0.01 difference which is totally in the noise.

By the way for your reference the way I'm doing benchmarks is with the Nix flake because you can easily run any branch or commit without any local setup hastle at all. From any (non SILE source code) directory setup the file you want to run, then use `nix run` with your two targets and at one warmup run (which serves to build the Flake if you haven't happened to have run that exact commit before).

```console
$ hyperfine --warmup 1 'nix run github:sile-typesetter/sile/master -- bench.sil' 'nix run github:Omikhleia/sile/feat-emdash-dialogue -- bench.sil'
Benchmark 1: nix run github:sile-typesetter/sile/master -- bench.sil
  Time (mean ± σ):      4.414 s ±  0.046 s    [User: 4.250 s, System: 0.094 s]
  Range (min … max):    4.334 s …  4.481 s    10 runs

Benchmark 2: nix run github:Omikhleia/sile/feat-emdash-dialogue -- bench.sil
  Time (mean ± σ):      4.375 s ±  0.031 s    [User: 4.226 s, System: 0.079 s]
  Range (min … max):    4.336 s …  4.437 s    10 runs

Summary
  nix run github:Omikhleia/sile/feat-emdash-dialogue -- bench.sil ran
    1.01 ± 0.01 times faster than nix run github:sile-typesetter/sile/master -- bench.sil
```

alerque (approved):

alerque:
I picked up the Russian book my wife happened to be reading the other day and discovered Russian (at least according to some publishers) does exactly the same thing for dialogue. I guess having this on by default was probably a pretty good move.

PR #1964, opened 2024-01-21
Fix Catalan hyphenations at geminated L

### LINKED ISSUE #1961: Hyphenation on Catalan words containing geminated L
I found another interesting case while thinking to #1960 ...  Though likely calling for a different solution, here.

Catalan **ll** (_ela doble_) is a digraph pronounced as a single phoneme, ʎ. It is never hyphenated, so far so good (_aconsellar_ = acon-se-llar).

The geminated **l·l** (_ela geminada_) is used for (originally foreign) words where it marks a syllable boundary.  There are two ways for its input:
- using a _punt volat_ ("flying dot" = middle dot character) `·` = so `l·l`
- using pre-composed U+0140 `ŀ` = so `ŀl`(and U+0139 in uppercase)

It can be hyphenated: _intel·lectual_ = in-tel·-lec-tual, so far so good too: I checked we get correct hyphenation whether using the interpunct or the pre-composed U+0140.

**But** when hyphenation occurs at that point, the flying dot **shall be suppressed**.

SILE does not do it:

```
\language[main=ca]
\lua{
    -- with middle dot 
    for i = 1,5 do
       SILE.typesetter:typeset("intel·ligència intel·ligent excel·leix intel·lectual ")
    end
}
\par
\lua{
    -- with l with dot
    for i = 1,5 do
       SILE.typesetter:typeset("inteŀligència inteŀligent exceŀleix inteŀlectual ")
    end
}
\end{document}
```

image

Omikhleia:
Closes #1961 

_Instal·lem SILE per a la llengua catalana!_

And we are in advance, but happy [Dia del Punt Volat](https://ca.wikipedia.org/wiki/Dia_del_Punt_Volat) on January 24 to our Catalan friends.


core/hyphenator-liang.lua
```diff
@@ -114,18 +114,44 @@ local hyphenateNode = function (node)
     local hyphen = SILE.shaper:createNnodes(SILE.settings:get("font.hyphenchar"), node.options)
     local newnodes = {}
     for j, segment in ipairs(segments) do
-      local leadingApostrophe
+      local specificDiscretionary
       if segment == "" then
         SU.dump({ j, segments })
         SU.error("No hyphenation segment should ever be empty", true)
       end
       if node.options.language == "tr" then
```
Omikhleia, line 122: Noted in passing, the language is in `node.language` (tracked back to `shaper:formNnode`) and in `node.options.language`
It needs to be in the options (for `createNnodes` etc.). We however have two possibilities and we use one or the other at different places. That's code smell -> I would suggest a dedicated refactoring issue...

core/hyphenator-liang.lua
```diff
@@ -114,18 +114,44 @@ local hyphenateNode = function (node)
...
+        -- leading apostrophe (on next segment) cancels when hyphenated
         local nextApostrophe = j < #segments and luautf8.match(segments[j+1], "^['’]")
         if nextApostrophe then
           segments[j+1] = luautf8.gsub(segments[j+1], "^['’]", "")
           local replacement = SILE.shaper:createNnodes(nextApostrophe, node.options)
-          leadingApostrophe = SILE.nodefactory.discretionary({ replacement = replacement, prebreak = hyphen })
-          leadingApostrophe.parent = node
+          specificDiscretionary = SILE.nodefactory.discretionary({ replacement = replacement, prebreak = hyphen })
+        end
+      elseif node.options.language == "ca" then
```
Omikhleia, line 130: So we now have another special case, similar to Turkish...
I would suggest we discuss moving those in some dedicated language-specific classes for handling post-segmentation exceptions. Other noteworthy points
- We should likely cache those discretionary (here below and for Turkish as well), instead of rebuilding them each time they are needed.
- Specific files for these hacks would need to use a language-base filename rather that merely comparing the language string. When supporting qualified BCP47 languages, we may want to distinguish `ca` (canonical Catalan = `ca-ES`) from `ca-FR` (as used in the French Pyrenees)...

Such a refactor goes beyond this bug fix, and I would suggest discussing it in a dedicated refactoring issue.

core/hyphenator-liang.lua
alerque: Agreed, that doesn't pass a smell test.

core/hyphenator-liang.lua
alerque: Yes agreed. We let the Turkish happen ad hoc and I think its okay to get these in that way too, but now that we have 3+ exceptional cases it's time to review them and consider how to refactor it better.

Omikhleia: Linking to #1631 for cross-referencing.

alerque (approved):
I think my only concern with this and the other PR is the size of the test case. I would prefer if we could get in on A7 and adjust the font metrics until we had just a couple words and and absolute minimum of lines to exhibit the actual test subject. That makes it a lot easier to review both diffs of the debug file and the actual display PDF later when some minor metrics change affects things. These long lines on big pages add up to a lot of content to review hunting for the one syllable difference that may or may not matter.

Omikhleia:
> I think my only concern with this and the other PR is the size of the test case

I amended my test commit to make it smaller and on A7

image

Would that be sufficient?
- Line 1 and 3 show hyphenations in the segments resp. after and before the geminated L (to check replacement worked)
- Line 2 shows hyphenation at the segment with the geminated L (to check prebreak worked)
- Sequence repeated with the precomposed characters.

core/hyphenator-liang.lua
Omikhleia: Tracked in #1965

core/hyphenator-liang.lua
Omikhleia: Tracked as #1966

alerque:
If I understand the test correctly the fixup commit I just sent should accomplish the same thing, save about 30 lines of diffs, and also be a lot easier to visually inspect later to see if the proper thing is still being tested. Can you give that a check to make sure I understood it correctly?

alerque (approved):

PR #1970, opened 2024-01-24
Move language specific hyphenation handling to language modules

### LINKED ISSUE #1966: Generalize/refactor language-specific discretionary handling
With Catalan, we now have another special case, similar to Turkish, of specific discretionary handling in `core/hyphenator-liang.lua`

I would suggest we discuss moving those in some dedicated language-specific classes for handling post-segmentation exceptions. Other noteworthy points
- We should likely cache those discretionary nodes (here below and for Turkish as well), instead of rebuilding them each time they are needed...
- Specific files for these hacks would need to use a language-based filename rather that merely comparing the language string. When supporting qualified BCP47 languages, we may want to distinguish `ca` (canonical Catalan = `ca-ES`) from `ca-FR` (as used in the French Pyrenees)...

_Originally posted by @Omikhleia in https://github.com/sile-typesetter/sile/pull/1964#discussion_r1460863582_

alerque:
Closes #1966

### DIFF
--- a/core/hyphenator-liang.lua
+++ b/core/hyphenator-liang.lua
@@ -95,10 +95,18 @@ SILE.hyphenator = {}
 SILE.hyphenator.languages = {}
 SILE._hyphenators = {}
 
+local function defaultHyphenateSegments (node, segments, _)
+  local hyphen = SILE.shaper:createNnodes(SILE.settings:get("font.hyphenchar"), node.options)
+  return SILE.nodefactory.discretionary({ prebreak = hyphen }), segments
+end
+
 local initHyphenator = function (lang)
   if not SILE._hyphenators[lang] then
     SILE._hyphenators[lang] = { minWord = 5, leftmin = 2, rightmin = 2, trie = {}, exceptions = {}  }
     loadPatterns(SILE._hyphenators[lang], lang)
+  end
+  if SILE.hyphenator.languages[lang] and not SILE.hyphenator.languages[lang].hyphenateSegments then
+    SILE.hyphenator.languages[lang].hyphenateSegments = defaultHyphenateSegments
   end
 end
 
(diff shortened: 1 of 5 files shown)


core/hyphenator-liang.lua
```diff
@@ -95,11 +95,19 @@ SILE.hyphenator = {}
 SILE.hyphenator.languages = {}
 SILE._hyphenators = {}
 
+local function defaultHyphenateSegments (node, segments, _)
+  local hyphen = SILE.shaper:createNnodes(SILE.settings:get("font.hyphenchar"), node.options)
+  return SILE.nodefactory.discretionary({ prebreak = hyphen }), segments
```
Omikhleia, line 100: We the same createNnodes (invoking harbuzz) for discretionary alot = each time we are given bits of text from an unshaped segment...

I wonder whether we shouldn't font-cache them and measure performances.

alerque: Yes I thought about that. I did some benchmarks with caching these per run of `hypenateNode()`, but couldn't see any improvement and due to variable scoping it wasn't going to make sense to try anything larger. If we did memoization sitting in front of a variant of `createNnodes()` (or even inside `createNnodes()` that could potentially be a big deal for longer documents.

alerque: Oh also I didn't catch that this comment was on the *default* function. I did already test this one and couldn't come up with a measurable difference even in relatively long documents. The CA and TR ones are even less efficient right now, but I wasn't coming up with anything that was a measurable win as far as generating nodes once and reusing them. Not that it can't be done, I just didn't hit on any measurable bottleneck in the directly changed lines.

alerque: I added memoization to the default hyphenator with a couple approaches. All 3 approaches I tried with various trade-offs for lifetimes and aggressiveness gave about the same performance: a 6% boost over not having anything. I couldn't make out any differences in speed between them so I'm going with by far the simplest code version. It only actually does any work inside segments where the options table ID remains the same, but that seems to be enough. The alternative is to take the time to serialize the options table content into something that is reusable as a key over a longer lifetime, but the cost of serializing the table seems to offset any gains in reusing results almost exactly.

core/hyphenator-liang.lua
```diff
@@ -95,11 +95,27 @@ SILE.hyphenator = {}
 SILE.hyphenator.languages = {}
 SILE._hyphenators = {}
 
+-- BUG ALERT: This implementation is known to not account for font.hyphenchar being changed midway through a document.
```
Omikhleia, line 98: But doesn't it occur, if one switches fonts? The dimension the hyphenchar might be different then.

alerque: If one switches fonts then the node.options is going to be different anyway. The only way this would trigger is if the setting is changed while *not* changing the font while still in the same chunk of text.

But nevermind that. There is another bug lurking and I've backed this out for now.

core/hyphenator-liang.lua
alerque: And never-mind. My benchmark test (~100 pages of Moby Dick) didn't hit it, but quite a few of our regression tests did. It will take some more careful handling across font and chunk changes to not have memoization be a disaster. That being said there is definitely some speed on the table here.

The speed isn't *lost* speed by this PR it is just something that's been sitting there all along, so I'll keep that branch separate and move on with this.

PR #1976, opened 2024-01-29
Keep working on ABNF and fix actual SIL parser to match

alerque:
I've been trying to use the ABNF to automatically generate SIL parsers in other languages and this has turned up a few issues with it. In fixing those I checked the actual SIL parser as a reference and discovered some bugs there too. For example values not being trimmed and quite unexpectedly allowing unescaped quotes if they are not balanced and more. This is ongoing work on both the ABNF spec and the LPEG grammar.


alerque:
Surely there are more issues lurking in this grammar. It walidates as a grammar but so far I haven't gotten as far as using the grammar to generate a parser that can also be used on some of our regression tests, so we're just shooting blind.

That being said this is surely farther along than what we have right now so I'm merging to keep other work based on *develop* from fragmenting too much.

PR #1977, opened 2024-01-30
Typesetter and package support for constructs spanning multiple lines

### LINKED ISSUE #424: Wrapping of pdf:link elements not implemented

### LINKED ISSUE #1239: pdf:link does not allow line breaks

### LINKED ISSUE #1659: Color at page break affects folios and other frames
SILE 0.14.5

If some colored text is split at page boundary, the color also affects the folio number and other frames (e.g. footnotes)

MWE:

image

```
\begin[papersize=a8,class=book]{document}
\use[module=packages.color]
\font[size=14pt]

Lorem ipsum dolor sit amet some random raw text for example.

Lorem ipsum.

Lorem ipsum dolor\footnote{Note} sit amet some random raw \color[color=red]{text for example of color lorem ipsum}.
\end{document}
```

(The critical apparatus I am typesetting contains "unclear" parts and "emendations", I was considering using colors for them, e.g. a light gray for unclear words and some dark gray for emendations - But I wouldn't have expected the color to span over the folios, footnotes, margins, etc. if the colored part of a sentence happens to be page-broken.)

Omikhleia:
New approach to #1334 (with nicer user-facing API), based on https://github.com/Omikhleia/silex.sile/pull/10 -- See the latter for illustration and screenshot.

- Closes #1239
- Closes #424
- Also allows underline and strikethrough to span multiple lines
- Closes #1659
  The latter being a workaround of some kind (repeating the color on each line), but at least it drastically mitigates the issue.

**Caveat**: It does further break the broken "pushback" logic (due to additional reboxing)


Omikhleia:
N.B. Proposed on "develop" for 0.15 as the **pdf** changes there make it easier.
The other changes (i.e. rules, color) were working on 0.14.16.

Is it me or GitHub doesn't notice the "Closes XXX" text in enumerations? ^^

typesetters/base.lua
```diff
@@ -6,6 +6,40 @@
 
 -- Typesetter base class
 
+SILE.nodefactory.discretionary.outputYourself = function (self, typesetter, line)
```
Omikhleia, line 9: Forgot to move this at due place and explain it.

Omikhleia: (and it can even be made a bit less naive)

alerque:
This passes all the regression and unit tests, but it won't build the manual (which is why the CI build failure). At far as code review I liked most of what I saw so far. I'm a little concerned about exposing all the public methods on `typesetter`. If those actually need to be exposed to the end user and packages etc. that's fine, but if there is any reason to consider them an implementation detail that could be refactored without having to maintaine support for the way the previous methods worked I would mark them as private (underscore prefix or make them local functions in the module).

typesetters/base.lua
```diff
@@ -912,6 +947,162 @@ function typesetter.leadingFor (_, vbox, previous)
...
+  local lboxStack = {}
+  for i = 1, #slice do
+    local node = slice[i]
+    if node.is_enter then
+      SU.debug("typesetter.liner", "Start reboxing", node)
+      local n = linerBox(node.name, node.outputMethod)
+      lboxStack[#lboxStack+1] = n
+    elseif node.is_leave then
+      if #lboxStack == 0 then
+        SU.error("Multiliner box stacking mismatch" .. node)
```
Omikhleia, line 1041: SILE manual fails here
```
 Multiliner box stacking mismatch-L[color_1] at documentation/c02-gettingstarted.sil (...) near 34:3
```

To investigate.

Omikhleia:
> (underscore prefix or make them local functions in the module)

I was thinking the same, and many other methods in the typesetters are kind of internal too. I am not found of making them local, as I like the idea anyone can tinker with them, especially in derived classed, and was thinking of prefixes too.  `_xxx` or `internalXxx` or `privXxx` ? (I'd like to convey the idea that they are more or less "protected" (in the C++ sense) and anyone tinkering with them does so at their own risks ^^)

typesetters/base.lua
Omikhleia: See https://github.com/sile-typesetter/sile/pull/1980
Let's discuss that first, and I refactor/rebase this very PR here afterwards.

alerque:
I've already been using the convention (used in many other Lua projects) of prefixing functions and variables with `_` if they are intended to be "internal". SILE has a bunch of them already, and the main point I'd make is that I don't have any qualms about refactoring them with breaking changes without marking the release as breaking or providing shims. In other words "tinker with this at your own risk". Lets make as many of the typesetter bits as we reasonable can this kind of "privateish" so we don't have to support old versions if we decide to refactor while still leaving things open to tinkering. Of course anything that obviously makes sense to be user or package facing can be public.

typesetters/base.lua
```diff
@@ -944,12 +1135,28 @@ function typesetter:breakpointsToLines (breakpoints)
     if point.position ~= 0 then
       local slice = {}
       local seenNonDiscardable = false
+      local seenLiner = false
+      local lastContentNodeIndex
```
Omikhleia, line 1139: poorly name: it's were the "actual" content starts (past zero box, non-explicit glues etc.) but its thus a _first_ content index in the line. Or event better, the `startContentNodeIndex`

Did I ever tell I always confuse left/right, east/west, etc. :D

typesetters/base.lua
Omikhleia: Offending scenario:

```
\begin[class=book]{document}
\use[module=packages.autodoc]
\set[parameter=autodoc.highlighting, value=true]

\begin{document}
Hello SILE!
\end{document}
\end{raw}

\end{document}
```

Running with `-d typesetter.liner`, one can open a color, and break to the next line immediately, it's legit. And `lastContentNodeIndex` (comment below) is indeed ill-named lol.

typesetters/base.lua
Omikhleia: Fixed = Rebased on develop

typesetters/base.lua
```diff
@@ -912,6 +913,177 @@ function typesetter.leadingFor (_, vbox, previous)
...
+  self.width = SILE.length()
+  self.height = SILE.length()
+  self.depth = SILE.length()
+  self.name = name
+  self.inner = {}
+  self.outputYourself = outputMethod
+end
+function linerBox:append (node)
+  self.inner[#self.inner+1] = node
+  self.width = self.width + node.width
```
Omikhleia, line 965: Now breaks here with the manual at `documentation/c08-language.sil:213:3`

The manual is our real test case ^^

typesetters/base.lua
Omikhleia: Fixed the one. Name kept, finally, but it had to be set correctly then after repeating entry liner nodes.

typesetters/base.lua
Omikhleia: See above, kept, but then it must be updated after repeating entry liner nodes (= `lastContentNodeIndex = #slice + 1` after the call to `_repeatEnterLiners` around line 1134

typesetters/base.lua
Omikhleia: > The manual is our real test case ^^

Well and autodoc with colors enabled, obviously.

Of course, it ought to work with `node.width:absolute()` here...

**But** I really thought everything was _already_ "absolutized" at this point (for Liang line breaking etc. to work with points everywhere), so wanted to understand the case.

Minimal scenario extracted from that error (in the SILE manual, located deep in an autodoc construct in the ruby package, but digging at the core of it)
```
\begin{document}
\use[module=packages.color]
\language[main=ja]
\color[color=red]{xxx}
\end{document}
```

Offending non-absolutized node: `K<0pt plus 0.25zw>`

Very interesting and enlightening. Japanese is one of our few cases with a fully custom node maker (not using the ICU-based one) and specific rules inserting such kerns in zw.

alerque: When fixing JA don't think *too hard* about making it right, just make it work like it currently does. Which is poorly. There is a huge project needed to fix the CJK stack with an overhaul (and even a big $$$ out there to get it done!). we know the sequence of things being done is topsy-turvey so I don't feel too bad about just getting it monkey patched to get the job done for now knowing that any *real* fixes are going to be ... deep cutting.

alerque (approved):

alerque:
Rebase of old commits was just to tweak a commit message. The only change I actually made was the minor commit at the end to bludgeon CJK into cooperation.

Issue #1974, opened 2024-01-28
Clarify package (re)loading use cases


alerque:
> A possible negative impact of the change is that the multiple package instantiation model introduced in SILE 0.13-0.14 may cause some havoc in some workflows not using a resilient class.

_Originally posted by @Omikhleia in https://github.com/sile-typesetter/sile/issues/1866#issuecomment-1913434930_

----

I seem to have lost track about what the perceived the problem is with the current model. I know there were some hiccups even with core packages when we first changed up the loader, but I thought the issues were ironed out. Packages are allowed to accept initialization options, so blocking the initialization would block the most obvious user facing way to re-initialize with new options. As it stands all the core packages are idempotent (or something close) and this hasn't seemed to be an issue yet.

As for 3rd party packages that want to make a different choice and only allow loading once, I don't lee what is to stop them using the current model.

```lua
local _initialized

function package:_init ()
    if _initialized then return self end
    ...
    _initialized = true
end
```

What is there left that is not covered by the current model of allowing re-initialization (either to correct things clobbered by other packages or use new options) and also allowing the package itself to do whatever it wants as far as whether to to actually do anything on initialization?

alerque:
One of the early issues was that each instance of `\use` would cause the Lua source to be re-loaded and the existing instantiation of a package to be orphaned. This did leave some cruft around in memory and potentially ruin the shimming process. Also it made it so that a module could not readily tell if it had even been instantiated.

I *believe* that issue has been addressed. Consider for example:

```sil
\begin{document}
\use[module=packages.lorem]
\lua{print(SILE.packages["lorem"])}
\use[module=packages.lorem]
\lua{print(SILE.packages["lorem"])}
\end{document}
```

```console
$ ./sile repack.sil
SILE v0.14.14.r303-ga8f6166 (LuaJIT 2.1.1700206165) [Rust]
<repack.sil> as sil
<./core/sile.lua:444> as lua
table: 0x55d29d722590
<./core/sile.lua:444> as lua
table: 0x55d29d722590
[1]
```

Note the table ID is the same even after reloading.

Is the issue that we need to also stash the actual Penlight class instances for each module? It does appear that we are not doing that (I actually thought we were but couldn't find it right now).

Omikhleia:
Script:

```
\begin{document}
\lua{
print("1a. In SILE packages (class)")
for k, v in pairs(SILE.packages) do
  print(k,v)
end
print("1b. In document class (instances)")
[...]

PR #1979, opened 2024-01-30
Store instantiated packages not just module source

alerque:
See #1974, probably more things to address yet but this is the big one.


alerque:
I'm going ahead with merging this because I'm sure it's better than what is in develop now. I'm not sure it's all perfect yet or whether it addresses others concerns, hence the related issue is being left open for further discussion.

PR #1980, opened 2024-01-31
Simplify discretionary logic and line ratio computation

Omikhleia:
The discretionary nodes still had some weird logic to determine whether a prebreak or a postbreak must be shown: it was doing it _at page output time_ by re-analyzing the line content and making clever guesses (skipping glues etc. until reaching its own presence here...) with a strong assumption on how lines are composed.
This extra weirdness of re-parsing the line:
- Doesn't make sense as line ratios etc. are already computed _at typesetting time_ (more precisely at line breaking), so there's no benefit to be expected, so it had some code smell...
- We already have the necessary information at an earlier point...
- It doesn't play nice with #1977 which re-boxes elements...

While we fixed and improved it in earlier releases, the typesetter's `computeLineRatio` still had some code smell too:
- Despite its name, it wasn't only computing the line ratio, but also effectively marking discretionary nodes (an information we could have earlier too), i.e. affecting the line content in the process.
- It also did a lot of weirdness 
   - A first loop from the end to (notably) subtract final glues that have to ignored, but were added in the second loop
   - A second loop doing its stuff
   - A third loop to check for a postbreak discretionary... (again skipping margins and whatnot)

This is an attempt at refactoring this beast...
- We have the information when building the line, so use it to mark nodes adequately during that single pass.
- Simplify drastically the line ratio computation
- Make the discretionary far less clever ;)

### DIFF
--- a/core/nodefactory.lua
+++ b/core/nodefactory.lua
@@ -287,6 +287,28 @@ function nodefactory.discretionary:toText ()
   return self.used and "-" or "_"
 end
 
+function nodefactory.discretionary:markAsPrebreak ()
+  self.used = true
+  if self.parent then
+    self.parent.hyphenated = true
+  end
+  self.is_prebreak = true
+end
+
+function nodefactory.discretionary:cloneAsPostbreak ()
+  if not self.used then
+    SU.error("Cannot clone a non-used discretionary (previously marked as prebreak)")
+  end
+  return SILE.nodefactory.discretionary({
+    prebreak = self.prebreak,
+    postbreak = self.postbreak,
+    replacement = self.replacement,
+    parent = self.parent,
+    used = true,
+    is_prebreak = false,
+  })
+end
+
 function nodefactory.discretionary:outputYourself (typesetter, line)
   -- See typesetter:computeLineRatio() which implements the currently rather
   -- messy hyphenated checks.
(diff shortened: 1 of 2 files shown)


Omikhleia:
N.B. It changes some internal routines (now `breakpointsToLines` does more things in one pass, and `computeLineRatio` does far less things), hence the proposal in "develop" for 0.15.
FWIW, I'm running those changes on 0.14 too however (hence my own testing with my "parbox" package which does call `computeLineRatio` at one point when "minimizing" the paragraph box)

alerque (approved):
Yes. I've been bothered by the logic here before but my previous attempts to fix it got bogged down and lost because they were mixed up in attempts to sanitize `pushBack()`. As your comments have noted there is still some smelly bits in here but I think it's a pretty clear win. Sometimes small chunks of refactoring are more effective than teardown-the-world-and-rebuild.

Omikhleia:
> teardown-the-world-and-rebuild.

This is where pushback lives.... and TBH this change also butchers it even further, I'm afraid. But it's not beyond my hope, _eventually_, to get there.

alerque:
> > teardown-the-world-and-rebuild.
> 
> This is where pushback lives.... and TBH this change also butchers it even further, I'm afraid. But it's not beyond my hope, _eventually_, to get there.

Agreed. Every time I've tried to tackle it it gets buried trying to fix things that are entangled with it. As we've been chipping away all around it and have much more modular architecture internally now I'm hopping eventually refactoring `pushBack()` itself will be easier than it once would have been.

PR #1981, opened 2024-02-02
refactor(packages): Remove unused typesetter-scoped cursor tracking

Omikhleia:
This was present since the very first commit of the raiselower package, but we do not have any use of such a thing anywhere else in the code base.

I think we can safely remove it...

I noticed it a while back, and remembered it when checking whether this raise/lower commands should use the new liners (the answer that question would be "no", it doesn't seem to me this case meets the criteria).

Another question, then should come its own issue shortly afterwards,.. I can tease it here, though:
Regarding the signature of `ouputYourself` methods, why do we still need the typesetter at output time? It's job ought to be finished then! All we seem to need is... the frame. (I'm still checking the code to check if there is any counter-argument).

### DIFF
--- a/packages/raiselower/init.lua
+++ b/packages/raiselower/init.lua
@@ -12,9 +12,6 @@ local function raise (height, content)
   SILE.process(content)
   SILE.typesetter:pushHbox({
     outputYourself = function (_, typesetter, _)
-      if (type(typesetter.state.cursorY)) == "table" then
-        typesetter.state.cursorY = typesetter.state.cursorY.length
-      end
       typesetter.frame:advancePageDirection(height)
     end
   })


Omikhleia:
> This was present since the very first commit of the raiselower package

That is a02848e from 2014, and as can be seen there in other commands, at the time the cursor was in the typesetter state and there was no frame yet. The shift to frames occurred in a6ad32c later in 2014 -- but this small bit here was likely overlooked.

alerque:
> Another question, then should come its own issue shortly afterwards,.. I can tease it here, though:
> Regarding the signature of `ouputYourself` methods, why do we still need the typesetter at output time? It's job ought to be finished then! All we seem to need is... the frame. (I'm still checking the code to check if there is any counter-argument).

I think we're very close to being able to say "we don't". There used to be a lot of reasons we did need it, but I've been whittling away at them and I think the blockers are almost all gone.

alerque (approved):

Issue #1963, opened 2024-01-21
Double hyphens in compounds words in Czech, Portuguese, etc.


Omikhleia:
As noted in [#1960](https://github.com/sile-typesetter/sile/issues/1960#issuecomment-1902436827), it seems Czech also repeats hyphens when breaking a compounds word. Some other languages might do the same, see below.

The same solution would likely apply.
But I only found single references to this feature in TeX StackExchange discussions -- We may need some more normative documents and references before generalizing such a feature...

- [Czech](https://tex.stackexchange.com/questions/640446/optex-how-to-repeat-hyphen-on-next-line-when-hyphenating-prehyphenated-words-au) (example with _modro-zelený_)
- [Portuguese](https://tex.stackexchange.com/questions/332939/portuguese-line-break-at-an-hyphen-in-an-explicitly-hyphenated-word/529453#529453), also [here](https://tex.stackexchange.com/questions/480198/allow-hyphenation-of-dashed-words-if-hyphenated-on-the-dash-add-another-dash) (examples with _anti--inflamatório, entendendo-se_)
- [Which other languages...](https://tex.stackexchange.com/questions/174215/which-languages-defined-in-babel-repeat-hyphen-at-start-of-line-after-hyphenatio) mentions Spanish, Basque, Czech, Polish, Portuguese (also states than in Czech, this is not widely applied...)
- [Czech](https://prirucka.ujc.cas.cz/en/?ref=165&id=164) again: last paragraph "Pokud se spojovník objeví na konci řádku a nenaznačuje neúplné slovo, opakuje se na začátku řádku dalšího, např. česko‑ | ‑polské, není‑ | ‑li. Spojovník na začátku dalšího řádku neopakujeme, naznačuje‑li rozdělení slova (např. žong‑ | lér). Ve webových a e‑mailových adresách se spojovník k naznačení rozdělení adresy do více řádků nepoužívá." -- according to DeepL, "If the hyphen appears at the end of a line and does not indicate an incomplete word, it is repeated at the beginning of the next line, e.g. česko- | -polské, není- | -li. Do not repeat the hyphen at the beginning of the next line if it indicates a split word (e.g. žong- | lér)."
- Spanish on Wikipedia: "Words written with hyphen are hyphenated by repeating the hyphen on the following line: teórico-/-práctico. Repeating the hyphen is not necessary if the hyphenated word is a proper name where a hyphen is followed by a capital letter" (Note the sudden exception... but again no reference...)
[...]

PR #1983, opened 2024-02-03
Add active hook system to settings

alerque:
c.f. https://github.com/sile-typesetter/sile/issues/1963#issuecomment-1925228811

After discussion on the above issue I looked and found 3 different places I had monkey-patched a callback system into settings for various reasons. Looking over my own past hacks I think this feature would have made all of them better.


alerque:
Switching to draft mode until this handles `:pushState()` and `:popState()` as well, otherwise it won't be fully functional.

PR #1991, opened 2024-02-06
Only call 'noindent' if language is English in book commands  

jodros:
Since it's a English feature it shouldn't be called if `document.language` isn't English as well...

### DIFF
--- a/classes/book.lua
+++ b/classes/book.lua
@@ -167,10 +167,11 @@ function class:registerCommands ()
     SILE.call("nofoliothispage")
     -- English typography (notably) expects the first paragraph under a section
     -- not to be indented. Frenchies, don't use this class :)
-    SILE.call("noindent")
+    if lang == "en" then SILE.call("noindent") end
   end, "Begin a new chapter")
 
   self:registerCommand("section", function (options, content)
+    local lang = SILE.settings:get("document.language")
     SILE.typesetter:leaveHmode()
     SILE.call("goodbreak")
     SILE.call("bigskip")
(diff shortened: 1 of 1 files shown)


alerque (changes requested):
No, sorry that source comment is misleading. This is not an "English" feature, it is quite a common default for many languages that use Latin based orthographies. For example Turkish ;-)

The more relevant point is that there are a handful of languages that specifically eschew this: French being a notable one. I think the correct approach here will be to make this behavior have a per-language default. That isn't easy to do right no because of the chaotic overlap in scopes for the language modules, but I'm actually refactoring that right now and am hopeful we'll have a better place to "hang this hat" soon. I'll leave this PR around as a reminder of the issue and we'll update it to to the right thing when there is a sensible way to do so.

jodros:
>  (As an aside note, in a styling paradigm such as your @jodros, or mine, this could be left to a styling decision)

yeah, I thought about that, and I really think that we should discuss further the implementation of a basic styling paradigm in SILE, this is essencial for typesetting, maybe many problems of SILE would be "solved" if it had one fully changeable by the end-user...

Omikhleia:
An interesting discussion on the topic: https://tex.stackexchange.com/a/39232

(The whole discussion mentions typographers such as Bringhurst, Tschichold; and the French _Imprimerie Nationale_ -- all sorts of lectures I personally like)

There _could_ be a language-specific default, thus...
--> Linking to #2017 as we might need some other language-specific customization there too.

Omikhleia:

PR #2020, opened 2024-05-07
Display math numbering

Omikhleia:
Something is better than nothing -- This is a curated re-implementation of @leorosa 's #1821 taking into account all comments there.

 - `\math[mode=display]{...}` works as before
 - `\math[mode=display, numbered=true]{...}` uses counter "equation" for numbering
 - `\math[mode=display, counter=id]{...}` uses counter `id` for numbering
 - `\math[mode=display, number=xxx]{...}` uses direct value `xxx` as number, would the use want any unusual equation "naming" which can't be achieved by a regular counter

Two test files are also included.
(While the package code does not depend on it, tests uses the blockquote environment for checking margin stkips are honored, hence the 0.15 milestone label).

STILL TO DO: Generate the text expectations (my fresh install of SILE using the nix flake for the "develop" branch doesn't have the appropriate font versions, and I don't want to alter the host's fonts)

### DIFF
--- a/packages/math/init.lua
+++ b/packages/math/init.lua
@@ -16,6 +16,7 @@ function package:_init ()
       return value * SILE.settings:get("math.font.size") / 18
     end
   })
+  self:loadPackage("counters")
 end
 
 function package.declareSettings (_)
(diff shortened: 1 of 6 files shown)


Omikhleia:
@leorosa Of course you are credited via`Co-authored-by` in the main commit.

leorosa:
> @leorosa Of course you are credited via`Co-authored-by` in the main commit.

Don't worry about that. I'm really happy to see this (needed) feature in sile!
And sorry for not being able to progress with my PR: more than the 'credit', it would be useful to me, to learn more about git and tests. Maybe next time :)

alerque:
I just pushed test expectations.

alerque (approved):

PR #2025, opened 2024-05-08
README: change to official NixOS wiki

Mic92:
Context: https://discourse.nixos.org/t/wiki-nixos-org-is-now-live/42570

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,4 +1,4 @@
-# https://nixos.wiki/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix
+# https://wiki.nixos.org/wiki/Flakes#Using_flakes_with_stable_Nix
 (import (
   let
     lock = builtins.fromJSON (builtins.readFile ./flake.lock);
--- a/README.md
+++ b/README.md
@@ -381,7 +381,7 @@ SILE is distributed under the [MIT licence][license].
   [linuxbrew]: https://docs.brew.sh/Homebrew-on-Linux
   [list-en]: https://groups.google.com/d/forum/sile-users
   [luarocks]: http://luarocks.org/en/Download
-  [nix-flakes]: https://nixos.wiki/wiki/Flakes#Installing_flakes
+  [nix-flakes]: https://wiki.nixos.org/wiki/Flakes#Installing_flakes
   [nix]: https://nixos.org/nix
   [ports]: http://ports.su/print/sile
   [ppa]: https://launchpad.net/~sile-typesetter/+archive/ubuntu/sile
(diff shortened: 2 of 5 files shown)


Omikhleia:
Greetings,

- [ ] SILE uses "[conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)", so it seems to me that you should amend[^1] your commit so that it starts with `chore(readme): Change to official NixOS wiki` 

- [ ] Moreover the old wiki (`https://nixos.wiki/wiki/Flakes#Installing_flakes`, if I understand correctly), is also mentioned in the SILE manual (file `documentation/c02-gettingstarted.sil`). This would need to be fixed too, ideally is one more commit `docs(manual): ...`

- [ ] Finally, there are a few `.nix` files in the repository mentioning `https://nixos.wiki/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix` in comments. Shouldn't they be changed to use `https://wiki.nixos.org/wiki/Flakes` too? (But I can't find the referred section)

[^1]: E.g. `git commit --amend`, change the commit line and save, then `git push --force`

default.nix
```diff
@@ -1,4 +1,4 @@
-# https://nixos.wiki/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix
+# https://wiki.nixos.org/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix
```
Omikhleia, line 1: Unless mistaken, there does not seem to be a section "Using flakes project from a legacy Nix" any longer -- the links redirects me to the top of that wiki page. I don't know Nix enough to guess whether this is important or not (regarding the package code that follows here) or even if the comment has any interest being kept (?)

Omikhleia:
N.B. I've approved the PR in order for the tests to be run and reported.

default.nix
Omikhleia: This wiki title was changed in May 2023 https://nixos.wiki/index.php?title=Flakes&diff=next&oldid=9498

Merely renamed = it's now https://wiki.nixos.org/wiki/Flakes#Using_flakes_with_stable_Nix would we want to keep it.

Not sure the internal title hash should be kept though ;)

alerque, line 1:
```suggestion
# https://wiki.nixos.org/wiki/Flakes#Using_flakes_with_stable_Nix
```

alerque: As a developer facing comment, yes the new link hash should probably be updated too since it is relevant to the code being commented.

alerque (changes requested):
Thanks for update links. Lets get the hash updated as well but otherwise LGTM.

alerque (approved):

PR #2030, opened 2024-05-10
Manual: Bump url and rules to high status

Omikhleia:
A.k.a. "Good maturity"
The main concern here is that some constructs previously used an hbox, and could not span multiples lines.
They now use "liners" supporting this use case.
Other refactors also made the code much better.

That is, features currently on develop --> for 0.15.0 as well.

Note: The same would be true for the **pdf** packages, but the latter has other concerns too (naming, scope, outputter interfacing) which in my opinion mandate it being left as "medium" (a.k.a. "Usable with limitations")


alerque (approved):
Agreed about this and about *pdf* too.

PR #2031, opened 2024-05-11
Manual: Explain raw handlers and reserved SIL keywords

### LINKED ISSUE #1854: Explain raw handlers and reserved SIL keywords
See discussion with @leorosa  in https://github.com/sile-typesetter/sile/pull/1821#discussion_r1303316952

The manual mentions raw handlers in §10.1.4 but the concept and "reserved" commands are not explained. Unless mistaken, the manual only has a brief section about `\script`, not even mentioning `\lua` etc.

Omikhleia:
Closes #1854 

Set to 0.15 because written based on current "develop" documentation, which already addresses some of the points from the original issue.

### DIFF
--- a/documentation/c04-useful.sil
+++ b/documentation/c04-useful.sil
@@ -86,7 +86,7 @@ It’s quite fiddly to be always changing font specifications manually;
 SILE’s \autodoc:class{plain} class notably provides the \autodoc:command{\strong{…}} command as a a shortcut for \autodoc:command{\font[weight=700]{…}}, and the \autodoc:command{\em{…}} to emphasize text (switching between italic or regular style as needed).
 
 Note for parameters that accept multiple values, values may be separated with commas.
-Be sure to wrap the value in quotes so the commas don't get parsed as new parameters.
+Be sure to wrap the value in quotes so the commas don’t get parsed as new parameters.
 For example \autodoc:command{\font[features="+calt,+ss01"]} will enable OpenType feature flags for both contextual alternatives and alternative style set 1.
 Similarly values that are themselves key=value pairs the quotation marks will keep them separate from other parameters.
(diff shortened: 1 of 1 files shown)


documentation/c04-useful.sil
```diff
@@ -317,4 +317,37 @@ This means that for the environment block version (\code{\\begin\{lua\}}) there
...
+Similarly, the content of the \code{\\math} command (for the \autodoc:package{math} package) falls outside the scope of the SIL language syntax and requires a different grammar.
+After all, its content follows the TeX math syntax, with commands with multiple arguments, special use of brackets, and so on.
+Therefore, we need to instruct the SIL parser that this content should not be interpreted,but rather extracted as a raw string.
+Later, it will be fed to another dedicated inputter for parsing.\footnote{In the case of math, it is currently a \em{pseudo}-inputter, but that is an implementation detail.}
+
+The SIL inputter reserves a few special keywords: \code{\\lua}, \code{\\script}; but also \code{\\ftl}, \code{\\math}, \code{\\sil}, \code{\\use}, \code{\\xml}; and finally \code{\\raw}, which we will discuss here.
+
+It is obvious that we can’t reserve too many keywords in advance.
+However, they must be known \em{before} parsing a file, which means they can’t be dynamic.
+The reserved keywords can’t be overridden or redefined afterwards.
```
alerque, line 342:
```suggestion
The reserved keywords can’t be overridden or redefined after document parsing has begun.
```

It is actually possible to add keywords and define your own, you just have to do it before parsing starts. The important thing is the dynamic nature can't be defined inside the document (or any module loaded by the document of course) it has to be done on the CLI or some other way. And there are ways. We don't have to explain those use cases, but I do think clarifying what the limitation actually is is worth it.

To my mind saying "afterwards" leaves open the question of "during".

alerque (changes requested):

PR #2042, opened 2024-06-05
Optimize typesetter mode handling

alerque:
This is awkward. I was working on something else and thought I spotted an easy way to optimize the typesetter look and save quite a bit of time. I threw it in and kept working on my other thing intenting to come back and benchmark it later.

It turns out this broke the world. Some tests started generating infinite pages and building the manual started hanging completely.

This means that `typesetter:leaveHmode()` is doing work even when already in vertical mode. I would have expected all the node queue processing to be a noop.

This is almost certainly intertwined with other `typesetter:pushBack()` issues that have surfaced, but perhaps the fact that this function is doing work outside of its labeled scope is a clue where the algorythm is breaking down.

### DIFF
--- a/typesetters/base.lua
+++ b/typesetters/base.lua
@@ -931,6 +931,7 @@ function typesetter:outputLinesToPage (lines)
 end
 
 function typesetter:leaveHmode (independent)
+   if self:vmode() then return end
    if self.state.hmodeOnly then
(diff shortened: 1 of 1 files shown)


Omikhleia:
Yep, I had seen this too: leaveHmode is also where the page building logic is triggered (and even if the horizontal queue is empty, we might have additional things in the vertical queue: penalties, vglues etc.)

PR #2047, opened 2024-06-07
Clear up what happens when at the end of paragraphs

### LINKED ISSUE #2045: Noindent not taking effect after after chapter in manual?
> Is losing noindent after chapter and section a new feature?

> [0.14.17 manual](https://sile-typesetter.org/manual/sile-0.14.17.pdf)

> image

> [0.15.0 manual](https://github.com/sile-typesetter/sile-typesetter.github.io/blob/71204d34eefcc157bee5f5fcdadea39f7e4c3d0f/static/manual/sile-0.15.0.pdf)

> image

_Originally posted by @Omikhleia in https://github.com/sile-typesetter/sile/issues/2044#issuecomment-2154283381_

alerque:
Closes #2045. At the moment it breaks a lot of tests, but the couple I've spot checked so far look *correct* and the previous expectations were wildly wrong (mostly applying parskip more than once when nothing in the input document suggested that should be happening). I'll keep reviewing test but I think this is saner than what we had before even if more bugs may be lurking.


alerque:
Nope, failing tests are a mix of correct and now broken in a different way. Yuck.

classes/base.lua
```diff
@@ -696,16 +696,18 @@ end
...
+   -- has handled that how they want, e.g. with a skip.
+   local queue = typesetter.state.outputQueue
+   local last_vbox = queue and queue[#queue]
+   local last_is_vglue = last_vbox and last_vbox.is_vglue
+   local last_is_vpenalty = last_vbox and last_vbox.is_penalty
+   if typesetter:vmode() and (last_is_vglue or last_is_vpenalty) then
+      return
+   end
    SILE.settings:set("current.parindent", nil)
+   typesetter:leaveHmode()
```
Omikhleia, line 709: The level of monkey-patching this does represent is _totally_ insane.

packages/autodoc/init.lua
```diff
@@ -389,31 +400,20 @@ function package:registerCommands ()
...
-            end)
-            SILE.call("smallskip")
+            SILE.typesetter:pushVglue(SILE.settings:get("document.parskip"))
+            pushline()
          end)
+         SILE.typesetter:leaveHmode()
       end)
    end, "Outputs its content as a standardized block of code")
 
-   self:registerCommand("autodoc:line", function (_, _)
```
Omikhleia, line 411: How unrelated to the PR?

PR #2048, opened 2024-06-08
Bibliography package improvements on BibTeX syntax

### LINKED ISSUE #2021: Bibtex implementation lacks crossref support for inheritance
SILE 0.14.17 or 0.15-develop

(Modern) bibtex/bitlatex/biber etc. define a special `crossref` field, indicating that the entry inherits any fields it is missing from the entry it cross-references.

The current implementation ignores it, resulting in incomplete references.

**Example**

```
@book{DragonDeBrume5,
  title     = {On some stars, flowers &amp; places in Middle-earth},
  booktitle = {On some stars, flowers &amp; places in Middle-earth},
  year      = 2023,
}

@inbook{LefevreMercury2023,
  author    = "Lefèvre, Alain",
  title     = "On the lost tale of Mercury",
  crossref  = "DragonDeBrume5",
  pages     = "7--12",
}
```

The second reference, when output, should use the missing fields from the first, and thus yield:
Lefèvre, “On the lost tale of Mercury.” In _On some stars, flowers & places in Middle-earth,_ 7–12. Le Dragon de Brume, 2023.

In other terms, it should be equivalent to:

```
@inbook{LefevreMercury2023,
  author    = "Lefèvre, Alain",
  title     = "On the lost tale of Mercury",
  booktitle = {On some stars, flowers &amp; places in Middle-earth},
  year      = 2023,
  pages     = "7--12",
}
```

**Rationale**

This inheritance mechanism is useful for collections, avoiding to repeat identical fields.

### LINKED ISSUE #2050: Correct handing of ampersands in bibliography entries
For strings such as _On some stuff & other things_, we currently have to format our bibtex files as follows for use with SILE:

```
@book{key,
  title = {On some stuff &amp; other things},
...
```

If we don't XML-escape the `&`, we get an error...

This is dues to SILE supporting XML entries in bibliography, which is non-standard... albeit interesting, e.g. if one wants to markup parts of entries in SIL XML. (The true use case however is not the user inserting markup, it's for the internal logic of rendering titles in italic, etc.)

It's not completely obvious, as:
- Bib(La)TeX would actually require `{On some stuff \& other things}` (and the big picture here is that it allows any (La)TeX constructs in field values, but also suffers from its own rules, hence the ampersand escaping I guess...)
- Most bibtex processors in other tools (citeproc, zotero, [web-based solutions](https://asouqi.github.io/bibtex-converter/), etc.) are just happy with no escaping at all in `{On some stuff & other things}`

Having to manually edit bibliographies to replace `&` by `&amp;` is cumbersome, we'd need to avoid it, or at least have some way to bypass it. (I'd also be interested in supporting Djot/Markdown in bibTeX files, but that's another hornet's nest :p )

### LINKED ISSUE #2051: Bibtex reader chokes at @preamble and @string
In (legacy) bibtex files, `@preamble` and `@string` are special entry types

- `@preamble` can contain (La)TeX instructions...
- `@string` allows defining abbreviations for use in other entries

Our reader errors if one or the other is present.
We should at best:
- Skip `@preamble`
- Manage `@string` or gracefully degrade if we don't want to support it.

Omikhleia:
Branch for some bibliography package fixes and improvements

Let's start gently and see where / how it goes.

Closes #2021
Closes #2051 
Closes #2050 
Closes #1860


packages/bibtex/init.lua
```diff
@@ -58,6 +68,56 @@ local parseBibtex = function (fn)
...
+
+   if not refs then
+      return
+   end
+   for _, ref in ipairs(refs) do
+      local parent = bib[ref]
+      if parent then
+         crossrefAndXDataResolve(bib, ref)
+         for k, v in pairs(parent.attributes) do
+            if not entry.attributes[k] and k ~= "crossref" and k ~= "xdata" then
```
Omikhleia, line 111: Not to safe: normally we resolved the parent, so the two extra "and"'s are not needed

packages/bibtex/init.lua
```diff
@@ -58,6 +68,56 @@ local parseBibtex = function (fn)
...
+-- to be defined before the entry that uses it, or even in the same bibliography
+-- file.
+-- Implementation note:
+-- We are not here to check the consistency of the BibTeX file, so there is
+-- no check that xdata refers only to @xdata entries
+-- Removing the crossref field implies we won't track its use and implicitely
+-- cite referenced entries in the bibliography over a certain threshold.
+-- @tparam table bib Bibliography
+-- @tparam string key Valid entry key
+local function crossrefAndXDataResolve (bib, key)
```
Omikhleia, line 86: note to self: would be more efficient if passing the entry rather than the key

Omikhleia: Also this function could be moved to a different support file.

packages/bibtex/init.lua
```diff
@@ -58,6 +68,56 @@ local parseBibtex = function (fn)
...
+   if not refs then
+      return
+   end
+   for _, ref in ipairs(refs) do
+      local parent = bib[ref]
+      if parent then
+         crossrefAndXDataResolve(bib, ref)
+         for k, v in pairs(parent.attributes) do
+            if not entry.attributes[k] and k ~= "crossref" and k ~= "xdata" then
+               entry.attributes[k] = v
```
Omikhleia, line 112: The Biblatex manual has an appendix B "Default Inheritance Setup" implying some fields are re-mapped depending on the target and source's types. This is also supposed to be fully configurable (with LaTeX commands, with "options" fields, etc.)
E.g. 
- the "title" of a `@periodiical̀` would become the "journaltitle" (an alias to "journal") of an `@article`
- the "title" of a `@book` would become the "booktitle" of an `@inbook`... but apparently not of an `@incollection`? Why?
 
This appendix table is really messy to read... Playing with a few JS-based bibtex engines, while they do supplement fields from `crossref` but do **not** seem to apply such field mappings... I sort of understand why!

The Biblatex manual is real hard to read... And it's never stated clearly if `xdata` follows the same inheritance mapping... Doh. This format is totally insane.

Omikhleia: Re: `crossref` mapping - Typst does some minimal/naive hard-coded inheritance:
https://github.com/typst/biblatex/blob/a43de9ca4fb5c981c39c4c162e6df54684015fc8/src/lib.rs#L630

Omikhleia: Minimal implementation done.

packages/bibtex/init.lua
```diff
@@ -133,6 +278,93 @@ To produce a full reference, use \autodoc:command{\reference{<key>}}.
...
+
+Some BibTeX implementations automatically include entries referenced with the \code{crossref} field in the bibliography, when a certain threshold is met.
+This implementation does not do that.
+
+Depending on the types of the parent and child entries, the child entry may inherit some or all fields from the parent entry, and some inherited fields may be reassigned in the child entry.
+For instance, the \code{title} in a \code{@collection} entry is inherited as the \code{booktitle} field in a \code{@incollection} child entry.
+Some BibTeX implementations allow configuring the data inheritance behavior, but this implementation does not.
+It is also currently quite limited on the fields that are reassigned, and only provides a subset of the mappings defined in the BibLaTeX manual, appendix B.
+
+Here is an example of a BibTeX file showing some of the abovementiond features:
```
Omikhleia, line 346: I can't spell "abovementiond" :p

packages/bibtex/init.lua
```diff
@@ -133,6 +278,93 @@ To produce a full reference, use \autodoc:command{\reference{<key>}}.
...
+Entries consist of fields, which are key-value pairs.
+The field names are case-insensitive.
+Spaces and line breaks are not important, except for readability.
+On the contrary, commas are compulsory between any two fields of an entry.
+
+String values shall be enclosed in either double quotes or curly braces.
+The latter allows using quotes inside the string, while the former does not without escaping them with a backslash.
+
+When string values are not enclosed in quotes or braces, they must not contain any whitespace characters.
+The value is then considered to ba a refererence to an abbreviation previously defined in a \code{@string} entry.
```
Omikhleia, line 323: to ba > to be

alerque: To “bah”, or not to “bah”, that is the question.

packages/bibtex/init.lua
```diff
@@ -7,37 +7,96 @@ local epnf = require("epnf")
...
+local function consolidateEntry (entry, label)
+   local consolidated = {}
+   for field, value in pairs(entry.attributes) do
+      consolidated[field] = value
+      local alias = fieldmap[field]
+      if alias then
+         if entry.attributes[alias] then
+            SU.warn("Duplicate field '" .. field .. "' and alias '" .. alias
+               .. "' in entry '" .. label .. "'")
+               SU.dump(entry)
```
Omikhleia, line 86: Forgot some debug stuff :)

packages/bibtex/init.lua
Omikhleia: Fixed and autosquashed

packages/bibtex/init.lua
Omikhleia: Fixed and autosquashed

Omikhleia:
So this first PR is focused low-handing fruits, regarding the support of more Bib(La)TeX syntax elements without breaking existing working behavior yet:
- Ignoring all content between entries (previous implementation mandated using `%` comments, but this is not correct)
- Ignoring `@preamble` elements (TeX-based systems)
- Supporting `@string` elements ("abbreviations") and string concatenation with `#`
- Supporting `@xdata` elements and `xdata` fields
- Supporting `crossref` fields with (an initial partial implementation) the biber/biblatex data inheritance rules
- Supporting `\&` as legacy alternative to `&` for BibTeX compatibility
- Supporting `~` as convenience for non-breaking space, and `\~` to get a tilde (BibTeX compatibility too)
- XML-escaping (`&`, `<`, `>`) in input, since we are later wrapping them in XML tags (for italics etc.) and processing the formatted output as XML

Later, I'll want to refactor some of the code and reorganize the files differently, in order to tackle properly with issues that _will_ change the output, but this first step should be a decent base for the work to come, and it remains minimally readable (which would become harder after reshuffling some of the logic in different files).

Ready for review.

packages/bibtex/init.lua
```diff
@@ -48,14 +106,88 @@ local parseBibtex = function (fn)
...
+         biblio[ent.label] = consolidateEntry(entry, ent.label)
+      end
+   end
+end
+
+--- Copy fields from the parent entry to the child entry.
+-- BibLaTeX/Biber have a complex inheritance system for fields.
+-- This implementation is more naive, but should be sufficient for reasonable
+-- use cases.
+-- @tparam table parent Parent entry
```
alerque, line 134: It warms my cold clammy heart to see function docs and LSP / doc generation hints. In general. In this case I do kind of wonder what the point is on a local function that isn't exported for use anywhere outside of this file. Or did I miss something?

alerque (approved):
This looks good to me. I'm fine with this as is, the comments I made were more because I didn't understand why you did something than because it has to change. You decide.

The commits I pushed are just automatically generated code formatting from `stylua` with `git absorb`. The fixup commits should apply to touchup the original commits if you rebase this branch for any reason, or I can too before merging.

I'll be out of pocket for the next week so I can't do much work (maybe a couple of mobile phone based scanning), but if having this merged will facilitate the ongoing work in this direction I can make that happen.

packages/bibtex/init.lua
```diff
@@ -7,37 +7,95 @@ local epnf = require("epnf")
...
 -- luacheck: push ignore
 -- stylua: ignore start
 ---@diagnostic disable: undefined-global, unused-local, lowercase-global
 local bibtexparser = epnf.define(function (_ENV)
+   local strings = {} -- Local store for @string entries
+
    local identifier = (SILE.parserBits.identifier + S":-")^1
    local balanced = C{ "{" * P" "^0 * C(((1 - S"{}") + V(1))^0) * "}" } / function (...) local t={...}; return t[2] end
-   local doubleq = C( P'"' * C(((1 - S'"\r\n\f\\') + (P'\\' * 1)) ^ 0) * '"' )
+   local quoted = C( P'"' * C(((1 - S'"\r\n\f\\') + (P'\\' * 1)) ^ 0) * '"' ) / function (...) local t={...}; return t[2] end
```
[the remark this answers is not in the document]
Omikhleia: To be honest, I refrain touching it too much, because I am not sure the initial rules are actually correct, regarding how the escaping is supposed to be done in the quoted strings. The code is using `P('\\" * 1)`... hinting towards `\"` (which I _think_ is indeed doable), but some manuals say otherwise, using internal braces:
image

And this is a bit tricky, as internal braces (in both cases) are supposed to mean something else too. (An currently we don't support them as intended). Maybe we should have a dedicated issue, and better check what other implementations and manuals recommend here?

packages/bibtex/init.lua
Omikhleia: I've the feeling that the bibtex package, providing `\cite`, `\reference`, `\loadbibliography` etc. is somewhat ill-named, it should be bibliography, and support for bibliography formats should be provided in their own files. Besides `.bib`, one could want to provide references in RIS, Refer, EndNotes... or even CSL-JSON, eventually...). It's a bit soon to say, maybe, and I didn't want to split things too early here.

Whatever, having the LSP instructed for type safety and parameter meanings (regardless of an eventual doc generation) is always nice, no?

packages/bibtex/init.lua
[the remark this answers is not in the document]
Omikhleia: By the way, the above screen shot is from "Tame the BeaST. The B to X of BibTEX,* version 1.4 (2009).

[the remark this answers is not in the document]
alerque: I think you're right the parser probably needs a more aggressive overhaul and I'm happy with just copying the existing function for now. Can you open and issue linking back to this so we come back to it some day? By the way have you run across any ABNF, EBNF, or other formal grammars for BibTeX stuff or are there just a hodge–podge of implementations we're copying that might not even all match each other?

packages/bibtex/init.lua
alerque: Yes I agree this should be a bibliography package with somewhat modular support for loaders.

As for the docstring hints, sure, and it should affect editors editing *that file*. Just be aware it won't get published since it isn't a public function.

alerque:
Just a side note on the Lua style thing.... when you see me push those kind of fixups just know there is absolutely no need to to make an effort to preserve them. If you happen to have other work in a commit or branch and my style fixups are causing rebase conflicts or anything like that just keep the work you have and blow the code formatting fixes away. They are 100% automated and next to zero work to redo at any point, so don't stress if they are causing conflicts. And don't stress about getting the formatting right while writing code either, whatever is easier to write you can write, the auto formatter can take it from any Lua code that runs.

packages/bibtex/init.lua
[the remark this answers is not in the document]
alerque: c.f. #2075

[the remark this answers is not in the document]
Omikhleia: >  Can you open and issue linking back to this so we come back to it some day?

Tracked as #2075 

> By the way have you run across any ABNF, EBNF, or other formal grammars for BibTeX stuff or are there just a hodge–podge of implementations we're copying that might not even all match each other?

For some elements (e.g. `@string`), I checked a few (non-grammar-based) implementation as well as a few ABNF grammars. But the rest is mostly from the biblatex manual and _Tame the Beast._ Problem is that most ABNF specifications I've found are incomplete... So yes, it ends up as a bit of a hodge-podge... (Hence also the "notes" on the format added in the documentation, as it's not clear to me that there's some full complete definition agreed by everyone and formally defined... Biber/biblatex keep adding stuff!)

packages/bibtex/init.lua
alerque: The TODO item in this discussion is now #2076.

Issue #2001, opened 2024-02-20
Hyphenation problems in Portuguese


jodros:
![изображение](https://github.com/sile-typesetter/sile/assets/15092760/85813cf1-08a0-4b83-8e44-7d9bbbb13b99)
![изображение](https://github.com/sile-typesetter/sile/assets/15092760/790e5058-70ee-45ea-9d90-11ced5a45663) ![изображение](https://github.com/sile-typesetter/sile/assets/15092760/b232edbb-2f75-41d4-8c7a-739b5eeb9b11) ![изображение](https://github.com/sile-typesetter/sile/assets/15092760/3e525029-15ab-4a44-8485-067c4c01f5e8)

One word I noticed to also have some trouble in being hyphenated is `quando`.

Yes, I know the first example isn't the best in terms of readability, but it's what I've right now since I'm trying the `parallel` package for now, I could give more examples for Russian soon...

jodros:
![изображение](https://github.com/sile-typesetter/sile/assets/15092760/63aec98e-bc62-459b-ae6e-fbcaf081ef1a)

Another format of the first example, this time I couldn't see the frames because `-d frames` isn't working well when I use `parallel`...

jodros:
![изображение](https://github.com/sile-typesetter/sile/assets/15092760/6bca1ed8-5b35-43ba-b7fd-240b8c857f6e)

alerque:
Lets split PT/RU into different issues because tracking down language-specific stuff doesn't always get resolved at the same time/via the same PR. Lets make this issue the PT one please.

For hyphenation issues, the first thing to check is if we even have break points to work with. Evidently not:

```console
$ ./sile
SILE v0.14.17.r373-g72965ad (LuaJIT 2.1.1700206165) [Rust]
> SILE.showHyphenationPoints("quando", "pt")
quando
> SILE.showHyphenationPoints("apaziguam", "pt")
apa-zi-guam
```

So at least for "quando", for some reason the patterns are not allowing any hyphenation there. According to PT language rules, where should the points be?

The screen shots are kind of hard to work with for this because I can't tell if the problem is other metrics (like not having any stretch available) might be contributing to poor break choices. Also I can't even be sure I'm typing the same text as you are entering in many cases. Can you post the actual XMl/SIL input files you're testing too?

Omikhleia:
Unless I misunderstood the screenshot, it doesn't look as an hyphenation issue, but rather a justification issue (overfull lines)

[...]

PR #2063, opened 2024-06-12
Add three missing syllables of Brazilian Portuguese

jodros:
This PR follows the problem in #2001... And yes, only three syllables, that's all when it comes to the pairs like ó/ô for PT-PT and PT-BR, respectively. Regarding to further differences between these two branches of Portuguese, they aren't related to anything this file has to do... And since the 1990 we have this Agreement among all the Lusophone countries, according to which:

> As for divergent spellings such as *anónimo* and *anônimo*, facto and fato, both will be considered legitimate, according to the dialect of the author or person being transcribed.[^*]

[^*]: (https://en.wikipedia.org/wiki/Portuguese_Language_Orthographic_Agreement_of_1990)


alerque (changes requested):
We need to document somewhere any adjustments we make to these files since I think they were imported at some point and we might regenerate them at some point too. I wouldn't want to sort through `git blame` to find out everything we've ever tweaked. Maybe even just a comment about the additions at the top of the list would suffice for now.

Omikhleia:
N.B. I don't think a "chore" commit is appropriate. It's either a fix or a feature (cool, in both cases).

PR #2064, opened 2024-06-12
Export env language var matching document state

### LINKED ISSUE #2038: Export locale for system commands run from Lua
One more for after we better work out how we're handling the locale code internally...

When I run `os.execute()` from Lua inside a SILE document, I would expect the document locale setting to have been exported as an env var ahead of time.

```sile
\begin{document}
\language[main=tr]
\fluent[name=Docker]{hello}
\eject
\begin{lua}
local date = os.execute("date")
local date = os.execute("env LC_ALL=tr_TR.UTF-8 date")
\end{lua}
\end{document}
```

```console
$ sile foo.sil -o - -b text
SILE v0.14.17.r414-g6382ac4 (LuaJIT 2.1.1700206165) [Rust]
<d/foo.sil> as sil
<./core/languages.lua:78> as xml
Merhaba Docker![1]  1
                     <./core/sile.lua:566> as lua
Fri May 31 01:47:20 AM +03 2024
Cum 31 May 2024 01:47:20 +03
[2]
2
```

Note the output of the first date command does *not* track the document context.

alerque:
Closes #2038. Needs to be merged after #2062 and #1641.


alerque:
This will need some re-arranging to confirm to whatever layout #2072 settles on for Rust based Lua module stuff.

core/settings.lua
```diff
@@ -28,6 +28,7 @@ function settings:_init ()
             ]]):format(language, language, language, language))
          end
          fluent:set_locale(language)
+         SU.setenv("LANG", language)
```
Omikhleia, line 31: While this could work for some language, the correct approach might be trickier:
- `LANG`expect a Unix locale (eg. `de_DE`)
- Our current `document.language` is only a two letter code (eg. `de`), which may be under-resolved
- If we move our `document.language` at some point to possibly fit either BCP47 or ICU locales, it still likely wouldn't be a valid Unix locale (e.g. consider e.g. BCP47 code `sr-Latn-RS`, ICU locale `sr_Latn_RS`, Unix locale `sr_RS@Latn`)

So we might have to consider conversion utilities (BCP47 -> ICU local -> Unix locale)

Omikhleia: While there are one-to-one canonicalization rules between BCP47 languages and ICU locales (which our justenoughicu wrapper even implements), I can't find any clear source for mapping to Unix/Posix locales (setlocale) as expected by LANG. So it might be a a hornest's nest.

(BTW, Why would we want it? For Lua `os.date` etc. to be localized? I'm not sure there's a good use case for it)

alerque: Yes, this whole maneuver is dependent on us starting to use proper locales (BCP47 or whatever) internally and on adding a mapping to the `setenv()` function that converts whatever than scheme is to a Unix locale.

I'm already using this is production with a small hack to map the languages I use. I have documents where one class is used to render inputs in multiple languages, and for example the inprint page has date information. This is rendered with the system `date` command, but in only outputs the correct localized string if the locale is set. This already works if I supply the right mapping. I also have other tools besides `date` that output content that gets ingested during the SILE render process (for figures, music, etc.) and is sensitive to the locale.

The real question is whether there is ever a case that setting the LANG will produce undesired results. For it to work the system tools need to have support for the expected locales compiled using `locale-gen` or similar, but would the user ever *not* expect it to work this way? I couldn't think of any time this would be objectively wrong.

alerque: Yes, except that instead of having our own custom "just enough" wrapper for ICU we're going to have access to the full ICU API surface via Rust crates (and of course can expose whatever bits of it we want on the Lua side, which is basically what we have now in `justenough...` except we won't need the C wrapper.

alerque:
This approach is going to de ditched (and this PR closed) when #2072 comes to fruition, but I see a few tidbits in here than need cherry picking.

core/settings.lua
alerque: > While this could work for some language

Still true, but I'm actually going to merge this before #1641 understanding that it's not going to work for all locales yet, but there is no point I see in waiting on proper local parsing and it might even make it easier to test locale parsing to have this in place.

PR #2071, opened 2024-06-15
Bibliography changes on authors, editors, publishers, dates

### LINKED ISSUE #2052: Editor and translator names in bibliography are passed as raw text
Entry:

```
@book{Kill,
  author = "Kill, Bill",
  editor    = "Doe, John and Smith, Jane",
  translator = "Pumpernickel, William",
  title     = "Another Book Entry",
  address   = "London",
  year      = 2022,
}
```

Expected Chicago Manual of Style 16th edition reference according to its CSL:

Kill, Bill. _Another Book Entry._ Edited by John Doe and Jane Smith. Translated by William Pumpernickel. London, 2022.

Obtained:

Kill, _Another Book Entry._ Edited by Doe, John and Smith, Jane, Translated by Pumpernickel, William. London: 2022.

There are a number of issues here (notably on the author), but the one I am reporting here is the following: editors and translators are the direct raw bibtex entry as text, not parsed as a name list (split for reordering depending on the citation/reference style).

Omikhleia:
Closes #2052 

**Features**
 - Some files split and reorganized (minimal refactor for now)
 - Support more Bib(La)TeX name fields (author but also editor, shortauthor, shorteditor, translator, holder) and parse/split names earlier (at entry parsing rather than on each use)
 - Partial support for BibLaTeX extended date field (ISO ETDF)
 - Month regularization (support 3-letter short code as BibTeX, but also 1-12 numbers)

**Known issue**
 - (See also in-code comment in the Chicago-style implementation file.) Date formatting obeys English only (as "Month Day, Year" with the month in English). This is still better that the previous behavior (raw as-is value of the BibTeX field), but obviously not general.

With the ongoing effort on CSL support, the above know issue might be acceptable for now and not worth addressing. The underlying question, however, is whether we want to consider still supporting the "legacy" style logic; and if so, how much efforts we are ready to involve in it...


packages/bibtex/styles/chicago.lua
```diff
@@ -1,5 +1,37 @@
 local Bibliography = require("packages.bibtex.bibliography")
 
+-- WORKAROUND FIXME
+-- We may want fluent strings for all languages, but for now we just use English.
```
Omikhleia, line 4: See #2074 :
It would be a shame to have all possible date composition rules in our ftl files, while CSL locales have them.
(Which doesn't imply that fluent would necessarily be out of the question -- Just that the input should be a CSL locale, however managed afterwards)

Omikhleia: But also not that CSL styles can override some of the CSL locales.
So proper handling of such rules = CSL locales + CSL styles.

Omikhleia: In-code comment improved.

alerque:
Rebased commits have only been modified by filtering the Lua through `stylua` to normalize formatting, no other changes.

alerque (approved):

PR #2072, opened 2024-06-15
Setup Lua module for exporting Rust functions to Lua

alerque:
To date we've only had a Rust wrapper with a few strings stuffed in then running an almost entirely Lua code base. An obvious benefit of having Rust at our finger tips is going to be access to different libraries and bindings and speed and type checking and such. But there hasn't been an ergonomic way to start reaching into that toolbox yet. I did some POC work several times before I went down the Rust road so I know it would be possible, but v0.15.0 just had the CLI wrapper.

The current draft of #2064 is the first real function from the Rust side being passed over to Lua. But that approach isn't very sustainable, and it completely breaks compatibility with Lua. We're closer than ever to being able to easily be used *as a library* from Lua (and with just a little more futzing, also Python, Node, and others). So I don't actually want to go down that road.

Instead I want to make all the Rust API surface we wast to use available on the Lua side via a module that can be built and installed on it's own.

Unfortunately the `mlua` library can't use the *vendored* feature at the same time as the *module* feature, so we have to split out a separate crate for it. This PR does that.

We'll probably want to keep all the code and logic in the sile create's lib and just use the rusile crate to export modules to Lua/Python/whatever. And we'll probably want proc macros to wrap the functions with the right bindings, but that can come later.

As of posting the Lua module builds and runs code, but isn't loaded back into the `SILE` runtime namespace.


alerque:
Just a brainstorm, I've been wondering how to setup Cargo to build two different variants of the same crate with different features. Maybe we don't need that for a regular SILE build, and the LuaRocks library version could inject it's own feature flag as part of the build. Just an idea to try — I don't know why I was fixated on needing to build a Lua module to load in Rust. As long as we feed the same functions in Rust (and inject in the CLI) the same functions that go into the module, there is no reason to load the module at all on the Rust side.

PR #2082, opened 2024-06-29
CSL support

### LINKED ISSUE #2022: Bibtex \cite errors on books without author
SILE 0.14.17 or 0.15-develop

`\cite[key=...]` exits on error in `commafy` (called by `andSurnames`) for an entry without author.

**Example**

```
@book{DragonDeBrume5,
  title     = {On some stars, flowers &amp; places in Middle-earth},
  year      = 2023,
}
```

**Rationale**

Such entries are however [legit](https://libguides.westsoundacademy.org/chicago-citation/book-with-no-author)

It may be used for instance for non-attributed collective works, or ancient works whose author is unknown, etc.

### LINKED ISSUE #2024: Bibliography always use curly quotes instead of language-dependent quotes
SILE 0.14.17 or 0.15-develop

While the bibliograpĥy package has provision for i18n via fluent, it always use curly quotes for article titles, etc.
That is, “'Title”

https://github.com/sile-typesetter/sile/blob/b2cc0841ff603abc335c5e66d8cc3c64b65365eb/packages/bibtex/bibliography.lua#L378-L382

This is naive, and wrong:
- If used in a book in another language, it would be expected that appropriate quotes are used, e.g. « Titre » (French), „Titel“ (German), etc.
- (If there are quotes in the title, they may also need to be shifted.)

### LINKED ISSUE #2026: Generating a bibliography
SILE 0.14.17 / 0.15-develop

The **bibliography** packages doesn't offer a way to generate a bibliography from cited works.

While one can achieve it by calling `\reference` for all cited works, this is tedious and sub-optimal. So we would need a `\bibliography` command.

Here is my first take at the topic:

- The `\cite` command should keep track of cited works in a scratch list
- ~~The `\reference` command should have an option to suppress the author, replaced by an horizontal rule (see rationale below)~~ Removed: This only makes sense in the context of the bibliography. The `\reference` command wouldn't make sense eventually. (If you want a full ref. to a book, use a citation styles that uses that for citations).
- A `\bibliography` command should be implemented
   1. Sort all entries (by default, by "author-date") with appropriate `SU.collatedSort` on names, then by date.
   2. Call the `\reference` logic on the sorted list,  with author suppressed if identical to the previous one (i.e. replaced by a rule)
   4. Clear the scratch list of cited works, so later use (in a different chapter or whatever) start afresh.

### LINKED ISSUE #2027: Month, day and date formatting in bibliography references
As of today, the **bibtex** package does it wrong

 1. The Chicago style implementation ignores the `day` although it shall be used when available for some formats (typically, journal articles)
 2. The `month` is output as set in the `.bib` file, but it should be replaced by a true month name (abbreviated or not depending on the citation style, Chicago if I read it correctly recommends the full month name). The problem is, BibTeX has a loose definition for `month`, which can be a number (e.g. `"3"`) or a full string (e.g. `"March"`) or a "standard" 3-letter code (= `"mar"`); the recommendation is to use the latter.[^1]

Moreover, the date shall likely be formatted according to language-specific rules.[^2]

Example: `@article{..., day= {14}, month = {apr}, year = {2024} ...}`
  - With the current implementation, we get `apr, 2024.`
  - We should actually get `April 14, 2024` in English Chicago citation style
  - And, say, `14 août 2024` in a French document[^3] -- however formatted, certainly not "apr" or "April" ;)

[^1]: Some references for `month` in BibTeX format:
         - https://www.bibtex.com/f/month-field/
         - Q13 in https://mirror.ibcp.fr/pub/CTAN/biblio/bibtex/contrib/doc/btxFAQ.pdf
         - p. 10 in https://ctan.mines-albi.fr/biblio/bibtex/base/btxdoc.pdf

[^2]: Note that **babel** for LaTeX has tables of month names and date rules for all supported languages... But it seems that language support for bibliographies comes with **babelbib** which also has month name translations. I'm a bit lost here, regarding how it works in the LaTeX world...

[^3]: At least according to some online citation tools I tried. I haven't looked deeply into CSL examples...

### LINKED ISSUE #2074: Bibliography: support CSL?
Some of the issues regarding bibliographies (eg. #2027, #2049) are hampered by the fact that SILE has it's own-made Chicago style implementation. I know of at least one similar attempt at implementing another style on the same grounds (@jodros 's [ABNT](https://github.com/jodros/abnt.sile/blob/main/packages/bibtex/styles/abnt.lua))

Question 1. What would be SILE's position on deprecating these hand-made styles, and switch to CSL?

 - Proper localization is provided by [CSL locales](https://github.com/citation-style-language/locales)
 - Lots of standard styles have a [CSL implementation](https://github.com/citation-style-language/styles), incl. various versions of the Chicago style.

Question 2. Would SILE be ok with its own implementation of (a decent subset) of CSL in Lua?
 
 - CSL is not that hard of a schema (compared, say, to USX and all its implications (#1398)...
 - The alternative is to use an existing CSL processor (but it is not that easier to hook back into SILE and it's far less fun IMHO)

Omikhleia:
Closes #2074 

It already does nice things (see screenshots in the referred issue).

In order to support CSL (Citation Style Language), we need to:
  - [x] Convert the BibTeX entries to CSL format.
  - [x] Use a CSL engine to format the citations and bibliography references. It boils down to:
     - [x] Support CSL locales
     - [x] Support CSL styles
     - [x] Implement the CSL processor/renderer (a "reasonable" subset at least)

Regarding the conversion of BibTeX entries, the mappings are not straightforward, but there is some prior art that we can check... None of the implementations I checked did the exact same things, so it's likely a bit messy...

Regarding the CSL engine, there are various existing implementations.
Yet, I had a look at them, and I am not really convinced by their code quality, so I went implementing the [CSL 1.0.2 specifications](https://docs.citationstyles.org/en/stable/specification.html)  from scratch. Because it's fun, and SILE has the guts to do it. And because I think I can.

Additionally, this would also close several other items.

Closes #2024 = The CSL locales takes care of it.

Closes #2022 = The CSL styles have appropriate fallbacks (substitutes, conditionals, etc.)

Closes #2027 = The CSL styles and locales define how to format localized dates in the selected citation or bibliography style.

Closes #2026 = Some CSL styles sort entries by citation order ("citation-number"), so keeping track of cited entries was needed anyhow.


csl/core/engine.lua
```diff
@@ -0,0 +1,967 @@
...
+   -- Not so sure it's the intended behavior from the specification?
+   if SU.boolean(self.locale.styleOptions["punctuation-in-quote"], false) then
+      text = luautf8.gsub(text, "([" .. rdquote .. rsquote .. "]+)%s*([.,])", "%2%1")
+   end
+   -- HACK: fix some double punctuation issues.
+   -- Maybe some more robust way to handle affixes and delimiters would be better?
+   text = luautf8.gsub(text, "%.%.", ".")
+   -- Typography: Prefer to have commas and periods inside italics.
+   -- (Better looking if italic automated corrections are applied.)
+   text = luautf8.gsub(text, "(</em>)([.,])", "%2%1")
```
Omikhleia, line 932: Suggested:

```
   text = luautf8.gsub(text, "(</em>)([%.,])", "%2%1")
   -- HACK: remove extraneous periods after exclamation and question marks.
   -- (Follows the preceding rule to also account for moved periods) 
   text = luautf8.gsub(text, "([!?])%.", "%1") -- Remove trailing punctuation
```

See https://github.com/sile-typesetter/sile/issues/2024#issuecomment-2198081835

Omikhleia: Not sufficient, we might also have to handled the quoted stuff when punctuation-in-quotes is not active: `« My Stuff! ».`

Real weird that CSL does not say a word about this...

alerque: Being familiar with the state of localization tech when CSL was developed, I don't find that oversight surprising at all. I'm surprised it covers as many edge cases as it does (even if my eyes want to bleed looking at how it was done).

Omikhleia: Checking some (non-bibliography related) French typography resources, `« My Stuff ! ». Blah...` is _sometimes_ accepted...
So I'll be opinionated: It's ugly and hurts my feelings.[^1] 

I certainly prefer `« My Stuff ! » Blah...` (but a comma _would_ look ok to me, my only concern is the period... Interesting how tastes and usage do affect things that do not have a "definitive" rule or convention...). 

So I'll also fix it.

(And my eyes want to bleed too, lol)

[^1]: I am a big advocate of not using "double marking" when not needed. It goes beyond the topic here, but for instance, some people write: 
     > This French guy said "_bonjour_" ("hello")...

    That's double marking (italic for foreign words, quotes for quotations). In _most_ situations, I tend to prefer:
    > This French guy said _bonjour_ ("hello")...

    There's no ambiguity, so a "single marking" is sufficient and remains typographically sound... and IMHO, it's more elegant (in addition to saving a few characters). Well... Actually, when I am the author, there are some contexts (e.g. in footnotes) where I may even prefer:
    > He actually said: "_bonjour tout le monde,_ hello everybody"...

    (It also boils down to removing some double marking: a parenthesized translation)

csl/core/engine.lua
```diff
@@ -0,0 +1,967 @@
...
+   return t
+end
+
+function CslEngine:_render_affixes (t, options)
+   if not t then
+      return
+   end
+   if options.prefix then
+      -- replace by localized prefix
+      local pref = options.prefix:gsub("[,;:]", function (c) return self.punctuation[c] or c end)
```
Omikhleia, line 184: We don't really care with SILE, as we do our own automated space handling for French (so we always get it right, removing and inserting non-breaking space automatically around concerned punctuation signs).

But _in a more global picture,_ this is probably "too clever" (same remark applies to the suffix and delimiter): 
 - chicago-author-date-fr _already_ encodes the non-breakable space as `&160;` in the delimiters, and affixes, so we now end up with two of them in the output... (one comes from the fr-FR locale for "colon", "semicolon"). As noted, SILE doesn't really care, the output will be corrected, but that's weird...)
 - If using other styles not ported to French (even the default generic chicago-author-date), if we remove this delim/affix localization, we won't get the non-breakable space in the output. (But again we don't care, SILE inserts them as appropriate)
 
On the other hand, there must be a reason for CSL local to have some punctuation terms. 'Cause if we remove this delim/affix localization, the terms aren't used anywhere -- and I don't understand why they exist in the first place!

Possible solutions:
 1. Remove this delim/affix punctuation localization, it wasn't supposed to exist...
 2. Or keep it only if the style default-locale exists and matches the locale lang... though I see the troubles coming, fr-CA etc. are going to be messy then...

We could question the CSL folks regarding the purpose of the "colon", "semicolon" terms etc. vs. the intended use of delimiters and affixes, but it's likely a hole in the specification... So the best course of action could be (1), simplifying the code, esp. since it has no effect with SILE's own good French support.

packages/bibtex/support/bib2csl.lua
```diff
@@ -0,0 +1,178 @@
...
+   csl['number-of-volumes'] = entry.volumes
+
+   -- BibTeX edition -- FIXME Can be a literal string or a number
+   csl.edition = entry.edition
+   -- BibTeX version
+   csl.version = entry.revision
+
+   -- BibTeX number and issue
+   csl.number = entry.number
+   csl.issue = entry.issue
```
Omikhleia, line 126: Not so trivial maybe:
https://github.com/JabRef/jabref/issues/8372#issuecomment-1023768144

csl/core/engine.lua
Omikhleia: > We could question the CSL folks regarding the purpose of the "colon", "semicolon" terms etc. vs. the intended use of delimiters and affixes, but it's likely a hole in the specification..

I [asked](https://github.com/citation-style-language/locales/issues/315) -- and the status seems to be that such replacements are not part of CSL 1.0.2.

csl/core/locale.lua
```diff
@@ -0,0 +1,259 @@
...
+
+--- Lookup a date format in the locale.
+-- @tparam string form The form of the date ('numeric' or 'text')
+-- @treturn table The date format as a table of cs:date-parts
+function CslLocale:date (form)
+   local dp = form and self.dates[form]
+   if not dp then
+      SU.error("CSL locale date format not found: " .. tostring(form))
+   end
+end
```
Omikhleia, line 170: Oops, missing `return dp`

csl/core/engine.lua
Omikhleia: Besides ? and !, also handle suppression after an ellipsis...

csl/core/engine.lua
```diff
@@ -0,0 +1,967 @@
...
+            -- NOTE: Seems the case may occcur, to select a specific date part from the date
+            -- I'm not sure I read this correctly in the CSL documentation!
+            -- That would imply usingg a copied date object, with only the relevant part?
+            SU.warn("CSL cs:date-parts in localized date not implemented")
+         end
+         -- Use locale date format
+         content = self.locale:date(options.form)
+         options.delimiter = nil -- Not supposed to exist
+      end
+      local t = self:_date_parts(options, content, date)
```
Omikhleia, line 477: I overlooked the `date-parts` _attribute_ on localized dates, as in `<date variable="original-date" form="text" date-parts="year" prefix="[" suffix="]"/>`
 - “year-month-day” - (default), renders the year, month and day
 - “year-month” - renders the year and month
 - “year” - renders the year

csl/core/engine.lua
Omikhleia: Fixed

csl/core/engine.lua
Omikhleia: Fixed (kept behind a non-standard extension)

csl/core/locale.lua
Omikhleia: Fixed

csl/core/engine.lua
Omikhleia: Fixed

csl/core/engine.lua
```diff
@@ -0,0 +1,1010 @@
...
+function CslEngine:_if (options, content, entry)
+   local match = options.match or "all"
+   local conds = {}
+   if options.variable then
+      local vars = pl.stringx.split(options.variable, " ")
+      local cond = false
+      for _, var in ipairs(vars) do
+         if entry[var] then
+            cond = true
+            break
```
Omikhleia, line 810: Testing with several style files and a variety of inputs, my reading of the spec here is wrong. here and below... I understood it as "evaluate each condition independently" and then "apply the match attribute" to each condition", but it should be applied to each elementary item (e.g. here, each variable).

E.g. seen with chicago-fullnote-bibliography (not with chicago-author-date, but it might just mean my inputs are not representative enough), on `<if variable="original-date issued" match="all">`

csl/core/engine.lua
```diff
@@ -0,0 +1,1010 @@
...
+      -- Keep silent: it's not a critical feature yet
+      -- SU.warn("CSL display not implemented")
+   -- end
+   return t
+end
+
+function CslEngine:_render_quotes (t, options)
+   if t and options.quotes then
+      -- Smart transform straight quotes in the input to localized inner quotes.
+      t = luautf8.gsub(t, "([%s%p])\"", "%1" .. self.punctuation.open_inner_quote)
```
Omikhleia, line 261: While this works for inner quotes surrounded by text (e.g. `Some "special" text` --> `Some ‘special’ text`), we also need to take into account quotes at the start or end of the input (e.g. `Some "special text"` --> `Some ‘special text’`)
```
      t = luautf8.gsub(t, "^\"", self.punctuation.open_inner_quote)
      t = luautf8.gsub(t, "\"$", self.punctuation.close_inner_quote)
```

csl/core/engine.lua
zepinglee: Note that the items in `type` attribute are always applied with `any` logic (i.e. "or"), which is not clearly stated either. For example, `<if type="chapter paper-conference" variable="container-title" match="all">` means `(type="chapter" or type="paper-conference") and variable="container-title"`.

I also recommend the <https://github.com/citation-style-language/test-suite> repository for testing.

Omikhleia: Thanks a lot for the explanation, @zepinglee ! It's really appreciated.

(I think the existing spec could be improved to mention such subtleties and be much more clear at some places -- But please take it as a bare statement, not a criticism: I know it's real hard to write good documentations and specifications on such topics.)

N.B. As for the test suite, yes indeed, I am aware of it, and at some point it would need to be addressed for complete conformance. But I'll be real honest here (and I'll clarify it in the PR description depending on the point I eventually reach), I am first of all thriving at getting enough features for processing my next book, incl. BibTeX-based bibliography of ~1500 entries. The "current" experimental bibliography solution in SILE completely fails at it, hence my attempt at rewriting it with CSL in mind. Clearly it's an iterative process with that goal in mind. (and I've only processed ~100 entries so far, with 2-3 style files and the two target locales only, noting pn the way where it's still wrong -- This is also my way at understanding progressively the expectations from the CSL spec, regardless of its test suite).

csl/core/engine.lua
Omikhleia: Note to myself if I need ever to need remember it: checked on `dragon:swank2017b`

Omikhleia: Done

csl/core/engine.lua
Omikhleia: Done

packages/bibtex/support/bib2csl.lua
Omikhleia: Reproduced, indeed. Fixed minimally as suggested in the above link. Seems to do the trick.

Omikhleia:
2024/07/14 "Stage 0" milestone: Successfully processed [1355 references](https://github.com/Omikhleia/awesome-sile-books/tree/main/bibliographies)
 - with en-US locale and styles chicago-author-date, chicago-fullnote-bibliography and apa.
 - with fr-FR locale and styles chicago-author-date, chicago-author-date-fr, chicago-fullnote-bibliography-fr and apa

Omikhleia:
2024/07/20 "Stage 1" milestone: Successfully processed [1508 references](https://github.com/Omikhleia/awesome-sile-books/tree/main/bibliographies),

- with entry sorting according to the CSL style.
- tested with fr-FR locale and styles chicago-author-date-fr, chicago-fullnote-bibliography-fr
- Some entries have belong to numbered "series" (a.k.a. collection-title and collection-number in CSL)

Omikhleia:
Soon leaving for vacations, so here are just some advancement notes to myself, in order to remember:
- [ ] Stage 2 = implement `subsequent-author-substitute` (I had it done more or less this week-end on an experimental ground, but I'm unhappy with the code so I didn't push it... I prioritized working on my bib files, now over 2000 references, and couldn't finish that code properly today...)
- [ ] Stage 3 = implement `page-range-delimiter` so page ranges would look decent...
- [ ] Stage 4 = Understand how to handle properly demoting/non-demoting particles in names (I've some of these in my bibliography files, so I guess it's time to dig into the topic...)
- [ ] Stage 5 = review package commands for multiple citations, and implement "locators".

That's a minimal set. There would still be a few missing features from the CSL spec, but at least all Chicago-styles would be covered fairly decently, and a first milestone would be passed.

Omikhleia:
Slowly back on track.
I rebased the branch, and added a commit with support for #2026 (see rationale in main description). Some tests performed with the American Chemical Society" (ACS) style, which uses the "citation-number".
We are not yet there, but it's a progress.
I also included silently ("in passing") a small refactor/fix for an issue I experienced with the Modern Language Association (MLA) style, which I used (with a few adaptations) for the 2600+ references in the book I made this summer, [_A bibliography of Tolkien studies in French & English_](https://sites.google.com/site/dragonbrumeux/bibliography) -- But there's still some code to clean-up and refactor from that work-in-progress ;)

Omikhleia:
Let's refactor a bit and support locators. It's a refactor, since none of this is release yet.

image

`This is demonstrated in \csl:cite[page=30-35]{FullInProceedings}; see also \csl:cite[fig=5, key=FullBook].`

Omikhleia:
I hate names with particles, definitively. :rofl:  -- Doh, it was hard for my tired brain. One checkbox ticked.

Omikhleia:
_The wild landscape of subsequent author substitutions_

(Or the reason why something started weeks ago and announced in a previous comment takes so long!)

There are lots and lots of styles in the CSL repository. They don't all use subsequent-author-substitute, but when the do (approx. 297 styles), there are around 35 different patterns. This doesn't mean much, so let's detail. It boils down to 7 different categories. With possibly some approximations and simplifications:
 - main case (210+): 1 to 7 **en-** or **em-**dashes (the majority of cases), possibly with trailing spaces or periods
 - empty = 21, mostly on styles with special display ("block" etc.)
 - 1 to 6 regular dashes = `-` (1), `---` (26), more dashes (4)
 - 3 to 10 underscores = three (4), more (15); with or without training spaces or periods
 - plain words (e.g. "id.") = 6
 - spaces only = 2
 - dubious = 2 ("???")

This brings a few considerations.
 - The empty cases are mostly used with block or column-like display. I haven't supported these cases in this initial implementation, so let's not bother with them for now...
 - Series of dashes, underscores, en-/em-dashes are rather idiosyncratic, probably a legacy from the early days of typesetting machines! Anyhow, unless one is very lucky with their choice of font, it never looks great on a typography point of view...
 - CSL "hanging-indent" is a mere boolean, unrelated with the length of substitution pattern. I mention it just in passing, as in my own preferences with MLA and Chicago styles, I'd like some consistency with the length of the substitution pattern and the hanging indent (e.g. 3em for both), heh.
 
Leaving aside the block/column cases for now, full/strict compliance is still complex and not that great typographically. I'll head therefore towards a few (debatable) "shortcuts"...

Then, there's the `subsequent-author-substitute-rule`... If I counted well, out of the 297 styles with author substitution, 16 are in "partial-each" mode, 4 in "complete-each", 0 in "partial-first". Let's say we can leave these aside for now, and assume the default "complete-all", this will cover a lot of standard cases.

But there's still another trap: "Substitution is limited to the names of the first `cs:names` element rendered. In the huge majority of cases, the first names are the authors (or their own substitutes when unknown), that is, a single list. The specification however does not forbid multiple variables in a `cs:names` (e.g. "author editor"), but as far as I can tell at a quick glance, it's not used much in the wild; that is, for the main contributors as first rendered field. The pattern occurs for secondary contributors etc. but of course it's not relevant for the substitution at stake. There's nothing that would prevent us from implementing the generic case, though it's kind of an extra refactor for a very uncertain benefit. So I'm heading again towards a "good enough" solution, for the sake of simplicity, which will work for "usual" styles such as Chicago and MLA.

The devil is in the details, as always. Yet, however many shortcuts are taken here, it's still superior to the "legacy" implementation we had. Simon's old 400-500 LoC experiment (682bbc51d8f2d0eadcdd6548dcf06c1f385da589) has lived long since 2015. Whether the present 2000-2500 LoC re-implementation will fare as long, only time will tell.

So it's a 80-20 situation. The 80% of the work is done, and the 20% remaining is the most complex to achieve. Whether it's really needed is hard to ascertain.
But it's a possible base for future improvements, maybe.
What I can sure say is that I am glad to see how good it performs for my bibliography project, compared to the older solution -- so it's a huge improvement already.

alerque:
Rebased commits just had `stylua` applied...

alerque:
Can you say a couple words about why you started off with this in `/csl` at the project root? I have some concerns about it being there, but am not sure where to put it. It could be an external LuaRock that SILE depends on by default, or it could be mixed around with some of it being a "utility" and some of it being in the bibtex package, or it could all just be bunged into the bibtex package (as I have at least temporarily done), or we could need some other top level asset handling space that hasn't occurred to me yet.

Am I correct in guessing part of the rational was to make a logical way for 3rd party packages to extend the CSL styles without feeling like they were reaching too deep into the bibtex package?

Omikhleia:
> Can you say a couple words about why you started off with this in /csl at the project root?

Nothing really fixed in my mind actually, but I don't like having too many things under "/packages" where they don't really belong. In my own classes and packages modules, I tend to have both "/packages/xxx" for the SILE package and "/xxx" for support or more general code.[^1] The latter could have been "/core/xxx", but what does "core" means anyway in such a context? And it's also easier to extract later, would this be moved in a separate component (= maybe SILE's core should eventually become lighter, and such things be part of an add-on package, under SILE's organization umbrella? E.g. if we ever want other formats than bib(la)tex being supported, for instance, without bloating the core distribution...)

> Am I correct in guessing part of the rational was to make a logical way for 3rd party packages to extend the CSL styles without feeling like they were reaching too deep into the bibtex package?

At this point, I don't fully know, to be honest.
Eventually, the bibtex package "as we know it currently" needs heavy refactoring.
There are things I didn't move yet outside of it in this PR, because it would have made the changes harder to follow and review.
I do think, though, that we need to discuss not only its very name, but how it would ideally be organized in the future.
My PR here is "somehow" conservative (I even kept the old implementation for now, but I think we'd need to discuss how to handle its deprecation in a major release at some point...). Eventually, we should envision a... er... "bibsile" or whatever better-named package, handling the user-facing commands (citations, biblio loading and printing) regardless of the format and lower-level support details. We are not here yet, so that might be just a dream... a nice one, but directions toward it might be straightforward ;)

In other terms, I didn't think that far yet. I organized a few things as I would have done in personal modules, but also kept lots of things where they currently were, even not liking the look of it[^2] -- just to get something working and still readable, considering a broader refactor is nevertheless needed at some point, where we have a clearer vision on the target and proper deprecation policy or the old stuff.

[^1]: E.g. in resilient.sile, I have "/packages/resilient/xxx" and "/classes/resilient/xxx", but also "/resilient" for some common code. It's my own way of trying to structure things, maybe, which are not that dependent on a given package or class implementation strictly speaking, and could be used outside of these, yet without cluttering a "core", "utilities" or whatever folder (which is far too generic, I'd tend to think).
[^2]: Simons' borrowed bits from nbibtex for instance... They offered a quick "working" solution, but it would need re-questioning the approach perhaps (the Lua port is hard to read and modify).

Omikhleia:
Oh, what I also mean here is that _maybe_ at one point we'd like to push the CSL engine as a standalone library? (It always sound interesting said this way, but extracting the rough.js port at a higher level in my own ptable.sile, with the idea that _eventually_ it could be proposed outside SILE, was never fully completed: These things look simple on the surface but always take more time that first considered when it comes to make it right)

Omikhleia:
That is, you can of course "Move CSL support module unter bibtex package", though I don't know what CSL and bibtex have in common :D

alerque:
>  hough I don't know what CSL and bibtex have in common

Just that CSL in only used in connection with bibliographies, and the only (poorly named) implementation of that we have is our bibtex package. Keeping all the related utilities together makes it much easier to package and maintain. If/when we do have other packages we can consider whether abstracting the CSL stuff under more general utilities makes sense.

Putting anything in the root of this project comes with some caveats for packaging and distribution and I'd rather not deal with that unless I really understand why that namespacing choice is warranted.

> These things look simple on the surface but always take more time that first considered when it comes to make it right)

Sure, but it also isn't *that* hard. I'm still seriously considering it for this module.

I looked through the sources and it seems like the only real issue is the use of SILE's `justenoughicu` for casing. That would create a circular dependency. But I think we could easily replace that with e.g. my own `decasify` Lua Rock. It can handle the lower, upper, title and even sentence casing.

Omikhleia:
> Putting anything in the root of this project comes with some caveats for packaging and distribution

I understand.
No problem, and I don't really care where files end up being located in the current implementation (I was just answering your question why I put them this way originally) -- At some point I'll open an "Epic" issue and list a few further ideas for the future of the bibliography support -- some of which also depends on the traction this feature gets (vs. the required effort), so I'm eager to see it live and start pondering how to build upon it further :smile:

alerque:
I think I'm going to go ahead with this namespacing (under bibtex) and we can refactor from there. I'd still consider extracting this to an external library, but to do that the interaction between SILE and the CSL engine interface should be much clearer, and potentially interchangeable with another implementation (e.g. a Rust one). A couple hundred kb of code we're still refactoring can live here until such a time as we have a clearly better plan, and also clearly moving ahead with this is better than what we have now.

When we get BCP-47 locale stuff straitened out the actual bundled styles can probably move to the language module, but that should be a non-breaking change later.

Omikhleia:
> I haven't looked for CSL engines yet

I have - there's Typst [Hayagriva](https://github.com/typst/hayagriva).

(_tongue-in-cheek on_) Typst has a huge active community, it also has a fairly good indexer, and plenty of advanced modules. (_tongue-in-cheek off_)

Last time I checked this summer, entry sorting was not language-dependend ... checking ... yep, the issue is still open. Well they don't have ICU yet under the hood.

alerque:
I hear you on the hype train. Don't get me wrong Typst has done some things *right* and *well*. It's fast and does some things very well. But I was quite surprised at the amount of attention it got before it supported basics like footnotes. Even now it doesn't support flexible vertical spaces. Supporting that alone is the cause of much of our nasty pushback problems and a huge chunk of our time, but it's widely used in publishing for good reason. Anyhow I am more convinced than ever that there is room for more than one approach here. Typst is doubling down on being a sandboxed environment with no tinkering (unless you build your own) and only its own dedicated syntax, and SILE allows you access to call out to anything external and tinker with anything internal and process whatever input formats you want.

Anyhooooo, yes Hayagriva also look like it took the approach of managing all it's own data in their own YAML format instead of existing bibliography data formats, so I'd view it more as an alternative backend we could support rather than a pathway to our primary support.

Omikhleia:
@zepinglee Indeed, thanks for pointing it! Seeing the online demo again, I'm pretty sure I looked at it too this summer, albeit very briefly, and had some concerns.
Well, there are philosophical and political decisions behind the possibilities of a shift to Rust evoked in the recent exchanges, and the use of such components in SILE - A meta-discussion would be welcome at some point, in the appropriate place.

Issue #2086, opened 2024-07-22
`SILE.scratch.headers` shouldn't be declared only in `twoside`?


jodros:
Just now I got: 

`Error: runtime error: /usr/local/share/sile/packages/twoside/init.lua:123: attempt to index field 'headers' (a nil value)`

Well, since this table is only used by `twoside` dependents, it should be declared there instead of having to declare whenever one creates a new class/package that relies on it.

Omikhleia:
Minimum working example and test?

'Cause we can go on monkey patching, sure, but you do realize "headers" are broken to their core, right? (#1842, #1291)

Omikhleia:
Also: "scratch" variables ought to get killed :)

jodros:
> Also: "scratch" variables ought to get killed :)

Wow, so the refactoring is bigger than I thought... What's going to be in its place?

> 'Cause we can go on monkey patching, sure, but you do realize "headers" are broken to their core, right? (https://github.com/sile-typesetter/sile/issues/1842, https://github.com/sile-typesetter/sile/issues/1291)

I'll take a look over this.

Omikhleia:
> > Also: "scratch" variables ought to get killed :)

Package A or class C tinkering with scratch variables from package B breaks all separation of concerns and deeply ties the implementations (making the scratch variable a part of the public API of a package, that any "equivalent" replacement would have to honor (beyond just providing commands). So the concept is somewhat defective by design.
 
> > 'Cause we can go on monkey patching, sure, but you do realize "headers" are broken to their core, right? (#1842, #1291)
> 
> I'll take a look over this.

Aside note: I have (partly) solved some of my concerns with my resilient.book class and my resilient.headers package -- though I know they still have some of above-mentioned issues regarding proper separation of concerns.

jodros:
> breaks all separation of concerns and deeply ties the implementations

Now I got it, but what's going to replace all those scratch variables?

Omikhleia:
> Now I got it, but what's going to replace all those scratch variables?

I'm afraid it's a case by case basis, due to the different use cases for them, and the lack of global styling approach in the core distribution.

Omikhleia:
See also https://github.com/sile-typesetter/sile/issues/1417

In many cases indeed, private members of the package might be a better approach.

alerque:
Yes, all `SILE.scratch` variables need to die. All of them. 100% no mercy.

[...]

PR #2087, opened 2024-07-22
Declare SILE.scratch.headers in twoside only

jodros:
#2086

### DIFF
--- a/classes/bible.lua
+++ b/classes/bible.lua
@@ -6,10 +6,6 @@ local plain = require("classes.plain")
 local class = pl.class(plain)
 class._name = "bible"
 
-if not SILE.scratch.headers then
-   SILE.scratch.headers = {}
-end
-
 class.defaultFrameset = {
    content = {
       left = "8.3%pw",
--- a/packages/twoside/init.lua
+++ b/packages/twoside/init.lua
@@ -72,6 +72,9 @@ function package:_init (options)
    end)
    self.class:registerHook("newpage", spreadHook)
    self.class:registerHook("newpage", switchPage)
+   if not SILE.scratch.headers then
+      SILE.scratch.headers = {}
+   end
 end
 
 function package:registerCommands ()
--- a/classes/book.lua
+++ b/classes/book.lua
@@ -52,9 +52,6 @@ function class:_init (options)
       insertInto = "footnotes",
       stealFrom = { "content" },
    })
-   if not SILE.scratch.headers then
-      SILE.scratch.headers = {}
-   end
 end
 
 function class:endPage ()


classes/bible.lua
```diff
@@ -6,10 +6,6 @@ local plain = require("classes.plain")
 local class = pl.class(plain)
 class._name = "bible"
```
Omikhleia, line 7: This class is a joke.

When will we kill it?

packages/twoside/init.lua
```diff
@@ -72,6 +72,9 @@ function package:_init (options)
    end)
    self.class:registerHook("newpage", spreadHook)
    self.class:registerHook("newpage", switchPage)
+   if not SILE.scratch.headers then
+      SILE.scratch.headers = {}
+   end
```
Omikhleia, line 77: The only place "twoside" uses headers scratch are in `\open-spread` -- but is it approached as it should, there?

Is even this command `\open-spread` in the right package/module? Is the solution general? (The answer is obviously "no" to me, so I'd like seeing good rationales for it...)

classes/bible.lua
jodros: > This class is a joke.

Why?

packages/twoside/init.lua
jodros: > Is even this command \open-spread in the right package/module?

I have no idea... Where do you think it should be?

classes/bible.lua
Omikhleia: Er? What is supposed to be a "bible" class?

Food for thought: I did typeset many bibles (https://github.com/Freely-Given-org/BibleTypesetter/pull/3 -- Still pending any sort of review and approval, which is sad, but anyway) without needing a "bible" **class**.

So what is supposed to be a "bible" class in layman terms?

RobH123:
On 23/07/24 09:24, Omikhleia wrote:
> Food for thought: I did typeset many bibles 
> (Freely-Given-org/BibleTypesetter#3 
> <https://github.com/Freely-Given-org/BibleTypesetter/pull/3> -- Still 
> pending any sort of review and approval, which is sad, but anyway) 
> without needing a "bible" *class*.
Yes @Omikhleia, I apologise for the slow reaction, but it's not 
forgotten and still on my to-do list. I'm still very interested in 
evaluating your work and hopefully extending/expanding it, it's just 
that other priorities still prevail for now.

packages/twoside/init.lua
Omikhleia: > I have no idea... Where do you think it should be?

My deep feeling is that only the **class** knows how to handle page breaks and what do to with "empty" pages, (based on the reason why it it introducing them with so-called spreads).

It's hard to explain with the "limited" book class provided by SILE, but I can try, based on my experience designing books with my own resilient.book class (again, it does have some flaws I am aware of, however)[^1]:

 - Consider a `\part` (yep, resilient has these) followed by a chapter, and both open on odd pages, so a blank even page is gonna be inserted in between. Are you in a "novel" or in a "technical documentation"? Why I am asking? 'Cause in a novel, usually, the book name is on even pages (so it's not a problem to repeat it, at least), but in a technical documentation, the chapter name is usually on even pages (so it has to be cleared after a part...)
 - Consider a `\chapter` followed by another chapter, also opening on an odd page[^2], and a blank even page might be needed. Same question, but trickier: to which chapter belongs the empty page? Should it have a header (not to say a footer), and which one? There's no simple answer, except that "it depends" on some publisher preferences...

And here we're indeed discussing parts, chapters, etc. -- that is, things that are (currently) a class-design topic.

[^1]: If you read its doc, you'll see for instance that resilient uses its own header logic, and that it has provision for "novel" and "technical" headers... Though the underlying implementation is a bit of a hack.
[^2]: Though that ought to be configurable: Some books _do_ allow chapters to begin on even pages... No wonder it took me years to design my resilient.book solution, as all styling decisions are configurable -- and suddenly, all sort of edge cases pop up :)

jodros: Thanks for your explanation, I got your point.

Well, I did this PR without ever thinking about `twoside` not being the most suitable place for `open-spread` or that scratch is about to deprecate... For me it was way simpler.

classes/bible.lua
alerque: Bible typesetting comes with a standard set of features over and above what most books require. Defaulting to grid typesetting, having headers that keep track of the verse at start and end of pages for use in headers, multiple types of footnotes/cross references, etc. The *bible* class should build on the book class with a few default features available that cover common needs that aren't so common in regular books. Of course you don't have to use it, it's just a starting point.

Additionally it's where I'd put stuff to cover default processing functions for OSIS/USFM/whatever tags that don't have any relevance to non-Bible content.

PR #2095, opened 2024-08-24
Make document state available during class's post init

### LINKED ISSUE #2083: Is there anything like a start hook?
I mean, something just like class:finish but to be run as the first thing just after the class is initialized?

I've tried `self:registerPostinit` but it doesn't seem to be what I thought it was for... BTW its actual function deserves to be in the documentation...

alerque:
Fixes #2083, although I'm unsure if this is actually a fundamental improvement or if it just makes a bigger mess of variable scopes. Catching the error and suggesting preambels might be a better route.

### DIFF
--- a/classes/base.lua
+++ b/classes/base.lua
@@ -281,6 +281,22 @@ function class:registerLegacyPostinit (func, options)
    end)
 end
 
+--- Register a callback function to be executed after the class initialization has completed.
+-- Sometimes a class or package may want to run things after the class has been fully initialized. This can be useful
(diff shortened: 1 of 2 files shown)


alerque:
I'm going to go ahead with this "fix". It still isn't ideal, but whether we're munging some global state variable from the inputter or the class doesn't seem to be consequential. The right answer is to stop munging global state, but this makes more sense to me and should make eventually refactoring it out entirely that much easier to accomplish.

I do **not** think this change is the right way to "fix" the issue in #2083, that should still be done with a preamble. But it does "fix" the issue in that we document what the related API is for.

PR #2096, opened 2024-08-24
Apply spelling fix suggestions from typos

alerque:


packages/bibtex/support/isodatetime.lua
```diff
@@ -98,7 +98,7 @@ local enddateonly = slash
    * date
    / function (t)
       return {
-         aproximate = t.approximate,
+         approximate = t.approximate,
```
alerque, line 101: @Omikhleia Not very many of these spelling fixes affects function in any way, but this one looks like it might amount to a bug fix, no?

Meanwhile there is another one `typos` is pointing out might be wrong that I can't verify. Can you look at this:

https://github.com/sile-typesetter/sile/blob/master/packages/bibtex/support/bibmaps.lua#L48

alerque: Never mind, I figured out this is working around a BibTeX oddity.

PR #2101, opened 2024-09-05
Escape Unicode so over-zealous code linters don't flip out

alerque:
Pun intended.

Looking at you Fedora, GitHub, and browser based editors.

c.f. https://artifacts.dev.testing-farm.io/0eb6559c-43aa-43fd-a0f7-b653dfd5013e/
c.f. https://bugzilla.redhat.com/show_bug.cgi?id=2149698#c37

### DIFF
--- a/outputters/text.lua
+++ b/outputters/text.lua
@@ -57,13 +57,13 @@ function outputter:setCursor (x, y, relative)
          if newy - cursorY > bs then
             outfile:write("\n")
          else
-            outfile:write("‫")
+            outfile:write("\u{202b}")
          end
       elseif newx > cursorX then
          if newx:tonumber() - cursorX:tonumber() > spc then
             outfile:write(" ")
          else
-            outfile:write("‫")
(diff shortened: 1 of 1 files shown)


hegjon:
I think that this change is good unrelated to the linters, since I would not see the difference between `202b` and normal parenthesis.

hegjon:
Forget my last comment, GitHub confused me

alerque:
@Omikhleia I did suggest Fedora users may be better off going back to their Linotype machines :wink:

@hegjon I will go ahead with this even if not required to pass lints because GitHub and even some code editors I tried are hopelessly confused about how to render this code. I've never considered this a problem working in the terminal or NeoVIM (even via GUIs like NeoVide) but clearly it would be tricky for some folks to even edit this code. Since it's a relatively easy fix...

alerque:
The Rust lint error here is a new one in Rust 1.81 which just dropped. I fixed it separately in *master*.

PR #2105, opened 2024-09-10
Extend SU.collatedSort for complex table sorting

Omikhleia:
I noticed when working on #2082 : Our `SU.collatedSort()` works on list of strings only:

```lua
local names = {"Charlie", "Bob", "Alice"}
SU.collatedSort(names)
--  gives: {"Alice", "Bob", "Charlie"}
```

My own bad at the time (#1632)... but we can't use it to sort structured (key-value) tables with a comparison callback, as one would do with `table.sort()`.

Of course there's a workaround, but it's a bit annoying (comparing for string equality, then performing a second ICU comparison for ordering)

So this PR is a proposal for supporting it, as:

```lua
-- Sort by age then name
local namesAndAges = { {name="Charlie", age=25}, {name="Bob", age=30}, {name="Alice", age=25} }
SU.collatedSort(namesAndAges, nil, function (a, b, stringCompare)
  if a.age < b.age then return true end
  if a.age > b.age then return false end
  return stringCompare(a.name, b.name) < 0
end)
-- gives { {name="Alice", age=25}, {name="Charlie", age=25}, {name="Bob", age=30} }

-- Sort by name then year
local namesAndYears = { {name="Alice", year=2005}, {name="Charlie", year=1995}, {name="Bob", year=1990}, {name="Alice", year=1995} }
SU.collatedSort(namesAndYears, nil, function (a, b, stringCompare)
  local nameCompare = stringCompare(a.name, b.name)
  if nameCompare < 0 then return true end
  if nameCompare > 0 then return false end
  return a.year < b.year
end)
-- gives { {name="Alice", year=1995}, { name = "Alice", year = 2005 }, {name="Bob", year=1990}, {name="Charlie", year=1995} }

```

In other terms:
 - Be less clever in the C code and just return -1, 0, 1 for (less, equal, greater)
 - Use a closure to pass ICU-enabled string comparison to the user callback.

### DIFF
--- a/core/utilities/sorting.lua
+++ b/core/utilities/sorting.lua
@@ -1,21 +1,30 @@
 --- Table sorting with language-dependent collation.
 -- @module SU.sorting
+--
 
 local icu = require("justenoughicu")
 
 local collatedSort = {
    -- No ICU for language "und", fallback to 'natural' table.sort
-   und = function (t, _)
-      table.sort(t)
+   und = function (t, _, comparator)
+      if comparator then
+         table.sort(t, function (e1, e2)
+            return comparator(e1, e2, function (s1, s2)
+               return s1 < s2 and -1 or s1 > s2 and 1 or 0
+            end)
+         end)
+      else
+         table.sort(t)
+      end
    end,
 }
 
 setmetatable(collatedSort, {
-   __call = function (self, t, options)
+   __call = function (self, t, options, comparator)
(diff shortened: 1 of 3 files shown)


Omikhleia:
Oops sorry for the multiple push force... Was on wrong branch :)

alerque (approved):

PR #2113, opened 2024-09-22
Fix syntax tree differences between commands and environments

### LINKED ISSUE #2106: AST differences between begin-end and command syntax
According to the documentation, the following should be equivalent:

Option 1:

```tex
\itemize{
	\item{Item 1}
	\item{Item 2}
}
```

Option 2:

```tex
\begin{itemize}
	\item{Item 1}
	\item{Item 2}
\end{itemize}
```

Option 3:

```xml
<itemize>
	<item>Item 1</item>
	<item>Item 2</item>
</itemize>
```

Indeed, options 2 and 3 provide identical output.  Option 1; however, crashes sile:

```sh
SILE v0.14.17 (Lua 5.4)
<test.sil> as sil

! Unexpected Lua error

error summary:
	Processing at: test.sil:2:2: in \itemize
	Using code at: ...Cellar/sile/0.14.17_1/share/sile/packages/lists/init.lua:76: attempt to concatenate a nil value (local 'cmd')

Run with --traceback for more detailed trace leading up to errors.
```

Full test case:

```tex
\document[class=book]{\use[module=packages.lists]
\itemize{
	\item{Item 1}
	\item{Item 2}
}
}
```

Full working version:

```tex
\document[class=book]{\use[module=packages.lists]

\begin{itemize}
	\item{Item 1}
	\item{Item 2}
\end{itemize}
}
```

Omikhleia:
Closes #2106 

Well, I hope, this SIL thing is tricky. All tests seem to pass, but more eyes are probably needed.

On that occasion:
 - I added a new pseudo-backend "ast":
    -  `sile somefile -b ast` dumps the raw syntax tree as produced by the inputter, and exits (skipping processing etc.)
    - It's quite handy for debugging inputters and comparing the generated tree,
    - It's actually something I long wanted in one form or another when I developed inputters (e.g. https://github.com/Omikhleia/markdown.sile/issues/15#issuecomment-1312460731)
    - [It is left as an exercice for developers of good will to implement a naive inputter for that syntax tree dump, so one could ensure full round-tripping and possibly insert filters to modify the tree outside SILE if they want ("à la Pandoc" filter concept)]
 - The produced tree is somewhat more compact than before:
    - Removing a level in case of content tables with just one element
    - Removing the `id = "content"` parsing artifacts entirely
 - The `-d inputter` debug produces a much more readable output


Omikhleia:
> All tests seem to pass

Well they passed on my branch. Seems master upstream got broken in the meantime due to b74deaaeabdd52e6102c1be1419ab3276058dc76 (= @alerque upgrading the Libertinus font, congrats for the new release by way.)

```
Passed in:
(string) 'Version 7.050;RELEASE'
Expected:
(string) 'Version 7.020;RELEASE'
```

inputters/sil.lua
```diff
@@ -80,49 +80,84 @@ local function getline (str, pos)
...
+      -- These nodes have multiple children, which need recursion.
+      SU.debug("inputter", sep and (sep .. "Processing command"), tree.command, #tree, "subtrees")
+      for i, node in ipairs(tree) do
+         if type(node) == "table" then
+            SU.debug("inputter", sep and (sep .. " -"), node.id or "table")
+            tree[i] = ast_from_parse_tree(node, doc, depth)
+            -- Simplify the tree if it's just a list
+            if type(tree[i]) == "table" and not tree[i].id then
+               SU.debug("inputter", sep and (sep .. " -"), "Collapsing subtree")
+               SU.splice(tree, i, i, tree[i])
```
Omikhleia, line 135: Erm. A splice in a for loop on the tree might not be wise.

inputters/sil.lua
```diff
@@ -80,49 +80,84 @@ local function getline (str, pos)
...
-      return tree[1]
-   elseif false or tree.id == "content" or tree.id == "environment" or tree.id == "command" then
-      SU.debug("inputter", "  - Massage in place", tree.id)
-      for key, val in ipairs(tree) do
-         SU.debug("inputter", "    -", val.id)
-         if val.id == "content" then
-            SU.splice(tree, key, key, massage_ast(val, doc))
-         elseif val.id then -- requiring an id discards nodes with no content such as comments
-            tree[key] = massage_ast(val, doc)
+      -- Thes nodes have only one child, which needs recursion
```
Omikhleia, line 117: Thes > These ;)

inputters/sil.lua
Omikhleia: Nah not wise at all. Funny no tests caught it.

alerque:
Any reason the AST outputter should be a hack and not a first class output option? And inputter for that matter.

Standby, hacking in progress.

Omikhleia:
> Any reason the AST outputter should be a hack and not a first class output option? And inputter for that matter.

Not sure what you mean, but there should be a way to bypasses all later usual steps (typesetter, shaper, outputter). Most of the time when dumping the AST, one doesn't want other outputs.

alerque:
> but there should be a way to bypasses all later usual steps (typesetter, shaper, outputter). Most of the time when dumping the AST, one doesn't want other outputs.

I could see a use case for both (see commit message notes) but for starters here is what I was thinking for the mode you describe that just dumps the initial parsed AST with no further processing.

Omikhleia:
Also this commit:
```
Additionally it is not a full serialization of the AST that could be
read back in with an input driver. We'd need a new dependency on
something like serpent for that, which would be a lovely next step.
```

Can you define "a full serialization of the AST" in a dedicated topic ?
(Reading the parse tree dumped by the inputter doesn't need a new dependency)

"Something like serpent" : the Python library? Something else? And what for?

alerque:
> I can't see any real use case.

This one.

The hack you put in is no different, this just allows other outputters to intercept the process at the same step and dump their own version of the input data. The real use case here is for the outputter to be able to shorten the processing loop and tell SILE to skip the command processing → typesetter → shaper → pagebuilder → framer → nodes etc.

If you have a better idea for naming or API that allowed the outputter to short circuit and get access to the output of the inputter I'm quite open to adjustment; but I would expect a 3rd party package to be able to access the same functionality we're using internally, hence special casing it with a hack didn't feel right.

alerque:
> Can you define "a full serialization of the AST" in a dedicated topic ?

I would expect a full serialization to be able to be saved to a file, then read back later and used to output the same document one would have gotten the first time. This would be something roughly equivalent to Pandoc's `Native` format.

> (Reading the parse tree dumped by the inputter doesn't need a new dependency)

Yes it does.

SILE's AST can contain anonymous Lua functions as a top level content type. A serializer would be able to dump the content of those functions as chunks that could be read back and run again later. Penlight's `pl.pretty.write()` is only a pseudo serialization that abbreviates and even entirely skips over things it doesn't understand. Compare it with Serpent's output:

```console
$ lua -e 'print(require("pl.pretty").write({ content = function () print("6") end }))'
{
  content = "function: 0x63035cb4d650"
}

$ lua -e 'print(require("serpent").block({ content = function () print("6") end }, { comment = false }))'
{
  content = ((loadstring or load)("\27LuaT\0\25\13\n\26\n\4\8\8xV\0\0\0\0\0\0\0\0\0\0\0(w@\1=(command line)\0\0\2\11\0\0\0\0\0D\0\2\1G\0\1\0\4print\46\0\0\0\0\0\0\0_ENV",'@serialized'))
}
```

The later can be deserialized and run again later, the former is useful for visually scanning what was going on but does not preserve the contents of the anonymous function.

Omikhleia:
> Presumably you meant to add a reply comment not edit & remove mine, right?

Doh yes, I must have misclicked "Edit" instead of "Quote".

Restored, I think ^^

My comment was supposed to be a reply with a quote:

> SILE's AST can contain anonymous Lua functions as a top level content type.

I am not going to rehearse #1834.
(_As of today_, we don't have any inputter that produces functions in its parse tree, _fortunately_.)



> Presumably you meant to add a reply comment not edit & remove mine, right?

alerque:
> _As of today_, we don't have any inputter that produces functions in its parse tree, _fortunately_.

Sure we do. In v0.13.3 we formalized a raw Lua inputter, but various less formalized ways of accomplishing it have existed since before I got involved in SILE. One of the things it can do is serve as an interface to typeset content being generated on the fly in another application (or even language) by feeding SILE a Lua "document" that is a closure with access to the bits it needs to process a data stream incrementally and getting started before the application terminates a known end point is reached, hence allowing a tree to be.

Whether this is a "fortunate" capability or not is debatable, but I use it extensively. One of SILE's strengths is allowing you to tinker with its internals. That's also a major contributing factor to it being so finicky sometimes.

More to the point, as I've been reviewing your work here I think what you want may not be an outputter at all (pseudo or otherwise). Is it possible that all you really wanted for this was an extra debug call or two such that just `sile -d inputter -d ast -b dummy document.sil` could show you what the inputter had produced pre-processing?

And what I what I have in mind as an outputter is something else entirely. I'll move that to another PR.

alerque:
Does ea58f4c accomplish what you had in mind with the pseudo backend? Pairing it with the dummy backend to get no other output should net you about the same affect:

```console
$ ./sile -d inputter -d ast -b dummy tests/absmin.xml
SILE v0.15.5.r60-g822a74f (LuaJIT 2.1.1700206165) [Rust]

[inputter] Running content type detection round 1 with xml<tests/absmin.xml> as xml

[inputter] Dumping AST tree before processing...
{
  "test",
  col = 1,
  command = "sile",
  id = "command",
  lno = 1,
  options = {
  },
  pos = 1
}
[ast] [tests/absmin.xml
[ast]   [test]
[ast] ][1]
[ast] [<nowhere>
[ast]   [1]
[ast] ]
```

If having the ast debugging stuff in there or needing two flags is cumbersome, maybe a new flag like `raw_ast` or whatever would work. Thoughts?

Omikhleia:
Let's go very simple.

We can remove the `-b ast` part from this PR. I moved too quickly, you moved too quiclky too.  Let's repurpose it on its own discussion, for later, and stay here with the _only_ part that was at stake: "Fix syntax tree differences between commands and environments"

Fair enough proposal?

Omikhleia:
> Did you look at the commit list?

I'll be real honest: nope. I am working tonight (er... or this weekend...) on releasing new versions of markdown.sile and resilientt.sile (leveraged with the cool stuff that went in SILE 0.15, now that I'm on it and can add new features it allows), so I'm a bit lost in between many things. Sorry if I might have sounded harsh (to the point of an unwanted misclisk, lol, it tells a lot) earlier; Let's go for what we know to fix something. Let's re-discuss the AST later, with no pressure: It's actually a very interesting topic on its own (and I'm sure we can come to a common understanding and improve things here alltogether)

PR #2116, opened 2024-09-27
Fix system locale availability in containers (and overhaul Docker builds)

alerque:


alerque:
Somehow the container image size doubled. It should have gotten a tiny bit smaller. Somewhere there is an oops.

alerque:
Blooper fixed. It still grew a little, but the locale database we now carry with matching localizations for system stuff to match SILE's locales is 99Mb by itself, so the gain we got from having a few less dependencies were canceled by having stuff we do want.

PR #2119, opened 2024-09-29
feat(math): Support square roots

### LINKED ISSUE #1604: MathML msqrt not supported
SILE 0.14.4

Document:
```
<document>
<use module="packages.math"/>

<mathml display="block"><mfrac><mrow><mo>−</mo><mi>b</mi><mo>±</mo><msqrt><mrow><mi mathvariant="normal">Δ</mi></mrow></msqrt></mrow><mrow><mn>2</mn><mi>a</mi></mrow></mfrac></mathml>

</document>
```

(I am not MathML-savvy, the MathML formula was generated with [TEMML](https://temml.org/) from LaTeX ` \frac{-b + \sqrt {\Delta}}{2a}`)

Expected:

image

Observed:
```
! Unknown math command msqrt at xmath.xml:50:1: in \mathml["display",display="block"]
```

Omikhleia:
Closes #1604 

**My dear math friends**... Do you really expect SILE will thrive without the square root? I don't know for other countries, but in France, at least when I was younger, the solution to second-degree equations was taught in school much earlier than integrals, matrices and other advanced stuff.

So we shall all want square roots, **even** if it has to be a hack of some sort.

Not supporting these is **detrimental** to SILE, when other **worse** solutions are proposed to **typists** (_pun intended and completely assumed_)

As I just said...

image

### DIFF
--- a/packages/math/base-elements.lua
+++ b/packages/math/base-elements.lua
@@ -1334,6 +1334,85 @@ end
 
 function elements.table.output (_) end
 
+elements.sqrt = pl.class(elements.mbox)
+elements.sqrt._type = "Sqrt"
+
+function elements.sqrt:__tostring ()
+   return self._type .. "(" .. tostring(self.radicand) .. ")"
+end
+
+function elements.sqrt:_init (radicand)
+   elements.mbox._init(self)
+   self.radicand = radicand
+   table.insert(self.children, radicand)
+   self.relX = SILE.types.length(0) -- x position relative to its parent box
+   self.relY = SILE.types.length(0) -- y position relative to its parent box
+end
+
+function elements.sqrt:styleChildren ()
+   self.radicand.mode = self.mode
+end
+
+function elements.sqrt:shape ()
+   local mathMetrics = self:getMathMetrics()
+   local scaleDown = self:getScaleDown()
+   local constants = mathMetrics.constants
+
+   self.radicalRuleThickness = constants.radicalRuleThickness * scaleDown
+   if self.mode == mathMode.display or self.mode == mathMode.displayCramped then
+      self.radicalVerticalGap = constants.radicalDisplayStyleVerticalGap * scaleDown
+   else
+      self.radicalVerticalGap = constants.radicalVerticalGap * scaleDown
+   end
+   self.extraAscender = constants.radicalExtraAscender * scaleDown
+
+   -- HACK: More or less ad hoc values, see output method for more details
+   self.symbolWidth = SILE.shaper:measureChar("√").width
+   self.symbolHeight = SILE.types.length("1.1ex"):tonumber() * scaleDown
+
+   self.width = self.radicand.width + SILE.types.length(self.symbolWidth)
+   self.height = self.radicand.height + self.radicalVerticalGap + self.extraAscender
+   self.depth = self.radicand.depth
+   self.radicand:shape()
+   self.radicand.relX = self.symbolWidth
+end
+
+local function _r (number)
+   -- Lua 5.3+ formats floats as 1.0 and integers as 1
+   -- Also some PDF readers do not like double precision.
+   return math.floor(number) == number and math.floor(number) or tonumber(string.format("%.5f", number))
+end
+
+function elements.sqrt:output (x, y, line)
+   -- HACK FIXME:
+   -- OpenType might say we need to assemble the radical sign from parts.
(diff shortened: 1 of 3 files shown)


Omikhleia:
Note: I would have wished for the "inline" (text mode) last formula to make the a<sub>n</sub> smaller and smaller. Would you help, oh great math friends?

Omikhleia:
Be my guests:

```
\begin[class=book, papersize=a6]{document}
\use[module=packages.math]
\font[family=Libertinus Serif]

We want square roots...
\math[mode=display, numbered=true]{\sqrt{-1} = i}

Display:
\math[mode=display, numbered=true]{x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}}

Or inline: \math{x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}}

\math[mode=display, numbered=true]{c = \sqrt{\frac{E}{m}}}

Or inline: \math{c = \sqrt{\frac{E}{m}}}

Let's go mad, for any sequence \math{a_n > 0}:
\math[mode=display, numbered=true]{\sqrt{a_1 + \sqrt{a_2 + \sqrt{a_3 + \sqrt{a_4 + \sqrt{a_5 ...}}}}}
 = e^{\sum_{n=1}^{\infty} \frac{\mi{ln}\, a_n}{2^n}}
}

Inline: \math{\sqrt{a_1 + \sqrt{a_2 + \sqrt{a_3 + \sqrt{a_4 + \sqrt{a_5 ...}}}}}
 = e^{\sum_{n=1}^{\infty} \frac{\mi{ln}\, a_n}{2^n}}}

\end{document}
```

packages/math/typesetter.lua
```diff
@@ -125,6 +125,12 @@ function ConvertMathML (_, content)
          SU.error("Wrong number of children in mfrac: " .. #children)
       end
       return b.fraction(children[1], children[2])
+   elseif content.command == "msqrt" then
+      local children = convertChildren(content)
+      if #children ~= 1 then
+         SU.error("Wrong number of children in msqrt")
+      end
+      return b.sqrt(children[1])
```
Omikhleia, line 133: Ah, I was mislead by mozilla's docs, but mathml says msqrt can have multiple children, considered in an implicit mrow. So likely we'd need to remove the condition and use: 
```
return b.sqrt(b.stackbox("H", children))
```

Omikhleia: Confirmed and fixed: Mozilla has even an example of it[^1], so their sentence "The square root accepts only one argument" is just imprecise. 

[^1]: Test 13 from their old [test page](https://fred-wang.github.io/MathFonts/mozilla_mathml_test/).

alerque (dismissed):

packages/math/base-elements.lua
```diff
@@ -1334,6 +1334,79 @@ end
...
+   local d = self.depth:tonumber()
+   local sw = self.symbolWidth
+   local dh = h - self.symbolHeight
+   local symbol = {
+      self.radicalRuleThickness, "w", -- line width
+      2, "j", -- round line joins
+      sw, self.extraAscender, "m",
+      sw * 0.4, h + d, "l",
+      sw * 0.15, dh,  "l",
+      0, dh + 0.5, "l",
```
Omikhleia, line 1396: We actually need a rounding helper here, some PDF readers do not like double precision numbers at all.

```
local function _r (number)
  -- Lua 5.3+ formats floats as 1.0 and integers as 1
  -- Also some PDF readers do not like double precision.
  return math.floor(number) == number and math.floor(number) or tonumber(string.format("%.5f", number))
end
```

Omikhleia: Fixed!
I'm done here with this (inventive but) poor man's square root ;)

khaledhosny: I see code in `elements.text:stretchyReshape` that uses math variants, so why not use it to draw the radical glyph?

alerque (approved):

packages/math/base-elements.lua
Omikhleia: > ... so why not use it to draw the radical glyph?

'Cause it didn't happen in 2+ years? It seems to me something quick is better than nothing, then...

Maybe too 'cause the stretchy reshape has a big TODO in its code: it doesn't do the glyph assembling where there's no variant available (e.g. no big enough delimiter), and this is going to be another hornets' nest on its own.

PR #2125, opened 2024-10-03
feat(languages): Add bibliography localizations for Swedish

illfygli:


alerque (approved):
Thanks for taking the time to contribute! I hope other Swedes can benefit. This should make it into the next release, which should be v0.15.6.

illfygli:
@alerque Nice, thank you!
I hope so too. I just found SILE, and ran into bibtex trouble due to the missing translation when using Swedish. Now that will work at least. :)

alerque:
Just FYI there is some more bibtex related stuff including basic CSL support in another pending PR (#2082), you might be interested in checking that out. Depending on how you are going about running SILE there are ways to try out development branches (Docker, Nix Flake, etc.). Feel free to open a discussion issue with your platform details if you need tips on the incantation for running develop versions prior to release.

alerque:
Also for this particular case you can just put the messages you want to set/override in your document until SILE support gets released:

```sile
\begin{ftl}
bibliography-and = och
bibliography-edited-by = Redigerad av { $name }
bibliography-et-al = m.fl.
bibliography-translated-by = Översatt av { $name }
hello = Hej <em>{ $name }</em>!
tableofcontents-not-generated = Kör om SILE för att bearbeta innehållsförteckningen!
\end{ftl}
```

illfygli:
@alerque Thanks!
That's definitely nicer than modifying the installed distro package files, like I did.

Omikhleia:
> @alerque Nice, thank you! I hope so too. I just found SILE, and ran into bibtex trouble due to the missing translation when using Swedish. Now that will work at least. :)

Welcome to SILE ! Note that regarding bibliographies and bib(la)tex support, we'll have a different solution based on CSL styles and locales, with many more features (The bibtex package is and remains experimental, but I'll have more proposals to it eventually). Feel free to check https://github.com/orgs/sile-typesetter/projects/12 to see what we have recently done and where we are heading.

illfygli:
@Omikhleia Thanks, checking it out. Good to see that bibliography support will be expanded!

PR #2127, opened 2024-10-03
Add AST dump as outputter and inputter

alerque:
Followup on code and discussion in #2113


alerque:
As of this moment this is just a rough POC. It can round trip at least some basic documents from XML/SIL → AST output -> AST input -> PDF.

No ETA on this because there are quite a few questions about how this should work and why and I haven't even given it a spin in my own workflows yet.

Issue #2111, opened 2024-09-19
SILE and XML - A Return on Experience


Omikhleia:
So Saith Simon: (https://github.com/sile-typesetter/sile/issues/508#issuecomment-354353242)
> SILE is fundamentally a typesetting system for arbitrary XML files."
>
> -- (https://github.com/sile-typesetter/sile/issues/508#issuecomment-354353242)

And we obviously mention similar things in the User Manual and other places (such as our Wiki, etc.)

**But is that actually true?**

And what sort of challenges a package-designer might face when trying to use SILE as a typesetter for "arbitrary" XML files? I have some experience with this,[^1] and I'd like to share it with you.

SILE ships with an XML streaming parser (Luaexpat a.k.a. lxp), and merely converts XML elements and their attributes to SILE syntax tree nodes, as a straightforward "exact" mapping (tags become commands, attributes become options, text nodes becomes content).

It doesn't do anything special with the XML, and notably:
 1. It has no knowledge of document model (old-school DTDs, or more modern XML schemas; both go beyond the question of the "validity" of the input, they also define default/fixed values for attributes, the interpretation of spaces in some elements, etc.)
 2. It doesn't care about namespacing, and potential conflicts between XML elements from different types of documents (preventing their use in the same document), and even the core SILE commands. Note that I am not specifically referring to XML namespaces in a document ('xmlns`, `xmlns:xxx`) specifically, though that is also a part of the equation; but rather to conflicts between different input documents. Let's call the former "external namespacing" and the latter "internal namespacing".
 3. Paragraphing is done by interpreting empty lines (precisely, double newlines) in the SILE typesetter, not in the inputter(s), which may be a problem for some XML documents.

_Thoughts on document models_

The problem in this space is that XML is a very broad format, with different applications.
But in practice, there are two different types of XML content (possibly, and actually often, mixed in a given schema/model, depending on context):
 - Presentational: XML that is meant to be rendered as-is, with minimal processing (i.e. more or less consisting in styled text, in the same order)
 - Stuctural: XML that is meant to be processed and rendered in a specific way (i.e. semantic information, where the order of elements is not necessarily the order of rendering)

This is a first-order approximation, but it's a useful distinction.[^2]

So beyond it's naive parser, what does SILE for us here?

Not much, so:
[...]

PR #2129, opened 2024-10-08
Test and fix verbatim line handling

### LINKED ISSUE #2128: Regression in \verbatim & \autodoc:codeblock handling of blank lines
As reported by @chrisvest [on mastodon](https://mastodon.social/@chrisvest/113263045558380660) the SILE manual has an issue with whitespace handling in code samples. We've lost all blank line to paragraph transformations in code samples. These days in the manual that's with `\autodoc:codeblock`, but the mechanism should be essentially the same as `\verbatium`. I've tried to get an MWE to bisect but I can't find an MWE that works/doesn't work in a reliable way. In fact at the moment even in versions where the manual appears to be correct my MWE's do *not* work.

Paging back through release manuals, we apparently completely broke all the code sample rendering in v0.14.10. That would make sense if you just looked at the fact that we used all now formatting macros for the whole manual starting in that release, but even restoring the v0.14.9 code for rendering code blocks or building a quick check with `\verbatium` still fails even in v0.14.9. And indeed all the way back to v0.11.0 at least.

Here is the v0.14.9 manual:

image

One release later in v0.14.10 the `parbreak` is gone:

image

And here it is today in v0.15.5:

image

The source is still to this day like I would expect:

```sile
Paragraph one.

Paragraph two.
This is not paragraph three.

This is paragraph three.
\end{raw}
```

Any ideas @Omikhleia?

alerque:
Closes #2128

...as of posting all I have is a test that shows its broken.


Omikhleia:
Would this meet the expectations (your test here + an additional check + the autodoc blurb from the manual cited in the issue):

image

If so, I can try to make it neat, perhaps this week-end. (Using a hack currently, but the path from that point seems to be the way to go, if it doesn't break anything else...)

alerque:
Yes that output looks right. The final test should have as little noise as possible but testing the double line scenario makes sense. Feel Free to push WIP to this branch or PR to target it.

My dev time is total chaos right now. Our son was born last week but had complications and is in the NICU for a while. I'm fiddling with FOSS here and there in waiting rooms to keep time moving.

Omikhleia:
> Our son was born last week but had complications and is in the NICU for a while.

I'm sincerely sorry to hear about the complications. I can't imagine how hard this must be for you and your wife. I’m keeping you all, and your baby in my thoughts and hoping for a smooth recovery.

> Feel Free to push WIP to this branch (...)

Just pushed a few commits on top of yours (the extra 2-line test, fixes to the verbatim/autodoc packages, and the crux of the issue in the typesetter). One test fails (bug-132) -- and it's actually a verbatim-related test, so that seems expected (I just looked at it very quicky, it does seem ok and it's likely caused by the parskip/baselineskip adjusments). This remains a WIP, though, I'll go on checking the changes and possibly backport them to my own environments to see how they behave here (the resilient.sile and markdown.sile have plenty of verbatim-like blocks too, so I'm eager to check how these changes would behave here...). But for now (3 AM lol), I need some sleep :p

alerque:
> > Our son was born last week but had complications and is in the NICU for a while.
> 
> I'm sincerely sorry to hear about the complications. I can't imagine how hard this must be for you and your wife. I’m keeping you all, and your baby in my thoughts and hoping for a smooth recovery.

Thank you. We and our son have had a rough go of it. His heart stopped during birth and he ended up with stage 2 hypoxic-ischemic encephalopathy, spent 10 days in ICU, and is still under close neurological monitoring. Physically he's recovering fairly well although brain activity on an EEG is still abnormal. But we're glad he's home with us now and we're praying for his eventual full recovery.

packages/verbatim/init.lua
```diff
@@ -13,20 +13,22 @@ function package:registerCommands ()
    self:registerCommand("verbatim", function (_, content)
       SILE.typesetter:pushVglue(6)
       SILE.typesetter:leaveHmode()
+      local lskip = SILE.settings:get("document.lskip") or SILE.types.node.glue()
```
alerque, line 16: Why are we stashing the setting here, but creating a new glue from just the length instead of just reusing the glue we saved? For that matter why are we fiddling with lskip anyway? And why does this work without the super stretchy rskip.

So many questions, one has to wonder if I'm up too late and my brain just isn't processing answers that might be in front of my face.

Omikhleia: If I remember right, we'll want to kill stretch/shrink on lskip/rskip (line 23-24) but keep the fixed part, so verbatim in a block quote for instance now still obeys the fixed margins on both sides. Before it would be back to full line = we couldn't quote code properly in indented blocks. They reverted to full line width.

Omikhleia: As for: "And why does this work without the super stretchy rskip."

'Cause we handle break as paragraphs now = we have our back saved by the parfillskip.

alerque: Roger and roger. Of course that makes sense.

Omikhleia: Now of course, one could argue that this whole `typesetter.parseppattern` stuff is _not_ sane. I'd have preferred the SIL parsers to know what a soft break is and insert that in a true AST, so the typesetter wouldn't have had to make assumptions on the way paragraphing is supposed to occur. But that's another "idiosyncrasy" of SIL that I'll leave to the future...

alerque: How is that an idiosyncracy of the SIL parser? Don't we have exactly the same issue in XML?

Omikhleia: > Don't we have exactly the same issue in XML?

Yes we do, and this work indeed for SIL XML. 
But I (briefly) tackled it in #2111 (§ Thoughts on paragraphing).
Not all XML schemas would expect two blank lines to become paragraphs.
E.g. HTML doesn't.
Delegating the decision on what is a paragraph in the input schema/language to the typesetter seems off, it should perhaps be an inputter decision, in the scope of arbitrary XML support.
There are workarounds for implementers, but it's more a question of semantics and design.

alerque: > Not all XML schemas would expect two blank lines to become paragraphs.

Of course not. How line breaks are handled is up to the content handler not XML itself. The schema might specify haw the content of a given tag should be handled. But for example given a SILE native schema I would expect my MWE above to treat newlines in `&lt;verbatim /&gt;` tags as breaks.

Hence why I'm suggesting this has nothing to do with the SIL language or parser. We don't really have an option of changing the parser to make newlines into nodes in the AST as long as we're trying to keep it interchangeable with an XML input.

> E.g. HTML doesn't. Delegating the decision on what is a paragraph in the input schema/language to the typesetter seems off

Even when the schema itself *says* that's how the content should be handled? We didn't invent XML or HTML.

Obviously we *could* just eschew SIL being sugar for an XML-ish schema and make it a totally unique language, but then I would question the sanity of having SIL as a markup language at all. We'd still have to have the typesetter handle this for actual XML, and SIL already isn't a *lightweight* markup language in the way Markdown/Djot/Asciidoc etc. are. So it would just be another complex idiosyncratic markup like USFM or others that only serve their niche and are frustrating to work with even in their niche.

> it should perhaps be an inputter decision, in the scope of arbitrary XML support.

I don't understand what you mean by this. What am I missing here?

Omikhleia: > I don't understand what you mean by this. What am I missing here?

Nevermind. It's orthogonal to verbatim, and _if time allows it,_ I'll eventually demonstrate a different approach to arbitrary XML input, based on some the ideas I discussed in #2111

PR #2135, opened 2024-10-16
fix(packages): Correct math greek symbols and their var-alternate

### LINKED ISSUE #2134: TeX-like phi, varphi and friends...
Formula: `\phi\varphi`

Expected: ϕφ

Observed:
 - Our phi is φ... so the **wrong** one as far as TeX-like compatibility is expected.
 - We don't have varphi...

Honestly?

```
diff --git a/packages/math/unicode-symbols.lua b/packages/math/unicode-symbols.lua
index 830a718c..5f140a8a 100644
--- a/packages/math/unicode-symbols.lua
+++ b/packages/math/unicode-symbols.lua
@@ -2583,7 +2583,8 @@ symbols.rho = "ρ"
 symbols.sigma = "σ"
 symbols.tau = "τ"
 symbols.upsilon = "υ"
-symbols.phi = "φ"
+symbols.phi = "ϕ"
+symbols.varphi = "φ"
 symbols.chi = "χ"
 symbols.psi = "ψ"
 symbols.omega = "ω"
```

Wait... Same kind of issue for epsilon... Our epsilon is ε = the **wrong** one, where TeX epsilon is ϵ and varepsilon ε

Our theta is the correct one θ... But we don't have the vartheta ϑ...
Our rho is correct ρ... But we don't have the varrho ϱ...
And there are yet other var-symbols...

Surely, it would be so simple to add/change (as show above), so there's must be some rationale behind the current implementation that eludes me?

Omikhleia:
Closes #2134 

Or does it? Who knows, it breaks test `tests/math-bigops.xml` due to the bogus epsilon. Surely whoever wrote this test checked again LaTeX, right?

EDIT: For the record, Typst deviates slightly from LaTeX https://qwinsi.github.io/tex2typst-webapp/cheat-sheet.html
- It has the `\varxxx..` symbols as `xxx.alt`
- Except that epsilon and phi are inverted (the .alt variants are the non-var ones)


OlivierNicole:
> Or does it? Who knows, it breaks test `tests/math-bigops.xml` due to the bogus epsilon. Surely whoever wrote this test checked again LaTeX, right?

I don’t remember, but it wasn’t deeply thought out. I think you should modify the test reference based on the decision that you take regarding the symbol names.

Omikhleia:
Fixed the broken test to use `\varepsilon` for ε, in compliance with TeX math's expectations (= output unchanged, an inline with the MathML case which uses an explicit ε).

alerque (approved):

PR #2139, opened 2024-10-17
TeX-math limits and other commands behaving as big ops

### LINKED ISSUE #2130: TeX-like math operator with sup-only part has the latter incorrectly place
**Input**

Formula in TeX-like syntax:

```
\prod_{n=1}^{\infty}
\prod_{p\,\in\,ℙ}
\prod^{ℙ} 
```

I.e. large operator, with (1) both sup and sub elements, (2) only sub elements, (3) only sup elements.

**Expected**

As obtained on https://latexeditor.lagrida.com/ and also https://saxarona.github.io/mathjax-viewer/

image

**Observed**

image

= The last case does not have the sup part above the operator.

**Analysis**

The math package code is insane and poorly commented, sorry to say.
Yet a line for the sup-only case (which differs from the sub-only case) looks dubious:

```
diff --git a/packages/math/texlike.lua b/packages/math/texlike.lua
index 5d1073fe..5abb03c6 100644
--- a/packages/math/texlike.lua
+++ b/packages/math/texlike.lua
@@ -328,7 +328,7 @@ local function compileToMathML_aux (_, arg_env, tree)
       tree.options = {}
    -- Translate TeX-like sub/superscripts to `munderover` or `msubsup`,
    -- depending on whether the base is a big operator
-   elseif tree.id == "sup" and tree[1].command == "mo" and tree[1].atom == atomType.bigOperator then
+   elseif tree.id == "sup" and tree[1].command == "mo" and symbolDefaults[tree[1][1]].atom == atomType.bigOperator then
       tree.command = "mover"
    elseif tree.id == "sub" and tree[1].command == "mo" and symbolDefaults[tree[1][1]].atom == atomType.bigOperator then
       tree.command = "munder"
```

And with the shown change, yay! (**EDIT** but see comment, I was a bit too naive here)

image

Dear math-aware friends, what do you think?

### LINKED ISSUE #2131: TeX-like math limits are not supported
_**My dear math friends**_...
Do you really expect SILE will thrive without the "limits"?
I don't know for other countries and recent school programs... but in my youth it was taught in "classe de terminale (scientifique)" (~ grade 12 maybe ?), that is to say indeed a scientific cursus, but not something _very_ advanced...

Surely all math friends would want limits natively supported in TeX-like syntax, wouldn't they?

**Input**

```
\lim_{x\to\infty} \frac{1}{x}
```

**Expected**

As obtained on https://latexeditor.lagrida.com/ (corresponding to display math mode):

image

**Observed**

Error: The `\lim` and `\to` commands aren't supported. (And I know `\to` comes from amsmath, I'm not even going to dicuss using `\rightarrow` as if amsmath hadn't existed for years...).

**Analysis**

First attempt at just adding these in my math formula

```
\def{lim}{\mi{lim}}% Note: crashes with \mo instead of \mi
\def{to}{\mo[atom=bin]{→}}
```

image

So we need something else to enforce `\lim` to be considered as a "big" operator...

Regarding the note "crashes with mo instead of mi, it crashes on the sub-handling line just following the dubious sup-handling line reported in #2130...)

Anyway, another attempt:

```
diff --git a/packages/math/texlike.lua b/packages/math/texlike.lua
index 5d1073fe..343b4e04 100644
--- a/packages/math/texlike.lua
+++ b/packages/math/texlike.lua
@@ -479,6 +479,8 @@ compileToMathML(
   \def{sqrt}{\msqrt{#1}}
   \def{bi}{\mi[mathvariant=bold-italic]{#1}}
   \def{dsi}{\mi[mathvariant=double-struck]{#1}}
+  \def{lim}{\mo{lim}}
+  \def{to}{\mo[atom=bin]{→}}
 
   % Standard spaces gleaned from plain TeX
   \def{thinspace}{\mspace[width=thin]}
diff --git a/packages/math/unicode-symbols.lua b/packages/math/unicode-symbols.lua
index 830a718c..32291b90 100644
--- a/packages/math/unicode-symbols.lua
+++ b/packages/math/unicode-symbols.lua
@@ -2649,6 +2649,7 @@ symbolDefaults["]"] = { atom = atomType.closeSymbol, stretchy = true }
 symbolDefaults["{"] = { atom = atomType.openingSymbol, stretchy = true }
 symbolDefaults["}"] = { atom = atomType.closeSymbol, stretchy = true }
 symbolDefaults["mod"] = { atom = atomType.binaryOperator }
+symbolDefaults["lim"] = { atom = atomType.bigOperator }
 symbolDefaults["%"] = { atom = atomType.binaryOperator }
 
 return {
```

Ya hey! Ya-harri-hey! Ya hoy!

image

Frankly, I have the feeling of monkey-patching a messy and undocumented piece of code... That must be me, so what do you think, O Great Ones Who Master Maths?

**Subsidiary question**

What about `\limsup` and `\liminf`?
Hint (which unfortunately cannot be shown in a GitHub comment): It _can_ be done if the operator symbols are entered as "lim inf" and "lim sub" with the space here being a narrow no-break space (U+202F). A regular space doesn't work due to the way parsing is done somewhere... AFAIK, it does look somewhat good, but it seems a bit dubious. Sure the Great Ones would frown?

Omikhleia:
Movin' forward...

One needs [incentives](https://en.wikipedia.org/wiki/Laplace%27s_method#Formal_statement_and_proof), so the target should be:[^1]

image

Closes #2131 

Closes #2130 

(The latter being... a side effect ;)

[^1]: This target also requires `\text` support... #2140

### DIFF
--- a/packages/math/texlike.lua
+++ b/packages/math/texlike.lua
@@ -252,6 +252,23 @@ local compileToStr = function (argEnv, mathlist)
    end
 end
 
+local function isBigOperator (tree)
+   if tree.command ~= "mo" then
+      return false
+   end
+   -- Case \mo[atom=big]{ops}
+   -- E.g. \mo[atom=big]{lim}
+   if tree.options and tree.options.atom == "big" then
+      return true
+   end
+   -- Case \mo{ops} where ops is registered as big operator (unicode-symbols)
+   -- E.g. \mo{∑) or \sum
+   if tree[1] and symbolDefaults[tree[1]] and symbolDefaults[tree[1]].atom == atomType.bigOperator then
+      return true
+   end
+   return false
+end
+
 local function compileToMathML_aux (_, arg_env, tree)
    if type(tree) == "string" then
       return tree
@@ -328,21 +345,13 @@ local function compileToMathML_aux (_, arg_env, tree)
       tree.options = {}
    -- Translate TeX-like sub/superscripts to `munderover` or `msubsup`,
    -- depending on whether the base is a big operator
-   elseif tree.id == "sup" and tree[1].command == "mo" and tree[1].atom == atomType.bigOperator then
+   elseif tree.id == "sup" and isBigOperator(tree[1]) then
       tree.command = "mover"
-   elseif tree.id == "sub" and tree[1].command == "mo" and symbolDefaults[tree[1][1]].atom == atomType.bigOperator then
+   elseif tree.id == "sub" and isBigOperator(tree[1]) then
       tree.command = "munder"
-   elseif
-      tree.id == "subsup"
-      and tree[1].command == "mo"
-      and symbolDefaults[tree[1][1]].atom == atomType.bigOperator
-   then
+   elseif tree.id == "subsup" and isBigOperator(tree[1]) then
       tree.command = "munderover"
-   elseif
-      tree.id == "supsub"
-      and tree[1].command == "mo"
-      and symbolDefaults[tree[1][1]].atom == atomType.bigOperator
-   then
+   elseif tree.id == "supsub" and isBigOperator(tree[1]) then
       tree.command = "munderover"
       local tmp = tree[2]
       tree[2] = tree[3]
@@ -479,6 +488,16 @@ compileToMathML(
   \def{sqrt}{\msqrt{#1}}
   \def{bi}{\mi[mathvariant=bold-italic]{#1}}
   \def{dsi}{\mi[mathvariant=double-struck]{#1}}
+
+  \def{lim}{\mo[atom=big]{lim}}
+
+  % From amsmath:
+  \def{to}{\mo[atom=bin]{→}}
+  % Those use U+202F NARROW NO-BREAK SPACE in their names
+  \def{limsup}{\mo[atom=big]{lim sup}}
+  \def{liminf}{\mo[atom=big]{lim inf}}
+  \def{projlim}{\mo[atom=big]{proj lim}}
+  \def{injlim}{\mo[atom=big]{inj lim}}
 
   % Standard spaces gleaned from plain TeX
   \def{thinspace}{\mspace[width=thin]}
--- a/packages/math/texlike.lua
+++ b/packages/math/texlike.lua
@@ -493,6 +493,11 @@ compileToMathML(
 
   % From amsmath:
   \def{to}{\mo[atom=bin]{→}}
+  \def{gcd}{\mo[atom=big]{gcd}}
+  \def{sup}{\mo[atom=big]{sup}}
+  \def{inf}{\mo[atom=big]{inf}}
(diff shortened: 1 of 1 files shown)


packages/math/texlike.lua
```diff
@@ -479,6 +479,10 @@ compileToMathML(
   \def{sqrt}{\msqrt{#1}}
   \def{bi}{\mi[mathvariant=bold-italic]{#1}}
   \def{dsi}{\mi[mathvariant=double-struck]{#1}}
+  \def{lim}{\mo{lim}}
```
Omikhleia, line 482: Checking Pandoc TeXmath, at a glance, other operators such as min, max, perhaps ker and others should also behavelike this. Ideas?

packages/math/unicode-symbols.lua
```diff
@@ -2649,6 +2649,9 @@ symbolDefaults["]"] = { atom = atomType.closeSymbol, stretchy = true }
 symbolDefaults["{"] = { atom = atomType.openingSymbol, stretchy = true }
 symbolDefaults["}"] = { atom = atomType.closeSymbol, stretchy = true }
 symbolDefaults["mod"] = { atom = atomType.binaryOperator }
+symbolDefaults["lim"] = { atom = atomType.bigOperator }
```
Omikhleia, line 2652: I would have preferred not doing this... But above, replacing `\def{lim}{\mo{lim}}` by `\def{lim}{\mo[atom=big]{lim}} crashes.
... At the same sup/sub stuff as https://github.com/sile-typesetter/sile/issues/2130#issuecomment-2419679652

Sounds there's something deeper to it...

packages/math/texlike.lua
Omikhleia: From the amsmath (§5.1) list of predefined operators: 
 arcos,  deg,    lg,     projlim,
 arcsin, det,    lim,    sec,
 arctan, dim,    liminf, sin,
 arg,    exp,    limsup, sinh,
 cos,    gcd,    ln,     sup,
 cosh,   hom,    log,    tan,
 cot,    inf,    max,    tanh,
 coth,   injlim, min,
 csc,    ker,    Pr

The 10 following ones (tested on https://latexeditor.lagrida.com/) have subscripts going under the operator:
projlim, lim, liminf, limsup, gcd, sup, inf, max, injlim, min

For the record, amsmath has 4 special operators varinjlim varliminf varprojlim varlimsup which show as "lim" with stacked long arrows (of the size taken by "lim") or bars. I'll leave these out of the current scope.

packages/math/unicode-symbols.lua
Omikhleia: And suddenly: Closes #2130 

The many Gods of Typography are with me tonight?

Omikhleia: image

alerque (approved):

Issue #2138, opened 2024-10-17
Epic: Unsupported MathML elements and attributes


Omikhleia:
This is an "Epic" (in the sense of a parent issue for related sub-issues)

Maybe this is lost in some old discussion on some of the MathML implementation issue? Anyhow, here's a list (to be enriched possibly).

Some constructs SILE's MathML implementation does not support as of **v0.15.5**:
 - [ ] multiscripts (f2) - and mprescripts
 - [x] mstyle (f6) - this one is legacy in MathML Core = layout as mrow (at least) = #2151; see below for styling
 - [x] mpadded (f22) - see #2151 
 - [x] mtext (f18, f19, f22) - honestly this one is _much_ needed and shouldn't be complex... see #2140
 - [x] mphantom (f22) - see #2143
 - [x] msqrt - see #2119
 - [x] mroot - see #2120 
 - [x] ms - see #2140
 - [x] maction - MathML core says it can be considered as an mrow where only the first child is rendered = see #2151
 - [x] menclose - not in MathML Core, only MathML 3-4, but in the absence of CSS it's useful. MathJax supports it. - see #2318

Some other unsupported things
 - [ ] Most attributes on elements, ex.
    - [x] "linethickness" on mfrac (f10), -- see #2151 
    - [x] "form" on mo (for mphantom esp.),
    - [ ] Consistent support of "displaystyle" and "scriptlevel" on several tags - interesting in several cases (e.g. TeX cfrac)
    - [x] mathcolor, mathbackground - see #2340 
    - etc. (but what else would we want?)
 - [x] Improper sizing of some parentheses (several cases) - #2151 addressed the stretchy attribute and I think that was it!
 - [x] Stretchy horizontal braces (f19; f22) - some naive solution proposed in #2151 (test-based, maybe not general)
 - [x] Very large stretchy operators (whether done by glyph assembly or **faked**, they must be improved)
     - [x] Very large vertical stretchy parentheses (f23) - see  #2151 (achieved with a **hack**/workaround)
     - [x] Very large vertical stretchy determiner bar (f24) - same **hack** comment as previous

Other topics
 - Styling ought to be discussed. MathML core makes many attributes and the mstyle element legacy, but uses CSS instead... A hornets' nest...

Notes
 - For cross-checking see also https://w3c.github.io/mathml-core/#mathml-elements-and-attributes
[...]

PR #2140, opened 2024-10-17
feat(math): Support MathML mtext and ms elements

Omikhleia:
Ticking one box in #2138 

I am on a quest, sort of...

So let's derive the [Quadratic Formula](https://developer.mozilla.org/en-US/docs/Web/MathML/Examples/Deriving_the_Quadratic_Formula).

image

### DIFF
--- a/packages/math/typesetter.lua
+++ b/packages/math/typesetter.lua
@@ -45,7 +45,7 @@ function ConvertMathML (_, content)
       local script = content.options.mathvariant and b.mathVariantToScriptType(content.options.mathvariant)
       local text = content[1]
       if type(text) ~= "string" then
-         SU.error("mi command contains " .. text .. ", which is not text")
+         SU.error("mi command contains content which is not text")
       end
       script = script or (luautf8.len(text) == 1 and b.scriptType.italic or b.scriptType.upright)
       return b.text("identifier", {}, script, text)
@@ -67,7 +67,7 @@ function ConvertMathML (_, content)
          end
       end
       if type(text) ~= "string" then
-         SU.error("mo command contains " .. text .. ", which is not text")
+         SU.error("mo command contains content which is not text")
       end
       return b.text("operator", attributes, script, text)
    elseif content.command == "mn" then
@@ -75,7 +75,7 @@ function ConvertMathML (_, content)
          or b.scriptType.upright
       local text = content[1]
       if type(text) ~= "string" then
-         SU.error("mn command contains " .. text .. ", which is not text")
+         SU.error("mn command contains content which is not text")
       end
       if string.sub(text, 1, 1) == "-" then
          text = "−" .. string.sub(text, 2)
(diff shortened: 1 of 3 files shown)


packages/math/typesetter.lua
```diff
@@ -136,6 +136,19 @@ function ConvertMathML (_, content)
       return b.mtr(convertChildren(content))
    elseif content.command == "mtd" then
       return b.stackbox("H", convertChildren(content))
+   elseif content.command == "mtext" or content.command == "ms" then
+      if #content > 1 then
+         SU.error("Wrong number of children in " .. content.command .. ": " .. #content)
+      end
+      local text = content[1] or "" -- empty mtext is allowed, and found in examples...
+      if type(text) ~= "string" then
+         SU.error("ms command contains " .. text .. ", which is not text")
```
Omikhleia, line 145: Copied for other similar cases from above. But this crashes if text can't be serialized to string.

Omikhleia: Fixed.

OlivierNicole:
This looks like a good working first step. Note that one day we probably want to be able to use SILE commands in `\text`, just like arbitrary TeX formatting can be applied inside a `\text{...}`…

Omikhleia:
> Note that one day we probably want to be able to use SILE commands in `\text`, just like arbitrary TeX formatting can be applied inside a `\text{...}`…

But I'd advocate _against_ it, as far as I have a word to say. It's not a "common" denominator for any real use case.

Omikhleia:
> But I'd advocate _against_ it, as far as I have a word to say. It's not a "common" denominator for any real use case.

"Note that one day we probably want to be able to use SILE commands"... Really? Folks, please be much more precise, do you mean SIL in its TeX-like syntax? (as `\em` and that sort of stuff?). Do you imply there's a real added value _in general_ for this to be true?

Had you said: "one day we probably want to be able to use Markdown, Djot or other markup in `\text`", I'd would have raised an eyebrow with interest.

How long does it take for you to understand that SIL is a dead end?

Omikhleia:
BUT let's go back on topic; there's a PR here, does it pass the bar? Something is better, sometimes, than an idealized nothingness.

alerque:
Hey guys can we take the rhetoric down a notch or two please? I should have caught this earlier but I've spent most of a couple weeks waiting in a hospital and haven't had the energy. This is an entirely volunteer project and interested parties are going to have different ideas and backgrounds. It's okay to disagree, but the last half dozen issues I've looked at today have a bit too much snark. I appreciate the contributions for sure and it's okay to think we're doing some things wrong, but lets take it easy.

Trying to ignore the tone now and address a little point:

> How long does it take for you to understand that SIL is a dead end?

Says who? You don't have to use it or like it, but given how much easier it is to read and write manually than XML I don't see it going anywhere. Also it is (or should be) 100% **just % syntactic sugar over a semantic document** that could just as well have been XML. It is easy to parse and very overhead to maintain, so it is not going anywhere.

That being said I don't see it as any different from any other input format, we should have 100% first class coverage for either XML or SIL (and potentially others in the future), the SIL syntax sugar is not some magic special "this is the SILE way" that gets special treatment.

> "Note that one day we probably want to be able to use SILE commands"... Really? Folks, please be much more precise, do you mean SIL in its TeX-like syntax?

I read this as SILE *commands*, probably matching whatever syntax flavor the parent document is using (XML-ish or TeX-ish or whatever other input module is being used. And yes I agree that text content inside an equation should support the typesetting features of the typesetter, not just plain text. There is precedent for this in every other Math layout system I've seen (including GUI ones used in desktop publishing, for better or worse).

That doesn't mean it has to be in this PR, just saying I'd be open for contributions in that direction.

alerque (approved):

PR #2142, opened 2024-10-19
fix(math): Fractions must have padding to avoid visual confusion

### LINKED ISSUE #1801: Consecutive math fractions are too close
Consecutive fractions in math display mode get garbled together, at the risk of being misinterpreted by the reader.

image

As a workaround, I can manually force a `\,` in between...

image

But it's annoying to have to do such manual fine tuning... Can't SILE do better here?

(And yes, the lack of proper square root really shows here, but that's a different issue #1604)

Omikhleia:
MathML Core 3.3.2: "To avoid visual confusion between the fraction bar and another adjacent items (e.g. minus sign or another fraction's bar)"

Closes #1801


Omikhleia:
Impacted tests:
- tests/math-bigops.xml
- tests/math-fractions.xml
- tests/math-macros.sil
- tests/math-stretchy.xml

--> Checking them atm.

Omikhleia:
> --> Checking them atm.

They all look good to me, the small extra padding is not too visible but fixes the issue:
For the record tests/math-fractions.xml experienced it, in its last test, i.e.

```
<math>
  \frac{i+1}{}
  \frac{}{i+1}
</math>
```

This was producing a single long continuous fraction bar.
Which might have been the wanted effect here? But this is not the correct way to handle alignments anyway, and would have been confusing without empty denominator/numerator...

Before/After:
image

Now I'd need to push the fixed test expectations.... And still don't have the correct version of Gentium(?) in my dockers...

alerque:
> Now I'd need to push the fixed test expectations.... And still don't have the correct version of Gentium(?) in my dockers...

This doesn't sound right. The test suite itself should be providing the correct font version for running (and updating) tests.

I'm happy to help update expectations, but in this case it should be pretty straight forward:

```console
$ make update_expecteds TESTSRCS='tests/math-bigops.xml tests/math-fractions.xml tests/math-macros.sil tests/math-stretchy.xml' CLEAN=true
```

Does that sort of thing not work for you in scenarios like this?

alerque (approved):

PR #2143, opened 2024-10-20
Support MathML mphantom and TeX-like phantom, hphantom, vphantom

Omikhleia:
On the quest to tick another box in #2138

 - MathML `<mphantom>`: MathML Core encourages the use of CSS for styling, but of course we don't have that. It's still a valid "full MathML" construct though.
   References:
    - https://w3c.github.io/mathml-core/#making-sub-expressions-invisible-mphantom
    - https://www.w3.org/TR/mathml4/#presm_mphantom

 - TeX/LaTeX `\phantom`, `\hphantom` and `vphantom` are no-brain implementations, with a small special extra attribute to MathML (`<mphantom>` only corresponds to `\phantom`).
   References:
    - https://latexref.xyz/_005cphantom-_0026-_005cvphantom-_0026-_005chphantom.html

**Note**: MathML has some extra considerations such as the "form" of operators in a phantom formula (e.g. `<mo form="infix">+</mo> ...`  - This is not addressed here, and would require a more general handling of attributes, which is another point in #2138...

### DIFF
--- a/packages/math/base-elements.lua
+++ b/packages/math/base-elements.lua
@@ -533,6 +533,43 @@ end
 
 function elements.stackbox.output (_, _, _, _) end
 
+elements.phantom = pl.class(elements.stackbox) -- inherit from stackbox
+elements.phantom._type = "Phantom"
+
+function elements.phantom:_init (children, special)
+   -- MathML core 3.3.7:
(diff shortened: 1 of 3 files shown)


Omikhleia:
In action:
image

PR #2147, opened 2024-10-21
Fix math mathvariants

### LINKED ISSUE #2146: MathML mathvariant support is incomplete
Our implementation of thee MathML `mathvariant` attribute only supports but a few of the possibilities from MathML3.[^1]

It's true that MathML4 and MathML Core both encourage the direct use of Unicode characters, making `mathvariant` somewhat legacy... but the logic is useful for parity with LaTeX in our TeX math like syntax...
 - [x] Support at least most of the mnemonics ("fraktur", "script", etc.)...
 - [x] Support numbers and greek when applicable (and not only capital and small letters)...[^2]
 - [x] Ensure TeX-like `\Gamma` etc. are upright (see comment below)
 - [x] In TeX-like math, simple macros for `\mathcal`, `\mathfrak` and `\mathbb` are then within reach... (on the principle of least astonishment, and for ease of integrating TeX-based formulas from other sources).

--> **I'm on this one.**

Fonts such as Libertinus Serif have a good coverage for these characters, so a show case demo even becomes easy:
 - Checking the mappings as I add them (checking the non-normative but really nice [MathML Core Appendic C](https://www.w3.org/TR/mathml-core/#new-text-transform-mappings))...
 - Generating a "ptable" generated by looping on variants and characters...

image

One always need a nice target, bits by bits... Eventually... Say the [Schwinger–Dyson equation](https://en.wikipedia.org/wiki/Schwinger%E2%80%93Dyson_equation#Derivation)

```
\left\langle\psi\left|\mathcal{T}\left\{\frac{\delta}{\delta\phi}F[\phi]\right\}\right|\psi\right\rangle = -\mathrm{i}\left\langle\psi\left|\mathcal{T}\left\{F[\phi]\frac{\delta}{\delta\phi}S[\phi]\right\}\right|\psi\right\rangle`
```

Should give what's expected...

image

(Firefox rendered MathML from Pandoc's output for Wikipedia's TeX-based formula).

We are not yet there (oh those `\left` and `\right`... on another day), but there's a way. Don't you think oh my dear "math friends"? -- or here, rather, my "theoretical physicist friends" :clown_face: 

[^1]: See our of `mathVariantToScriptType` and related code: 5 out of 18 (we might not support all of them, see next note, but we _can_ support most.
[^2]: There's some arabic stuff too, but I'll likely suggest to keep it out of this issue for now.

Omikhleia:
Closes #2146

~~WARNING: Currently on top of a raw merge of various branches.~~
 - [x] To rebase after those other branches are merged
 - [x] Some math tests are expected to fail, review them and check the new expectations are met --> See comment below

### DIFF
--- a/packages/math/base-elements.lua
+++ b/packages/math/base-elements.lua
@@ -2,6 +2,8 @@ local nodefactory = require("types.node")
 local hb = require("justenoughharfbuzz")
 local ot = require("core.opentype-parser")
 local syms = require("packages.math.unicode-symbols")
+local mathvariants = require("packages.math.unicode-mathvariants")
+local convertMathVariantScript = mathvariants.convertMathVariantScript
 
 local atomType = syms.atomType
 local symbolDefaults = syms.symbolDefaults
@@ -19,32 +21,6 @@ local mathMode = {
    scriptScriptCramped = 7,
 }
 
-local scriptType = {
-   upright = 1,
-   bold = 2, -- also have Greek and digits
-   italic = 3, -- also have Greek
-   boldItalic = 4, -- also have Greek
-   script = 5,
-   boldScript = 6,
-   fraktur = 7,
-   boldFraktur = 8,
-   doubleStruck = 9, -- also have digits
-   sansSerif = 10, -- also have digits
-   sansSerifBold = 11, -- also have Greek and digits
-   sansSerifItalic = 12,
-   sansSerifBoldItalic = 13, -- also have Greek
-   monospace = 14, -- also have digits
-}
-
-local mathVariantToScriptType = function (attr)
-   return attr == "normal" and scriptType.upright
-      or attr == "bold" and scriptType.bold
-      or attr == "italic" and scriptType.italic
-      or attr == "bold-italic" and scriptType.boldItalic
-      or attr == "double-struck" and scriptType.doubleStruck
-      or SU.error('Invalid value "' .. attr .. '" for option mathvariant')
-end
-
 local function isDisplayMode (mode)
    return mode <= 1
 end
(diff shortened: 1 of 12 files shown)


Omikhleia:
Visual inspection of the new test results = meeting my expectations, lowercase greek symbols are now in italic

 - tests/bug-1495-inline-math-layout.sil = pi, theta
 - tests/feat-math-display-numbered.sil = pi
 - tests/feat-math-display-unnumbered.sil = pi
 - tests/math-bigops.xml = several (epsilon etc.)
 - tests/math-macros.sil = sigma
 - tests/math-stretchy.xml = zeta
 - tests/math-subsup.xml = zeta
 
No other change noted on that visual inspection, and the diff files seem to be ok (we get "nil" for those math expressions, so it's hard to tell, but the changes are around the place they'd be expected.

Omikhleia:
> Doh it seems I fixed my Gentium setup this time, but I probably had an earlier Libertinus Math font in that config. Refreshing it...

Oh yay! Now I have another setup with Gentium Plus 5.0 from 2014 instead of 6.101, _but_ Libertinus 7.050 instead of 7.040. I feel good :stuck_out_tongue:

alerque:
> Now I have another setup with Gentium Plus 5.0 from 2014 instead of 6.101, _but_ Libertinus 7.050 instead of 7.040. I feel good 😛

Living on the wild side might be fun...

...but if in doubt about the exact right incantation of font versions, it should be easily achievable by nuking the cached versions (`rm -rf .fonts .sources`) from your source tree and then the next time `make test` or `make update_expecteds` or whatever is run it will go download the exact versions used in the tests.

PR #2149, opened 2024-10-23
fix(math): Math font size must follow the document font size

### LINKED ISSUE #2145: Math font size does not follow the current normal text size
Up to SILE 0.15.5, the math font size is defaulting to 10 (pt) via the `math.font.size` setting.

While I can understand the need sometimes to _adjust_ the math font size (e.g. based on the look of the target math font), this is a bad default and solution.
 - My booklet is typeset in Libertinus 12pt (... getting older, I use bigger font sizes), and uses Libertinus Math for math = Inline formulas are too small...
 - One may also expect formulas:
   - in block-quotes (e.g. a quotation from a book, generally typeset in a small font)...
   - or in footnotes...  (idem)
   - or even in headers (and hence in the ToC etc.)

So the font size ought to be the same as the normal text size by default, and possibly be adjusted by some ratio, but certainly not in a different and independent size setting (or _all_ commands that tweak the font size would need to also tweak the math font size).

Omikhleia:
Closes #2145

### DIFF
--- a/packages/math/init.lua
+++ b/packages/math/init.lua
@@ -42,9 +42,13 @@ function package.declareSettings (_)
    })
    SILE.settings:declare({
       parameter = "math.font.size",
-      type = "integer",
-      default = 10,
+      type = "number or integer",
+      default = SILE.settings:get("font.size"),
    })
+   SILE.settings:registerHook ("font.size", function (size)
+      -- Follow document font size changes
+      SILE.settings:set("math.font.size", size)
+   end)
    -- Whether to show debug boxes around mboxes
    SILE.settings:declare({
       parameter = "math.debug.boxes",
--- a/tests/feat-math-display-numbered.sil
+++ b/tests/feat-math-display-numbered.sil
@@ -4,6 +4,11 @@
 \use[module=packages.lorem]
 \use[module=packages.math]
 \font[size=11pt]
+% Side effect of #2145: the math font size now follow the current document
+% font size, so whould now be 11pt.
+% The previous behavior was an hard-coded 10 pt.
+% Restore it here, so the test passes unchanged.
+\set[parameter=math.font.size, value=10]
 % Test numbered display math for all options
 % and in all alignments
 
--- a/tests/feat-math-display-unnumbered.sil
+++ b/tests/feat-math-display-unnumbered.sil
@@ -9,6 +9,11 @@
 % and in all alignments
 
 \begin{blockquote}
+% Side effect of #2145: the math font size now follow the current document
+% font size, so whould now be some ratio of 11pt (due to the blockquote)
+% The previous behavior was an hard-coded 10 pt.
+% Restore it here, so the test passes unchanged.
+\set[parameter=math.font.size, value=10]
 Begin justified
 \math[mode=display]{e^{i\pi} = -1}
 End justified.


packages/math/init.lua
```diff
@@ -42,9 +42,13 @@ function package.declareSettings (_)
    })
    SILE.settings:declare({
       parameter = "math.font.size",
-      type = "integer",
-      default = 10,
+      type = "number or integer",
```
alerque, line 45: I think we should consider whether we ought to allow relative values here such that formulas would track changes to the document font size in a relative way.

packages/math/init.lua
```diff
@@ -42,9 +42,13 @@ function package.declareSettings (_)
...
+      default = SILE.settings:get("font.size"),
    })
+   SILE.settings:registerHook ("font.size", function (size)
+      -- Follow document font size changes
+      SILE.settings:set("math.font.size", size)
```
alerque, line 50: If we do allow a relative value (e.g. `1.2em`) then this would need to adjust to compute the difference (or just not change it at all, but we would have to review if we are using `:absolute()` where necessary) not just hard code a value.

Even if not, we probably want to figure out how to *remove* this hook if the user resets `font.size` after having set `math.font.size` at any point.

alerque (changes requested):
This definitely needs fixing, but just hard coding a link to `font.size` seems too aggressive and might trip up in some cases.

packages/math/init.lua
Omikhleia: The more I think about it, the more I tend to believe that `math.font.size` should be killed and we use just the current font size.
Let me know you feeling (whether we address it in 0.15.8, whether we don't, and what we foresee) -- So I can document it (or not) in an upcoming booklet.

alerque: > The more I think about it, the more I tend to believe that `math.font.size` should be killed and we use just the current font size.

I don't see how that would work. We're automatically switching font faces for Math and it may or may not be from the same font family or appropriately sized to be matched with the body text. I don't see how we can get away with *not* allowing separate control over the size when we are changing the face.

> Let me know you feeling (whether we address it in 0.15.8, whether we don't, and what we foresee) -- So I can document it (or not) in an upcoming booklet.

I'm happy to move this forward as soon as we agree on how it *should* be handled. I just don't want to push out a change that we know has just a different set of downside to the current one. If perchance the solution we come up with ends up being a significant breaking change that might make it more complicated and delay things as well, but as long as this is a setting issue we can probably shim it in the *retrograde* package and not worry about it some reflows being breaking.

Omikhleia: Noted!
I removed my assignment and shifted the original issue to 0.x.y -- It might be reconsidered indeed, but that won't occur as part of my on-going 2024 effort on math which is soon reaching its end.

Omikhleia: With support for ex-height adjustment in 0.15.10, we could now kill the `math.font.size` entirely, and use that feature.

I would be tempted doing it without compatibility shim. It's use was totally buggy and also problematic in other places where we use measurement casts for some elements (e.g. mpadded). Having a compatibility shim to keep a broken behavior

A good mention in release notes should be sufficient IMHO. I don't think many existing users depend on that bad setting. A good mention in release notes should be sufficient IMHO. I don't think many existing users depend on that bad setting.

I can draft it an update PR.

Omikhleia: > we could now kill the `math.font.size` entirely, and use that feature.

Or keep it and use both possibilities (manual setting or automated adjustment).

Omikhleia:
I am closing my own PR here in favor of #2307 -- The discussions above lead to a slightly better approach.

As noted, the approach here had a few concerns
 - (major) It wouldn't adapt relatively to the eye (a.k.a. ex-height) of the document font and match it perfectly. That's not really could typography...
 - (minor) The math font size would follow the document font size on each change (whether used or not)
 - (minor) Using a setting hook doesn't seem to be the most adequate way
 - ...

Issue #2150, opened 2024-10-25
Math "accents" (for vectors, hats etc.) are not properly supported


Omikhleia:
_Dear SILE 0.15.5 math enthusiasts,_

Let's say you want to typeset a vector $\vec{v}$ (with a nice top arrow) in SILE, or a variable with a fancy hat $\hat{n}$, or a long vector $\overrightarrow{AB}$...

How would you do that?

Quick answer: **you can't.**

I am not discussing here the issue of stretchy operators, which is a different topic[^1]
... but the correct positioning of accents on top of or under other elements...

And as shown in this issue description here, GitHub is not bad at it. So? :pig_nose: 
How hard can it be for such an _advanced_ typesetting system as SILE? :clown_face: 

[^1]: See #2148 (although the case here would be on horizontal stretching... erm...)

**MathML**

One is supposed to use `<mover>`, `<munder>`, `<munderover>` to attach accents to other elements, with specific attributes ("accent" and/or "accentunder", see references below). EDIT: And one might not even have to use them for operator symbols in a pre-defined table, where defaults may apply.

For instance,
 - with U+20D7 ("vec"): `<mover accent="true"><mi>x</mi><mo>&#x20D7;</mo></mover>`
 - with U+2192 ("rightarrow"): `<mover accent="true"><mi>x</mi><mo>&#x2192;</mo></mover>`

Try these, and see how SILE currently handles them.
The "accent" attribute is not supported, and the position of the accent somewhat off....
From a quick analysis, SILE's OpenType parser is handing the corresponding MATH features, so it would be at hands...

References:
 - MathML 4: https://www.w3.org/TR/mathml4/#presm_mo_dict_attrs 
    - "accent=true" on `<mo>` (but this is not part of MatML Core)
    - "accentunder=true" on `<munder>`
    - "accent=true" on `<mover>`
    - "accent=true" / "accentunder=true" on `<munderover>`
- MathML Core describes the layout rules for accent
   - notably using OpenType MATH features for top accent attachment, e.g. https://www.w3.org/TR/mathml-core/#dfn-top-accent-attachment
   - bottom accent attachment is probably addressed without needing OpenType, but I haven't checked yet.

**TeX-like math**

In LaTeX, one would use `\vec{v}`, `\hat{n}`, `\overrightarrow{AB}` etc.

What does these produce in SILE?
Let's try.
The commands are recognized, defaulting to atoms.
The argument is lost without warning, and the output is not what one would expect.
[...]

Issue #2148, opened 2024-10-22
Math operators: stretchy or not stretchy


Omikhleia:
Let's take again the Schwinger-Dyson equation, from some Joe's [test page](http://eyeasme.com/Joe/MathML/MathML_browser_test.html) which gives both a TeX formula and a MathML one...

**Expectation**:

![image](https://github.com/user-attachments/assets/e5bc0705-08de-43a4-aafb-c26cdb426300)

**Current SILE results** (SILE 0.15.5 + most of my previous math-related PRs, notably #2146 slightly extended)

MathML:

![image](https://github.com/user-attachments/assets/6e0e60b8-7368-4312-80fc-eb8ba4ef5959)

TeX-like = the TeX formula with the `\left` and `\right` commands set to nothing.

![image](https://github.com/user-attachments/assets/90cd478f-45b7-498e-b554-5f2af6313f20)

**Analysis**

On the bright side, we are quite close in both cases. But the some operators are not stretched (in both cases); and (in the TeX-like case) some are stretched while they shouldn't.

 - MathML:
   - The operator that are stretched are those having an explicit "symbolDefaults" override in `packages/math/unicode-symbols.lua` (a small list)
      - The square brackets are there too, but stretching occurs _within_ the context of the surrounding mrow, which is expected.
      - The vertical bar is here too but is not marked as stretchy
      - The angle brackets are missing
   - So our problem concerns the angle brackets and the vertical bar... and a large number of other symbols actually. The MathML guys were kind enough to provide [a non-normative but good list of such symbols](https://www.w3.org/TR/MathML/appendixc.html), see especially the table in §C.4....

 - TeX-like
    - The issue is very similar and part of the solution is the same...
    - ... But with a few extra considerations if we target a reasonable support of TeX math
       - In TeX many of these symbols are not stretchy by default, and this is where `\left` and `\right` come into play.
       - TeX knows _at parsing_ (i.e. using something similar to our "symbols" table) which atoms are mathopen/mathclose etc., and can use that to wrap the content in a box... And Pandoc knows that too, actually.
          - Running with `-d texmath`, we see that we converted the TeX to ...`\mi{F}\mo{[}\mo{ϕ}\mo{]}`...
            i.e. `<mi>F</mi><mo>[</mo><mo>ϕ</mo><mo>]</mo>`
          - Compare with Pandoc:
[...]

PR #2151, opened 2024-10-26
Math more fixes and features

### LINKED ISSUE #1293: Some MathML that fails
SILE 0.12

Math input:
```
<sile>
<mathml>
<mtable><mtr><mtd><mi>A</mi></mtd></mtr><mtr><mtd><mi>B</mi></mtd></mtr></mtable>
<mi>C</mi>
<mtable><mtr><mtd><mi>D</mi></mtd></mtr><mtr><mtd><mi>E</mi></mtd></mtr></mtable>
</mathml>
</sile>
```

Expected:
```
A D
 C
B E
```

Observed

```
Error detected:
	/usr/share/sile/core/length.lua:96: attempt to index local 'self' (a nil value)
stack traceback:
	/usr/share/sile/core/length.lua:96: in function '__add'
	packages/math/base-elements.lua:486: in function 'shape'
	packages/math/base-elements.lua:309: in function 'shapeTree'
	/usr/share/sile/packages/math/typesetter.lua:117: in function 'handleMath'
...
```

At line 486, `self.children[i - 1].relY` is nil for i == 2 (and indeed not initialized above in the case i == 1 ? However setting it there to 0 avoids the crash, but all items are then stacked vertically...

(MathML is hard, maybe I am doing it wrong ? the above MathML snippet was generated with the demo at https://demo.wiris.com/mathtype/en/developers.php and and also shows up correctly in https://www.mathmlcentral.com/Tools/FromMathML.jsp)

### LINKED ISSUE #1799: Parentheses or brackets size adaptation in display math
Standalone parentheses or brackets in (display) math automatically adopt a "big" scale if needed around content:

```
\math[mode=display]{(\frac{a}{b})}
```
image

However, they lose this capability if the expression gets (subscript or) superscripts, e.g. gets elevated to some power:

```
\math[mode=display]{(\frac{a}{b})^2}
```
image

Adding extra brackets around the parenthetical part sort of works, but the power is not raised higher (as e.g. TeX would do after a `\bigg)`) so it doesn't seem to be a fully satisfying workaround:

```
\math[mode=display]{{(\frac{a}{b})}^2}
```
image

My expectation (image manually tweaked) would have been:
image

Is there a proper way to achieve it?

image

### LINKED ISSUE #2122: Wrong position of sub/superscripts in math on big content
**Observed**
 (the square root comes from #2119, but this is orthogonal to the issue)

image
image

The exponents on "big" blocks (bracketed blocs with fractions, etc.) are really off.

**Note**

There's some commented out code in our implementation, let's check what restoring it does:

```diff
diff --git a/packages/math/base-elements.lua b/packages/math/base-elements.lua
index ea58071e..18af78cb 100644
--- a/packages/math/base-elements.lua
+++ b/packages/math/base-elements.lua
@@ -610,7 +610,7 @@ function elements.subscript:shape ()
       self.sub.relX = self.width + subShift
       self.sub.relY = SILE.types.length(math.max(
          constants.subscriptShiftDown * scaleDown,
-         --self.base.depth + constants.subscriptBaselineDropMin * scaleDown,
+         (self.base.depth + constants.subscriptBaselineDropMin * scaleDown):tonumber(),
          (self.sub.height - constants.subscriptTopMax * scaleDown):tonumber()
       ))
       if self:is_a(elements.underOver) or self:is_a(elements.stackbox) or self.base.largeop then
@@ -629,7 +629,7 @@ function elements.subscript:shape ()
       self.sup.relY = SILE.types.length(math.max(
          isCrampedMode(self.mode) and constants.superscriptShiftUpCramped * scaleDown
             or constants.superscriptShiftUp * scaleDown, -- or cramped
-         --self.base.height - constants.superscriptBaselineDropMax * scaleDown,
+         (self.base.height - constants.superscriptBaselineDropMax * scaleDown):tonumber(),
          (self.sup.depth + constants.superscriptBottomMin * scaleDown):tonumber()
       )) * -1
       if self:is_a(elements.underOver) or self:is_a(elements.stackbox) or self.base.largeop then
```

**Result**:

image
image

--> Maybe can we bisect why this code was commented out? It seems it was doing something more or less expected!

Omikhleia:
**On An Adventure**[^1] - Part One

 - fix(math): Default table row/column spacing must depend on the math font size = previously unreported
 - feat(math): Support MathML maction (basic) and mstyle (partial) = tick some more check-boxes in #2138
 - fix(math): Paired open/close atoms shall be forced in a mrow in TeX-like math = tick one small check-box in #2148 and #1799
 - fix(math): Handle LaTeX-like math top accent commands correctly = tick one check-box in #2150 
 - feat(core): Add px unit as 0.75pt as per CSS3 = needed by the following commit...
 - fix(math): Support linethickness attribute on MathML mfrac = tick one more small check-box in #2138
 - feat(math): Support MathML mpadded and refactor TeX-like phantoms =tick some another check-box in #2138 and make thing better
 - fix(math): Wrong condition on moveable limits (**workaround**) = previously unreported, seen during testing, in-code comment for the future...
 - refactor(math): Prepare for horizontal reshaping of stretchy symbols = just some generalization of a method
 - fix(math): Honor stretchy attribute on MathML mo elements = tick one (small) concern in #2148
 - fix(math): Stretchy symbols should work on large content (**workaround**) = use (bad) re-scaling until we support complex glyph assembly eventually. It's still looks more decent than not doing anything...[^2]
 - fix(math): Wrong vertical alignment of integral signs = previously unreported, seen on the Cauchy Formula (MathML "complex1" torture test from the Test Suite)[^3]
 - fix(math): Support basic operator stretching along the inline axis = probably a lame implementation, but working on most common cases[^4]
 - fix(math): Empty mrow's in munderover should not affect the baseline = "Grandpa, why are these called torture tests?" "My child, the MathML followers of the Hydra wanted us to suffer..."
 - fix(math): Correct sub/superscript position for subformulas vs. symbols] = #2122 
 - fix(math): Top-level math element must behave as an (horizontally stacked) mrow = #1293 
 
Closes #2122 

Closes #1799 

Closes #1293 

[^1]: "Grandpa, how did you end up on such an adventure in the perilous Land of Math?"
      "Well, it all began with a simple, amusing task involving square roots, nth roots, and the like..."
      "And that went smoothly, didn’t it?"
      "At first, yes. But then I started picking out random formulas, sure I could typeset anything that came my way. Little did I know, I had wandered into the lair of the Hydra."
      "The Hydra? That beast with countless heads that regrows two for every one that's cut off?"
      "Exactly. The great monster, relentless devourer of formulas"
      "And did you manage to defeat it?"
      "I can’t say for sure. But I stopped complaining and making issues. I just drew my sword and kept fighting the beast. Pouring commits into the lair, one after another."

[^2]: So I'm taking a shortcut here -- one which would likely make math font designers frown, as formerly for the msqrt/mroot support. I'll explain something about it in the booklet I'm also preparing as I progress on this PR... (I haven't counted, but I've assembled over 50 formulas at least, all problematic in an area or another. Frankly, glyph assembly from bits is insane, is the _least_ of the problems, and is far too costly).

[^3]: Previously unreported... Just a bad confusion between largeop and big operators, a mere coding slip likely... but it's a very funny one, we even show it, unnoticed, on our "demo" examples on the web-site :rofl: - Yes that circled-integral is too high with respect to the baseline, look folks! :rabbit: 
     image

[^4]: I couldn't make sense of MathML Core §3.4.2.2 so I just made it work for a series of tests. The results seem acceptable for a first try, and one may perhaps understand the implied logic if we ever meet more complex cases: 
image


Omikhleia:
Failed regression tests assessment so far:
 - tests/bug-1495-inline-math-layout.sil = sizing of paired parentheses = consequence of fix(math): Paired open/close atoms shall be forced in a mrow in TeX-like math
 - tests/feat-math-display-numbered.sil = table spacing = fix(math): Default table row/column spacing must depend on the math font size
 - tests/math-bigops.xml = integral positioning = fix(math): Wrong vertical alignment of integral signs
 - tests/math-macros.sil = paired parentheses (as above) + position of superscripts = fix(math): Correct sub/superscript position for subformulas vs. symbols
 - tests/math-tables-mathml.xml = table spacing, as above
 - tests/math-tables-tex.sil  = idem previous
 - tests/math-variants.xml = position of superscripts + size of parenthesis
    - Size of parenthesis: I had a little doubt, esp. for the first space, but my browser shows the same things for this MathML, so it's meeting the expectations.

= All tests failed where expected.

Omikhleia:
I pushed the fixed test expectations.
These 17 commits (+646 −390 changes) conclude Part One of _Grandpa's adventures in the Land of Maths_ :rofl: 
There are still 4 issues I'd like to address, but it will be for another day in another PR.

packages/math/base-elements.lua
```diff
@@ -1044,7 +1130,21 @@ function elements.text:output (x, y, line)
    -- There should be no stretch or shrink on the width of a text
    -- element.
    local width = self.width.length
-   SILE.outputter:drawHbox(self.value, width)
+   -- HACK: For stretchy operators, MathML Core and OpenType define how to build large glyphs
```
alerque, line 1133: Do we have a dedicated tracking issue for someday addressing this?

alerque (approved):

packages/math/base-elements.lua
Omikhleia: Not yet, hence my bold **workaround** notes here in the list above, and some bold too in the parent Epic (fixed). If time allows at the end of the journey, I'll want to have a refactoring session, ideally.

alerque: #2155

Issue #2160, opened 2024-11-04
Display math mode page break and paragraph indenting issue


Omikhleia:
Two related issues with SILE 0.15.5's implementation of display math mode:
 - page breaks may occur before or after a display math formula.
 - if the formula is in a paragraph, the next line is always paragraph indented

**Example**

```
We have:
\math[mode=display]{formula1}
hence:
\math[mode=display]{formula2.)
```

**Obtained**

```
(indent)We have:
        (potential page break here)
                     formula
        (potential page break here)
(indent)hence:
        (potential page break here)
                     formula2
        (potential page break here)
```

**Expected**

```
(indent)We have:
        (no page break unless wanted)
                     formula
        (potential page break here unless not wanted)
hence: (not indented)
        (no page break unless wanted)
                     formula2
        (potential page break here)
```

For reference, (La)TeX has finer control here:

 - predisplaypenalty (default 10000 = keep with previous lines)
 - postdisplaypenalty (default 0 = allow normal break; I remember tweaking it to negative for some personal preferences)
 - displaywidowpenalty (default 50)

It's less obvious than it might seem at a glance
 - In the current implementation, pre/postdisplaypenalty would be easy, but this is likely a wrong approach as....
 - ... displaywindowpenalty would required being at paragraphing (line node boxing),
 - ... and the noindent also need to know whether we are in the same paragraph or not.

Omikhleia:
N.B. added "additional notes" to the initial description.

Omikhleia:
To demonstrate the effect (though I wouldn't push the code as-is ^^)

![image](https://github.com/user-attachments/assets/22591749-f73e-4a8e-b22c-cc43fa1e8a34)

Omikhleia:
I am suggesting a conservative workaround in #2162, at least discouraging page breaks before display math equations and tuning breaks after them,  introducing TeX's pre/post display penalties and doing the minimal things needed in our current implementation to use them adequately...

But the are various ways a finer-grain approach could be considered.
[...]

PR #2162, opened 2024-11-09
fix(math): Avoid page breaks before display math equations

Omikhleia:
See #2160.
This is a conservative workaround.
Introducing the predisplay/postdisplay penalties is the right thing to do, but a more general solution to the full issue requires a more subtle handling of in-paragraph display math equations, which is not considered here.

### DIFF
--- a/packages/math/init.lua
+++ b/packages/math/init.lua
@@ -56,6 +56,31 @@ function package.declareSettings (_)
       type = "VGlue",
       default = SILE.types.node.vglue("2ex plus 1pt"),
    })
+
+   -- Penalties for breaking before and after a display math formula
+   -- See TeX's \predisplaypenalty and \postdisplaypenalty
+   SILE.settings:declare({
+      parameter = "math.predisplaypenalty",
+      type = "integer",
+      default = 10000, -- strict no break by default as in (La)TeX
+      help = "Penalty for breaking before a display math formula",
+   })
+   SILE.settings:declare({
+      parameter = "math.postdisplaypenalty",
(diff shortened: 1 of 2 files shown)


alerque:
v0.15.6 is a bit delayed, I'm chasing an unexpected panic in some builds and some build concerns. At first blush this seems to make sense to me but I don't have time to dig into alternatives. If you think this should go out in the current batch of math fixes and features please feel free to target the milestone and I'll review before releasing.

Omikhleia:
> If you think this should go out in the current batch of math fixes and features please feel free to target the milestone and I'll review before releasing.

Yes, it's still better than nothing here -- so adding the 0.15.6 milestone if it's still doable.
I commented the original issue for the "eventual" right way to fix this properly at some point, but I might never reach that point any time soon, while at least here we can have something fairly decent working with the current implementation, without taking too much risks.
More generally, I'm nearly done with what I wanted in the "first" batch of mathematical "fixes & improvements":
- I see you moved #2145 to 0.15.7 and I agree we need to better address it (for now, I can go for the workaround in my math booklet in progress, and re-consider this point later)
- I'm working on the "operator table" from MathML, but it's a "big" thing to get right... It's on tracks, but I can't say when I'll complete it (this week? next one? later? doh!)
- And the next topic would be a proper handling of (at least top) accents and stretchy operators, not a big thing but it's highly depending on the previous point.

So I'm still adding stuff to it, but feel safe cut the release at your convenience :)
-

alerque (approved):

PR #2165, opened 2024-11-11
feat(math): Support MathML bevelled fractions

Omikhleia:
Bevelled fractions (also known as skewed or diagonal fractions) are displayed with a slanted line (slash) instead of a horizontal bar.

Although this feature is still included in MathML4, it has been [excluded from the MathML Core](https://github.com/w3c/mathml/issues/29#issuecomment-2276354003) specifications.
As a result, most browsers are unlikely to support bevelled fractions...
But SILE is not a browser -- so it has the freedom to implement a broader standard! :cake: 

As stated in the in-code comments, though, MathML4 does not exactly specify how to compute the layout, and the related OpenType math parameters are, at best, unclear[^1] -- and not pleasant looking; moreover there's no clear distinction for display / cramped / text modes... :cactus: 

So I went for my own interpretation, also looking at various TeX packages (nicefrac, xfrac) for references. I do like the result:[^2]

image

Oh, as a final note -- I'm going to enter the debate on the ~~insane~~ glyph assembly logic once again... _So yes,_ the proposed code draws a line. A mere line, of appropriate thickness (incl. the linethickness attribute when specified) :lollipop: 

[^1]: After writing this, I found a ConTeXt manual (Hans Hagen & Mikael P. Sundqvist, _Mathematics in ConTEXt,_  August 7, 2024 edition):  SkewedFractionHorizontalGap and SkewedFractionVerticalGap "do not make sense (for us) so we do not use them." -- It's an interesting reading, by the way.

[^2]: I don't think lowering the fraction with respect to the baseline is the _right_ thing to do. But see previous note: the mentioned ConTeXt book does it. Well, with plenty of other options etc. -- Likely beyond our own scope for now.

### DIFF
--- a/packages/math/base-elements.lua
+++ b/packages/math/base-elements.lua
@@ -1523,7 +1523,7 @@ function elements.sqrt:output (x, y, line)
    local symbol = {
       _r(self.radicalRuleThickness),
       "w", -- line width
-      2,
+      1,
       "j", -- round line joins
       _r(sw + s0),
       _r(self.extraAscender),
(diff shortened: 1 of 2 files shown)


khaledhosny:
> the related OpenType math parameters are, at best, unclear

https://github.com/notofonts/math/blob/main/documentation/building-math-fonts/index.md#skewedfractionhorizontalgap-skewedfractionverticalgap

khaledhosny:
> 1. After writing this, I found a ConTeXt manual (Hans Hagen & Mikael P. Sundqvist, _Mathematics in ConTEXt,_  August 7, 2024 edition):  SkewedFractionHorizontalGap and SkewedFractionVerticalGap "do not make sense (for us) so we do not use them." -- It's an interesting reading, by the way.

Always take ConTeXt’s opinions about OpenType with a big grain of salt.

alerque (approved):

PR #2167, opened 2024-11-14
feat(math): Support the MathML operator dictionary and many TeX-like aliases

### LINKED ISSUE #1642: binary/unary operator in TeX-like math mode
SILE 0.14.5

Using the regular minus dash (-) or the minus unicode symbol U+2212 (−) in TeX-like math:

```
\begin{document}
\use[module=packages.math]

\math[mode=display]{\pi=\sum_{k=0}^\infty\frac{1}{16^k}(\frac{4}{8k+1} - \frac{2}{8k+4} - \frac{1}{8k+5} - \frac{1}{8k+6})}

\math[mode=display]{\pi=\sum_{k=0}^\infty\frac{1}{16^k}(\frac{4}{8k+1} − \frac{2}{8k+4} − \frac{1}{8k+5} − \frac{1}{8k+6})}

\math[mode=display]{e^{i\pi} = -1}

\math[mode=display]{e^{i\pi} = −1}

\end{document}
```
image

Both cases apparently use U+2212 for the rendering of the formula (which seems great).

However:
- In the case U+2212 is used, it is better-looking as a binary operator, but badly looking as a unary operator.
- Inversely for the regular minus (which, theoretically at least, is probably the only valid option in TeX), looking better as unary operator, but being too close to the content as binary operator.

More generally perhaps, both `\pm` and `+` work well as binary operators, but are not so nicely placed as unary operator.

Experimenting with online renderers to MathJax and LaTeX, all cases with regular `-`, `+` or `\pm` look consistent with my expectations, so perhaps there would be a way to better handle unary vs. binary use of such operators in SILE's math implementation?

### LINKED ISSUE #2169: Outdated math example in user manual 0.15.6
The new pairing of opening/closing operators in TeX-like math makes this part of the package documentation outdated:

image

But the effect of braces for grouping would still have to be explained and is still valid for solving certain cases ("un-paired" stretchy operators). We should revise this part of the documentation, possibly depending on how other topic in #2148 are eventually addressed.

Omikhleia:
~~WIP for now, still things to check and fix, but most of the stuff is starting to work well... So it's good "saving  checkpoint" to share.~~

Good to go, IMHO :rocket: 

**First commit**
- Lots of refactoring in the atom / types logic... and some code clean-up.
- The operator dictionary table (for MathML), and most of the TeX-like commands (for that syntax), are now based on the W3C xml-entities [unicode.xml](https://raw.githubusercontent.com/w3c/xml-entities/gh-pages/unicode.xml) with the diff patch from [this issue](https://github.com/w3c/xml-entities/issues/11#issuecomment-2471104698) applied.[^1]
   - It ticks yet another checkbox in #2148 :cat: 
   - Lots of "big"-operators and extra integrals are now supported
     Eg. `\coprod`, `\bigcupdot`, `\intclockwise` and all their friends do the expect magic.
   - Lots of additional TeX-like commands (`\circ`, `\pod`, etc. to just mention a few) are supported.
   - More "stretchy" operators work (... _but not all yet, and proper support for accents will come later_....)
   - "form" attribute on MathML mo and a first take at "embellished" operators (paving the way for another checkbox in #2138 eventually, though we are not there yet...)

**Subsequent commits**

Ah, the operator table makes many things better, but others are worse...

  - fix(math): Spacing rules must distinguish binary and unary operators --> Needed consequence of making the minus-hyphen (U+002D) equivalent to the minus sign (U+2212) for math purposes. An occasion to also kill an old issue...
  - fix(math): The (escaped) percent is an ordinary atom in TeX-like syntax --> The former behavior had it encoded as a binary operator, it's now aligned with what (La)TeX would produce.
  - fix(math): A period must be allowed in TeX-like math syntax for numbers --> Decimals 0.123 later seen as `<mn>0.123</mn>` instead of `<mn>0</mn><mo>.</mo><mn>123</mn>` (with period as punctuation in MathML, possibly affecting spacing)
  - Some tests have been re-aligned / cleaned after analysis = all math tests are passing again.

Then..
  - fix(math): Improve spacing rules on limit-like operators --> Includes a refactor...
  - ... and more refactoring follows on those atom types with too many names ^^ = KISS :kissing_cat: a thing shall have one name only.
  - feat(math): Add pre-defined TeX-like operator functions (cos, sin, etc.) --> Bonus cake :cake: 
  - docs(math): Update description of atoms and TeX-like pre-defined constructs

**Issues addressed**

Closes #1642 

Closes #2169

Parts of some other issues are mentioned above. We're getting closer.
  
[^1]: The converter/extractor (in XSLT 1.0 i.e. xsltproc-compatible) from unicode.xml to our Lua structures is included too in the PR.


packages/math/atoms.lua
```diff
@@ -0,0 +1,38 @@
...
+   relationalOperator = 3,
+   openingSymbol = 4,
+   closeSymbol = 5,
+   punctuationSymbol = 6,
+   inner = 7,
+   overSymbol = 8,
+   underSymbol = 9,
+   accentSymbol = 10,
+   radicalSymbol = 11,
+   vcenter = 12,
```
Omikhleia, line 14: Just for the record, keeping those unused atom types for now until more things are cleared up, but eventually some will be removed or reassigned.

Omikhleia:
@alerque Good for review, in my opinion.
I have a few other commits that build upon this nice baby, but they are more controversial and I'll likely open dedicated issues before pushing a proposal.

Omikhleia:
> @alerque Good for review, in my opinion. I have a few other commits that build upon this nice baby, but they are more controversial and I'll likely open dedicated issues before pushing a proposal.

Oh wait, on a second thought (when I annotated the "unused" atom types above...) I just realized we can clear it up a bit here, with a good gain: bigOperator is ill-named from the start, it's TeX mathop and all of a sudden something becomes much clearer.

packages/math/atoms.lua
```diff
@@ -0,0 +1,38 @@
+local atomType = {
+   ordinary = 0,
+   bigOperator = 1,
```
Omikhleia, line 3: Very misleading from the start: we initially conflated big ops and other "mathop". It's true that both _may_ have largeop or movablelimits... But there's more to it here: `\sin` in TeX would be a mathop too, and proper spacing then occurs with but a small fix (due to how paired open/close logic, TeX does ofc things differently)
image

Omikhleia: Done, and a test case added.

packages/math/atoms.lua
Omikhleia: Cleaning performed

packages/math/tools/unicode-xml-to-sile.xsl
```diff
@@ -0,0 +1,201 @@
+<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+<!--
+  Stylesheet to convert the unicode.xml file to a SILE Lua file:
+  xsltproc unicode-xml-to-sile.xsl unicode.xml > ../packages/math/unicode-symbols-generated.lua
+  Where unicode.xml is:
+  https://raw.githubusercontent.com/w3c/xml-entities/gh-pages/unicode.xml
```
alerque, line 6: Is the data we need not available anywhere in UCD/CLDR? It seems like this should be in lua-cldr not here. I'm working on fixing this up with build targets so it can be regenerated for now, but this doesn't feel like the right home.

alerque: Bah. I did some digging an this xml-entities mapping seems to be maintained on it's own, and referenced in the MathML spec but not kept as part of UCD. Why do spec oversight orgs make things so convoluted? Bah.

Omikhleia: The W3C xml-entities repository mixes several sources (HTML entities, UCD and Unicode Math Technical Record, MathML specs, TeX / AMS / unicode-math names etc. ) into a single XML document.
It is used (according to Carlisle, the main editor, if I read that correctly) to prepare some of the tables in the MathML specs and elsewhere.
It's some sort of aggregation of specs from those various groups, converted (with lots of scripts and/or XSLT stylesheets) into one common file -- So it's a "derived" source indeed, but I found it neat to get most information directly, rather than doing the "assembly" work from all those scattered committee-designed sources ourselves, or rather than just "copying" some incomplete part of LaTeX's unicode-math's mappings (what SILE's code was doing, without providing a path for automating it).

Omikhleia: >  I'm working on fixing this up with build targets so it can be regenerated for now

Maybe wait -- I'll rebase this PR onto master (probably some small conflicts with the ssty / primes stuff on tests) and work on accent support, but I haven't checked yet that our Lua conversion was including all the necessary details (I did use some property to extract the combining classes, but didn't check it yet). We might need to map things slightly differently (as top/bottom and wide/regular accents). 
Eventually when we have it right, I don't expect these files to be regenerated often -- it's supposed to be fairly stable, though I did found some issue (now fixed by Carlisle).

Omikhleia: By the way, it obviously has some overlaps with our `char-def.lua` file used for linebreaking (notably), and if I understand it correctly, "stolen" from ConTeXt by Simons years ago. I haven't checked how those could be reconciled...

alerque: Yes, and I agree what you have is much better than not having any automation at all. While I still might move *where* we put the generated data, generating it from relevant canonical sources (aggregated ones or otherwise) is clearly the right move.

alerque: https://github.com/sile-typesetter/sile/issues/2176

alerque (approved):

PR #2170, opened 2024-11-14
Use en-US instead of en-GB spelling for math types

alerque:
I missed this in reviewing #2165, but we've been using en-US spelling though most of the code base. Two Ls in bevelled appears to be an en-GB thing.

Is there any particular reason *not* to stick with US spellings in math? Obviously this has some impact on people using this feature, so if there is any precident in the science and math publishing world for one over the other I'd be happy to follow it.

### DIFF
--- a/packages/math/typesetter.lua
+++ b/packages/math/typesetter.lua
@@ -147,8 +147,8 @@ function ConvertMathML (_, content)
       if #children ~= 2 then
          SU.error("Wrong number of children in mfrac: " .. #children)
       end
-      return SU.boolean(content.options.bevelled, false)
-            and b.bevelledFraction(content.options, children[1], children[2])
+      return SU.boolean(content.options.beveled, false)
+            and b.beveledFraction(content.options, children[1], children[2])
          or b.fraction(content.options, children[1], children[2])
    elseif content.command == "msqrt" then
       local children = convertChildren(content)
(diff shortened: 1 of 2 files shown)


packages/math/typesetter.lua
```diff
@@ -147,8 +147,8 @@ function ConvertMathML (_, content)
       if #children ~= 2 then
          SU.error("Wrong number of children in mfrac: " .. #children)
       end
-      return SU.boolean(content.options.bevelled, false)
-            and b.bevelledFraction(content.options, children[1], children[2])
+      return SU.boolean(content.options.beveled, false)
```
Omikhleia, line 150: So **NOT** here in any case. The [MathML standard](https://www.w3.org/TR/mathml4/#presm_mfrac_att) has the attribute with UK spelling: 
image

Omikhleia: (No strong opinion or objection for our own code and documentation - Just the attribute name must be the one from the W3C standard)

alerque: Okay, thanks for the reference. I'll add an exception to `typos` instead of changing the spelling. As much as it looks wrong to me all of our references are going to be in regard to MathML, and if that's the spelling used in the standard I don't see any sense in mixing and matching.

PR #2180, opened 2024-11-25
Patch up autodoc code blocks

### LINKED ISSUE #2168: Fix the \autodoc:codeblock I botched during the last release
While jumping through all the hoops releasing v0.15.6 last night (c.f. #2157) it became apparent that the verbatim fix for #2128 as addressed in #2129 looked great in the regression test but actually blew up the manual. The rule used above and below code blocks was wildly miss-spaced in some places and bloated the PDF by 40 some pages.

I probably should have just given up and gone to bed, but instead I tinkered with the autodoc package until the manual generation looked sane and forged ahead. In the light of the morning it is apparent that fix might have gotten the job done for the manual but it wasn't the *right* fix and it has broken the related regression test.

alerque:
Closes #2168


packages/autodoc/init.lua
```diff
@@ -371,6 +371,11 @@ function package:registerCommands ()
 
    -- Homogenizing the appearance of blocks of code
    self:registerCommand("autodoc:codeblock", function (_, content)
+      -- The way we use this command in the SILE manual is both on it's own *and* nested inside
+      -- a \raw environment. For this purpose we don't care about and don't want to bother with
+      -- making sure the leading and trailing whitespace is minimized in the SIL source, so
+      -- we trim it off here.
+      content = SU.ast.trimSubContent(content)
```
alerque, line 378: Does this change sound right to you @Omikhleia? The whitespace being passed in the `\raw` blocks now being (legitimately) treated as actual lines seems to be the main problem that blew up the manual. My previous fix was obviously wrong, this seems right now. I did tweak the alignment of the rule just a little bit and used raise instead of manually pushing VGlue because sorting out when to add spacing or not was much easier that way. Hence the final test expectations did change just a small shim. I also did *not* place `\novbreak` before and after the cod blocks as that seems to introduce yet another problem we'd have to cancel out.

Omikhleia: As far as I can say viewing the code (not tested yet), it seems ok.
What is the problem your are having with novbreak, if I may ask? I've some regression (it seems) on them too, though I've not been able yet to track its origin. (Something changed in paragraphing, possibly unrelated however.)

PR #2184, opened 2024-11-30
Math overset/underset and munder/mover spacing

### LINKED ISSUE #2133: TeX-like "def" macro doesn't allow inverting the arguments order (a.k.a. "overset" not supported)
Formula:

```
\def{overset}{\munderover{#2}{}{#1}}
R^n \overset{A}{→} R^m
```

This issue should have been a proposal: "Let's support TeX-like `\overset` which is a nice convenience"... and here would have been expected to produce the arrow with a smaller A on top of it... and surely a simple macro is enough, right?

... But the above crashes with "Wrong number of arguments (2) for command overset (should be 1)"

Extra test: as in the code, fractions are defined as `\def{frac}{\mfrac{#2}{#1}}` and `\frac{x}{y}` gives x / y...
Let's try `\def{invfrac}{\mfrac{#2}{#1}}` and expecting it to be produce y / x...
Same failure as above.

So we have these def-macro thing "to save (us) some typing"...
How can we achieve `\overset`? "When macros are not enough, creating new mathematical elements is quite simple" (... followed by lots of not-so-simple explanation for a newcomer). Sure. It's typically a case where a macro would have been neat, though.

:cactus:  *My dear math friends*, yes, again... Will you start to typeset some real maths one day? Just askin'

Subsidiary remark: Nowhere does the documentation says how little of the TeX-like math syntax[^1] is currently supported... and if there are any real goals in making it more complete.... But the reality check leads to serious doubts... :grimacing: 

[^1]: Read "commands". But you got the point.

Omikhleia:
Closes #2133

Did I say that my remaining math tasks would be slightly controversial? 
Well, this one isn't really, provides a good TeX-like feature and fixes a spacing weirdness, along with a nice test.

_Yet,_ the code has some smell. Not really what I added, the smell was just here unrevealed, and now we are all eyes on it.
I've spent literally hours to try finding the root cause, with no result, so I went for the most minimal things I could do to make it work without breaking something else (esp. MathML tables)... I cannot spent more efforts here on such a "small" topic, while "accents" are still awaiting and the last big missing block in our math experience...

Still, I'd advocate for the PR to be accepted, it does it's job supporting a missing feature nicely, and maybe one day someone more clever will pinpoint the underlying design issue. Advice is welcome...

### DIFF
--- a/packages/math/base-elements.lua
+++ b/packages/math/base-elements.lua
@@ -732,8 +732,25 @@ local function isNotEmpty (element)
    return element and (element:is_a(elements.terminal) or #element.children > 0)
 end
 
+local function unwrapSingleElementMrow (elt)
+   -- CODE SMELL.
+   -- For \overset or \underset in LaTeX, MathML would use <mover> or <munder>.
+   -- It would need to inherit the base's atom type, especially if the later is an operator
+   -- (binary, relational etc.), which is a fairly common case, e.g.
+   --   \overset{R}{=} (equality with a R above the equal in some Ramanujan summations),
+   -- but we can't remove 1-element mrow's in the math typesetter, or have them inherit
+   -- their base's atom type here above, because it breaks tables for some reasons
+   -- that I couldn't figure out.
+   if elt:is_a(elements.stackbox) and elt.direction == "H" and #elt.children == 1 then
+      return unwrapSingleElementMrow(elt.children[1])
+   else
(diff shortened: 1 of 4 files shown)


Omikhleia:
Just for making it easier to understand without checking the test file, is here what it allows:
image

alerque (approved):

PR #2186, opened 2024-12-01
Sync Flake with nix derivation changes for easier extension of available Lua modules

alerque:
Hey @doronbehar, my attempt to port the upstream packaging changes to the sile derivation to the flake here did not go quite as planned. It builds and runs, but it doesn't *work* as it should as evidenced by the fact that I can delete stuff in the `luaPackages` array and the build doesn't fail the configure stage where it should be detecting dependencies. That tells me the upstream package one is being used, not the locally defined one. I couldn't figue our the Nix language incantation to override the array in the passthru table and use it in the luaEnv at the same time.

### DIFF
--- a/build-aux/ax_lua_require.m4
+++ b/build-aux/ax_lua_require.m4
@@ -54,7 +54,7 @@ AC_DEFUN([AX_LUA_REQUIRE],[
         ACTION_IF_FOUND
     ], [
         AC_MSG_RESULT([unable to load])
-        m4_ifset([ACTION_IF_NOT_FOUND][ACTION_IF_NOT_FOUND],
+        m4_ifset([ACTION_IF_NOT_FOUND],[ACTION_IF_NOT_FOUND],
             [AC_MSG_FAILURE([cannot find Lua module MODULE])])
     ])
 
--- a/build-aux/pkg.nix
+++ b/build-aux/pkg.nix
@@ -16,6 +16,7 @@
   cargo,
   rustc,
   rustPlatform,
+  luarocks,
 
   # buildInputs
   cargo-edit,
@@ -28,7 +29,6 @@
   stylua,
   taplo,
   typos,
-  darwin,
   # FONTCONFIG_FILE
   makeFontsConf,
   gentium,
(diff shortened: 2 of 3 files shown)


doronbehar:
I ran:

```
nix eval --json .\#sile-luajit.buildInputs | jq --raw-output '.[]' | grep -- '-env'
```

On the `nix-sync` branch, and got:

```
/nix/store/8w44qpgqn7f3aa4plyfnzngn4rzsr0qw-luajit-2.1.1713773202-env
```

Then I applied:

```diff
diff --git i/build-aux/pkg.nix w/build-aux/pkg.nix
index 5d774f8e..79ae2d11 100644
--- i/build-aux/pkg.nix
+++ w/build-aux/pkg.nix
@@ -136,7 +136,6 @@ stdenv.mkDerivation (finalAttrs: {
     luaPackages =
       [
         # modules used at runtime
-        "cassowary"
         "cldr"
         "fluent"
         "linenoise"

```

And got a different hash:

```
/nix/store/0xc26lxpqgxwfndd2z3j4igplsmq14c8-luajit-2.1.1713773202-env
```

doronbehar:
I also see:

```
sile> Makefile.am:70: error: 'rusile.so' is not a standard library name
sile> Makefile.am:70: did you mean 'librusile.a'?
```

Which maybe related to https://github.com/NixOS/nixpkgs/issues/361005 .

doronbehar:
And the build succeeds, but running the executable fails to find the modules I removed.

alerque:
Yes, I got that too (per my message above you might have missed, being busy composing one yourself). I just identified why I didn't figure that out sooner: at runtime it finds the cassowary installation from my host system when it can't find it in the flake. As SILE is designed for users/projects to add or override their own dependencies at runtime that is expected, it just made debugging this hard.

That doesn't explain why it succeeds to build when it should not, but I'll look into that more.

alerque:
Your output is also suggesting that the LuaRocks manifests are not getting installed properly via `nix`. I suspect `luarocks list` is going to show nothing even when various Lua rocks *are* provided by Nix packages. This is probably a general ecosystem wide problem with Lua in Nix.

alerque:
Sooo, that turned out to be a bug in my autoconf macro code, not the Flake at all.

PR #2187, opened 2024-12-03
feat(math): Minimal support for accents in MathML and TeX-like commands

### LINKED ISSUE #2150: Math "accents" (for vectors, hats etc.) are not properly supported

Omikhleia:
Closes #2150

**Description**

**1**. So let's first be positive, regarding what this PR brings:
 - MathML accent/accentunder on mover/munder/munderover elements
 - Many top and bottom accents "quite well placed" (no longer floating off the formula): accents, arrows, etc.
 - Proper syntax support for these in TeX-like math syntax
 - TeX-like "variant" limits (varlimsup, varliminf, varprojlim, varinjlim)
 - Horizontal braces, brackets and parentheses, with appropriate TeX math support (over/underbrace, over/underbracket, over/underparen)

That's not too bad, from a user point of view.
IMHO, we wouldn't fare long without accents, and this is exactly what this PR solves on the surface, leaving room for further improvements (see below).

**2.** The implementation is _slightly_ more controversial:
 - Use of combining vs. non-combining Unicode code points (ideally, we likely need both, with additional conversion tables).
 - Use of horizontal re-scaling when glyph construction is not possible (partly depends on the previous point).
 - Incomplete implementation of MathML Core's layout recommendations. (Well, this was already the case, but it needs to be mentioned: This is not fully closing the gap)
 - Some internal flags introduced as MathML specifications still seems in the flux regarding these topics...

This being said, we are closer to the goal, and there's a limit on how much one can do with unfinished and evolving standards. I tend to believe something towards the goal is better than nothing at all, and that the benefits for being able to showcase the above features is greater than waiting for a better implementation that might never come?

What do you think?

### DIFF
--- a/packages/math/base-elements.lua
+++ b/packages/math/base-elements.lua
@@ -732,6 +732,18 @@ local function isNotEmpty (element)
    return element and (element:is_a(elements.terminal) or #element.children > 0)
 end
 
+local function getAccentMode (mode)
+   -- Size unchanged but leave display mode
+   -- See MathML Core §3.4.3
+   if mode == mathMode.display then
+      return mathMode.text
+   end
+   if mode == mathMode.displayCramped then
+      return mathMode.textCramped
+   end
+   return mode
+end
+
 local function unwrapSingleElementMrow (elt)
    -- CODE SMELL.
    -- For \overset or \underset in LaTeX, MathML would use <mover> or <munder>.
@@ -748,10 +760,13 @@ local function unwrapSingleElementMrow (elt)
    end
 end
 
-function elements.underOver:_init (base, sub, sup)
+function elements.underOver:_init (attributes, base, sub, sup)
    elements.mbox._init(self)
    base = unwrapSingleElementMrow(base)
    self.atom = base.atom
+   self.attributes = attributes or {}
+   self.attributes.accent = SU.boolean(self.attributes.accent, false)
+   self.attributes.accentunder = SU.boolean(self.attributes.accentunder, false)
    self.base = base
    self.sub = isNotEmpty(sub) and sub or nil
    self.sup = isNotEmpty(sup) and sup or nil
@@ -771,10 +786,10 @@ function elements.underOver:styleChildren ()
       self.base.mode = self.mode
    end
    if self.sub then
-      self.sub.mode = getSubscriptMode(self.mode)
+      self.sub.mode = self.attributes.accentunder and getAccentMode(self.mode) or getSubscriptMode(self.mode)
    end
    if self.sup then
-      self.sup.mode = getSuperscriptMode(self.mode)
+      self.sup.mode = self.attributes.accent and getAccentMode(self.mode) or getSuperscriptMode(self.mode)
    end
 end
 
@@ -816,7 +831,10 @@ function elements.underOver:_stretchyReshapeToBase (part)
 end
 
 function elements.underOver:shape ()
+   local constants = self:getMathMetrics().constants
+   local scaleDown = self:getScaleDown()
    local isMovableLimits = SU.boolean(self.base and self.base.movablelimits, false)
+   local itCorr = self:calculateItalicsCorrection() * scaleDown
    if not (self.mode == mathMode.display or self.mode == mathMode.displayCramped) and isMovableLimits then
       -- When the base is a movable limit, the under/over scripts are not placed under/over the base,
       -- but other to the right of it, when display mode is not used.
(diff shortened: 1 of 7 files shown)


packages/math/base-elements.lua
```diff
@@ -795,7 +814,7 @@ function elements.underOver:_stretchyReshapeToBase (part)
       if elt:is_a(elements.text) and elt.kind == "operator" and SU.boolean(elt.stretchy, false) then
          elt:_horizStretchyReshape(self.base.width)
       end
-   elseif part:is_a(elements.underOver) then
+   elseif part:is_a(elements.underOver) and (self.attributes.accent or self.attributes.accentunder) then
```
Omikhleia, line 817: This condition is dubious, and I might have messed up a rebase conflict here. Under check.

Omikhleia: So dubious that it breaks f19 and f22 in the Mozilla Test Suite :rofl:

Omikhleia: Lol.

packages/math/unicode-symbols.lua
```diff
@@ -128,6 +144,27 @@ symbols.Digamma = "Ϝ" -- Supported by LaTeX's unicode-math
...
+end
+
+-- In xml-entities's unicode.xml, we derived "ord" for U+034D COMBINING LEFT RIGHT ARROW BELOW
+-- as is doesn't have a "D" class (diacritic)
+-- Actually it's not the fault of xml-entities, such a character is not covered in the latest
+-- Unicode MathClass file: https://www.unicode.org/Public/math/revision-15/MathClassEx-15.html
+-- It should clearly be a "botaccent" however, for \underleftrightarrow to work as do other accents.
+-- Note the U+20E1 COMBINING LEFT RIGHT ARROW ABOVE has s a "D" and we thus mapped to an "accent",
+-- used for \overleftrightarrow, so the asymmetry is at least weird.
+-- For future reference, see also https://github.com/w3c/xml-entities/issues/12 butit's probably
```
Omikhleia, line 164: "butit" lol, was it late.

Omikhleia:
> What do you think?

For an in-depth understanding of the challenges, your can also check my [_SILE and the Hydra of Maths. Showcase & Critical Assessment 2024_](https://drive.google.com/file/d/1kGHIHUU9dsAsMBrGYotPUO0XyDLwD-CF), 
 - §2.3 pp. 34-37 for the intricacies regarding accents in MathML and Unicode...
 - Table 10 p. 49 and §3.3.12 pp. 53-56 regarding the results achieved in this PR.

alerque:
Yes I saw that. And read about ⅔ of it last night. And [posted it to lobste.rs](https://lobste.rs/s/4j7btu). Somebody else [posted it to Hacker News](https://news.ycombinator.com/item?id=42315621). PDFs are hard to get traction in the news with, but I do think that deserves to be read for anybody interested in the math typesetting ecosystem (and not just SILE's tiny part in it).

alerque:
Force push is just `stylua`.

alerque (approved):

PR #2193, opened 2024-12-09
Warn when asked to measure a character not shaped in a font

### LINKED ISSUE #1706: shaper:measureChar can return the width of a .notdef character
I had some surprise getting a width from calling SILE.shaper:measureChar() with a character not existing in my font. (It existed in another font I previously used.)

Investigating, the measured character is a `.notdef` one in that font, i.e. the function _does_ shape something and silently returns a value, but is it not what one may expect (but, likely, the width of a default square or of a :skull_and_crossbones: or whatever the font returns here).

See:

https://github.com/sile-typesetter/sile/blob/607dcf7b3d8a83547779758cb70f6285a07c4848/shapers/base.lua#L67-L69

Shouldn't this be:

```
  if #items > 0 then
    if items[1].gid == 0 or items[1].name == ".null" or items[1].name == ".notdef" then
      SU.error("Unable to measure character", char)
    end
    return { height = items[1].height, width = items[1].width }
  else
```

N.B. The proposed extra check is what the fallback shaper does...

https://github.com/sile-typesetter/sile/blob/607dcf7b3d8a83547779758cb70f6285a07c4848/shapers/fallback.lua#L104

... to actually ensure the shaped token(s) lead to valid glyph(s).

(EDIT: cleaned up the proposed code, I had left some stuff from earlier debugging)

alerque:
Closes #1706

### DIFF
--- a/spec/shaper_spec.lua
+++ b/spec/shaper_spec.lua
@@ -12,4 +12,34 @@ describe("SILE.shapers.base", function ()
       assert.is.truthy(space.width.stretch > SILE.types.measurement(0))
       assert.is.truthy(space.width.shrink > SILE.types.measurement(0))
    end)
+
+   describe("measureChar", function ()
+      SILE.settings:set("font.family", "Libertinus Serif", true)
+
+      it("should measure simple characters", function ()
+         local measurements, found = SILE.shaper:measureChar("a")
+         assert.is.truthy(found)
+         assert.is.truthy(measurements.width > 0)
+         assert.is.truthy(measurements.height > 0)
+      end)
+
+      it("should measure multiple characters", function ()
+         local measurements, found = SILE.shaper:measureChar("ab")
+         assert.is.truthy(found)
+         assert.is.truthy(measurements.width > 0)
+         assert.is.truthy(measurements.height > 0)
+         -- Composite character should be taller than base character
+         assert.is.truthy(measurements.width > SILE.shaper:measureChar("a").width)
+      end)
+
+      -- TODO, we also need a test for composite characters, but I couldn't finde one in our test fonts
+      -- it("should measure composite characters", function ()
+      --    local measurements, found = SILE.shaper:measureChar("â")
+      --    assert.is.truthy(found)
+      --    assert.is.truthy(measurements.width > 0)
+      --    assert.is.truthy(measurements.height > 0)
+      --    -- Composite character should be taller than base character
+      --    assert.is.truthy(measurements.height > SILE.shaper:measureChar("a").height)
+      -- end)
+   end)
 end)


shapers/base.lua
```diff
@@ -67,8 +67,18 @@ function shaper:measureChar (char)
    local options = SILE.font.loadDefaults({})
    options.tracking = SILE.settings:get("shaper.tracking")
    local items = self:shapeToken(char, options)
-   if #items > 0 then
-      return { height = items[1].height, width = items[1].width, depth = items[1].depth }
+   SU.dump({ char, items })
```
Omikhleia, line 70: This is going to be very verbose without any real need for it, no?

shapers/base.lua
```diff
@@ -67,8 +67,18 @@ function shaper:measureChar (char)
...
+   if #items >= 1 then
+      local item = items[1]
+      if item.name == ".notdef" then
```
Omikhleia, line 73: Would checking item.gid be more robust than a name check? Normally, if I read that well, .notdef is supposed to be GID 0 according to [Microsoft Typography](https://learn.microsoft.com/en-us/typography/opentype/spec/recom#glyph-0-the-notdef-glyph)
I am no font expert -- my initial suggestion in the original issue was checking the GID 0, and also for "additional safeguard", the .notdef name, and the .null name -- I am not sure regarding the latter and how I sort of guessed it there (from what I read now it seems the NULL / .null was something from Apple's TTF, but no longer applies to OTF -- Perhaps I still have some old TTF's ? Dunno).

shapers/base.lua
```diff
@@ -67,8 +67,18 @@ function shaper:measureChar (char)
...
+         SU.warn(
+            ("No glyph available for '%s' in %s, returning measurements of '.notdef' instead"):format(
+               char,
+               options.family
+            )
+         )
+      end
+      return { height = item.height, width = item.width, depth = item.depth, name = item.name }
```
Omikhleia, line 81: Do we really need returning the name? Could be the GID, or just a boolean "undefined = true", so the caller wouldn't have to perform a ".notdef" explicit check. The name might not be meaningful or portable any for most extended characters (ligatures, etc.), would it?

shapers/base.lua
```diff
@@ -67,8 +67,18 @@ function shaper:measureChar (char)
    local options = SILE.font.loadDefaults({})
    options.tracking = SILE.settings:get("shaper.tracking")
    local items = self:shapeToken(char, options)
-   if #items > 0 then
-      return { height = items[1].height, width = items[1].width, depth = items[1].depth }
+   SU.dump({ char, items })
+   if #items >= 1 then
+      local item = items[1]
+      if item.name == ".notdef" then
+         SU.warn(
```
Omikhleia, line 74: But should we really warn, or let the caller warn if needed?
Ex. in the zenkaku case below, we'll have a specific warning, to this one comes as extraneous, no?

shapers/base.lua
Omikhleia: N.B. I am not fond of warnings is such utilities. The caller could try to measure char X, and fallback to char Y if X is not defined, without warnings.

shapers/base.lua
alerque: Um. Yes of course. That's just debug output leftover from development and should not have been in the commit at all.

shapers/base.lua
alerque: Probably. I am not sure there is a 100% solution with one check, we probably need to combine a bunch of possible combinations of results. he thing is don't have the energy to ferret them all out right now and I'm willing to start with something. Adding checks that cover more possible failure modes should be easy as long as we do it with some boolean return value.

I did just change the default check to checking the GID for 0.

shapers/base.lua
alerque: Yes I think we need to return the name. That way downstream users of the API can make their own call about whether to use the results or not. Actually I just took it a step farther and passed the entire table of info we got back from the shaper. That includes all the keys currently used and also some information about offsets that could be useful for some measurement scenarios. Also it includes the GID and name in case the consumer wants to check them.

My rational for including the warning was that we have better access to the font name and such here in the function than the user does easily when using the function so it was easier to create a meaningful warning. I'm not set on it though, so lets try a different shape ... returning a boolean status flag might be pretty useful too. How does that look?

shapers/base.lua
alerque: I reworked this, see previous comment for discussion points on throwing a warning or not.

alerque:
The fact that this failed a test may have just revealed a big lurking bug, but I don't have time to pull it apart right now and I urgently need a tagged release so I can send a book to press with a known version. I'm going to bump this and release dot-8, but I'm open to a quick follow up patch release if we figure this or anything else out. Right now macOS being 2 versions behind because I broke Darwin builds is a big deal and that needs to get fixed ASAP.

shapers/base.lua
```diff
@@ -67,8 +67,9 @@ function shaper:measureChar (char)
    local options = SILE.font.loadDefaults({})
    options.tracking = SILE.settings:get("shaper.tracking")
    local items = self:shapeToken(char, options)
-   if #items > 0 then
-      return { height = items[1].height, width = items[1].width, depth = items[1].depth }
+   if items and items[1] then
+      local item = items[1]
+      return item, item.gid ~= 0
```
khaledhosny, line 72: This code (even before this PR) seems to assume that a character will be mapped to only a single glyph, but this is not true in general.

Omikhleia: Exactly. So it would need to return the **total** witdth, height and depth of the measured content, not to break existing code (and meet the most common use case).

alerque: Done. Incidentally it can also be used to measure multiple characters now.

Issue #2202, opened 2024-12-19
Can't run either `\printbibliography` or `\csl:reference`. 


jodros:
Both give me this same error:

```
 Error: runtime error: /usr/local/share/sile/packages/bibtex/csl/engine.lua:106: bad argument #1 to 'gsub' (string expected, got nil)
stack traceback:
        [C]: in ?
        [C]: in function 'gsub'
        /usr/local/share/sile/packages/bibtex/csl/engine.lua:106: in function 'init'
        .../local/share/sile/lua_modules/share/lua/5.1/pl/class.lua:38: in function 'call_ctor'
        .../local/share/sile/lua_modules/share/lua/5.1/pl/class.lua:171: in function 'CslEngine'
        /usr/local/share/sile/packages/bibtex/init.lua:487: in function </usr/local/share/sile/packages/bibtex/init.lua:476>
        /usr/local/share/sile/core/sile.lua:630: in function 'call'
        /usr/local/share/sile/core/sile.lua:421: in function 'process'
        /usr/local/share/sile/core/sile.lua:500: in function 'processString'
        /usr/local/share/sile/core/sile.lua:550: in function </usr/local/share/sile/core/sile.lua:513>
```

I've been trying with different `.bib` and `.csl` files as well...

Omikhleia:
Line 106 is  `self.subsequentAuthorSubstitute = self.inheritable["bibliography"]["subsequent-author-substitute"]` right?
Which CSL file are you using?

Omikhleia:
Actually, MLA and Chicago have `subsequent-author-substitute` (which are the ones I mostly used, or derivatives of thereof)... but not a bunch of other styles that don't do author substitution.

So the code lacks a little condition check there indeed. Should be easy to fix.

Omikhleia:
(Still, let me know which style you primarily target, and I'll do some extra tests upon my return from X-mas vacations)

jodros:
Now, after a new attempt, I've found that with [this one](https://github.com/citation-style-language/styles/blob/6b7b611908b20c91f34110d1c9489fb3278e0ef5/universidade-federal-do-espirito-santo-abnt.csl) it at least generates the PDF, even though the command `\printbibliography` still doesn't work, while `\csl:reference` does

Those that don't work are (for now):
[...]

PR #2204, opened 2025-01-01
CSL fixes (subsequent-author-substitute, LDoc typings)

Omikhleia:
See #2202


alerque:
Test failure in LuaJIT run is [this shenanigans](https://github.com/starwing/luautf8/pull/55).

alerque:
I *believe* the Ubuntu build failure is also the same issue although it is less obvious. It seems to be picking up a pre-built set of our vendored LuaRocks that were built in a previous CI run. We seem to have cached the build results in spite of the actual failure to build the module. This is somewhat confirmed by the LuaJIT test run failing **not** on building the `luautf8` module where the error happens but later when we try to use it.

alerque (approved):

PR #2206, opened 2025-01-02
Fix use of legacy \script function with non-modules

alerque:
The current implementation assumes that a table is returned. I checks to see if that table is a proper SILE module, but it doesn't stop to think what happens if just a function, string, or other data is returned.


alerque:
Test failure in LuaJIT run is [this shenanigans](https://github.com/starwing/luautf8/pull/55).

PR #2207, opened 2025-01-02
Overhaul indexer

### LINKED ISSUE #1339: Enhancing the indexer package

Omikhleia:
Closes #1339 

 - Automatically register the indexer on endpage class hook = using the 0.14.x way so any document class can use it (vs. the 0.12.x way where classes had to explicitly call a method)
 - Use language-dependent sorting (collation) in indexer = Sorting based on ICU for the current language
 - Support page ranges in indexer = either `none` (all pages are listed without page range collapsing) or `expanded` (consecutive pages are collapsed in full, e.g. 301-305)
 - Support links to pages in the indexer = hyperlinks if the `pdf` package is previously loaded
 - Add indexer options `minimal`/`minimal-two`  for page range collapsing = consecutive pages are collapsed in minimal ranges (e.g. 301-5) or minimal ranges with at least two digits (e.g. 301-05). N.B. the "minimal", "minimal-two" and "expanded" names are loosely derived from those defined in CSL.[^1]
 - Add option for filler between item and page in indexer = `dotfill` (leaders) (default), `fill` (space as before), `comma` (content is not line-flushed). These are the most usual ways of presenting an index, i.e.
    - `item . . . . . 50, 60-63`
    - `item           50, 60-63`
    - or `item, 50, 60-63`
  - Add new styling hooks to the indexer = introduces hooks on entry and pages, (BREAK) removes hard-coded verbatim styling (@davidchisnall see discussion below)
  - Allow indexes to be used anywhere in a document = As for table of contents, use an intermediary file, so the index(es)
may be used (at the cost of a SILE re-run) anywhere, and not only on their own page at the end of a document... (This does fix several items from #1339, notably lost index entries if there's no page break before the `\printindex` and index entries are on that page[^2])
 - refactor(packages): Change `index:item` into an indexer package method = It's no longer callable from SIL/XML, so a package method is certainly better than a command.
 - feat(typesetters): Add content to text utility to the base typesetter = **at last**[^3]
 - refactor(packages): Use typesetter content to text in indexer and tableofcontents
 - docs(packages): Improve indexer package documentation = As it's name says :)

[^1]: For some consistency, although indexes are not related to CSL. Note that some language may prefer a certain format (e.g. French typography rules usually recommend full expanded ranges, while English rules may prefer a minimal variant), but the package leaves that decision to the user.

[^2]: It's also more general towards supporting indexes that are not at the end of document on their own pages. I do have cases for indexes early in the document (as for table of contents, one may want them this way, in some books...). Reading the index from an intermediary files also allows for other packages to use it (e.g. one could implement a fancyindex as as did for the ToC in [fancytoc.sile](https://github.com/Omikhleia/fancytoc.sile), and do their own rendering based on another logic while reusing most of the "common" code)

[^3]: As noted in the commit log, there are good reasons to do that (I always was convinced of it, at least! Doh)... @davidchisnall Just for the notice, I'll need to backport that to resilient's (actually silex's) modified typesetter, if that proposed change is accepted.

### DIFF
--- a/packages/indexer/init.lua
+++ b/packages/indexer/init.lua
@@ -187,6 +187,31 @@ function package:formatPages (pages)
    return _addDelimiter(ret, self.config['page-delimiter'])
 end
 
+-- Output an index entry.
+-- @tparam table options The index entry options (passed to the style hooks).
+-- @tparam table entry The index entry as a SILE AST.
+-- @tparam table pages The formatted pages as a SILE AST.
+function package:outputIndexEntry (options, entry, pages)
+   SILE.settings:temporarily(function ()
+      if self.config.filler ~= "comma" then
+         SILE.settings:set("typesetter.parfillskip", SILE.types.node.glue())
+      end
+      SILE.settings:set("current.parindent", SILE.types.node.glue())
+      SILE.call("index:entry:style", options, entry)
+      if self.config.filler == "dotfill" then
+         SILE.call("dotfill")
+      elseif self.config.filler == "fill" then
+         SILE.call("hss")
+      elseif self.config.filler == "comma" then
+         SILE.typesetter:typeset(", ")
+      else
+         SU.error("Unknown filler: " .. self.config.filler)
+      end
+      SILE.call("index:pages:style", options, pages)
+      SILE.call("smallskip")
+   end)
+end
+
 function package:registerCommands ()
    self:registerCommand("indexentry", function (options, content)
       if not options.label then
@@ -242,35 +267,11 @@ function package:registerCommands ()
       end
       SU.collatedSort(sortedIndex)
       SILE.call("bigskip")
-      for _, k in ipairs(sortedIndex) do
-         local pages = self:formatPages(index[k])
-         SILE.call("index:item", { pages = pages, index = options.index }, { k })
+      for _, entry in ipairs(sortedIndex) do
+         local pages = self:formatPages(index[entry])
+         self:outputIndexEntry({ index = options.index }, { entry }, pages)
       end
    end, "Print the index")
-
-   self:registerCommand("index:item", function (options, content)
-      -- Unconventional: options.pages is an AST
-      local pages = options.pages
-      options.pages = nil
-      SILE.settings:temporarily(function ()
-         if self.config.filler ~= "comma" then
-            SILE.settings:set("typesetter.parfillskip", SILE.types.node.glue())
-         end
-         SILE.settings:set("current.parindent", SILE.types.node.glue())
-         SILE.call("index:entry:style", options, content)
-         if self.config.filler == "dotfill" then
-            SILE.call("dotfill")
-         elseif self.config.filler == "fill" then
-            SILE.call("hss")
-         elseif self.config.filler == "comma" then
-            SILE.typesetter:typeset(", ")
-         else
-            SU.error("Unknown filler: " .. self.config.filler)
-         end
-         SILE.call("index:pages:style", options, pages)
-         SILE.call("smallskip")
-      end)
-   end, "Output an index item (normally an internal command)")
 
    -- Hooks for styling the index
    self:registerCommand("index:entry:style", function (_, content)
--- a/packages/indexer/init.lua
+++ b/packages/indexer/init.lua
@@ -3,9 +3,9 @@ local base = require("packages.base")
 local package = pl.class(base)
 package._name = "indexer"
 
-function package:buildIndex ()
+function package.buildIndex (class) -- Called from the class, not as a package method
    local nodes = SILE.scratch.info.thispage.index
-   local thisPage = self.class.packages.counters:formatCounter(SILE.scratch.counters.folio)
+   local thisPage = class.packages.counters:formatCounter(SILE.scratch.counters.folio)
    if not nodes then
       return
    end
@@ -22,21 +22,14 @@ function package:buildIndex ()
       end
    end
 end
--- if content then
---   for i = 1, #content do
---     if not SILE.scratch.index.commands[content[i].label] then
---       SILE.scratch.index.commands[content[i].label] = {}
---     end
---     SILE.scratch.index.commands[content[i].label][class:formatCounter(SILE.scratch.counters.folio)] = 1
---   end
--- end
 
 function package:_init ()
    base._init(self)
+   self:loadPackage("infonode")
+   self.class:registerHook("endpage", self.buildIndex)
    if not SILE.scratch.index then
       SILE.scratch.index = {}
    end
-   self:deprecatedExport("buildIndex", self.buildIndex)
 end
 
 function package:registerCommands ()
@@ -60,7 +53,6 @@ function package:registerCommands ()
    end)
 
    self:registerCommand("printindex", function (options, _)
-      self:buildIndex()
       if not options.index then
          options.index = "main"
       end
@@ -99,7 +91,6 @@ The entry can be styled using the \autodoc:command{\index:item} command.
 
 Multiple indexes are available and an index can be selected by passing the \autodoc:parameter{index=<name>} parameter to \autodoc:command{\indexentry} and \autodoc:command{\printindex}.
 
-Classes using the indexer will need to call its exported function \code{buildIndex} as part of the end page routine.
 \end{document}
 ]]
 
--- a/packages/indexer/init.lua
+++ b/packages/indexer/init.lua
@@ -19,7 +19,7 @@ local function _isNotSamePage (p1, p2)
 end
 
 -- Group pages into ranges of consecutive pages.
--- @tparam table pages A list of pages with pageno and link fields.
+-- @tparam table pages A list of pages with pageno (counter) and link (internal string) fields.
 -- @treturn table A list of ranges, each containing a list of pages.
 local function _groupPageRanges(pages)
    local ret = {}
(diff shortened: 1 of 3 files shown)


davidchisnall:
I'm not sure if it's a side effect of something else I've done, but this appears to print the index in a monospaced font.

Omikhleia:
> I'm not sure if it's a side effect of something else I've done, but this appears to print the index in a monospaced font.

It always did. But I'll remove that formatting, as part of my work here -- there's no reason for monospace to be used.

davidchisnall:
I've copied the file into my project and am using it there until it's part of a release (yay for modularity!).  It seems to be working nicely.  Ideally I'd want to be able to split index entries into letter headings and make it use a two-column layout, but I think the latter is independent of this package (though I'm struggling to make it work) and the latter is not vital (and maybe can be done in the style hook?).

Thanks!

Omikhleia:
> It seems to be working nicely.

Thanks, glad it's helpful.
I still have a few commits to propose, but I am close to what I intended here (the rest is mostly some clean-up refactor, and documentation).

> Ideally I'd want to be able to split index entries into letter heading

I actually wanted the same, but problem is that we don't have enough ICU support for that. Checking for differences on the first character to insert a "letter milestone" is not sufficient: E.g. In French "Éric" and "Ethan" should possibly both be under E, not under E and É). It might be doable with ICU (diacritic removal, via a transform `NFD; [:Nonspacing Mark:] Remove; NFC.`), but
 - It would need some extra C code for this in our ICU wrapper...
 - The solution might to be general either, as some languages expect some different rules
 - And even diacritic removal might not be enough for a general solution (e.g. in French œ would be under the "O" letter, but some languages expect some characters like this to be in their own section)

--> On such a topic, a dedicated issue would be required, and one would have to check the litterature, regarding how this is addressed in general.

> ... make it use a two-column layout

Wholly independent from this package indeed (and we don't really have robust multicolumn layouts _yet_).

davidchisnall:
For the sectioning, a hook that provides the name of the current and previous (empty string / nil for the first one) and lets you return nothing or a line break, with per-language defaults and a warning if your language didn't have one would probably be a nice start.  The implementation for English is fairly simple but, as you say, for others it can be different (for some languages, I suspect, it's trivial because there's no equivalent notion and everything ends up in a single section).

Omikhleia:
> For the sectioning, a hook that provides the name of the current and previous 

As said, wait for this to PR to be finished, reviewed and merged, and feel free to propose something... (But as far as I am concerned, I'd be against any hook that works only for English.)

Omikhleia:
@alerque Ready for review.
I have no idea why the tests fail here (on a busted test for the font manager); they visibly passed on my own fork.

Omikhleia:
Some dumb illustration of what it does (entries are kinda stupid, but heh)

image

alerque:
> but problem is that we don't have enough ICU support for that

If you tell be exactly what function(s) you want exposed (preferably from icu4x) I can expose it to Lua via the `rusile` module. I'm still mulling over various approaches to a public facing API for that, but as long as we consume it internally in a package we can play around with how it will work because the only breaking change will be the public feature set of the package, not *how* we reached into ICU to accomplish it.

alerque:
> For the sectioning, a hook that provides the name of the current and previous (empty string / nil for the first one) and lets you return nothing or a line break

A hook is fine if we can come up with something we think will have wide utility, but this doesn't sound like a robust way to define such a hook. Also if the current implementation has a callback function for outputting each line, *implementing* such a sectioning device yourself should be trivial (just stash the *last* entry in a local variable and check it when outputting an entry and output a header first if desired.

alerque:
> * (BREAK) removes hard-coded verbatim styling

@Omikhleia You mention this as breakage, but it doesn't sound to me like something we need to hold off on for a major release bump since I would consider the existing weird styling a bug not a feature. What you you think? I'm targeting this for the next minor (patch level since we're stuck in ZeroVer mode) release unless I hear suggestion to the contrary.

Omikhleia:
> but it doesn't sound to me like something we need to hold off on for a major release bump

I think the same, IMHO we could just go for it in a minor/path release and not bother too much.
 - It's not _that_ breaking and your Release Notes usually give enough information to potential users :cat: 
 - This package moreover was probably never really used for real:
     - it was still using the 0.12 way to be used (= calling a method from the class, which implies a custom class, which is likely not the case for most users)
     - no one (but me) complained about me
     - no one had ever noticed (but me) that sorting wasn't language-aware etc.
     - same, no one mentioned the hard-coded style behavior...
 - It's marked as "experimental" for some time now in the manual, and I don't think we should necessarily be afraid of slight breaking change in highly experimental packages
   image

Should we change the "experimental" status in the manual, by the way? I am not even sure, we still lag behind LaTeX's capabilities here (with its "see also" entries, multi-level indexes, etc.). We can just move forward, and if it starts being used, they we have interesting topics to address.

alerque:
Oh yes of course, as long as it's still marked experimental I wouldn't even consider making a breaking release for changes. Just change it for the better and hope for the best ;-)

I agree it probably hasn't seen much use. My own projects have their own implementations (two of them I see). I'll have to work on bringing them in closer to the source by basing on and extending this. That will help get it some exercise and also make it easier to contribute improvements.

In the mean time I see no reason not to move ahead with this sooner rather than later. We can even consider removing the experimental tag after this PR (anytime before the release or whatever release follows when we decide the usage should be considered somewhat stable).

alerque (approved):

PR #2216, opened 2025-01-15
Enhance parallel package with overflow frame alignment, footnote handling and improved documentation

no-vici:
### Summary
This pull request introduces significant enhancements to the `parallel` package, including improved footnote handling, reasonable alignment logic, and expanded documentation.

### Key Features and Changes
1. **Footnote Handling:**
   - Added `parallel_footnote` and related commands for managing footnotes in parallel frames.
   - Implemented `generateFootnoteId` and `getFootnoteHeight` functions to uniquely identify and measure footnotes, supporting efficient layout and splitting.
   - Introduced the `typesetFootnotes` function to process footnotes, handle overflow, and manage their layout dynamically across frames.

2. **Frame Alignment:**
   - Improved `sync` command to include `parskip` handling and modularized balancing logic.
   - Enhanced `parallelPagebreak` function to handle overflow content, dynamic frame balancing, and proper synchronization of frames and footnotes.

3. **Documentation:**
   - Expanded the `package.documentation` section with detailed explanations, usage notes, and examples.
   - Highlighted challenges and best practices for maintaining alignment, managing footnotes, and working with SILE’s parallel typesetting capabilities.

### Why This Change is Important
These updates improve the usability and flexibility of the `parallel` package by addressing key limitations and providing tools for managing complex layouts. The enhanced documentation also helps users understand and utilize the package more effectively with what SILE can offer.

### Testing and Compatibility
- The updates have been tested with standard SILE setups using the `Gentium Plus` font family at a 12pt font size.
- Known limitations include unsupported custom `document.parskip`, `document.lineskip` and potential alignment challenges with varying font sizes or custom spacing settings.

### Next Steps
- Additional testing with more complex layouts and document classes.
- Gathering user feedback to refine and extend the package’s functionality.

### DIFF
--- a/packages/parallel/init.lua
+++ b/packages/parallel/init.lua
@@ -3,7 +3,7 @@
 ----------------------------------------------------------------
 --
 --  This file can simply be dropped into ⟨project⟩/packages
---  or placed in a $SILE_HOME search path.  The only external
+--  or placed in a $SILE_HOME search path. The only external
 --  requirement is plain (pl) which ships with SILE.
 --
 ----------------------------------------------------------------
@@ -22,24 +22,24 @@ local log = function(...)
 end
 
 ----------------------------------------------------------------
---  Two pools of typesetters: main‑text frames & footnote frames
+--  Two pools of typesetters: main text frames & footnote frames
 ----------------------------------------------------------------
 local typesetterPool, footnotePool = {}, {}
 
 ----------------------------------------------------------------
---  Misc. shared state tables
+--  Shared state tables
 ----------------------------------------------------------------
-local calculations = {} -- running height / mark bookkeeping
-local folioOrder = {} -- frame order on the physical page
+local calculations = {} -- Running height / mark bookkeeping
+local folioOrder = {} -- Frame order on the physical page
 
 ----------------------------------------------------------------
---  A no‑op typesetter (used once during class initialisation)
+--  A no-op typesetter (used during class initialization)
 ----------------------------------------------------------------
 local NullTypesetter = pl.class(SILE.typesetters.default)
 NullTypesetter.outputLinesToPage = function() end
 
 ----------------------------------------------------------------
---  Footnote manager (all footnote logic is encapsulated here)
+--  Footnote manager: encapsulates all footnote logic
 ----------------------------------------------------------------
 local FootnoteManager = {}
 FootnoteManager.__index = FootnoteManager
@@ -58,27 +58,30 @@ end
 function FootnoteManager:get(frame)
    return self.frames[frame]
 end
+
 function FootnoteManager:clear(frame)
    self.frames[frame] = {}
 end
+
 function FootnoteManager:processNotes(frame, fn)
    for _, n in ipairs(self.frames[frame]) do
       fn(n)
    end
 end
+
 function FootnoteManager:generateId(frame, note)
    return frame .. ":" .. note.marker
 end
 
---  Measure the height of one footnote (with caching)
+-- Measure the height of one footnote (with caching)
 function FootnoteManager:height(frame, note, ts)
    local id = self:generateId(frame, note)
    if self.footnoteHeightCache[id] then
       return self.footnoteHeightCache[id], nil
    end
 
-   local q = {}
    ts:pushState()
+   local q = {}
    ts.state.outputQueue = q
    SILE.call("parallel_footnote:constructor", { marker = note.marker }, note.content)
    ts:popState()
@@ -92,9 +95,9 @@ function FootnoteManager:height(frame, note, ts)
 end
 
 ----------------------------------------------------------------
---  Utility helpers (frame‑height accounting, phantom glue, …)
+--  Utility Helpers
 ----------------------------------------------------------------
--- Utility function: Iterate over typesetters in a pool and apply a callback function.
+-- Iterate over typesetters in a pool and apply a callback function
 local function iterateTypesetters(pool, callback)
    local old = SILE.typesetter
    for frame, ts in pairs(pool) do
@@ -104,12 +107,9 @@ local function iterateTypesetters(pool, callback)
    SILE.typesetter = old
 end
 
--- Utility function: Calculate the height of new material for a given frame.
+-- Calculate the height of new material for a given frame
 local function calculateFrameHeight(frame, ts)
    local h = calculations[frame].cumulativeHeight or SILE.types.length()
-   -- typesetter.state.outputQueue now holds actual content reflecting the real layout of lines.
-   -- Therefore, we can calculate the height of new material by adding the height of each line
-   -- in the queue.
    for i = calculations[frame].mark + 1, #ts.state.outputQueue do
       local ln = ts.state.outputQueue[i]
       h = h + ln.height + ln.depth
@@ -117,12 +117,11 @@ local function calculateFrameHeight(frame, ts)
    return h
 end
 
--- Utility function: Manage the heights of frames.
+-- Manage the heights of frames
 local manageFrameHeights = function(frames, callback)
    local frameHeights = {}
    local maxHeight = SILE.types.length()
 
-   -- Calculate the height of new material for each frame
    iterateTypesetters(frames, function(frame, typesetter)
       local height = calculateFrameHeight(frame, typesetter)
       frameHeights[frame] = height
@@ -131,17 +130,15 @@ local manageFrameHeights = function(frames, callback)
       end
    end)
 
-   -- Apply the callback function to each frame
    iterateTypesetters(frames, function(frame, typesetter)
       callback(frame, frameHeights[frame], maxHeight - frameHeights[frame])
    end)
 end
 
--- Balances the height of content across frames by adding glue to the shorter frame.
+-- Balances the height of content across frames by adding glue to shorter frames
 local addBalancingGlue = function()
    manageFrameHeights(typesetterPool, function(frame, currentHeight, heightDiff)
       if heightDiff:tonumber() > 0 then
-         -- Add vkern (vertical glue) to balance the frame height
          local typesetter = typesetterPool[frame]
          table.insert(typesetter.state.outputQueue, SILE.types.node.vkern({ height = heightDiff }))
          log("Added balancing glue of", heightDiff, "to bottom of frame", frame)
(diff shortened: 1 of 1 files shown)


packages/parallel/init.lua
```diff
@@ -3,65 +3,394 @@ local base = require("packages.base")
...
+   local typesetter = typesetterPool[frame]
+
+   -- Calculate precise line height by typesetting a sample line
+   -- local lineHeight = calculateLineHeight("hp")
+
+   -- log("Precise lineHeight after simulation = ", lineHeight)
+
+   -- If lineHeight could not be calculated, fall back to baselineSkip and lineSkip of the document
+   if not lineHeight then
+      local baselineSkip = SILE.settings:get("document.baselineskip").height or SILE.types.length({ length = 0 })
```
Omikhleia, line 82: `SILE.types.length({ length = 0 })` > `SILE.types.length()` (zero by default)

packages/parallel/init.lua
```diff
@@ -3,65 +3,394 @@ local base = require("packages.base")
...
+      local lineSkip = SILE.settings:get("document.lineskip").height or SILE.types.length({ length = 0 })
```
Omikhleia, line 83: The problem here is that depending on the linespacing algorithm (the default one is more or less the TeX one), the lineskip is used conditionally (when boxes are greater than the baselineskip and would collide vertically), not always. In most cases (i.e. regular text with decent settings for the baselineskip), the lineskip isn't applied.

packages/parallel/init.lua
```diff
@@ -99,75 +438,196 @@ function package:_init (options)
...
+
+   self:registerCommand("parallel_footnote:constructor", function(options, content)
+      local markerText = options.marker or "?" -- Default marker if none provided
+
+      SILE.settings:temporarily(function()
+         -- Set font footnotes, 80% of current font size
+         SILE.settings:set("font.size", SILE.settings:get("font.size") * 0.80)
+
+         -- Measure the marker width
+         local markerWidth = measureStringWidth(markerText)
```
Omikhleia, line 552: You are computing the width by shaping the marker text. This will be shaped _again_ below when you typeset it (`SILE.typesetter:typeset(markerText)`). I'd suggest rather using a box: you'd get the width and can just re-use it as-is afterwards, without re-computing costs under the hood.
```
local markerBox = SILE:typesetter:makeHbox({ markerText })
local markerWidth = markerBox.width
...
SILE.typesetter:pushHbox(markBox) -- don't re-typeset, just push our box

File: packages/parallel/init.lua [hunk 4]
```diff
@@ -99,75 +438,196 @@ function package:_init (options)
...
+
+         -- Set hanging indentation
+         local hangIndent = SILE.types.length("14.4pt"):absolute()
+         SILE.settings:set("current.hangAfter", 1) -- Indent subsequent lines
```
Omikhleia, line 556: Mere remark: note that hangAfter/hangIndent is touchy (on when it resets etc.) and overkill here.
For a 1-line indent, it can most of the time be avoided:
- set the whole left skip to your amount (document.lskip)
- set the current paragraph indent to the _negative_ of that amount ;)

packages/parallel/init.lua
```diff
@@ -99,75 +438,196 @@ function package:_init (options)
...
+         SILE.settings:set("current.hangIndent", hangIndent)
+
+         -- Calculate the gap after the marker
+         local markerGap = hangIndent - markerWidth
```
Omikhleia, line 560: Tricky: what to do if the markerGap is negative here (i.e. footnotes reach such a number that it's larger that your note indent (if I got it right, the hard-coded 14.4pt above... Why _this_ peculiar value by the way?)... Or even close to 0?
This is something I addressed in resilient.footnotes, basically the idea is the following (simplified)
  -  decentSpacing is some space >= interword space (usual greater by a small amount)
  - If markerGap is > decentSpacing  then do you usual stuff:
     - push marker, 
     - add kern,
     - process content
   - otherwise, we are too close or past the available size: 
      - push marker,
      - push glue with size interwordSpace
      - process content

BTW interwordSpace = `SILE.shaper:measureSpace(SILE.font.loadDefaults({}))` (respecting other adequate settings, and it has stretch and shrink)

packages/parallel/init.lua
```diff
@@ -99,75 +438,196 @@ function package:_init (options)
...
-         )
       end)
+
+      -- Add balancing glue to align frame heights.
       addBalancingGlue(maxheight)
+
+      -- Handle parskip (spacing between successive frames).
+      local parskip = SILE.settings:get("document.parskip")
+
+      if not parskip.length then
```
Omikhleia, line 501: When does it occur? document.parskip is always defined no?

packages/parallel/init.lua
```diff
@@ -99,75 +438,196 @@ function package:_init (options)
...
+         -- Use user-defined parskip value.
+         addParskipToFrames(parskip)
+      end
+   end)
+
+   self:registerCommand("smaller", function(_, content)
+      SILE.settings:temporarily(function()
+         local currentSize = SILE.settings:get("font.size")
+         SILE.settings:set("font.size", currentSize * 0.75) -- Scale down to 75%
+         SILE.settings:set("font.weight", 800)
```
Omikhleia, line 514: - "0.75em" is what you probably want without needing to perform the computation yourself.
- Weight 800 is very heavy...
- Of course textsubsuper.sile would allow not bothering here. But your are in the core distribution. Bah...

Omikhleia: My "Bah!" above = never mind. People like reinventing the wheel ever and ever ;)

packages/parallel/init.lua
```diff
@@ -63,43 +63,49 @@ end
...
    offset = offset or 0
-   if offset >= numLines then
+   if numLines <= offset then
       SU.warn("Offset is larger than the number of lines available; no dummy content will be generated.")
       return
    end
 
-   -- Fill the frame with dummy content using white-colored text to avoid visible output.
+   -- Add dummy content to fill the frame
    SILE.call("color", { color = "white" }, function()
```
Omikhleia, line 101: Not sure I understand what you are trying with this dummy content, but "white-colored" text is a dubious solution.
 - If you really want N black lines, one way would be to use zerohbox nodes instead of some text (i.e. IIRC, `SILE.typesetter:pushHorizontal(SILE.types.node.zerohbox()`). It should work, normally... But if failing report a bug and in the meantime, typeset a non-breaking space in that case (`luautf8.char(0x00A0)` or one of his friend). No need for coloring ;)
 - _Theoretically,_ the better way would likely to push a `SILE.types.node.vkern()` with the appropriate height, directly to the _output queue_. These nodes exists in the code base, but don't have a decent API and are only used a bad package (linespacing). I had some notes about them and vglues in general (there's more that meets the eyes here)... I'll look for them and open an issue. Another option in the meantime might be to hack a vglue so that its "discardable" and "explicit" properties are _both_ false. That what a vkern should be, with some extra considerations...

packages/parallel/init.lua
Omikhleia: N.B. Unless I misunderstood what you were trying here, regarding this type of footnotes, and my above discussion, I didn't invent it. It's (one of) Jan Tschichold's recommendation(s).
As long as the marker fits, it's in the available space:
image
Once the marker no longer fits, the first line shall flow as a regular line:
image

Tschichold provides other indications for different ways of formatting things (footnotes, etc.) and he is often quite opinionated on many topics. But what to expect from a great typographer, designer and writer? ;) 
Probably the best things I've read while working on my modules. Robert Bringhurst comes a bit after, his book is dull or imprecise at times, though still commendable ;) -- it doesn't say anything on that matter, anyhow.

packages/parallel/init.lua
Omikhleia: > I'm an old man, so a bold text would allow me to spot it more easily ;) Will adjust it afterward.

[textsubsuper.sile](https://github.com/Omikhleia/textsubsuper.sile) adopts weight **600** (medium) for fake footnotes (raised+scaled text), following Bringhurst in that matter (".. smaller numbers in semibold ...")

packages/parallel/init.lua
Omikhleia: >  I may be wrong there, but somehow it works for me.

Hm... `not parsikip` is still false when parskip is set to 0pt. It's always false ;)

packages/parallel/init.lua
Omikhleia: > Looked into your resilent.footnotes but did not find it,

Hehe.... The logic is entirely abstracted and buried  deep into the `style:apply:number` logic (in the resilient.styles package), 'cause every styling decision is customizable in _re·sil·ient_, and the footnote package doesn't have to care, it just applies a style where needed.

packages/parallel/init.lua
Omikhleia: Re: ...  `SILE.types.length("2.5pt"):absolute()`

Absolutizing converts any relative unit (e.g. em) to points. Not needed here, thus, you are already in pt in your length (but why an hard-coded 2.5, one wonders).

Omikhleia: Re: 
>  SILE always converts my hangIndent to pt.

'Cause it's being "absolutized" (see previous comment) at some point. (The question of SILE absolutizing things too soon or too late is for another day...).

`SILE.types.length("1em"):absolute()` is the current font size in pt. Knowing this, you shouldn't need an hard-coded 14.4 if it's supposed to be depended on the current font size. Either use the `font.size` (always converted to points as a number), or express things in 1em and absolutize() them to points.

packages/parallel/init.lua
Omikhleia: > I guess, SILE must supports that straight out of the box.

There's a 3rd-party package for that, which does have a bunch of options:
 - using natural superscripts (some fonts have very nice ones, e.g. Brill)
 - using fake superscripts (raised, smaller) = more or less like you do here, but with extra capabilities (of course fiddling with the weight, but also performing some re-scaling)

I am not sure whether bringing it to the core ("out of the box") distribution is the right time... I am not opposed to it -- I don't care, actually -- but there was sort of a meta-discussion long ago on what was expected in the core distribution, and the conclusions are dubious at best.... This being said, I do tend to think that a diglot class and a parallel package would be best addressed outside of the core distribution. But heh, it's not my call. :rabbit:

packages/parallel/init.lua
Omikhleia: > parskip.length is always nil regardless of whatever the document.parskip is set to.

'Cause the parskip is not a **length** object, it's a **vglue** (= with height/depth/width fields, which are lengths)

PR #2218, opened 2025-01-17
Sync flake with nixpkg

alerque:
Does 91ca9878 look right @doronbehar?

c.f. https://github.com/NixOS/nixpkgs/pull/373209

### DIFF
--- a/build-aux/pkg.nix
+++ b/build-aux/pkg.nix
@@ -79,19 +79,18 @@ stdenv.mkDerivation (finalAttrs: {
     sed -i -e 's/tarball-version/flake-version/' configure.ac
   '';
 
-  buildInputs =
-    [
-      finalAttrs.finalPackage.passthru.luaEnv
-      cargo-edit
-      harfbuzz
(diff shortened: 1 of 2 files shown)


build-aux/pkg.nix
```diff
@@ -79,19 +79,18 @@ stdenv.mkDerivation (finalAttrs: {
...
+    cargo-edit
+    harfbuzz
+    icu
+    fontconfig
+    libiconv
+    libxslt
+    stylua
+    taplo
+    typos
+  ];
```
doronbehar, line 93: This is one of my painful points in the new nix formatting, but I guess I'll have to live with it :).

doronbehar (approved):

build-aux/pkg.nix
alerque: My beef is it did not apply the same style to all elements, e.g. `luaPackages` and `configureFlags` did not get formatted the same way `buildInputs` and others did. :shrug:

PR #2221, opened 2025-02-04
feat(typesetters): Discourage page breaks after hyphenated lines

Omikhleia:
When typesetting a real book, sooner or later, the following occurs:

image

Most typographers would frown upon hyphenated words at the end of a page (even so more if the page ends being a right page).

Is this a bug? Well yes if SILE reportedly pretends being as good as TeX already.
Otherwise it's a "feature", implementing the missing TeX "brokenpenalty".

So here it is...

**Simplified MWE**

```
\begin[papersize=a7]{document}
\neverindent
\nofolios
\font[family=Libertinus Serif, size=12pt]
\language[main=en]

Check page break.

% Ad hoc skip to trigger the issue, with some flexibility to allow for lines to
% move around.
% Some lines of text, our sentence with a long word ("desintructionalizations")
% that we expect to be hyphenated, and some more text.
% The game is for our word to be broken, but not conflict with orphan/widow lines.
\skip[height=6.5cm plus 1cm minus 1cm]

Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam nec purus nec nunc ultricies ultricies.
Desintructionalizations may lead to revolutions in the future.

\end{document}
```

**Observed**

image

**Expected (fixed)**

image

It should be a good thing to start with.[^1]

[^1]: To be honest, there's more to be discussed on the topic...
          - Esp. with footnotes, where this could be more "acceptable" (say some) (so what value should be active with footnotes is a question for the footnote logic)
          - And... Er... on the main text content with footnotes again, due to the way they can be split or not, etc. and how the current pagebuilder computes penalties and badness.

### DIFF
--- a/packages/retrograde/init.lua
+++ b/packages/retrograde/init.lua
@@ -12,6 +12,9 @@ end
 
 -- Default settings that have gone out of fashion
 package.default_settings = {
+   ["0.15.10"] = {
+      ["typesetter.brokenpenalty"] = 0,
+   },
    ["0.15.0"] = {
       ["shaper.spaceenlargementfactor"] = 1.2,
       ["document.parindent"] = "20pt",
--- a/typesetters/base.lua
+++ b/typesetters/base.lua
@@ -89,6 +89,13 @@ function typesetter.declareSettings (_)
    })
 
    SILE.settings:declare({
+      parameter = "typesetter.brokenpenalty",
+      type = "integer",
+      default = 100,
+      help = "Penalty to be applied to broken (hyphenated) lines",
+   })
+
+   SILE.settings:declare({
       parameter = "typesetter.orphanpenalty",
       type = "integer",
       default = 3000,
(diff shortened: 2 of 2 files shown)


Omikhleia:
NOTE.
This **will** obviously affect existing books. If needed, one can set `typesetter.brokenpenalty` to 0 and the output will then be the same as before. I'm not sure we should put that in "retrograde" or not.

alerque (changes requested):
> NOTE. This **will** obviously affect existing books. If needed, one can set `typesetter.brokenpenalty` to 0 and the output will then be the same as before. I'm not sure we should put that in "retrograde" or not.

Yes definitely, any time settings or behaviors change default values and the implementation is trivial enough that we expect the old behavior to be achievable it should be added to `retrograde` at the relevant version.

alerque (approved):

alerque:
I'm not yet sure why CI has blown up, but that's probably on me to fix.

PR #2224, opened 2025-02-05
fix(math): Script 'math' is needed for the ssty feature to work properly

### LINKED ISSUE #2223: Primes with ssty not working with STIX Math Two
"Math layout engine are expected to set **script** to **math** ... May be that is what SILE is missing."[^1]

See https://github.com/alerque/polytype/pull/10#issuecomment-2636960991 for full context

Thanks @khaledhosny for bringing that explanation.

[^1]: I'll need to amend and update [_SILE and the Hydra of Maths. Showcase & Critical Assessment 2024_](https://drive.google.com/file/d/1kGHIHUU9dsAsMBrGYotPUO0XyDLwD-CF), Part II §2.1 "Primes gone wild" with a few additional paragraphs full of extra ramblings. 😁

Omikhleia:
So that primes etc. use the correct glyphs.
Libertinus Math v7.051 does not require it, but STIX Two Math 2.13 b171 and TeX Gyre Termes Math v1.543 do, for instance.

---

Closes #2223 

Yay!

I long wanted to test TeX Gyre Termes Math, so gave it shot. I had the same problem as STIX Two Math, with ssty having no effect. Both were indeed fixed by enforcing the script to "math".
We could wonder why Libertinus Math respected ssty without the "math" script while other didn't, but heh. It doesn't complain either if the script is set, so I am gonna pass on that one.

### DIFF
--- a/packages/math/base-elements.lua
+++ b/packages/math/base-elements.lua
@@ -214,6 +214,11 @@ function elements.mbox:_init ()
       size = SILE.settings:get("math.font.size"),
       style = SILE.settings:get("math.font.style"),
       weight = SILE.settings:get("math.font.weight"),
+      -- https://learn.microsoft.com/en-us/typography/opentype/spec/math#opentype-layout-tags-used-with-the-math-table
+      --   "Script tag to be used for features in math layout.
+      --   The only language system supported with this tag is the default language system."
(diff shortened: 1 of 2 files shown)


Omikhleia:
Just for the show effect (double prime on the lowest denominator in the formula, "Laplace's method")

image

alerque (approved):

Issue #1008, opened 2020-08-03
Expand characters allowed in commands to full‌ XML spec


alerque:
See #721 and particularly [this comment](https://github.com/sile-typesetter/sile/issues/721#issuecomment-666371204). Currently we only allow a relatively limited subset –nearly ANSI– of characters in command names. This does theoretically limit what valid XML we can render. Practically speaking I don't know how much XML in the wild uses full spectrum UTF-8 in tag names, but...

Since this all the other points discussed in that issue are resolved and I have no plans of implementing this right away, I'm moving it to it's own issue.

Omikhleia:
And what the reciprocal? Currently we allow `:` in commands (e.g. `verbatim:font`, `pullquote:font` &c).
But in XML, it would conflict with namespaces.
(Well, we don't _really_ support XML namespaces, but we ought to.)

PR #2230, opened 2025-02-16
More CSL features: multiple citations and paragraph format

### LINKED ISSUE #2023: Chicago-style bibliography ignores the volume and subtitle
SILE 0.14.17 or 0.15-develop

Example:

```
@book{DragonDeBrume1,
  title     = "Tolkien, le façonnement d'un monde",
  subtitle  = "Botanique &amp; Astronomie",
  year      = 2011,
  volume    = 1,
  number    = 2,
}
```

Obtained reference: _Tolkien, le façonnement d'un monde._ Le Dragon de Brume, 2011.

Expected reference: _Tolkien, le façonnement d'un monde._ Vol. 2, _Botanique & Astronomie._ Le Dragon de Brume, 2011.

From template:
Author Surname, First Name. _Book Title: Subtitle of Book._ Vol. #, _Title of Volume._ Edited by First Name Surname. Place of Publication: Publisher, Year.

See also https://libguides.tru.ca/chicago/books-multivolume

N.B. Additional subtleties appear here, whether one should use title/subtitle or (rather) maintitle/title -- Our support is currently fairly limited. But the important thing here is to get something correctly identifying the work.
I'd be equally satisfied with _Tolkien, le façonnement d'un monde: Botanique & Astronomie._ Vol. 2._ Le Dragon de Brume, 2011 (which is what we'd likely get with title/subtitle, the devil is in the details...)

### LINKED ISSUE #2196: Citing multiple bibliography entries
SILE's 0.15.7 CSL support mimicked the `csl:cite` command on the legacy `cite` command.
Both thus accept only one key (as content or `key` parameter). To quote our documentation:
> To produce an inline citation, call `\csl:cite{⟨key⟩}`, which will typeset something like “(Jones 1982)”. If you want to cite a particular page number, use `\csl:cite[page=22]{⟨key⟩}`. Other “locator” options are available (article, chapter, column, line, note, paragraph, section, volume, etc.) – see the CSL documentation for details. Some frequent abbreviations are also supported (art, chap, col, fig…).

However, the need to cite multiple entries, possibly each with its own locators, exist.
These would need to be in a single call to the CSL engine, as:
 - The style would add whatever delimiters it uses around the whole citation.
 - The style could even reorganize them. (We don't implement entry collapsing yet, but eventually we might get there)

In (Pandoc-flavored) Mardown, for instance, one may have:[^1]
 - `See [@doe;@smith]` = rendered as, e.g. "See (Doe, 2022; Smith, 2024)"
 - Or with locators, `See [@doe page 1; @smith chap 4]` = "See (Doe, 2022, p. 1; Smith, 2024, chap. 4)"
 - Or eventually with also collapsing, `See [@doe2023;@doe2024]` = "See (Doe, 2023, 2024)"

I refrained proposing something in our initial implementation, because as noted above the legacy cite command compatibility was the first thing to get first of all, but also because the API in SIL TeX-like and/or SIL XML might not be straightforward and "natural"...

But it's certainly time to discuss it.

Would this be the "right" way?

```
\begin{citation}
  \cite[key=doe, ...]
  \cite[key=smith, ....]
`\end{citation}
```
= `\citation{\cite[key=doe, ...]\cite[key=smith, ...]`
= `<citation><cite key="doe"/><cite key="smith"/></citation>`

What do you think? The key issue here is that it introduces a "structured" top-level tag, where other content should likely be rejected (and all spaces ignored) -- We do already have something similar for lists, but the pattern is not that common in SILE's code base.
Also note that in such constructs, 
 - the `\cite` node might even just be an AST command node without underlying command provided.
 - Or the `\cite` command may exist "outside" of a citation, and become a mere shortcut to it (i.e. `\cite[...]` at higher level is just processed as `\citation{\cite[...]}` would be.

Let's decide for a satisfying design first (the coding is easy afterwards).

[^1]: N.B. (I am not discussing here "author-in-text" `@doe` and "suppress-author" `[-@doe]` which would require a separate topic on their own. AFAIK, it's not really covered in CSL 1.0.2 and I don't really understand how this is supposed to "magically" work if ones follows those specifications. Clarifications would be welcome (in an appropriate issue?), if any knowledgeable reader passes by.

Omikhleia:
~~Currently based on #2228 - to rebase accordingly~~ REBASED

**Objectives**

 - [x] Properly grouped multiple citations
 - [x] Handling of the CSL indentation rules (`hanging-indent`, `second-field-align`) in order to support more styles effectively (stretch goal: with IEEE as main target).
 - [x] I think it will then also be the right time to switch to `csl` mode by default, while still keeping the legacy stuff for now.

**Changes**

 - Some code refactors
 - Some fixes and improvements
 - `\nocite` command for marking an entry as cited with out actually citing it (= force its inclusion in the list of cited works, as with LaTeX)
 - `\cites` command for grouped (multiple) citations

**Other notes**

On the way, I tested a few more styles, now guaranteed not to crash :rofl: and produce something at least decent :rabbit: 
 - (EDIT) (obviously still Chicago author-date)
 - IEEE (for the indentation rules) w/ and w/o URL.
 - ISO690 (author-date) and ISO690 (numeric), both English and French styles (I was told this one is often used in French papers, so heh)... And there was a bug in our locale logic (fix: Apply default locale on locale overrides without xml:lang). Doh.
 - APA, as several French styles are APA-derivatves (... and I thought APA was for Psychology-related journals, but apparently is used in many other contexts, incl. Humanities).
 - ACS
 - (EDIT) MLA
 - (EDIT) _Acta philosophica_ (see discussion below, after further fixes)
 - (EDIT) ABNT (see discussion below)
 - (EDIT) Harvard / Cite them right (see discussion below)

Closes #2196 

Closes #2023

### DIFF
--- a/packages/bibtex/csl/engine.lua
+++ b/packages/bibtex/csl/engine.lua
@@ -831,6 +831,28 @@ function CslEngine:_name_et_al (options)
    return t
 end
 
+local function initialize (options, entry)
+   if SU.boolean(options.initialize, true) and options['initialize-with'] then
+      -- FIXME TODO Quick and dirty:
+      -- Major styles abbreviate the given name with ". " as initialize-with
+      -- and don't set initialize-with-hyphen to false.
+      -- So here we just use the short name already derived by our BibTeX parser.
+      -- ... But this is not general, obviously.
+      return entry['given-short'] .. options['initialize-with']:gsub("%s*$", "")
+   end
+   return entry.given
+end
+
+function CslEngine:_format_a_name (options, name)
+   if not options then
+      return name
+   end
+   return self:_render_formatting(
+      self:_render_textCase(name, options),
+      options
+   )
+end
+
 function CslEngine:_a_name (options, content, entry)
    if entry.literal then -- pass through literal names
       return entry.literal
@@ -875,36 +897,71 @@ function CslEngine:_a_name (options, content, entry)
       return table.concat(name, " ")
    end
 
+   -- Name-parts for formatting:
+   -- If set to “given”, formatting and text-case attributes on cs:name-part affect
+   -- the “given” and “dropping-particle” name-parts.
+   -- FIXME TODO Affixes surround the “given” name-part, enclosing any demoted name
+   -- particles for inverted names.
+   -- If set to “family”, formatting and text-case attributes affect the “family” and
+   -- “non-dropping-particle” name-parts.
+   -- FIXME TODO Affixes surround the “family” name-part, enclosing any preceding name
+   -- particles, as well as the “suffix” name-part for non-inverted names.
+   -- N.B. Very few styles use affixes, hence the FIXME TODO, not implemented yet.
+   if not options["__name-parts"] then
+      -- First the options from the cs:name node
+      local formattingAndCaseOptions = {
+         ["text-case"] = options["text-case"],
+         ["font-style"] = options["font-style"],
+         ["font-variant"] = options["font-variant"],
+         ["font-weight"] = options["font-weight"],
+         ["text-decoration"] = options["text-decoration"],
+         ["vertical-align"] = options["vertical-align"],
+      }
+      local namePartOptions = {
+         given = formattingAndCaseOptions,
+         family = formattingAndCaseOptions,
+      }
+      for _, namep in ipairs(content) do
+         -- Override with the options for the cs:name-part explicit nodes, if any
+         if namep.command == "cs:name-part" then
+            namePartOptions[namep.options.name] = pl.tablex.merge(
+               namePartOptions[namep.options.name],
+               namep.options,
+               true
+            )
+         end
+      end
+      options["__name-parts"] = namePartOptions -- memoize
+   end
+   local namepartOptions = options["__name-parts"]
+
    local form = options.form
-   local nameAsSortOrder = options["name-as-sort-order"] or "first"
-
-   -- TODO FIXME: content can consists in name-part elements for formatting, text-case, affixes
-   -- Chigaco style does not seem to use them, so we keep it "simple" for now.
-
    if form == "short" then
       -- Order is: [NDP] Family, e.g. van Gogh
       if entry["non-dropping-particle"] then
          return table.concat({
-            entry["non-dropping-particle"],
-            entry.family,
+            self:_format_a_name(namepartOptions.family, entry["non-dropping-particle"]),
+            self:_format_a_name(namepartOptions.family, entry.family),
          }, " ")
       end
-      return entry.family
+      return self:_format_a_name(namepartOptions.family, entry.family)
    end
 
-   if nameAsSortOrder ~= "all" and not self.firstName then
+   local nameAsSortOrder = options["name-as-sort-order"]
+   local familyFirst = nameAsSortOrder == "all" or (nameAsSortOrder == "first" and self.firstName) or false
+   if not familyFirst then
       -- Order is: [Given] [DP] [NDP] Family [Suffix] e.g. Vincent van Gogh III
       local t = {}
       if entry.given then
-         table.insert(t, entry.given)
+         table.insert(t, self:_format_a_name(namepartOptions.given, initialize(options, entry)))
       end
       if entry["dropping-particle"] then
-         table.insert(t, entry["dropping-particle"])
+         table.insert(t, self:_format_a_name(namepartOptions.given, entry["dropping-particle"]))
       end
       if entry["non-dropping-particle"] then
-         table.insert(t, entry["non-dropping-particle"])
+         table.insert(t, self:_format_a_name(namepartOptions.family, entry["non-dropping-particle"]))
       end
-      table.insert(t, entry.family)
+      table.insert(t, self:_format_a_name(namepartOptions.family, entry.family))
       if entry.suffix then
          table.insert(t, entry.suffix)
       end
@@ -916,24 +973,24 @@ function CslEngine:_a_name (options, content, entry)
       -- Order is: Family, [Given] [DP] [NDP], [Suffix] e.g. Gogh, Vincent van, III
       local mid = {}
       if entry.given then
-         table.insert(mid, entry.given)
+         table.insert(mid, self:_format_a_name(namepartOptions.given, initialize(options, entry)))
       end
       if entry["dropping-particle"] then
-         table.insert(mid, entry["dropping-particle"])
+         table.insert(mid, self:_format_a_name(namepartOptions.given, entry["dropping-particle"]))
       end
       if entry["non-dropping-particle"] then
-         table.insert(mid, entry["non-dropping-particle"])
+         table.insert(mid, self:_format_a_name(namepartOptions.family, entry["non-dropping-particle"]))
       end
       local midname = table.concat(mid, " ")
       if #midname > 0 then
          return table.concat({
-            entry.family,
+            self:_format_a_name(namepartOptions.family, entry.family),
             midname,
             entry.suffix, -- may be nil
          }, sep)
       end
       return table.concat({
-         entry.family,
+         self:_format_a_name(namepartOptions.family, entry.family),
          entry.suffix, -- may be nil
       }, sep)
    end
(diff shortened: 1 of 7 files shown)


Omikhleia:
Stretch goal achieved:

image

Omikhleia:
Bonus point achieved:[^1]

image

[^1]: Current implementation does not honor `display` (left-margin, right-inline etc.) but at least doesn't crash and looks "decent" :joy_cat:

Omikhleia:
> One small thing (highlighted) not investigated yet.

Asouqi's Online Bibtex Converter gets it right:
image

Interesting, "if" blocks should probably inherit the upper group delimiter:
```
  <macro name="issue">
    <group delimiter=", ">
      <text variable="volume" prefix="Vol.&#160;"/>
      <choose>
        <if variable="volume">
          <text variable="issue" prefix="no.&#160;"/>
          <text variable="page" prefix="p.&#160;"/>
        </if>
        <else-if variable="issue">
          <text variable="issue" prefix="No.&#160;"/>
          <text variable="page" prefix="p.&#160;"/>
        </else-if>
        <else>
          <text variable="page" prefix="P.&#160;"/>
        </else>
      </choose>
    </group>
  </macro>
```

I'll see if it can get on this row of changes, or open a dedicated issue.

Omikhleia:
- ISO 690 is now good (also with uppercase authors)
- IEEE is better (correct name ordering, and abbreviation of given names)
  For the record the following issue is not ours:
  image
  It was introduced in the CSL style between late 2021 (what Asouqi's online translator uses for its predefined toggle) and 2024 (confirmed by using the latest CSL in that same tool; I haven't tried to bissect the commit). This is what happens with too complex standards and no proper regression tests...

Omikhleia:
- APA looks better too (consistent superscripts across entries).

I think I am done with the things I wanted to see here.

Omikhleia:
Juste rebased, now that #2228 got merged.

packages/bibtex/csl/engine.lua
```diff
@@ -434,7 +436,7 @@ function CslEngine:_layout (options, content, entries)
       -- citations, affixes and formatting apply on the whole layout.
       -- Affixes are arround the delimited list, e.g. "(Smith, 2000; Jones, 2001)"
       -- Rendering is done after, so vertical-align, etc. apply to the whole list,
-      -- e.g. <textsuperscript>1,2</textsuperscript>
+      -- e.g. <bibSuperScript>1, 2</bibSuperScript>
```
alerque, line 439: I don't think our command naming scheme is robustly case sensitive. If you are trying to use a package specific naming scheme we have a precedent for using a `:` separator, as in `bibtex:superscript`.

alerque (changes requested):

packages/bibtex/csl/engine.lua
Omikhleia: The ":" separator is problematic if we ever go for XML namespaces at one point. A dash, maybe?

packages/bibtex/csl/engine.lua
```diff
@@ -1,6 +1,6 @@
 --- A rendering engine for CSL 1.0.2
 --
--- @copyright License: MIT (c) 2024 Omikhleia
+-- @copyright License: MIT (c) 2024, 2025 Omikhleia
```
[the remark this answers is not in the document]
Omikhleia: Agreed, can be removed.

packages/bibtex/csl/engine.lua
Omikhleia: See also https://github.com/sile-typesetter/sile/issues/1008#issuecomment-2654762731

packages/bibtex/csl/engine.lua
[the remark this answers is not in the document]
alerque: No particular need to for this PR, there are a few others I noticed around already. This was just a heads up that I'd be normalizing them all at some point.

packages/bibtex/csl/engine.lua
Omikhleia: XML tags are case-sensitive (HTML tags aren't not, but that's a different topic/history), so we'd probably be safer considering commands to be case sensitive.
XML namespaces allow for only one level (_ns:tag_ with only one colon). Anyway we are ages from reaching that point...

But another problem here with the colon is that the CSL engine's logic on spaces and punctuation (just doing string replacements with a bunch of gsub's in the post-processing stage) is a bit dumb and the colon, quotes etc. are going to be nasty in command names - the reason why I stick with lame tags (without arguments etc.) and delegated them to SILE. So if we rename `<bibXxx/>` to `<bibtex:xxx>` we likely break something (at least in French), same if we add arguments (e.g. `<bibRule>3</bibRule>` to  `<bibRule len="3"/>` would likely break (for all languages).

Omikhleia:
Moving back to draft: checking an additional CSL file (acta philosophica) = missing case in name-parts for formatting, should be an easy fix.

Omikhleia:
> Moving back to draft: checking an additional CSL file (_acta philosophica_) = missing case in name-parts for formatting, should be an easy fix.

image

As easy as 3 commits...
 - Formatting/case options can be on `cs:name` and not only on (optional) `cs:name-part` elements.  Now we get the small caps here.
 - An overlooked boolean conversion on the plural flag (... which would have lead to "vols. 1" here instead of "vol. 1"), doh.
 - Some additional subtleties with empty macro suppressions (previously added for ABNT etc. in #2228 but causing havoc here in another group/macro. The nesting rules are not that clear when reading the specs...

The bibliography output looks as expected (from other renderers) above... But there's still something weird in the citation output. Investigating it --- it's also one of the rare styles I tested with a "note" citation format (vs. "in-text"), so it's a good extra check...

packages/bibtex/csl/engine.lua
```diff
@@ -449,7 +463,10 @@ function CslEngine:_layout (options, content, entries)
    -- quite logically as we force a paragraph break between entries.
    for _, entry in ipairs(entries) do
       self:_prerender()
-      local elem = self:_render_children(content, entry)
+      local elem = self:_render_children(content, entry, {
+         secondFieldAlign = self.inheritable.bibliography["second-field-align"] and true or false,
+         delimiter = options.delimiter
```
Omikhleia, line 468: Useless: There shouldn't be a delimiter on the bibliography layout as per the specs.

packages/bibtex/csl/engine.lua
```diff
@@ -424,7 +436,9 @@ function CslEngine:_layout (options, content, entries)
    if self.mode == "citation" then
       for _, entry in ipairs(entries) do
          self:_prerender()
-         local elem = self:_render_children(content, entry)
+         local elem = self:_render_children(content, entry, {
+            delimiter = options.delimiter
```
Omikhleia, line 440: Wrong! There delimiter on the citation layout applies between entries (taken care of just below) and shouldn't be passed as context for an individual entry (=> breaks e.g. with acta-philosophica that starts with an immediate `<choose>`)

Omikhleia: Fixed

packages/bibtex/csl/engine.lua
Omikhleia: Fixed

Omikhleia:
> But there's still something weird in the citation output. Investigating it (...)

Doh. Ugly regression fixed (with fixup + rebase).
Recap comparing before/after big PDFs ... my eyes hurt now :rofl:
 - Chicago author-date :heavy_check_mark: 
 - IEEE :heavy_check_mark: (some problems in the CSL style confirmed with other renderers, not our fault)
 - ISO690 (author-date) and ISO690 (numeric) :heavy_check_mark: 
 - MLA :heavy_check_mark: 
 - ACS :heavy_check_mark: 
 - APA :heavy_check_mark: 
 - Dragon de Brume MLA (my modified MLA style for my books) :heavy_check_mark: (and found an issue in my CSL style in one of my changes)
 - _Acta philosophica_ :heavy_check_mark: 
 - ABNT :heavy_check_mark:  (some problems in the CSL style confirmed with other renderers, not our fault)

alerque (approved):

PR #2234, opened 2025-03-07
fix(typesetters): Special punctuation spaces need better italic corre…

### LINKED ISSUE #2233: Slight improvement to italic correction in French
The automated italic correction heuristics, which we added in 0.15.0, works quite well as far as I can tell, and our current implementation might even possibly be better than other existing solutions, as shown on [polytype.dev](https://polytype.dev/italic-correction/). With English text, for instance, I find it quite neat (and I enabled it in most of my books).

However, it includes a "special" case for French, which uses extra non-breaking spaces before "high punctuation" (= different spacing rules before colon, semi-colon, exclamation/question marks and quotations marks).

Without the special case (let's call this the "no special rule" for this discussion), italic correction would have occurred in these situations, moving the said punctuation a bit further. At the time, I couldn't find a definitive/authoritative source on that rather obscure topic (and an edge case), but visually it wouldn't look very nice... and doesn't seem to be what we find in books (although it's always hard to know what they did, depending on the font, etc.)

So the "current" implementation cancels the italic correction there. The assumption was naive when introduced at the time: since we already have those special punctuation spaces, we shouldn't need the italic correction there most of the time, should we?.

And in most cases, it looks look... But depending on fonts, there are still cases where slight overlaps may occur in fr-FR. And it is worse for variants that would change, tweak or even cancel (some of the) special punctuation spaces, e.g. playing with settings `languages.fr.thinspace` or `languages.fr.guillspace`...

A "better" solution here would likely be to take into account that space in the computation, and check if it needs some partial compensation to avoid overlaps.
It makes fr-FR a bit better IMHO, and accounts for variants where tweaking the punctuation spaces (fr-CH or fr-CA might want to do that for various historical reasons, whether they follow the German or English usage, etc.) -- which would likely consider the current behavior to be a bug actually!

Illustration (animated GIF).

Image

Going the "better" way would obviously change the spacing a bit, and could thus result in different output (regarding line breaks and paragraphing) in some cases... Yet, my view here would be that is acceptable. Automated italic correction remains an "experimental" feature (behind a setting), and even if my own previous French books _could_ be impacted, I wouldn't mind here and would accept the _possibly_ slightly different output -- I would view it as a bug fix, and not really a breaking change.

Omikhleia:
…ction

The heuristics previously considered that special punctuation spaces (as used in French) would cancel italic automated correction. This could however still lead to character overlaps even with decent fonts, so a slightly better logic is needed here: by default, ensure minimal italic correction if the space is not sufficient, for partial compensation.
A new setting is also added to fully apply italic correction.

Closes #2233

### DIFF
--- a/tests/feat-italic-correction.expected
+++ b/tests/feat-italic-correction.expected
@@ -167,28 +167,28 @@ T	12 w=4.7534	())
 Mx 	14.8819
 My 	122.6951
 T	14 w=6.1523	(+)
-Mx 	25.0138
+Mx 	25.0137
 T	169 w=7.2803	(«)
-Mx 	35.4699
+Mx 	35.4698
 Set font 	Gentium Plus;15;400;Italic;normal;;;LTR
 T	739 88 747 w=23.2910	(fluff)
-Mx 	61.9367
+Mx 	61.9366
 Set font 	Gentium Plus;15;400;;normal;;;LTR
 T	170 w=7.2803	(»)
-Mx 	73.1709
+Mx 	73.1963
 Set font 	Gentium Plus;15;400;Italic;normal;;;LTR
 T	73 85 72 81 70 75 w=35.9619	(french)
-Mx 	112.7506
+Mx 	112.7759
 T	83 85 82 82 73 w=29.1431	(proof)
-Mx 	143.8756
+Mx 	144.3722
 Set font 	Gentium Plus;15;400;;normal;;;LTR
 T	4 w=4.0796	(!)
-Mx 	151.9347
+Mx 	152.4313
 T	11 w=4.7534	(()
-Mx 	157.1638
+Mx 	157.6604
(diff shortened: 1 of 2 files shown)


Omikhleia:
Test `tests/feat-italic-correction.sil` obviously fails, but the differences extend beyond the French cases only, and affects other languages as well: When computing the italic correction "delta", I had forgotten that a character depth or height can be negative... So the computation could result in a negative correction, rather than always inserting a positive one as was intended by design...

It seems to me it is better now, but I am reviewing the diff carefully and performing some extra tests.

Omikhleia:
> It seems to me it is better now, but I am reviewing the diff carefully and performing some extra tests.

Looks better in the latest iteration. (The differences are very slight, but significant)

alerque (approved):

PR #2237, opened 2025-03-15
feat(core): Support ex-height and cap-height font adjustment

### LINKED ISSUE #2235: Bad use of font size hard-coded adjustments offsets
In the code base, there are at least two instance of font size tuning of the following kind;

https://github.com/sile-typesetter/sile/blob/2544c7da62806f1928fdb60abb3b245d87b18385/classes/plain.lua#L265

https://github.com/sile-typesetter/sile/blob/2544c7da62806f1928fdb60abb3b245d87b18385/packages/verbatim/init.lua#L9

This is a "really bad"™ implementation - absolute "hard" correction in points is not the way to go depending on the initial font size; and the pseudo-"adjustments" is plain wrong with _any_ other default fonts at _any_ other sizes than the default 10pt...

On a first thought, everything ought to be in a **font relative unit** here at least -- this would also solve discussion #2182 ...

... But, but... why do we have such "corrections" by the way? Because one font (here Hack for instance) doesn't pair well with another font (say main body, e.g. in Gentium)...

So the adjustments here is _with respect to one specific font vs. another specific font_... It cannot be hard-coded, and it cannot either really be just in a font-relative unit, if we do care about **proper typography**. It's a matter of "eye", i.e. in these cases matching "ex". CSS there would rely on "font-size-adjust"... So we'd need something similar (e.g. _a least_ `\font[family=xxxx, adjust=ex-height]` see [this reference](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size-adjust#font-metric) for other possible metrics...).

Omikhleia:
Basic proposal for #2235 as used in the example shown there, i.e.

```
normal: Lorem \font[family=...]{Ipsum}.

x-adjusted: Lorem \font[family=..., adjust=ex-height]{Ipsum}

caps-adjusted: Lorem \font[family=..., adjust=cap-height]{Ipsum}
```

Here the `adjust` option can be either "ex-height" or "cap-height", optionally preceded by a ratio (1 assumed otherwise)

Still to do:

 - [ ] Confirm the syntax for this `adjust` option suits everybody interested
 
   In the original ticket, I wrote:
   > Whether we allow any unit there (...) or restrict it to sensible expressions

   The proposal here follows the second option (more or less "à la CSS") so other expressions might be considered eventually, without needing to introduce weird units.

 - [ ] Confirm whether we'd want or not to use the OS/2 table when the metrics are available[^1]
 - [x] Add a fix to use the new font adjustment in the hard-coded manual font corrections mentioned in the original issue. Probably impacts some existing tests...
    - `\code` from plain class
    - (default) `\verbatim:font` from "verbatim" package
    - (default) `\urlstyle` from "url" package (as it relies on `\code`)
 - [x] User manual update to document that new option
 - [x] User manual update for its own customization of the `\code` command...

Closes #2235 

[^1]: (Comment) I can add the impacts to the open-type parser if requested. We may want to keep both possibilities, as I don't know how reliable are font designers there, so this could be for instance two more adjust metrics, e.g. "ex-height-from-font" and "cap-height-from-font" (defaulting to the "ex-height" (x) and "cap-height" (H) measurements when not available).


core/font.lua
```diff
@@ -4,10 +4,58 @@ local icu = require("justenoughicu")
...
+   SU.error("Unknown font adjust metric " .. metric)
+end
+
+local function adjustedFontSize(options)
+   local adjust = options.adjust
+   local parsed = adjustment:match(adjust)
+   if not parsed then
+      SU.error("Couldn't parse font adjust value " .. adjust)
+   end
+   local baseOpts = pl.tablex.copy(options) -- shallow copy
```
Omikhleia, line 37: If need be, add a comment for the rationale for the shallow copy vs. in-place modification:
We shouldn't alter the original AST, as these things may end-up been propagated elsewhere (e.g. to running headers etc.).

Omikhleia: Comment added.

alerque (approved):

PR #2243, opened 2025-03-20
Clean up ICU shim

### LINKED ISSUE #359: unnecessary mallocs in justenoughicu.c
See https://github.com/simoncozens/sile/pull/358#discussion_r74957358

basically, you're using "pure preflight mode", which does:

* call `func()` with a 0-length output buffer
* work is done, and the output length needed is returned
* `malloc()` new buffer on the heap
* call `func()` again with a real buffer
* use buffer 
* `free()` buffer

Instead, you can allocate an output buffer on the stack if you think the text might fit in it, and then in the best case you only call `func()` once, and don't malloc ever.  Performance, heap use, threading wins here.

alerque:
Long term this shim might go away, but first I have to figure out what it is doing and make sure it is doing it right so we even have something to compare against.

Closes #359

### DIFF
--- a/justenough/justenoughicu.c
+++ b/justenough/justenoughicu.c
@@ -249,19 +249,21 @@ int je_icu_bidi_runs(lua_State *L) {
     UBiDiDirection dir = ubidi_getVisualRun(bidi, i, &start, &length);
     lua_newtable(L);
     // Convert back to UTF8...
-    int32_t l3 = 0;
-    char* possibleOutbuf = malloc(4*length);
-    if(!possibleOutbuf) {
+    // Each UTF-16 code unit can expand to at most 4 UTF-8 bytes
+    const int buflen = (length * 4) + 1;
+    char* utf8buf = malloc(buflen);
+    if(!utf8buf) {
       return luaL_error(L, "Couldn't malloc");
     }
-    u_strToUTF8(possibleOutbuf, 4 * length, &l3, input_as_uchar+start, length, &err);
+    int32_t utf8len;
+    u_strToUTF8(utf8buf, buflen, &utf8len, input_as_uchar+start, length, &err);
     if (!U_SUCCESS(err)) {
-      free(possibleOutbuf);
-      return luaL_error(L, "Bidi run too big? %s", u_errorName(err));
+      free(utf8buf);
+      return luaL_error(L, "UTF-8 conversion failed: %s", u_errorName(err));
     }
     lua_pushstring(L, "run");
-    lua_pushstring(L, possibleOutbuf);
-    free(possibleOutbuf);
+    lua_pushstring(L, utf8buf);
+    free(utf8buf);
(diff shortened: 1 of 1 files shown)


justenough/justenoughicu.c
```diff
@@ -249,19 +249,20 @@ int je_icu_bidi_runs(lua_State *L) {
     UBiDiDirection dir = ubidi_getVisualRun(bidi, i, &start, &length);
     lua_newtable(L);
     // Convert back to UTF8...
-    int32_t l3 = 0;
-    char* possibleOutbuf = malloc(4*length);
-    if(!possibleOutbuf) {
+    // Each UTF-16 code unit can expand to at most 3 UTF-8 bytes
+    char* utf8buf = malloc(length * 3 + 1);
```
srl295, line 253:
```suggestion
    // Each UTF-16 code unit can expand to at most 4 UTF-8 bytes
    const auto buflen = (length * 4) + 1;
    char* utf8buf = malloc(buflen);
```

justenough/justenoughicu.c
```diff
@@ -249,19 +249,20 @@ int je_icu_bidi_runs(lua_State *L) {
...
+    if(!utf8buf) {
       return luaL_error(L, "Couldn't malloc");
     }
-    u_strToUTF8(possibleOutbuf, 4 * length, &l3, input_as_uchar+start, length, &err);
+    int32_t utf8len;
+    u_strToUTF8(utf8buf, length * 3 + 1, &utf8len, input_as_uchar+start, length, &err);
```
srl295, line 258:
```suggestion
    u_strToUTF8(utf8buf, buflen, &utf8len, input_as_uchar+start, length, &err);
```

srl295 (commented):
3-byte UTF-8 excludes non-BMP chars! (> U+FFFF )

alerque:
Thanks for the obviously correct byte count @srl295.

What isn't obvious to me is whether the original issue your reported has been properly and completely fixed. The cheese has just moved too much for my lack of C knowledge. It does look like there is one less `malloc()` in that look and the writes to the Lua side make more sense to me, but still...

The original you reported the issue on would have been from just before the v0.9.4 release: 
https://github.com/sile-typesetter/sile/blob/v0.9.4/src/justenoughicu.c#L74

srl295:
> Thanks for the obviously correct byte count @srl295.

+1

> What isn't obvious to me is whether the original issue your reported has been properly and completely fixed. The cheese has just moved too much for my lack of C knowledge. It does look like there is one less `malloc()` in that look and the writes to the Lua side make more sense to me, but still...

I replied on that issue that yes it does seem fixed.

> The original you reported the issue on would have been from just before the v0.9.4 release: https://github.com/sile-typesetter/sile/blob/v0.9.4/src/justenoughicu.c#L74
> 
> If it is done it would be nice to _know_ and get the issue closed with this final tweak.

PR #2247, opened 2025-03-25
Remove stdlib shims and normalize typesetter module layout a bit

alerque:
Just cleaning up cruft I bumped into...

### DIFF
--- a/classes/jplain.lua
+++ b/classes/jplain.lua
@@ -9,7 +9,6 @@ class._name = "jplain"
 
 function class:_init (options)
    tplain._init(self, options)
-   SILE.languageSupport.loadLanguage("ja")
    SILE.settings:set("document.language", "ja", true)
    SILE.settings:set("font.family", "Noto Sans CJK JP", true)
 end
--- a/core/utilities/init.lua
+++ b/core/utilities/init.lua
@@ -860,4 +860,11 @@ function utilities.stripContentPos (content)
    return SU.ast.stripContentPos(content)
 end
 
+function utilities._avoid_base_class_use (obj)
+   if type(obj) == "table" and obj._name and obj._name ~= "base" then
+      local type_group = "SILE." .. obj.type .. "s"
+      SU.deprecated(type_group .. ".base", type_group .. ".default", "0.15.11", "0.16.0")
+   end
+end
+
 return utilities
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -13,6 +13,7 @@ packages/pandoc*	@alerque
 packages/hanmenkyoshi*	@sile-typesetter/lang-cjk
 packages/ruby*	@sile-typesetter/lang-cjk
 packages/tate*	@sile-typesetter/lang-cjk
+typesetters/*tate*	@sile-typesetter/lang-cjk
 
 # Distro Packaging
 *.nix	@alerque @doronbehar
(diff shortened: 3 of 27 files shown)


classes/jplain.lua
```diff
@@ -9,7 +9,6 @@ class._name = "jplain"
 
 function class:_init (options)
    tplain._init(self, options)
-   SILE.languageSupport.loadLanguage("ja")
    SILE.settings:set("document.language", "ja", true)
```
Omikhleia, line 12: Just a question here. Should we rather always use `SILE.call("language"..."` rather than using the low-level setting? The difference is not that clear. Formally we use a bunch of things indeed (languageSupport, setting), but why not the high level command directly?

Omikhleia: (While at it, same remark for the next line, what behavior do we expect being different from setting the font or using a call to `\font`?  That's also a matter of showing what we expect developers to use in their own classes/packages, i.e. which paradigms are expected)

alerque: The difference is not clear because it's an absolute scramble. At the moment they are both effectively equivalent. We'll have a clearer answer to this after we sort out the whole language support module layout. For now about the only comment I have is that manually loading the language support module before using it in a setting or font/language command call should *not* be necessary any more.

core/utilities/init.lua
```diff
@@ -860,4 +860,11 @@ function utilities.stripContentPos (content)
    return SU.ast.stripContentPos(content)
 end
 
+function utilities._avoid_base_class_use (obj)
```
Omikhleia, line 863: Trying to see the impacts: Does this mean that I'll have a deprecation here when this lands?
https://github.com/Omikhleia/ptable.sile/blob/33e26b6a8a9e00f4018d14e5ef1b91dacbae0e5a/packages/parbox/init.lua#L59

If so, shouldn't we rather target 0.16 rather than 0.15.11?
On this occasion, I started working on a few language related things - e.g. right/left min handling - which might be for a 0.16 rather than a 0.15.x = maybe we should discuss our plans here (e.g. would we need a dedicated branch for these things, as we did initially for 0.15.0? What do you thing is the best path?)

alerque: I think the answer is that targeting 0.15.11 for warnings is fine, but it might be possible to extend the breaking point to 0.17 so we can have a full release cycle. Unfortunately I couldn't *prove* that yet because the typesetter init sequence is still broken in other ways. I figured that unless I figured out the other breakage, being more aggressive about waning was better because it will bypass the other mess people can get themselves into with half-initialized typesetters. (Note even our test expectations had a mistake related to this.)

As soon as we start working on anything that won't be 15.x and needs a major breaking release we can use the *develop* branch again. Currently I'm trying to pick away at some lower hanging fruit that doesn't nede a breaking release. I actually started trying to sort out language modules but kept running into other stuff that can/should be cleaned up first.

If you have any WIP for language/setting stuff I'd be happy to collaborate on it.

alerque: > Trying to see the impacts: Does this mean that I'll have a deprecation here when this lands? [Omikhleia/ptable.sile@`33e26b6`/packages/parbox/init.lua#L59](https://github.com/Omikhleia/ptable.sile/blob/33e26b6a8a9e00f4018d14e5ef1b91dacbae0e5a/packages/parbox/init.lua#L59)

Yes it will show a deprecation warning, but that code is already broken for other reasons. It's using a hacky workflow that we haven't actually supported at all (at least since moving away from the stdlib model to he Penlight model).

Instead of instantiating a new typesetter and then hacking on it, make a new typesetter class inheriting from the base one, then modify the bits you want to change, **then** instantiate it. The "hot patching" it after it is instantiated is problematic already, this deprecation warning is just to catch existing cases of this exact never-properly-supported pattern.

Issue #2232, opened 2025-02-28
Color fonts do not seem to work


Omikhleia:
```
\begin{document}
Hello \font[filename=Nabla-Regular.ttf]{World} of \font[filename=Nabla-Regular.ttf]{SILE}.

\font[filename=Nabla-Regular.ttf]

Lorem ipsum.

\end{document}
```

Nabla font from https://fonts.google.com/specimen/Nabla?categoryFilters=Technology:%2FTechnology%2FColor

Running with `-d color-fonts`, we do see some actions:

```
[color-fonts] Switching to color font Shaper
[color-fonts] Switching from color fonts shaper back to previous shaper
```

But:
 - Dubious line breaks (and the relevant code does indeed invoke leaveHmode when switching shapers...)
 - No colors

![Image](https://github.com/user-attachments/assets/014bcea2-8eb7-4aa7-a647-bbc6af2db108)

Did I miss something?

Omikhleia:
N.B. Totally killing the bidi package (= making its code no-op) does not change the result, so this is probably unrelated to #501

Omikhleia:
Also relating to color fonts: #691

alerque:
c.f. #1261 (other colored fonts stuff)
c.f. https://github.com/sile-typesetter/sile/pull/2249#discussion_r2030012913 and https://github.com/sile-typesetter/sile/pull/2249#discussion_r2030013381 (getting rid of the *package* and making it a regular *shaper*)

Omikhleia:
For the record, _some _ color-enabled font do work...

```
[fonts] Looking for ;10;400;;normal;+ss02;;LTR;./temp/Eadui-CPAL.otf
[fonts] Resolved font family  -> ./temp/Eadui-CPAL.otf
[color-fonts] Switching to color font Shaper
[typesetter] Leaving hmode
```

<img width="704" height="73" alt="Image" src="https://github.com/user-attachments/assets/d5bc8c83-c092-4a2b-ae8e-fa9e26888ff3" />

(With the [Eadui](https://github.com/psb1558/Eadui-Font) font, and the ss02 "blue capitals style".

(Well we still have the dubious leaveHmode when switching fonts)

khaledhosny:
Nabla uses version 1 of COLR table, SILE only supports version 0 (the code predates version 1).

Issue #1965, opened 2024-01-21
nnodes contain the language in several fields


Omikhleia:
The language is in `node.language` (tracked back to `shaper:formNnode`) and in `node.options.language`
It needs to be in the options (for `createNnodes` etc.). We however have two possibilities and we use one or the other at different places. That's code smell -> I would suggest a dedicated refactoring issue...

_Originally posted by @Omikhleia in https://github.com/sile-typesetter/sile/pull/1964#discussion_r1460859798_

alerque:
Continuing from [this PR review comment thread](https://github.com/sile-typesetter/sile/pull/2249#discussion_r2030136128)...

Yes, not only do we have duplicate data, but the data is not as smart as it can/should be. We are stuffing two strings with the language tag that potentially needs parsing and then a huge lookup chain to actually find a language module, hyphenation patterns, or whatever.

I would propose that when we create nodes, instead of having a string property with the language we initialize it with a reference to the actual language module instance it should be handled with. At the same time we can also give it convenience methods like `node:hyphenate()` that can be used to hyphenate the node using the correct language's hyphenator.

Omikhleia:
> ... instead of having a string property with the language we initialize it with a reference to the actual language module instance it should be handled with.

Nodes are low-level construct and should probably not have a _direct_ reference to other objects. It would be a very drastic and deep architecture change at best, if I understand it well. 
t might be an inversion of logic. I am really not sure about it ^^
Also, I am not sure SILE is in a position where changing internals so deeply in each 0.x for little actual (typographically visible) gain. Not that is is not mandated sometimes, but we'd need to be very careful.
(I had some suggested changes to the nodes actually, incl. documentation and questions on some of them - I'll try to clean up that stuff, it could be part of my "wrap-up" of old things that annoyed me,[^1] maybe I'll have a better view afterwards)

[^1]: This issue here is on some nnodes' weirdness, but I have concerns with several other little used nodes.

alerque:
> Also, I am not sure SILE is in a position where changing internals so deeply in each 0.x for little actual (typographically visible) gain. 

[...]

PR #2249, opened 2025-03-26
Make language support bits into first class modules

alerque:
Not working, not finished, will be rehashed...

PR based on #2248, originally branched from `v0.15.10^`, with a merge from the `class-warfare` branch after some initial hacking.


documentation/c10-classdesign.sil
```diff
@@ -261,8 +261,6 @@ The latter acts as a setter or getter, so a minimal code will usually look as fo
 
 \begin[type=autodoc:codeblock]{raw}
 function class:declareOptions ()
-  base.declareOptions(self) -- Note: support parent class options
```
Omikhleia, line 264: base --> plain -> A --> B
B no longer has to call the parent method?

Omikhleia: (Also I don't see how it related to "language support")

alerque: > base --> plain -> A --> B B no longer has to call the parent method?

Correct. One should never use base (prototype) modules directly, but always something that inherits from them. Anything inheriting *directly* from a base class doesn't have to worry about calling the parent method(s) other than `base._init(self)`, all other methods will just be stubs if they are designed for regular use. Second/third level inheritors will have to figure out their own thing. Should `B` call `A`'s method? That's up to `B`. And `A` can decide what if anything it overrides vs. extends from `plain`. The only difference is that `plain` doesn't have to worry about calling stuff from `base`.

alerque: > (Also I don't see how it related to "language support")

It doesn't, all that stuff is in #2248 for merging to master and unrelated to the language module re-org. The commits are showing up here because I have this targeting develop for now, although that actually may change if I can work out shims.

Omikhleia: But then, here, in this very example, base -> plain ->  myclass.

Shouldn't our myclass invoke the parent options (= we inherit from plain, not base)

Omikhleia: (I mean, the example called `base.declareOptions` here, but this was probably overlooked for `plain.declareOptions` no? The `base`variable isn't even defined earlier, unless mistaken)

alerque: Correct on both counts. I didn't catch it because it was incorrectly marked as running `base.<method>`, but the one it actually needed to run was the actual parent class `plain.<method>`. Fixed (here for the moment, but that fixup will end up in the other PR eventually, not this one).

alerque: Fix cherry picked and applied as fixup in #2248, thanks for the catch.

alerque:
This is back together enough to pass about half the regression tests (notable the `en` ones). Its still just a pile of wires as I sort out scopes and init sequences. A quick question for @Omikhleia — off the top of your head do you think it makes more sense to integrate the "nodeMaker" into the language class where there would only ever be one nodeMaker used with each language, or would we ever need alternative nodeMaker classes that could be mixed and matched with different languages? It seems we mostly just use the Unicode nodeMaker across the board so I'm having trouble theorizing what might be useful. For example it makes sense to totally mix and match any combination of inputters, outputters, classes, packages, languages, etc. The question is does it make any sense to mix in different languages with different nodeMaker modules?

Omikhleia:
@alerque Re: nodeMaker

We have a few specific node makers that do fancy things (repeating hyphens for some languages, special handling of punctuation spaces for French...) -- These would be active for normal text in these languages, but there's may be cases where one wants these special rules disabled.
URL links come to mind, when allowed additional language-dependent breakpoints, then ideally:
 - we'd at least cancel repeating hyphens there; we'd also cancel punctuation space -- actually we have a hack of sorts to that effect for French, with a fake language `_fr_noSpacingRules` (with some comment around it) but it's not very general and pre-dates our repeating hyphens for a bunch of languages)...
 - to be complete, we should likely cancel the hyphenchar there too...

I'm not sure how these would fit in the new scheme proposed here, but it would be an occasion to sort them out for generalization (= one language might use different node breaking strategies depending on context).

As an aside note (at a cursory glance of the PR only), the (Liang-) hyphenator is not really tied to the language. It uses language patterns, but it's an algorithm in itself and I would have kept it separate from the "language" class (if I read the proposed changes right). One may even want to replace or extend it -- So as to avoid an inversion of logic, I would have kept it in its own file (perhaps even remade as a proper class).

alerque:
> @alerque Re: nodeMaker

Thanks for the thoughts and processing, it helps...

> We have a few specific node makers that do fancy things (repeating hyphens for some languages, special handling of punctuation spaces for French...) -- These would be active for normal text in these languages, but there may be cases where one wants these special rules disabled.

What I'm hearing here is that node makers are *tightly* linked to language (so stashing the current one under the language module system makes sense) but not absolutely defined by (picking en-US over en-GB shouldn't preclude you from being able to use one specific to handling URLs) .

> URL links come to mind, when allowed additional language-dependent breakpoints, then ideally:
> 
>     * we'd at least cancel repeating hyphens there; we'd also cancel punctuation space -- actually we have a hack of sorts to that effect for French, with a fake language `_fr_noSpacingRules` (with some comment around it) but it's not very general and pre-dates our repeating hyphens for a bunch of languages)...
> 
>     * to be complete, we should likely cancel the hyphenchar there too...

So again the node maker being setup by default by the language makes sense, but again there might be interchangeable parts used between them or activated by packages or whatever.

So again the route I'm headed down is *not* splicing nodemaker functions into the language but keeping them as their own module type, but also *not* elevating that module to our top level types (things that are setup as inputs, e.g. by CLI flags).

> I'm not sure how these would fit in the new scheme proposed here, but it would be an occasion to sort them out for generalization (= one language might use different node breaking strategies depending on context).

Sure, and it should have access at rut time to its parent language and from there to the parent typesetter.

> As an aside note (at a cursory glance of the PR only), the (Liang-) hyphenator is not really tied to the language. It uses language patterns, but it's an algorithm in itself and I would have kept it separate from the "language" class (if I read the proposed changes right). One may even want to replace or extend it -- So as to avoid an inversion of logic, I would have kept it in its own file (perhaps even remade as a proper class).

Yes, I was kind of headed the same direction with the hyphenator stuff — keeping it separate but not promoting it to a top level module type.

typesetters/base.lua
```diff
@@ -46,16 +46,35 @@ end
 -- @param frame A initial frame to attach the typesetter to.
 function typesetter:_init (frame)
    SU._avoid_base_class_use(self)
+   -- TODO: make class first arg of typesetter init, ditch globals hack
+   self.class = SILE.documentState.documentClass
```
Omikhleia, line 50: Why would the typesetter(s) have to know anything about the document class, by the way?

alerque: Among other reasons, because it needs access to call the `newPar()`, `endPar()`, `newPage()`, and `endPage()` functions that are defined in the class.

...and I'm working towards removing all the global state stuff (`SILE.scratch`, `SILE.documentState`, etc.).

alerque: To save you the trouble of searching, see 2fab290d2. Those are example of accessing global state variables that I'm trying to eliminate. Not having that kind of nonsense will make actually using and extending SILE **much much** easier because each piece should actually be modular with known inputs and outputs and no side effects.

Omikhleia: (endPage/newPage are _page builder_ concepts = inherently not tied to the typesetter. Just saying, _this is nonsense in a layered architecture._)

alerque: Agreed. Lots of stuff is in the wrong layers.

I'm leaning towards giving *all* module types some kind of access to navigate the entire tree of modules that make up the active SILE session rather than a minimalist approach of only having what our current implementation uses. Right now this is already the case because we just reach into global state when we want something, but in the process of removing that it isn't always clear what *may be* needed by some alternative implementation. I'm currently proposing all modules have some link into the current process internals: not necessarily as a list of all current modules, but as it's most logical parent and most relevant children.

alerque:
100% regression tests passing, next up unit tests. Some of the number formatting stuff is not wired back up yet.

After that I have to look at shims and see what we can or can't shim and hence what level of breakage this will be.

Overall the new language module layout is starting to make *much* more sense than it ever did. We can still adjust too.

languages/base-hyphenator.lua
```diff
@@ -0,0 +1,222 @@
...
+   self.leftmin = 2
+   self.rightmin = 2
+   self.trie = {}
+   self.exceptions = {}
+   self:registerCommands()
+   self:loadPatterns()
+end
+
+function hyphenator:loadPatterns ()
+   local code = self.language:getShortcode()
```
Omikhleia, line 20: What is a "language short code"? (I guess I know and won't be happy with the answer.)

alerque: Ground work towards introducing full proper BCP-47 language tags where they belong while still having some crufty homegrown stuff under the hood....

That might be the wrong method name and it is only transitional while we implement proper locale parsing, but my thought was that we can start getting actual settings such as `document.language` using BCP-47 while downcasting the info to load our current module structure. Otherwise We have a cart before the horse problem refactoring the code where I can't figure out how to even work on refactoring *anything* because both ends are broken.

Why wouldn't you be happy with that?

Omikhleia: Below it's used for e.g. `languages.%s.hyphens` so I wonder how at last we can get the en-GB patterns etc.

alerque: With a two stage loaeder:

```lua
-- Suppose this is a language model for en-GB:
local language = require("languages.en-GB")

local status, hyphens = pcall(require, ("languages.%s.hyphens"):format(language._name))
if not status then
   status, hyphens = pcall(require, ("languages.%s.hyphens"):format(language:getShortcode()))
end
```

This would load `en-GB` patters if they exist and fall back to the `en` patterns if not.

We can wrap that logic is a much nicer API for determining whether SILE has support for a specific locale or not, but that's the general idea.

And at the very least `:getShortcode()` should be a private undocumented `:_getShortcode()` method because it's unlikely that's actually the interface we want to expose for parsing language tags, it is just something I was using while I try to get everything wired back together and shimmed for backwards compatibility.

Omikhleia: (But go ahead, let's see what it gives eventually)

alerque:
This is coming along. I'm not sure if I want to include the changes to the test suite yet because we actually want to pass the old tests first to prove we're shimming things right. Also this repeats some bits from #2253 and there is an open question on if the fix is correct.

We might be able to target this for v0.15.x too because I don't think I've broken anything beyond the point of no shimability.

alerque:
With a little bit of shimming, all regression tests test from v0.15.10 pass on these refactored internals, and all the busted unit tests pass in their old form except for one that used a private prefixed API anyway.

tests/hyph-except.sil
```diff
@@ -2,14 +2,20 @@
 \use[module=packages.retrograde,target=v0.14.17]
 \nofolios
 \language[main=tr]
-\lua{SILE.typesetter:typeset(SILE.showHyphenationPoints("rab'bin", "tr"))}
+\begin{lua}
+function typeset_hyphens (word, lang)
+   local hyphenator = SILE.typesetter.language.hyphenator
```
Omikhleia, line 7: Why should the hyphenator be accessed via the typesetter? Isn't this an inversion of control (vs. the typesetter invokes a hyphenator, which could a first class citizen)

core/font.lua
```diff
@@ -233,10 +227,10 @@ local font = {
       local font = ot.parseFont(face)
       if font.cpal then
          SILE.require("packages.color-fonts")
-         if SILE.shaper._name ~= "harfbuzzWithColor" then
+         if SILE.shaper._name ~= "harfbuzz-color" then
```
Omikhleia, line 230: We can rename it, but it won't make it work! (#2232)

languages/ja/nodemaker.lua
```diff
@@ -0,0 +1,340 @@
...
+   if bc == 5 then
+      return "0.25zw"
+   end
+   if bc == 7 then
+      return "0.5zw"
+   end
+   return 0
+end
+
+-- local okbreak = SILE.types.node.penalty(0)
```
Omikhleia, line 295: Present in the previous code base, but what is it?

shapers/base.lua
```diff
@@ -134,11 +134,11 @@ function shaper:createNnodes (token, options)
    if #items < 1 then
       return {}
    end
-   local lang = options.language
-   SILE.languageSupport.loadLanguage(lang)
-   local nodeMaker = SILE.nodeMakers[lang] or SILE.nodeMakers.unicode
+   -- TODO this shouldn't need a private interface to a different module type
+   local language = SILE.typesetter:_cacheLanguage(options.language)
```
Omikhleia, line 138: Same kind of remarks, why does the shape has to know the typesetter? (N.B. We do already have one co-dependency here, the shaper accesses the typesetter to get its frame... just to know whether we are in TTB. But this is dubious too, and having all these co-dependencies make it nearly impossible to tweak one without tweaking the other. Introducing more co-dependencies rather than fewer will make it even harder, perhaps even impossible. Hard to know without a clean architecture design / diagram, and so much poor undocumented code in so many places.

core/font.lua
Omikhleia: Besides, it's still and hard-coded reference to a shaper by name. Weird design.

packages/color-fonts/init.lua
```diff
@@ -3,108 +3,6 @@ local base = require("packages.base")
...
-               outputYourself = function ()
-                  SILE.outputter:popColor()
-               end,
-            })
-         end
-      end
-      return nodes
-   end
-end
-
```
Omikhleia, line 107: If I read these deletions correctly, the color-fonts package does not have any code now, but a documentation below that says it does something, which must be wrong. Yet, the package is still loaded in core.font. Weird design.

languages/my/init.lua
```diff
@@ -1,3 +1,10 @@
+local unicode = require("languages.unicode")
+
+local language = pl.class(unicode)
+language._name = "my"
+
+-- TODO tokenizer never used or tested!
```
Omikhleia, line 6: Indeed, not called anywhere in the previous code base either.
--> Do we have an open issue on the topic?

languages/base.lua
```diff
@@ -0,0 +1,167 @@
...
+   fluent:set_locale(lang)
+   os.setlocale(lang)
+   setenv("LANG", lang)
+end
+
+function language:_getLegacyCode ()
+   return self._name
+end
+
+-- TODO: not about hyphenation rules per warning....
```
Omikhleia, line 58: Lot's of TODO in this file, not all very clear. No idea for instance what this one is supposed to mean.

languages/base-nodemaker.lua
```diff
@@ -0,0 +1,96 @@
+local chardata = require("char-def")
```
Omikhleia, line 1: On the naming of this file: why is it `languages/base-nodemaker.lua` and not e.g. `languages/nodemakers/base.lua` (as we have for other typesetters, classes, shapers, etc.)

core/break.lua
```diff
@@ -709,7 +709,9 @@ function lineBreak:doBreak (nodes, hsize, sideways)
          self.threshold = inf_bad
       end
       if self.pass == "second" then
-         self.nodes = SILE.hyphenate(self.nodes)
+         -- TODO give line break modules proper access to typesetter
+         local hyphenator = SILE.typesetter.language.hyphenator
```
Omikhleia, line 713: The line breaker should not have to know the typesetter.

core/font.lua
alerque: I agree it is a weird design, but there are several places where the shaper(s) reach into things they probably don't need. I was mainly focused on refactoring the language modules, not *all* the modules. Since we have a tracking issue for the broken color shaper and this isn't making the design *worse*, only focusing on isolating the language modules I'll leave this alone for now.

languages/kn/init.lua
```diff
@@ -1,3 +1,11 @@
-SILE.languageSupport.languages.kn = { defaultScript = "Knda" }
-local hyphens = require("languages.kn.hyphens-tex")
-SILE.hyphenator.languages["kn"] = hyphens
+local unicode = require("languages.unicode")
+
+local language = pl.class(unicode)
+language._name = "kn"
+
+function language.declareSettings (_)
+   -- TODO get this *unset* when switching to other languages
```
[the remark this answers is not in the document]
alerque: This probably needs fixing before this PR merges. Even though it isn't a regression, the "how" of fixing this might affect the actual shaper of the module interface.

packages/color-fonts/init.lua
alerque: Agreed, it is totally obsolete. The color fonts support is a shaper module, not a package module. I'm adding a deprecation notice and removing the automatic loader.

languages/base.lua
alerque: Fair enough. This was a note that the message in `SU.warn()` needed updating to reflect the actual function, not mixed with notes about hyphenation support. The trouble is I fixed the task without removing the TODO note. (Also working through other TODOS over here...)

languages/my/init.lua
alerque: #2255

languages/base-nodemaker.lua
alerque: Good question. Because all the directories in `languages/*` are language codes and I wanted to be able to iterate them to get a list of languages we support at some level without bumping into things that were not languages.

That also means these should probably go somewhere else entirely. I'll think about what to do with supporting module types that are not part of our core module loader system. Maybe they should just be promoted to top level, I dunno.

core/break.lua
alerque: The line breaker *does* need access to the hyphenator, and the hyphenator is a child of the language, and the language is a child of the typesetter. How else do you propose it get access the hyphenator without introducing more globals or circular dependencies.

I wasn't worried about leaving this one here still since the active typsetter is still a global at `SILE.typesetter` so this is not a regression, it just drops the direct access to instantiated hyphenators as a global.

languages/ja/nodemaker.lua
alerque: I don't know, a smudgy leftover from early development. It look like it was actually used in 2aaf16a9f then ditched in 1e2e4a40d and never actually used again.

tests/hyph-except.sil
alerque: Vs. what not? In this layout the typesetter **does** invoke the language which then invokes it's own hyphenator. The typesetter is the most logical "close" thing for the document itself to reach for. There isn't a global for the currently active hyphenator, and we don't have any kind of sane state API (`SILE.documentState` being a disaster zone I'm chipping away at removing). Hence if the currently active typesetter (e.g. from here via `SILE.process()`) wants to find the current hyphenator for some reason, asking the typesetter for it's currently active language module and asking that language module what it's active hyphenator is seems like the only logical way.

Does that make sense or am I overlooking something?

languages/kn/init.lua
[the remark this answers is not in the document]
alerque: How does fb83af911 sit with you as a work around until we finish refactoring our settings interface so that the settings have a scope that follow their active modules (e.g. switching to previously used typesetters or languages brings in the settings last used in that instance)?

core/break.lua
Omikhleia: > The line breaker does need access to the hyphenator,

Correct

> and the hyphenator is a child of the language, 

Or phrase otherwise, the current hyphenator is an instance of a language dependent hyphenator.

> and the language is a child of the typesetter.

That's where I think something might be wrong. The "typesetter" ought to be language agnostic. Shaper agnostic too (or should it also has the shaper as a child). Line breaker agnostic too. (And I hope to show it should also be page-builder agnostic too... We'll see on that one).

alerque: > > The line breaker does need access to the hyphenator,
> 
> Correct (EDIT: that's the only requirement for the Knuth-Plass algorithm, when "pass 2" kicks in.)

:+1: That's all we're using, but we do need to give it that. The only question I guess is the shape of the API & instances of stuff for it to get there.

Notably we can't just pass a hyphenator on instanciatiaion because the *active* one will change following the language (and possibly other factors). We can potentially pass it one by attaching it to the nodes themselves. That's probably the right way to do this, but I didn't refactor that yet because my main goal wasn't to fix the break module but to get language modules to be a thing in the first place.

> > and to the hyphenator is a child of the language,
> 
> Or to phrase otherwise, the _current_ hyphenator is an instance of a language dependent hyphenator.

Yes, exactly.

> > and the language is a child of the typesetter.
> 
> That's where I think something might be wrong. The "typesetter" ought to be language agnostic. Shaper agnostic too (or should it also has the shaper as a child?). Line breaker agnostic too. (And I hope to show it should also be page-builder agnostic too... We'll see on that one).

I'm listening.

I do think the typesetter should be language agnostic. And it is: but at any given moment when asked to typeset something it does so using it's currently active language. It also keeps a cache of active instances of all languages it has used, and it can switch between them.

I think ultimately the shaper should also be a child of the typesetter (currently is is just a global), and yes probably the line and page break mechanisms too. Yes the shaper should be agnostic of those components, I 100% agree with you there.

The place it gets a bit tangled up is with frames. We have a bit of a circular dependency going on with the frames vs typesetter vs break modules.

Omikhleia: > We can potentially pass it instances by attaching them to the nodes themselves. 

The nnodes know in which language they are (but see also #1965, they even know it too much lol).

alerque: Continuing discussion in the dedicated issue. If I missed anything important in this comment thread not applicable to discuss in that issue or elsewhere don't hesitate to bring it to my attention. Otherwise I think we're on the same page with this, and since fixing the linebreak, pagebreak, and node creation functions is a bit of scope creep for this PR I'll mark this as resolved in favor of dedicated issues.

Omikhleia: That's still a lot of changes for a 0.15.11 (= whether a patch level or minor).
What is the immediate gain for users?

shapers/base.lua
alerque: Already touched on in other threads but for completeness...

> why does the shaper has to know the typesetter?

It doesn't, but what it has in a string with the language tag, and what it actually needs is an instance of that language module. The only sane way (as of this PR) to get that is by pulling the correct instance out of the active typesetter's language cache. If as [mentioned in #1965](https://github.com/sile-typesetter/sile/issues/1965#issuecomment-2781398331) we attach the language instance to the node itself, we would no longer need to ask the typesetter anything about this. (There would still be an issue where it is asking the frame about the writing direction, but that is also likely wrong, we probably need to pass that direction information in with the node, not query it from the frame later.)

I'm marking this as resolved in favor of the dedicated issue.

PR #2256, opened 2025-04-07
refactor(types): Code cleaning and LDoc comments on node types

Omikhleia:
FWIW @alerque Some attempt at documenting and slightly reshuffling the code in `SILE.types.node`...

WIP as doing this, many comments popped up, I'll annotate the PR later, there's a lot of code smell that I'd like to sort out "file by file"

### DIFF
--- a/types/node.lua
+++ b/types/node.lua
@@ -925,74 +925,23 @@ end
 
 -- DEPRECATED FUNCTIONS
 
-function box:isBox ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.type == "hbox"
-      or self.type == "zerohbox"
-      or self.type == "alternative"
-      or self.type == "nnode"
-      or self.type == "vbox"
+local function _deprecated_isX ()
+   SU.deprecated("node:isX()", "is_X", "0.10.0", "0.16.0")
 end
 
-function box:isNnode ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.type == "nnode"
-end
-
-function box:isGlue ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.type == "glue"
-end
-
-function box:isVglue ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.type == "vglue"
-end
-
-function box:isZero ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.type == "zerohbox" or self.type == "zerovglue"
-end
-
-function box:isUnshaped ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.type == "unshaped"
-end
-
-function box:isAlternative ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.type == "alternative"
-end
-
-function box:isVbox ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.type == "vbox"
-end
-
-function box:isInsertion ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.type == "insertion"
-end
-
-function box:isMigrating ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.migrating
-end
-
-function box:isPenalty ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.type == "penalty"
-end
-
-function box:isDiscretionary ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.type == "discretionary"
-end
-
-function box:isKern ()
-   SU.warn("Deprecated function, please use boolean is_<type> property to check types", true)
-   return self.type == "kern"
-end
+box.isBox = _deprecated_isX
+box.isNnode = _deprecated_isX
+box.isGlue = _deprecated_isX
+box.isVglue = _deprecated_isX
+box.isZero = _deprecated_isX
+box.isUnshaped = _deprecated_isX
+box.isAlternative = _deprecated_isX
+box.isVbox = _deprecated_isX
+box.isInsertion = _deprecated_isX
+box.isMigrating = _deprecated_isX
+box.isPenalty = _deprecated_isX
+box.isDiscretionary = _deprecated_isX
+box.isKern = _deprecated_isX
 
 local _deprecated_nodefactory = {
    newHbox = true,
--- a/types/node.lua
+++ b/types/node.lua
@@ -1,5 +1,8 @@
---- SILE node type.
--- @types node
+--- This module defines the standard node classes used in SILE.
+-- Some packages or other modules may define their own node classes for specific purposes,
+-- but this does not have to be covered here.
+--
+-- @module SILE.types.node
 
 local nodetypes = {}
 
@@ -19,19 +22,28 @@ end
 
 local _dims = pl.Set({ "width", "height", "depth" })
 
-nodetypes.box = pl.class()
-nodetypes.box.type = "special"
+--- Base abstract box class used by the other box types.
+--
+-- Other node classes derive from it, adding or overriding properties and methods.
+-- It should not be used directly.
+--
+-- @type box
 
-nodetypes.box.height = nil
-nodetypes.box.depth = nil
-nodetypes.box.width = nil
-nodetypes.box.misfit = false
-nodetypes.box.explicit = false
-nodetypes.box.discardable = false
-nodetypes.box.value = nil
-nodetypes.box._default_length = "width"
+local box = pl.class()
+box.type = "special"
 
-function nodetypes.box:_init (spec)
+box.height = nil
+box.depth = nil
+box.width = nil
+box.misfit = false
+box.explicit = false
+box.discardable = false
+box.value = nil
+box._default_length = "width"
+
+--- Constructor
+-- @tparam table spec A table with the properties of the box.
+function box:_init (spec)
    if
       type(spec) == "string"
       or type(spec) == "number"
@@ -63,23 +75,27 @@ function nodetypes.box:_init (spec)
 end
 
 -- De-init instances by shallow copying properties and removing meta table
-function nodetypes.box:_tospec ()
+function box:_tospec ()
    return pl.tablex.copy(self)
 end
 
-function nodetypes.box:tostring ()
+function box:tostring ()
    return self:__tostring()
 end
 
-function nodetypes.box:__tostring ()
+function box:__tostring ()
    return self.type
 end
 
-function nodetypes.box.__concat (a, b)
+function box.__concat (a, b)
    return tostring(a) .. tostring(b)
 end
 
-function nodetypes.box:absolute ()
+--- Create an absolute version of the box.
+-- All Dimensions are based absolute (i.e. in points)
+--
+-- @treturn box A new box with the same properties as the original, but with absolute dimensions.
+function box:absolute ()
    local clone = nodetypes[self.type](self:_tospec())
    for dim in pairs(_dims) do
       clone[dim] = self[dim]:absolute()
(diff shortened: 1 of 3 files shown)


types/node.lua
```diff
@@ -90,105 +106,64 @@ function nodetypes.box:absolute ()
...
 
-function nodetypes.box:lineContribution ()
-   -- Regardless of the orientations, "width" is always in the
-   -- writingDirection, and "height" is always in the "pageDirection"
+--- Returns either the width or the height of the box.
+-- Regardless of the orientations, "width" is always in the writingDirection,
+-- and "height" is always in the "pageDirection"
+--
+-- @treturn SILE.types.length  The width or height of the box, depending on the orientation.
+function box:lineContribution ()
```
Omikhleia, line 114: I never understood this despite multiple attempts.
  - "mistfit" is a weird concept
  - When this method is used in `hbox:scaledWidth` the second argument is always self.width.
Franky incomprehensible, one can never know what is the expected code (all the more with such self-targeting calls to `SU.rationWidth`, another very obscure concept everywhere...

types/node.lua
```diff
@@ -90,105 +106,64 @@ function nodetypes.box:absolute ()
...
    return self.misfit and self.height or self.width
 end
 
-function nodetypes.box:outputYourself ()
+--- Output routine for a box.
+-- This is an abstract method that must be overridden by subclasses.
+--
+function box:outputYourself (_, _)
```
Omikhleia, line 121: Ah luacheck warning here. I should hould have used `box.outputYourself (_, _, _)`

Or really? Is there any weird benefit for this, rather than ignoring "212/self" in our .luacheckrc, so methods look always the same (with an implicit self, but they be called with it anyway).

types/node.lua
```diff
@@ -686,16 +907,99 @@ function nodetypes.vbox:append (box)
...
+--- A migrating node is a node that can be moved from one frame to another.
+-- Typically, footnotes are migrating nodes.
+--
+-- Derived from `hbox`.
+--
+-- Properties ìs_migrating, is_hbox and is_box are true.
+--
+-- @type migrating
+-- @usage
+-- SILE.types.node.migrating({ material =  ... })
```
Omikhleia, line 919: This class can be passed anything (no constructor to enforce anything), and is thus _very_ loosely defined.

What material is supposed to be is also unclear (a list of vboxes? insertions-specific nodes)?

types/node.lua
```diff
@@ -271,50 +271,88 @@ function nodetypes.nnode:outputYourself (typesetter, line)
...
    self.width = nil
 end
 
-function nodetypes.unshaped:__tostring ()
+function unshaped:__tostring ()
    return "U(" .. self:toText() .. ")"
 end
 
-getmetatable(nodetypes.unshaped).__index = function (_, _)
+getmetatable(unshaped).__index = function (_, _)
```
Omikhleia, line 308: What is this?
The commented code is not helpful.

types/node.lua
```diff
@@ -662,7 +878,12 @@ function nodetypes.vbox:unbox ()
    return { self }
 end
 
-function nodetypes.vbox:append (box)
+--- Added a box or several to the current vbox.
+-- The box height is the height is the total height of the content, minus the depth of the last box.
+-- The depth of the vbox is the depth of the last box.
+--
+-- @tparam box|table box A box or a list of boxes to add to the vbox.
+function vbox:append (box)
```
Omikhleia, line 886: So the parameter is sometimes a single box, sometimes a vboxlist. Unclear API?

It _could_ take both, but the check below is that if it is a box, it can be unboxed. Lot's of undocumented assumptions.

- "insertions" use a vboxlist (material to split) or a box (via its own vbox-derived class).
- If not mistaken, the only other use case is `pagebuilder.collateVboxes` with a vboxlist. This method by the way has _nothing_ to do with the pagebuilder. It's a fairly dumb convenience, and see above comment on the constructor.

types/node.lua
```diff
@@ -479,157 +568,277 @@ function nodetypes.alternative:deltas ()
...
-function nodetypes.vbox:_init (spec)
+local vbox = pl.class(box)
+vbox.type = "vbox"
+vbox.nodes = {}
+vbox._default_length = "height"
+
+--- Constructor
+--
+-- @tparam table spec A table with the properties of the vbox.
+function vbox:_init (spec)
```
Omikhleia, line 826: I remember having suffered here. As with all of these classes, one can pass almost anything and the box class makes it a field. So if called with `SILE.types.node.vbox({ nodes = ... }` then the nodes will be set after the parent class super call. Ok... and it's indeed used this way with lines in the typesetter.
  - So why does pagebuilder.collateVboxes uses an empty constructor and an explicit append()?
  - ... Oh wait, the depth/height logic then differs (here, the max, but with append it's  "the box height is the height is the total height of the content, minus the depth of the last box."

If the discrepancy is legit (?) then some explanation is needed, 'cause again one can't make any sense of it currently.

types/node.lua
alerque: I objected to this a long while back but I think I've come around and would be fine with making that an automatic exception. It does make it easier when copying code from prototypes when some implementations may need self and others don't.

types/node.lua
```diff
@@ -204,17 +177,34 @@ function nodetypes.hbox:outputYourself (typesetter, line)
...
+--
+-- Derived from `hbox`.
+--
+-- Properties is_zerohbox (and convenience is_zero) and is_box are true.
+-- Note that is_hbox is NOT true: zerohbox are used in a specific context
+--
+-- @type zerohbox
+local zerohbox = pl.class(hbox)
+zerohbox.type = "zerohbox"
+zerohbox.value = { glyph = 0 }
```
Omikhleia, line 190: I wondered what this standalone use of `glyph` in this module here.

I _think_ it comes from the pango shaper (regex-searched for occurrences, so maybe I haven't been exhaustive): It's `shaper:shapeToken` does return a "glyph" (vs. value.glyphString etc. with harfbuzz).

But I've a **serious** concern here. Nodes ought to be shaper-agnostic, at least at some level. Or is every potential shaper implementation put its crap in the nodes? And how can we expect using them correctly in packages etc. then?

types/node.lua
```diff
@@ -204,17 +177,34 @@ function nodetypes.hbox:outputYourself (typesetter, line)
...
-nodetypes.nnode.pal = nil
-nodetypes.nnode.nodes = {}
-
-function nodetypes.nnode:_init (spec)
+--- A zerohbox is a special-kind of hbox with zero width, height and depth.
+--
+-- Derived from `hbox`.
+--
+-- Properties is_zerohbox (and convenience is_zero) and is_box are true.
+-- Note that is_hbox is NOT true: zerohbox are used in a specific context
```
Omikhleia, line 185: "in a specific context" = I wrote this here but I have no idea when a zerohbox is used and should differ from an hbox with 0-dimensions !!!

types/node.lua
```diff
@@ -204,17 +177,34 @@ function nodetypes.hbox:outputYourself (typesetter, line)
...
+--
+-- Derived from `hbox`.
+--
+-- Properties is_nnode and is_box are true.
+--
+-- @type nnode
+local nnode = pl.class(hbox)
+nnode.type = "nnode"
+nnode.language = ""
+nnode.pal = nil
```
Omikhleia, line 202: Let me die in utter pain :rofl: What is this? Ahhh might have found it. The Pango shaper again! (See another comment relating to it.) OMG.

types/node.lua
```diff
@@ -686,16 +907,99 @@ function nodetypes.vbox:append (box)
...
+migrating.value = {}
+migrating.nodes = {}
+
+function migrating:__tostring ()
    return "<M: " .. tostring(self.material) .. ">"
 end
 
+-- DEPRECATED FUNCTIONS
+
+function box:isBox ()
```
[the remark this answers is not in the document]
alerque: A long time. Sadly it predates our `SU.deprecated` system that tracks when to warn vs. when to error, but I think it has been more than enough time. These warnings can be dropped, but lets target 0.16.0 for that since to date they still have returned the shim functions. A separate PR to the develop branch will handle that.

types/node.lua
Omikhleia: So the next review we should do is the shapers, to see if the API can be made more generic, no?

If we *can't* avoid enriching/annotating (n)nodes with shaper-specific stuff, at least we ought to put them in a subobject. `nnode.pal` --> `nnode._shaper_specific.pango.pal` or anything similar so one _knows_ that the data is somewhat internal and tied to a shaper... for reasons...

types/node.lua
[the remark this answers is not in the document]
alerque: Nevermind there is no need to put this off until develop, we can hard error now. It's been since v0.10.0 that these were throwing warnings ever if they were also still working. I'm okay with just breaking them now if you are. If you'd rather keep the shim hot until v0.16.0 that's fine with me too.

I nuked them, but if you think keeping the shim is more reasonable I can adapt it.

types/node.lua
```diff
@@ -271,50 +269,88 @@ function nodetypes.nnode:outputYourself (typesetter, line)
...
 end
 
-function nodetypes.unshaped:shape ()
+--- Shapes the text of the unshaped node.
+-- This is done by calling the current shaper with the text and options of the node.
+-- The result is a list of nnodes inheriting the parent of the unshaped node.
+-- The notion of parent is used by the hyphenation logic and discretionaries.
+--
+-- @treturn table A list of nnodes representing the shaped text.
+function unshaped:shape ()
```
Omikhleia, line 319: In relation to that [discussion](https://github.com/sile-typesetter/sile/issues/1965#issuecomment-2782731163)

Unless mistaken, that the only method (that is, besides `SILE.outputter` in output routines, a topic worth discussing separately) when we invoke an external (global) object currently from a node, the `SILE.shaper`. The later will eventually need the language.

types/node.lua
```diff
@@ -324,19 +360,34 @@ function nodetypes.discretionary:__tostring ()
...
 
-function nodetypes.discretionary:cloneAsPostbreak ()
+--- Clone the discretionary node for postbreak use.
+-- This is used to create a new discretionary node that is used in postbreak context.
+-- The discretionary must previously have been marked as used.
+--
+-- When breaking compound words, some languages expect the hyphen (prebreak) to be
+-- repeated in the postbreak context, typically at the beginning of the next line.
+--
+-- @treturn SILE.types.node.discretionary A new discretionary node with the same properties as the original, but marked form use in postbreak context.
```
Omikhleia, line 389: ... "marked **form** use" > Read "for"

types/node.lua
```diff
@@ -1,7 +1,8 @@
---- SILE node type.
--- @types node
-
-local nodetypes = {}
+--- This module defines the stadard node classes used in SILE.
```
Omikhleia, line 1: stadard > standard, my bad.

types/node.lua
Omikhleia: And no specific constructor, so if passed with dimensions etc. it's a zerohbox with dimensions. Maybe a zerohbox is **not** a hbox with zero width, height and depth, after all, but something wholly alien. Who knows what was the intent of the Architect.

types/node.lua
```diff
@@ -271,50 +269,88 @@ function nodetypes.nnode:outputYourself (typesetter, line)
...
+--
+-- @type unshaped
 
-function nodetypes.unshaped:_init (spec)
+local unshaped = pl.class(nnode)
+unshaped.type = "unshaped"
+
+--- Constructor
+--
+-- @tparam table spec A table with the properties of the unshaped node.
```
Omikhleia, line 295: Again those "undefined" `spec`.
The _only_ meaningful use is `SILE.types.node.unshaped({ text = "string", options = { ... }})`

We should make it clear what all these nodes take as contructor argument. Again I understand we can pass any key, and they'll end up in the instance via the generic box... But it's not clear what to expect, and it's certainly not a good object-oriented design, if _any_ node can be abused in so many ways.

types/node.lua
```diff
@@ -271,50 +269,88 @@ function nodetypes.nnode:outputYourself (typesetter, line)
...
+
+function unshaped:_init (spec)
    self:super(spec)
    self.width = nil
```
Omikhleia, line 299: See above. The only reason we may have to clear `self.width` is _if_ a width was passed to the super constructor. The box class is too clever for its own.

types/node.lua
```diff
@@ -19,19 +20,28 @@ end
...
+
+local box = pl.class()
+box.type = "special"
+
+box.height = nil
+box.depth = nil
+box.width = nil
+box.misfit = false
+box.explicit = false
+box.discardable = false
```
Omikhleia, line 38: To read last :rofl:

So we have this concept of explicit/discardable. 

The explicit property is only set "directly" from the calling code (direct affectation, or sometimes passed to the constructor).

Two booleans = four combinations...

Let's try to see what we have:

| Node         | discardable | explicit |
|--------------|-------------|----------|
| normal glue    | true        | false    |
| explicit glue  | false       | true     |
| kern           | false       | false    |
| normal vglue   | true        | false    |
| explicit vglue | false       | true     |
| vkern          | false       | false    |
| penalty        | true        | false    |
| other box      | false       | false    |
| (insertion)    | true        | false    |

Though 4 combinations, but (true, true) is not possible _apparently_, so 3 combinations only.

I thought I knew my old (TeX-derived) box and glue model, but can't find something similar in the TeXBook, unless it has a very different name. What does it mean and imply? 

Why are smallskip/medskip etc. "explicit" for instance?

In the code base, discardable nodes are sometimes skipped. Sounds logical.
Sometimes explicit nodes are skipped, regardless of being discardable or not. I don't get the concept here. What I am sure of is that parts of the code (e.g. the typesetter) does things depending on these properties, and other parts (e.g. the pagebuilder) don't even test exactly the same thing. Either this is very clever, or we are very lucky. I don't want to rely on some uncertain luck.

types/node.lua
Omikhleia: Are there even builds of SILE with the pango shaper? 'Cause keeping a software maintainable is also the take, sometimes, the decision to cut some branches in the tree, if they never blossomed.
I am not even sure it is useful as an "example" of anything, since it probably breaks everywhere it can.

types/node.lua
Omikhleia: > Sometimes explicit nodes are skipped

E.g. at the top of a page.
Hence the necessary TeX-like weird consequence: having to `\hbox{}\par\skip[...]` (or vfill, etc.). Which is not without side-effect, depending on the linespacing strategy (as the empty hbox ends in a vbox line of its own, indeed of 0-height/depth, but tell that to something grid-based for instance...)

What I want to say here is that the desperately obscure design is where I failed at all my attempts of proper page vertical spacing, new page builder strategies and grid-baseline strategies). Might be a skill issue, I can hear it. Might be it's just so hard to guess what this things are supposed to do.... that after all, I am right to feel desparate too, vs. a solution that _kinda_ worked, despite all its `\expandafter` `\csname` `\z@` ugly constructs. :man_shrugging:

Omikhleia: > The explicit property is only set "directly" from the calling code (direct affectation, or sometimes passed to the constructor).

Or... in script, `\skip[discardable=false]` = make me an _explicit_ glue. I'm totally lost, if one boolean parameter is enough, then why two boolean properties?

types/node.lua
```diff
@@ -479,157 +568,277 @@ function nodetypes.alternative:deltas ()
...
+--
+-- @type vssglue
+local vssglue = pl.class(vglue)
+function vssglue:_init (spec)
    self:super(spec)
    self.height = SILE.types.length(self.width.length, infinity, infinity)
 end
 
-nodetypes.zerovglue = pl.class(nodetypes.vglue)
+--- A zerovglue is a standard vglue with zero height and depth.
```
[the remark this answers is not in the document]
alerque: I tentatively agree with immediate deprecation and eventual removal on one condition: `zerohbox` can go away with it. Conceptually they seem the same even if one is more utilized than the other, they just break the immediate sibling relationship of other nodes by stuffing something in the queue that later processing can short-circuit and not bother looking deeper. If this is a valid use case in horizontal mode, it should be valid in vertical mode too. The processing logic isn't that different.

types/node.lua
alerque: Misfit seems like a case of bad naming. This is only used for CJK tate, and I assume means an element whose internal contents and hence dimensioning are rotated 90 degrees from the current flow.

types/node.lua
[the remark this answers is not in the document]
Omikhleia: Well, zerohglue is not really used or anything.

Unfortunately zerohbox is used, and I haven't yet passed the point I understand _why_ it is needed. (Sometimes, just to kill parindent, for instance... So something _else_ is deeply broken as far as concepts are concerned, but we are not yet there... (And I would _tend_ to think that vertical processing is a very different beast. We don't need to kill all monsters on the same day, says Grandpa (returning from the Land of Maths, he ought to know that some beasts are slow to slay properly :rofl: )

types/node.lua
Omikhleia: No idea. Would be tempted to have it removed, and see where SILE breaks, if it does.

types/node.lua
alerque: Fixed.

types/node.lua
alerque: It does. A vast majority of our regression suite outputs blank space if I remove this node type and adjust anywhere unshaped pushes happen to push an hbox node. Notably much of the BiDi segmenting falls apart, your italic correction code would be removed, etc. Somehow most of our nodes are starting out as unshaped.

types/node.lua
alerque: This is exactly why I've been nursing along the Pango shaper at various points just to verify that our shaper API is sane. I don't know anything about this particular case, but in order to be shaper agnostic our nodes do need enough information attached to please any reasonably sane shaper, no?

If we were to drop down to just one shaper our API surface area would be smaller, but SILE itself would be *far* less robust.

types/node.lua
Omikhleia: Wow ^^

alerque:
There are some obvious wins in commits here already, and just as obvious areas that still need attention. Unfortunately I don't think all the open questions are going to get resolved before I have a chance to cut a release, likely tomorrow. I'll likely cherry pick some bits from here so they make the cut for v0.15.11, but keep this PR with discussion running.

types/node.lua
Omikhleia: > This is only used for CJK tate

Hey, that's right. Must relate to  this feature then (screenshot from an Adobe documentation). -- Latin rotated or not in TTB writing.

image

alerque:
BTW the misfit→orthogonal commit here isn't because that idea is finalized (and it it were to be it probably needs a regression shim) but just because I needed something to keep this PR and review comment chains alive after doing a partial merge of the branch locally.

types/node.lua
Omikhleia: An update from reading more of the code from the page builder. It _would_ seem:
 - discardable vglues are removed at the top and bottom of a frame
 - explicit vglues are removed at the top of a frame but kept at bottom

I've not checked yet whether horizontal glues work in a similar way.

Omikhleia: And another update, reading the TeXBook ch. 14 p. 95: "The last four types (glue, kern, penalty, and math items) are called discardable, since they may change or disappear at a line break (...)'. The math items here refers to math-on/math-off kerns, let's not discuss these here... The point is that in TeX-lingvo, a kern is "discardable". But not in SILE -- our kern has "discardable false".

An implication is that the "dirtry tricks" (TeXBook app. D)  for "hanging punctuation" (p. 395) do not work in SILE. Trying to implement a "naive" punctuation protrusion based on that trick (albeit not so naively, but adding  "clever" kerns in the Unicode node breaker)... Of course, one could tweak computeLineRatio() with a hack that would be working...

image

But that's a bad hack and the devil in the the detail -- let's say (I might be wrong but) SILE doesn't seem to follow the exact rules from the TeXbook here either for the Knuth-Plass line breaking.

Anyhow, the conclusion would then be that SILE's notion of "discardable" node here is therefore also different from that of TeX. Which, as usual in such a situation, might be legit or not... But with no explanation and discussion of the original intents, it's real hard to understand the expectations and the rationale behind all these stubble differences. 

I keep running into dead ends in my investigation.
Do you think it would make sense for us to revisit the underlying concepts entirely, and suggest a rebuild from scratch?

Omikhleia: Ah! But while the TeXbook mentions kerns as being discardable (ch. 14), but only consecutive kerns are ignored (app. D). So they are sort of non-discardable, unless consecutive. Er. Knuth's wording is rather opaque here.

Issue #2104, opened 2024-09-09
SILE vs. TeX hyphenation patterns


Omikhleia:
Things not addressed in #2102:

- Things where we have our own upgrades, so likely won't change?
   - Turkish (**tr**) : many changes by @alerque 
   - Esperanto (**eo**): solution by @ctrlcctrlv differing from what TeX currently has
- Things that might have to wait for a better implementation of BCP47...
   - en-GB / en-US (but see also just below, last item)
   - zh-Latn-pinyin
   - mul-ethi (Ethiopic)
   - nb/nn variants (Norsk, we do have them though a refactor could be neat eventually)
   - several la variants (liturgic/classic Latin)
   - mn (Mongolian in Cyrl script or Cyrl-x-lmc for Xalx Mongolian variant)
   - sh (Cyrl vs. Latn, Serbo-Croatian, now deprecated?)
- Things I am not sure of...
  - our "el" vs. TeX's "grc" is problematic
- Things where we differ but may change for better
  - pt (Portuguese): We have @jodros 's updates, but an incoming  PR might improve upon it: https://github.com/hyphenation/tex-hyphen/pull/62 **Should we wait or go for an early adoption?** = **0.16.x** or earlier ?
  - es (Spanish): Lots of differences with our patterns = **0.16.x** ?
  - th (Thai): Lots of differences with our patterns, and no idea where the latter come from. = (?)
  - bg (Bulgarian) = Lots of differences with our patterns, and no idea where the latter come from. = **0.16.x** ? Note the TeX patterns were updated in 2017 so whatever we had might be outdated (?)
  - de (German) = We have something, TeX has several variants depending on orthography reforms, frankly I don't know.
  - en (English, see also above): Currently we are based on en-US, but we lack a whole "additional patterns" that were not present in the original TeX file and were added later (though at an earlier stage than our import). My understanding is that "old" TeX being memory constrained, these were absent from the original implementation. But it's pretty unclear why we don't have the extra patterns... = **0.16.x** ? as it could make rendered documents different...

_Originally posted by @Omikhleia in https://github.com/sile-typesetter/sile/issues/2102#issuecomment-2336669874_

jodros:
> pt (Portuguese): We have @jodros 's updates, but an incoming PR might improve upon it: https://github.com/hyphenation/tex-hyphen/pull/62 Should we wait or go for an early adoption? = 0.16.x or earlier ?

Glad to know about @leolca's work! I can't see reasons to wait.

[...]

PR #2259, opened 2025-04-07
Fix hyphen minima logic and update several hyphenation patterns

### LINKED ISSUE #2017: Hyphenation minimun left/right constraints should be language-specific

Omikhleia:
Closes #2017 

Well the simplest way, but closing it nevertheless -- One can implement better user-defined preference on top of that at least, but the behavior by default ought to be sane.

Addresses some of #2104 (but not all yet, one can only move at a speed that doesn't exceed _c_ in this universe). 

Breaks stuff? Depend on what one calls "breaking". Changing hyphenation minima (in the core logic), and patterns (for some languages here), **will** make the output different. I know it: Respecting (2, 2) instead of (2, 3) for the minima **will** imply all my SILE-made French books are ~~possibly~~ surely impacted. That's annoying. That's also proper typography now, so I am all for it, _despite the annoyance._

**One** busted test failing so far:

```
Passed in:
(string) 'Nor-we-gi-an'
Expected:
(string) 'Nor-we-gian'
```

Expected, IMHO, and https://typst.app/tools/hyphenate/ would also lead to 'Nor-we-gi-an'. So we should accept that too, for the best, and let other tests run. --> EDIT fixed in a dedicated commit


Omikhleia:
So as of French (change in default hyphenation minima), I'd vote for it, despite the annoyance and all my former books being affected (= I don't care if it's 0.15.x or 0.16, my books are supposed to last longer than that...)

As of Portuguese, these are the newest changes that even @jodros though they were good.

Unaddressed: English (very old incomplete patterns) and German (ibid.). On the latter, some Typists may laugh at us. They would be in their right.

Omikhleia:
Ah, now that the busted test is fixed, **14** regular tests are failing. Time to grab a shovel and check all of these.

Omikhleia:
Take this one - same visible ouput, but...

image

As for width 19,3438+16,7188 = 36,0626 = 36,0625 ± very tiny rounding approximation in the text ouput = I'm going to push the fixed expectations for tests falling in the same case.
It's okay. It just shows the (not-that-long-ago fixed) discretionary stuff works, though it's a bit sub-optimal... One problem a day is enough (and this isn't really one).

Omikhleia:
So besides those with same output and just an artefact of the discretionaries in the debug output, there are those where the additional discretionary  - if the language patterns allow for something else than (2, 3) that is - is being used, leading to a possibly different paragraphing: `tests/footnote.sil` is one of these. 

Before:

image

Now:

image

Legit, as `la/hyphens-tex.lua` says Latin is (2, 2). I am going to commit these too tomorrow or later, it's annoying to fix a bug so old and which remained unnoticed so long, but this is correct... I need some sleep though...

alerque:
> it's annoying to fix a bug so old and which remained unnoticed so long, but this is correct...

I understand, but also realize that this bug did *not* affect English or Turkish, by far the most utilized languages in SILE's history. The bugs evident in the text expectations are mostly due to our heavy use of Latin there as filler!

I have tooling for visually reviewing test differences, I'll help with that.

Omikhleia:
> I have tooling for visually reviewing test differences, I'll help with that.

Thanks, you rock as usual!

I only saw your remark after I completed my own review of the failing tests. No problem: it's a sort of a double-check in blind mode, heh, and I wholly agree with the conclusions. There might _still_ be something  weird on some of the debug outputs and our discretionary handling, but I found the same signature in non-affected tests, and the visual output was pretty neat nevertheless. Couldn't pinpoint whether I was hallucinating or there's really a weirdness, but in any case, it's not caused by these changes. I think we are safe to go indeed.

alerque:
> There might _still_ be something weird on some of the debug outputs and our discretionary handling, but I found the same signature in non-affected tests

I had the same thought. Notably it wasn't affecting all instances of the same words in the tests. It only happened with words *near* the edge of the frame, presumably because the word triggered a request to *try* hyphenation points. Even if none of them were actually used, the output seems to get split.

That in itself might have been a regression, but I found the same thing happening before too. My conclusion was that the different hyphenation rules were causing different words to get considered for hyphenation than might have been eligible before (min 2 instead of 3) and hence a few more syllables were showing up shaped separately even though they ended up together instead of split.

Omikhleia:
> It only happened with words near the edge of the frame,

Yep that's exactly it. And we had it before, but it's a bit weird indeed...

PR #2267, opened 2025-04-11
Feat: Support CSL position condition

Omikhleia:
~~Work in progress - pushed for information and to safeguard it before vacations~~

Supporting "position" in `csl:if` construct is one of the things that weren't implemented yet.

It's mostly used in "notes" citation styles, to evaluate and render _ibid._, _cit._ and the likes, rather than repeating the whole entry. Some "in-text" styles do use it too, but they are a minority, and unless mistaken, none of the more usual styles of that kind I checked needed it.

But of course, Chicago (fullnote), Chicago (note) or _Acta Philosophica_ etc. would be much better with it, for people who like to have those shorter references (normally to be put in some foot-or-end notes) rather than to follow author-date, numeric etc. conventions. So I wanted to address it after the completion of other refactors (that eventually went in 0.15.10), and there we are now.

There ought to be two parts of it:
 - [x] Support in the CSL engine[^1]
 - [x] Support in the upper processor (in our case, the SILE package)

[^1]: On my way doing it, I fixed an overlooked boolean coercion issue; and tried to fine-tune a bit the "space/punctuation clean-up" hacks (regexp) at post-rendering (though these are fairly fragile...); and also a small issue with "punctuation-in-quotes" not being applied due the post-rendering being invoked to soon (not noticed earlier as it happens mostly for "notes" citation styles, and... I had checked some output but with French, which does not have punctuation in quotes) :rofl:

### DIFF
--- a/packages/bibtex/csl/engine.lua
+++ b/packages/bibtex/csl/engine.lua
@@ -437,7 +437,6 @@ function CslEngine:_layout (options, content, entries)
       for _, entry in ipairs(entries) do
          self:_prerender()
          local elem = self:_render_children(content, entry)
-         elem = self:_postrender(elem)
          if elem then
             table.insert(output, elem)
          end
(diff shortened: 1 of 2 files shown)


Omikhleia:
Tested with Chicago (fullnote), Chicago (note) and _Acta Philosophica_. Here is what it does, for mere illustration, shown with the latter style:[^1]

image

I.e. notice the "cit.", "_ibid_." in shorter entries of already cited works.

[^1]: Don't ask me why this otherwise nice style does not add a period at the end of the cites while Chicago would. Lol.

alerque (approved):

alerque:
I force-pushed to rewrite commit messages with [imperative verbs](https://cbea.ms/git-commit/#imperative) and amended with some minor style fixups from `stylua`. No other changes were made.

Issue #2268, opened 2025-04-12
Proposal: move BIDI into core


mimbrown:
I've run into the same issue as described in #1796, where hboxes don't work in RTL environments because the BIDI algorithm never gets applied. I had already started prototyping my own fix, which involved a `bidiMakeHbox`, before diving into the issues and finding the author of that issue had [already proposed a similar fix](https://github.com/sile-typesetter/sile/pull/1822) (which never got merged). I'm willing to try to get that going again, but tbh, it's always seemed really strange to me that `bidi` is a package. I feel like bi-directional text is something that should "just work", and that's currently not the case in SILE, which really makes RTL languages feel second-class. So before going for the hotfix/patch version (involving another ugly override of a method on a core class), I'm proposing that BIDI get moved into the core typesetter class. Is there any scope for this?

alerque:
I agree, but "core" is a bit misleading. Everything in this repository including the packages are part of the core SILE distribution, and this particular thing has no business in the un-aptly named "core" directory. Rightly it's a typesetter variant, so it should be either methods on our base / default typesetters or be made into a first class typesetter variant with reasonable default detections put in place for when to enable it.

mimbrown:
True, "core" was probably a poor choice of words. By that I meant, it should be directly and unashamedly part of `typesetters/base`. If you really want to, it could continue to support the "bidi-on" and "bidi-off" commands, which would just set a flag on the typesetter that the typesetter respects, but I would think that should probably be deprecated right? I don't think anyone **needs** bidi turned off, it's such a basic expectation. I'm planning to work on a PR, but let me know if you think this is off course from where you want to go.

mimbrown:
> or be made into a first class typesetter variant with reasonable default detections put in place for when to enable it.

I want to push back on this a bit, I really don't think bidi support is something that should be *enabled*. RTL text shouldn't be any less of a first class citizen than LTR, I think that's part of the ethos of this project. The only thing I can think of is that, if you know you just have LTR text, then you could try to micro-optimize by telling the typesetter "no need to compute the runs on this". But I think we should just make it smart enough that those kinds of optimizations are practically worthless. Looking at the bidi package, I think there's few places we could make the implementation faster. For example, after the text is split into runs by `icu`, if there's only one run and its level matches the ambient level, the function could just abort right there and return the node list unchanged.

TLDR: I want bidi support to be something no user of SILE ever needs to think about, even for optimization purposes.

alerque:
[...]

Issue #1987, opened 2024-02-04
Regarding the signature of outputYourself method


Omikhleia:
So here it is, from https://github.com/sile-typesetter/sile/pull/1981#issue-2115697365

The signature of the `ouputYourself` method is `(self, typesetter, line)`

I'd question, with the rationale of **dis-entangling concepts**, why the typesetter is still needed _at output time_.

Normally, the typesetter's job ought to be finished then, as we are pushing completed (= fully typeset) content to the page...

Unless mistaken, in all cases, what we only need... is the **frame**, to keep track of cursor positions.

I couldn't find other use cases in the code base than `typesetter.frame.direction` (e.g. in boustrophedon), `typesetter.frame:advancePageDirection` or `typesetter.frame:advanceWritingDirection`, `typesetter.frame:newLine()`
`typesetter.frame:newLine()`, `typesetter.frame.state.cursorX` and/or `typesetter.frame.state.cursorY` -- visibly, unless I missed something, it's always using the frame.

There's a reason why the typesetter was passed, as shown in #1981: at a much earlier stage, the frames were not here and the cursor tracking was done in the typesetter -- but that was in 2014...

So it would seem we now pass the typesetter to a method _only_ needing the typesetter's frame, which does indeed sound logical...

_Is this correct and should we proceed forward changing the signature and passing only the typesetter's frame?_
(A compatibility shim should be straightforward, with a `frame.frame` field pointing to itself)

PR #2278, opened 2025-05-03
WIP: Initial change to move bidi to core, get bidi working in hboxes

mimbrown:
The PR has a couple changes:
* Moves most of the bidi functionality to the `core` folder (for lack of maybe a better place). Minimal tweaks to some bidi functions to take, for example, a `direction` instead of a `typesetter`, to reduce direct dependency on the typesetter.
* Move the `splitIntoRuns` and `reorder` calls directly into `typesetters/base.lua#boxUpNodes`
* Add the `splitIntoRuns` and `reorder` calls into `typesetters/base.lua#makeHbox`
* Strip the `bidi` package down to be mostly just deprecated exports.


typesetters/base.lua
```diff
@@ -1489,22 +1491,29 @@ function typesetter:makeHbox (content)
       depth = d,
       value = recentContribution,
       outputYourself = function (box, atypesetter, line)
-         local _post = _rtl_pre_post(box, atypesetter, line)
          local ox = atypesetter.frame.state.cursorX
          local oy = atypesetter.frame.state.cursorY
          SILE.outputter:setCursor(atypesetter.frame.state.cursorX, atypesetter.frame.state.cursorY)
          SU.debug("hboxes", function ()
+            local isRtl = atypesetter.frame:writingDirection() == "RTL"
```
Omikhleia, line 1498: Just a comment in passing. At outputting stage (= here in outputYourself), we'll eventually want not to invoke the typesetter (first argument currently of the method). See also #1987 = it's far too late and out of context, and only the frame should be needed here for cursor positioning (so we would then be able to start untangling those concepts and lean toward a cleaner layered approach). Well, that's my opinion at least, but those concepts are really too much entwined currently.

mimbrown:
@alerque this does not yet include any test updates or anything like that, just the basic code change. The thing that's controversial is, I've also updated the output logic for hboxes, since I've found them to be wrong by default in RTL mode.

Given the content:

```
Testing, testing, \hbox{Test \font[family = "Al Tarikh", direction = "RTL"]{مفتاح معايير الويب!} another}
```

## When the document is LTR

### Current (RTL text is laid out backwards)

image

### New

image

## When the document is RTL

### Current (RTL text is laid out backwards, content in the hbox starts from the *left edge* of the box)

image

### New

image

typesetters/base.lua
Omikhleia: Ah but I read to quickly, you are using `atypesetter.frame` here, so forget my comment then, it's still going via the frame, and the approach _might_ be legit indeed

mimbrown: @Omikhleia thanks for the comment. I agree with what you've pointed towards in #1987.

Is it fair to say that, eventually, `debugHbox` should be aware of the `writingDirection` (or entire frame)? It currently just gets a box and a scaledWidth, but should probably know the direction in order to draw itself.

mimbrown: For the record, I'm not using any info that the current outputYourself function isn't already using.

Omikhleia:
This does break a lot of tests, including unexpected ones.

mimbrown:
Coming back to this after a while. @Omikhleia I've tried to visually inspect the difference between the current typesetter and the changes I've made, and can't make anything out. If I'm reading the diffs correct, it's looking like a lot of these tests are differences of fractions of points, but maybe I'm missing something. Do you have any idea what's going on?

Omikhleia:
> Do you have any idea what's going on?

Sorry for the delay @mimbrown  I failed to see earlier that I was pinged here ^^

I retraced your steps, it turns out that with in the original code, `self:shapeAllNodes()` was invoked twice (once in the bidi logic, then in the regular typesetter) at different times (on different node lists).

I daresay it's really dubious... It might also explain that when I once tried to remove the whole logic bidi (e.g. in the old code, just commenting out the content of the enabling/disabling methods), the debug output also had some subtle differences of the same sort... My feeling is that the old code was wrong, despite being "active" all the time by default... Doh.

Omikhleia:
> (...) but I just put it back to see if that fixes a lot of the random test failures.

Darn, all tests failed but for some font fetching issue in the CI. Let's invoke our great @alerque here :)

> I agree, it's troubling that the behavior would be different, it doesn't seem like it _should_ be, but I'm not prepared to break all tests and stand by the new behavior as better!

I'll add a dedicated comment on the code, but still, this indeed has heavy code smell... Well let's see if it does the trick and confirms the root cause when the CI is repaired.
Yet, personally I'd prefer SILE to address its issues _even_ at the cost of breaking many tests. 

_If the decision is to keep these moved lines_, my opinion is that it would need an extra in-code comment to note the code small and refer to this issue (or a new one), so people who maintain or investigate the code in the future are aware of the potential issue... But let's wait for a result, discussion and decision.

mimbrown:
Yes I just pulled in the upstream, possibly in between me making this PR and now something in the test setup changed.

typesetters/base.lua
```diff
@@ -644,6 +646,9 @@ function typesetter:boxUpNodes ()
    if #nodelist == 0 then
       return {}
    end
+   nodelist = bidi.splitNodelistIntoBidiRuns(nodelist, self.frame:writingDirection())
+   self:shapeAllNodes(nodelist)
```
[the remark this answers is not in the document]
mimbrown: I agree, I'll add the comment (if this works). Just want to validate that it does before going to all the trouble 🙃

typesetters/base.lua
```diff
@@ -1446,10 +1446,13 @@ function typesetter:makeHbox (content)
    self.state.hmodeOnly = true
    SILE.process(content)
 
+   self.state.nodes = bidi.splitNodelistIntoBidiRuns(self.state.nodes, self.frame:writingDirection())
```
Omikhleia, line 1449: by the way, to we need to put that back into `self.state.nodes`? There's a lot of back-and-forth stuff of that kind, I know, in the code base, but it would be best to avoid it, and just use locals.

typesetters/base.lua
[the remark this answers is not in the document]
mimbrown: 🤷‍♂️ it definitely does look weird. If we figure out why this changes subtly when calling `shapeAllNodes` either twice or before/after discarding nodes (not sure which is the culprit yet), that would probably help us think through how it should be fixed.

Looking at the function, it seems like calling it multiple times _shouldn't_ have any effect, unless calling `shape()` on a node can somehow return more unshaped nodes, which would be problematic. So I would guess that calling `shape()` _after_ removing the discardable nodes is what is causing the difference?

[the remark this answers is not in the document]
Omikhleia: My best guess for the difference in the logs is that it is introduced by a glue, and the the line ratio (affecting its shrinking/stretching) occurs to be slightly different -- but the amount is small and distributed on all glues, so in the end its just a few decimals (but this would affect "Mx" lines in the logs in the way we observe).

Hmm... I wonder if the issue isn't what we put pack into the `state.nodes` or not, during all that chain of calls...

See: 

```
function typesetter:breakIntoLines (nodelist, breakWidth)
   self:shapeAllNodes(nodelist)
   local breakpoints = SILE.linebreak:doBreak(nodelist, breakWidth)
   return self:breakpointsToLines(breakpoints)
end
```

Then breakPointsToLines works on `self.states.nodes`, not the `nodelist`. Ouch. That's the big assumption. They might be the same or not and shapeAllNodes works "in place" "by default", so it all depends on what the `nodelist` originally was. It might have been `self.states.nodes` initially... but `nodelist = bidi.splitNodelistIntoBidiRun` returns a _new_ nodelist, and do we put it back to `self.states.nodes`? Otherwise later at some point this will be inconsistent.

Frankly these user of the internal states and the multiple side-affecting methods are a mess to follow. As as wrote in another comment, the constant hesitation on what we work (putting things back and forth in the states) is a likely source of error. The code is quite a mess there.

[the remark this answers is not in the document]
Omikhleia: When I see such things, I am frightened:

https://github.com/sile-typesetter/sile/blob/b3b17ad99a5d2bee7999ab4b8b9e5e3d6c54c59a/core/break.lua#L713

Another cross-boundary side-effect here... Ideally linebreak:doBreak() should do that, but should return the nodelist it worked on...

typesetters/base.lua
```diff
@@ -705,6 +710,12 @@ function typesetter:boxUpNodes ()
          vboxes[#vboxes + 1] = SILE.types.node.penalty(pageBreakPenalty)
       end
    end
+   for i = 1, #vboxes do
+      local v = vboxes[i]
+      if v.is_vbox then
+         v.nodes = bidi.reorder(v.nodes, self.frame:writingDirection())
```
Omikhleia, line 716: Hello @mimbrown

As I am working on my own rewrite of a typesetter ([here](https://github.com/Omikhleia/resilient.sile/pull/145)), I tried to port your proposed changes to it.

It made me realize there's an issue here, when the document uses "liners" (which I introduced earlier to SILE), i.e. for underlining, striking out, etc. portions of text that can be line broken.[^3] The liner logic eventually wraps part of lines if liner boxes (responsible for applying the wanted effect), and these can be deeply nested...

So here, bidi re-ordering -- which as far as I can tell, expects a flat list of shaped nodes, comes far too late, and will not handle these liner boxes.

This raises a few questions:
 - Why is it done here so late on the final line vboxes (via another repeated loop on all lines), while it could perhaps have been done on each line individually before it's finalized (i.e. just before the liner reboxing logic applies).
 - But why is it done, actually, after line-breaking has occurred? Is that due to some interpretation with respect to the Knuth-Plass algorithm?
 - And what exactly is it doing? The bidi code is totally undocumented (and even has some code commented out, never a good sign), with opaque variables (what's `rv` !? why to we have to `SU.compress` it for holes[^2]); and weird sequences[^1]...

I'd totally be in favor of having bidi also made "standard" in my re-implementation, but it's a tricky piece of code, and I don't even know what to test, without knowing the languages :)

[^1]: E.g. sequences of `SU.flip_in_place` / `reverse_each_node`, where the later browses the node list again to `SU.flip_in_place` internal hboxes, knowing their inner structure. Whether such code is legit is hard to say, for one not knowing the details. 

[^2]: That's the only use for this compress utility in all the SILE code base, if not mistaken...

[^3]: E.g. `\underline{....}` in SIL.

Omikhleia: Linked to a 10-year old issue: https://github.com/sile-typesetter/sile/issues/76

mimbrown: @Omikhleia I definitely don't have the answers to all the "why" questions. My goal with this PR was to change things as minimally as possible, and only remove/change things that I was logically positive shouldn't have an effect. BiDi is really painful... but, I am pretty sure, we can't reverse *before* the line-breaking happens. Here's an example of why:

Imagine a string of glyphs (`'` means the character is RTL):
```
ABC DEF A'B'C' D'E'F'
```

If we want to apply the BiDi algorithm, it becomes:
```
ABC DEF F'E'D' C'B'A'
```

Easy. But let's say we need a line break in the middle of the RTL text. In that case, we want the output to look like this:
```
ABC DEF C'B'A'
F'E'D'
```

But if we reorder *before* breaking into lines, we'll end up with:
```
ABC DEF F'E'D'
C'B'A'
```

Off the top of my head, I don't know why we couldn't do the reordering before the reboxing. Maybe we should! But here's why we can't do it before line-breaking.

Omikhleia:
Linked #2268

PR #2280, opened 2025-05-11
Small improvements to the Norwegian translations

turtlegarden:
This pull request improves the Norwegian translations (especially Nynorsk) slightly, and shifts to the preferred[^1] _et al._ translation in Norwegian.

[^1]: Although _et al._ can be used, it is most often an Anglicism, and native Norwegian abbreviations are better, especially in Nynorsk.

### DIFF
--- a/languages/nb/messages.ftl
+++ b/languages/nb/messages.ftl
@@ -10,6 +10,6 @@ book-chapter-title = Kapittel { $number }
 
 hello = Hei <em>{ $name }</em>!
 
-tableofcontents-not-generated = Kjør SILE om igjen slik at innholdsfortegnelsen blir behandlet!
+tableofcontents-not-generated = Kjør SILE om igjen for å behandle innholdsfortegnelsen!
 
 tableofcontents-title = Innhold
--- a/languages/nb/messages.ftl
+++ b/languages/nb/messages.ftl
@@ -2,7 +2,7 @@ bibliography-and = og
 
 bibliography-edited-by = Redigert av { $name }
 
-bibliography-et-al = <language main="la">et al</language>.
+bibliography-et-al = mfl.
 
 bibliography-translated-by = Oversatt av { $name }
 
--- a/languages/nn/messages.ftl
+++ b/languages/nn/messages.ftl
@@ -10,6 +10,6 @@ book-chapter-title = Kapittel { $number }
 
 hello = Hei <em>{ $name }</em>!
 
-tableofcontents-not-generated = Køyr SILE om igjen slik at innhaldslista vert behandla!
+tableofcontents-not-generated = Køyr SILE om igjen for å behandle innhaldslista!
 
 tableofcontents-title = Innhald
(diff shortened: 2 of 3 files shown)


alerque (dismissed):

alerque (commented):
Thanks for taking the time to contribute!

I suspect the `no` locale could also use a similar update, at least to the `bibliography-et-all` message. Are you is a position to comment on or implement that too?

If not we can proceed without it.

Omikhleia:
> Are you is a position to comment on or implement that too?

As an aside note, checking the "standard" CSL locales may help, e.g. "Oversatt av" is indeed used there too for ["nb"](https://github.com/citation-style-language/locales/blob/e27762505af6bfeedb68e0fb02c444b5f310b4e2/locales-nb-NO.xml#L565) vs. "Omsett av" for ["nn"](https://github.com/citation-style-language/locales/blob/e27762505af6bfeedb68e0fb02c444b5f310b4e2/locales-nn-NO.xml#L565) -- So the suggested change here is in line with the expectation.

(I don't know if/when SILE will eventually remove the legacy bibliography implementation, and hence also remove those no longer needed keys from the fluent i18n files)

Omikhleia:
N.B. Likewise, CSL locales indeed have "mfl." for the "et-al" term (for both ["nb"](https://github.com/citation-style-language/locales/blob/e27762505af6bfeedb68e0fb02c444b5f310b4e2/locales-nb-NO.xml#L80) and ["nn"](https://github.com/citation-style-language/locales/blob/e27762505af6bfeedb68e0fb02c444b5f310b4e2/locales-nn-NO.xml#L80))

Omikhleia:
However, "Redigert av" in our "no" file seems wrong for the translator, no? But I don't remember what the default interpretation of "generic" Norwegian" should be (and CSL here only had nb-NO and nn-NO, so it doesn't help). Unless mistaken, the generic "no" might be assumed to be equivalent to "nb" (Bokmål), so should have the same changes -- but surely it would be preferable to have it checked by someone more knowledgeable!

turtlegarden:
Oh, you're right— I'll fi that!

turtlegarden:
Also updated the “regenerate contents” message because it seemed a tad unnatural and did not convey the message properly.

alerque:
I have refactored the commits a bit to use the proper commit messages (so they turn up in our automatically generated release notes) and also to *actually* just alias no→nb messages.

Unfortunately since you opened this PR from your fork's *master* branch instead of a branch dedicated to this PR I can't push those modified commits here and mere. I'll merge locally and you'll be credited with your name on the actual commits, but this PR won't show as merged. Next time make a new branch for each PR you submit and that won't be an issue. Not a big deal, just a workflow nicety.

alerque:
Merged at 39ecffef8.

turtlegarden:
Thank you! Apologies for all the noise and inconvenience, I'm not very used to GitHub.

PR #2281, opened 2025-05-17
Overhaul frames

alerque:
Sanitize the frame type like other things we're passing around, keep track of them attached to modules via a registry, etc.


alerque:
Hmmm, where exactly *was* I in this?

image

PR #2282, opened 2025-05-21
Add simple poetry example

berattrkz23:
This PR adds a minimal but formatted poetry typesetting example using the poetry package. It demonstrates title, author, and structured stanzas. Thanks for the awesome project!

### DIFF
--- a/examples/simple-poetry.sil
+++ b/examples/simple-poetry.sil
@@ -1,0 +1,20 @@
+-- examples/simple-poetry.sil
+
+SILE.require("packages/poetry")
+
+SILE.call("document", {}, function ()
+  SILE.call("title", {}, "A Poem About the Moon")
+  SILE.call("author", {}, "Ada Lovelace")
+
+  SILE.call("poem", {}, [[
+    The moonlight bathes the silent trees,  
+    With silver breath upon the leaves.  
+    A hush descends across the land,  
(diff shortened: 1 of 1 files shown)


Omikhleia:
@berattrkz23 Greetings,

I'd be interested in knowing how you came up with that example, and with which version of SILE you tested it. A few things things seem to be off:
 - This is not a SIL file, but some Lua code (calling SILE internal methods - doing this is not _necessarily_ wrong, but not the best way for a standard example, and not in SIL syntax anyway...).
 - Using `SILE.require` this way, for loading packages, was deprecated several years ago...
 - Examples are _usually_ provided in another repository (= the repository of the [web site](https://github.com/sile-typesetter/sile-typesetter.github.io), not as part of this repository (= the SILE engine)
 - **More importantly,** the SILE standard distribution does not have a `poetry` package[^1], so something else is needed for this example to even work with it...

[^1]: Some 3rd-party module do provide some poetry support, but with a different syntax.

Omikhleia:
@berattrkz23 I am moving this PR to "Pending closure" as none of the above questions were answered and there does not seem to be a way for making it work, unless I missed something. Please feel free to comment, you or other contributors, as a last change to eventually say their word here. Otherwise, after some period of time, this inactive PR will be closed without further notice.

Issue #2283, opened 2025-05-30
Bibliography CSL "substitute" element does not always work as expected


Omikhleia:
When a `<csl:substitute>` element is processed, "substituted variables are suppressed in the rest of the output to prevent duplication"

Well, our logic here is somewhat broken :(

Sometimes they are duplicated, sometimes there's too much suppression. It highly depends on the CSL style (i.e. how it has nesting, groups, choose conditions etc.), so an MWE is hard to extract.

I probably introduced a regression via commit e805a2c (so since 0.15.10) which affects some of my workflows (= duplicated items, due to wrapping macros in a pseudo-groups) with my own styles.

But the logic was likely already broken from the origin... Currently, we might even suppress a variable at sorting, and we then don't have it any longer at rendering... Oops. It's not very clear how to behave here. But even cancelling suppression at sorting, some problem remain afterwards.
I also experimented a quick hack with my main (custom MLA-based) style, but it's probably not a generic fix, and I'm afraid we'd have to track substitutions better than we currently do, the initial stuff was naive there...

So I am just reporting the issue for now, at least, with no idea yet how to fix it properly.
One of the problems is that we don't have tests yet for this CSL light implementation to check both for correctness and regression, but crafting them would be hard (and the CSL tests in their repository are fairly messy too to start with....).

Omikhleia:
For the record here, some of the "problems" are repaired in #2294 but the changes might not be general enough to address all cases. I still need to work on a MWE for some other styles that I don't use as much as Chicago[^1] or MLA, with entries that would trigger the issue and help understanding where the implementation errs.

[^1]: As we have it ("vendored") in our package. The style on the CSL repository was recently upgraded (to a newer edition of the Chicago Manual of Styles) and seems problematic (at a quick glance). It uses a lot more nested constructs in macros and name substitutes than the previous version, so it's not the easiest thing to debug ^^

PR #2294, opened 2025-07-13
Some more bibliography improvements

### LINKED ISSUE #2085: Support ids field in bibtex files
Bib(La)Tex has an interesting `ids` field:

```
  ids = {alias1, alias2}
```

"Citation key aliases for the main citation key. An entry may be cited by any of its aliases ... the citation (will be treated) as if it had used the primary citation key. This is to aid users who change their citation keys but have legacy documents which use older keys for the same entry."

### LINKED ISSUE #2126: bibliography entry filtering strategy
With #2082 we'll have a `\printbibliography` command, outputting a sorted bibliography (with, currently as proposed, an option `cited=boolean`)...

Let's think further.

I don't remember the name of the corresponding (La/Bib)TeX command and its details, and I like to have my mind clear from influences... but what I remember from my younger self: for sure, there's some feature for filtering part of the bibliography.

Example use case:
- I've citations to  books, articles from journals, in-book chapters, book reviews, &c.
- I'd possibly want to split the bibliography according to some criteria (cited books, cited articles, &c), with my own headings and introduction text before these bibliography subsets.

--> This issue is for tracking the idea, compare it with other solutions and propose something interesting.

(Bib/La/Tex would for instance likely allows to filter on its "document types" (the `@type` in .bib entries), but what would this mean in a grander scheme if our target is CSL types? And how one can distinguish primary sources to secondary sources, etc. or what does it imply for "reviews" which are at the same time "regular" articles in some publication?) = Some analysis of existing solutions might be a prerequisite if we don't just want to "mirror" what Bib/La/TeX does, without questioning it.

Omikhleia:
~~**Work in progress** - progressively back-porting upstream some things from my (private) fork of the bibliography components.~~

An Overview Of The Changes So Far...

**(Slightly) Breaking changes**

 - The legacy bibliography implementation is removed.

   The CSL-based bibliography engine, introduced in SILE 0.15.7 at the end of Nov. 2024, was foreshadowing the eventual deprecation of the earlier home-grown legacy implementation of (a very limited and wrong at places subset) of the Chicago author-date style.

   It's now the only one supported, so the code base can be simplified and extended with new features (see below).

   This also encompasses the removal of the i18n keys (in language files) that were used by the legacy implementation.

**Features**

 - `\printbibliography` supports new cool options:

    -  `filter`: A space separated list of filters to apply to the bibliography, when `cited` is false.

        It allows filtering by entry type (ex. `type-book` or `not-type-book`), by keywords (ex. `keyword-linguistics`), or issue date (ex. `issued-2020` for entries issued in 2020, or `issued-2023-2025` for entries issued between 2023 and 2025).

        Entry types and keywords are to be understood in terms of CSL here (not BibTeX).

        Besides those simple built-in filters (which, in my experience at least, cover most use cases), a (very and somewhat internal) low-level Lua API is provided to allow users to implement their own named filters.

       (The reasons for these filters being allowed only when `cited` is false are that they are meant to filter the complete bibliography. Filtering cited entries would make tracking and resetting more complex, for little gain in normal use cases, in my opinion - at least for now, with things still in the flux.)

    - `related`: A boolean option to include "related" entries in the bibliography (in an indented block under the main entries).

       With a Bib(La)TeX bibliography, it correspond to entries which have a `related` field (a comma-separated list of entry keys). So, say, if you have a book B and an article A (e.g. possibly a review, commentary, etc.) with a `related` field containing the key of B, A will be listed nested under B in the bibliography (independently of being cited or not, on its own).

 - By default, ISBNs and ISBNs in bibliographies can now break at dashes, which is useful for justification and line breaking (as these elements are quite long).
 
    This can be disabled by setting the `breakISBN` option to false in the bibliography options.

 - Bib(La)TeX support improvements:

    - The `ids` field for citation key aliasing is now supported.

    - Several fields (`ids`, `keywords`, `related`, and `xdata`) are interpreted as lists of comma-separated values, as they should be.

 - `\nocite` elements are now accepted in `\cites` constructs for grouped citation, e.g., `\cites{\cite[key1]\nocite[\key2]` -- it's a convenience, allowing one to "skip" entries in grouped citations, while keeping them in the
bibliography of cited works, without much rewriting of that block.

**Fixes**

  - Sorting is more consistent in its attempt at a strict weak ordering, avoiding random exceptions.

  - Superscript "folding" is now applied to ordinal numbers in bibliographies, as it was already done for other terms.

  - State is better handled across multiple bibliography runs, avoiding issues with author substitution (e.g., dashes) in per-chapter bibliographies (or when the new "related" references are enabled).

  - Smart straight quotes transformation plays better with italic text (when this extension is enabled, which is the default).

  -  Rework how sorting and substitutions are done in bibliographies. This is an attempt at fixing #2283 but it still seems that something is amiss with certain styles. Well, it might be partial, either due to a misinterpretation of the specification or a real bug. I haven't been able to pinpoint it, despite hours of attempts :crying_cat_face: -- But it does repair something that I broke in SILE 0.15.13, and the code should be much better as new refactored.

**Other changes**

 - Welcome the new `CslProcessor` class.

   It's a big refactor, splitting most of the CSL processing logic out of the `bibtex` SILE package, which now just implements the command "layer" for SILE as a typesetting engine, delegating the work to that processor for the bibliography processing.

   This also allows using the CSL processing logic outside of SILE's regular processing flow (e.g., using SILE as a "Lua-on-steroids" toolkit with ICU, etc. but without going through the typesetter, PDF output, etc.)[^1]

 - Improved in-code documentation, with better LDoc annotations.

[^1]: For instance, I am using it to generate HTML versions of my bibliographies, soon to be released. It's still a bit rough, but it paves the way for better abstractions.

Closes #2126 

Closes #2085 

Regarding #2283 - the changes here might be a partial / imperfect fix, as noted above, but the code makes more sense now.


packages/bibtex/csl/engine.lua
```diff
@@ -975,7 +992,7 @@ function CslEngine:_a_name (options, content, entry)
    end
 
    local nameAsSortOrder = options["name-as-sort-order"]
-   local familyFirst = nameAsSortOrder == "all" or (nameAsSortOrder == "first" and self.firstName) or false
+   local familyFirst = nameAsSortOrder == "all" or (nameAsSortOrder == "first" and self.state.groupState) or false
```
Omikhleia: Oops, wrong replacement here.

Omikhleia:
For the record:
My ETA for my work on this topic is around end of August 2025 (where I'll use the new features and fixes for a new edition of one of my books). I'm mostly done (and could have moved the PR to "Ready", but I have a short break the two next weeks, and will check this and other topics upon my return.

Omikhleia:
> My ETA for my work on this topic is around end of August 2025

Main objectives achieved, and PR moved to Ready.

(I had two other very secondary objectives in my wish-list. I'll probably open dedicated issues for them. This PR here is already a nice baby.)

(Le Dragon de Brume will soon communicate regarding a book update, and an online web site. Both were made with SILE, and the code from this PR.)

I am suggesting this should go in 0.15.14, nearly one year after the work that eventually went in 0.15.7 was initiated. True, it's a _slightly_ breaking _assuming_ some people might have relied on the legacy bibtex implementation in anything serious... I cannot believe such users really exist; and even so, the update should be quite transparent, or at least easy and understandable.

alerque (approved):

Issue #2297, opened 2025-07-15
CI/tests: fonts SBL_Hbrw and AmiriQuran's addresses changed and now CI breaks


Omikhleia:
At this date, the CI pipelines for tests currently fail:

```
/usr/bin/curl -fsSL https://www.sbl-site.org/Fonts/SBL_Hbrw.ttf -o .fonts/SBL_Hbrw.ttf
curl: (22) The requested URL returned error: 404
```

This has notably been seen in PR #2294 and #2278 - esp. for the latter, impending further work on it.

I think the "new "URL for this font is now: https://www.sbl-site.org/wp-content/themes/basket/fonts/bib-fonts/SBL_Hbrw.ttf

Do these folks have a proper/stable way to distribute that font? Can we question why SILE's CI and tests may be using such a "weird" font, while there might be more stable alternatives, distributed in a more stable wyay?

Omikhleia:
FWIW, this font was introduced in 2019 via commit 399b31b085f397573e077d4cc3d83b998c30fb86

Omikhleia:
Now the CI also seems to break for font AmiriQuran (URL in the makefile is to https://github.com/alif-type/amiri/ which has no 0.113 version, has only one version, much more recent, actually contributed by @khaledhosny if I read that well. It seems to be a fork of https://github.com/aliftype/amiri which has more versions... EDIT: Note the slightly different name, confusing...) -- Same questions as above:
 
- Do these folks have a proper/stable way to distribute that font
- Can we question why SILE's CI and tests may be using such a font, while there might be more stable alternatives (?)
- Why is SILE's makefile referring to a fork? (or what changed?)

Omikhleia:
>  EDIT: Note the slightly different name, confusing...


And https://github.com/alif-type/amiri/ redirect to https://github.com/alicloud-sqa/amiri/ while  https://github.com/aliftype/amiri still exists.
Very messy... And frightening...

khaledhosny:
> > EDIT: Note the slightly different name, confusing...
> 
> And https://github.com/alif-type/amiri/ redirect to https://github.com/alicloud-sqa/amiri/ while https://github.com/aliftype/amiri still exists. Very messy... And frightening...

Very frightening indeed. I originally had the organization named `alif-type`, then I renamed it `aliftype`. Apparently github let someone create an `alif-type` organization again, and that person renamed it `alicloud-sqa` so `alif-type` now redirects to that.

alerque:
[...]

PR #2302, opened 2025-08-01
fix(math): Fix placement of math stretchable delimiters and accents

### LINKED ISSUE #2300: Vertical stretchable math delimiters aren't composed as they should
Dear math friends,

See https://github.com/alerque/polytype/issues/17 where this was found, with detailed examples.

Adding some debug:

Image

All boxes seem to have correct height and depth, and correct baselines.

What we see, however, is that the stretched delimiters are not placed where they should -- they are stretched to the right size actually, I would say, but end up offset compared to their target box.

My analysis so far leads to two points:
 - We pick the _closest_ delimiter in the glyph variants. It may be too short, rather than too big -- I'd perhaps question if this is the best choice when variants are available...
 - Then, and that's the most visible and important effect here, we use PDF-rescaling to match the expected height + depth, but that symmetric vertical rescaling can only be "naive" around the center of the glyph, while in a formula such as this one, the glyph would have to be asymmetric.

We _could_ compute a "better" offset to move a rescaled element in its intended box. This still wouldn't be totally right as it would move their middle point. This might not be "too" visible for parentheses and bar-like delimiters, but this would be quite problematic for elements such as curly braces.

Of course, the real solution would be to implement the glyph assembly logic properly (#2155), with different segments for the top and bottom parts so as to match the asymmetry of the intended height and depth. The PDF-rescaling trick was a quick solution, but it cannot work well with such asymmetric sub-formulas.

### LINKED ISSUE #2301: Incorrect placement of horizontally stretched math accents in justified lines
Accidentally noticed when checking solutions for #2300 -- just looking quickly (and not exhaustively yet) for possible impacts on my booklet _SILE and the Hydra of Maths_ -- but the problem was already in the original edition.

It's just slightly more visible here (with the last "attempt" on #2300, resulting in slightly scaled parentheses also raising the overline bar...), but it was still present and I just failed to notice it earlier...

Image

... And that overline bar is actually misplaced -- it should have been as MathJax renders it with the exact same formula (`\textstyle` added to switch to inline mode):

Image

**Minimal reproduction**

```
Inline $`\mathcal{L}_\phi=\overline{\left(\partial_\mu\phi\right)}\partial^\mu\phi-\frac{\mu\phi-m_h^2\left[\bar{\phi}\phi-\frac{\nu^2}{2}\right]^2}{2\nu^2}`

$`\mathcal{L}_\phi=\overline{(\partial_\mu\phi)}\partial^\mu\phi-\frac{\mu\phi-m_h^2[\bar{\phi}\phi-\frac{\nu^2}{2}]^2}{2\nu^2}`

Lorem ipsum dolor sit amet, consectetur adipiscing elit.
$`\mathcal{L}_\phi=\overline{\left(\partial_\mu\phi\right)}\partial^\mu\phi-\frac{\mu\phi-m_h^2\left[\bar{\phi}\phi-\frac{\nu^2}{2}\right]^2}{2\nu^2}`
Sedeiusmod tempor incididunt.

Lorem ipsum dolor sit amet, consectetur adipiscing elit.
$`\mathcal{L}_\phi=\overline{(\partial_\mu\phi)}\partial^\mu\phi-\frac{\mu\phi-m_h^2[\bar{\phi}\phi-\frac{\nu^2}{2}]^2}{2\nu^2}`
Sedeiusmod tempor incididunt.
```

Image

**Analysis**

 - It's correct when the line is short.
 - It's independent from `\left...\right` constructs being explicitly used or not
 - It occurs when the line is stretched or shrunk due to justification. The formula is also stretched or shrunk, everything seems at its intended place, except the accents.

   Image

**Solution**

It turns out there's just a missing scaling for the x position in the call to scaleFn() in method `elements.text:output`

```
SILE.outputter:scaleFn(x, y, xratio, yratio, function ()
```

Should have been:

```
SILE.outputter:scaleFn(scaleWidth(x, line), y, xratio, yratio, function ()
```

(Well, this `scaleWidth(x, line)` is invoked a few lines earlier, so could be stored in a variable of course, and just reused)

And hurray! This does fix too in my booklet some _slightly_ offset accents on p. 5 (formula above the table of contents), which I had noticed but did not understood at the time.

Omikhleia:
Closes #2300 

Closes #2301 

For #2300, it implements the solution presented in my last [comment](https://github.com/sile-typesetter/sile/issues/2300#issuecomment-3143855759)

CI tests are currently failing due to #2297 -- I have **not** checked whether some of the existing tests are affected or not.

What I _did_ check is:

 - The small scenario covering both formulas mentioned in the issues

   ```
   $$`\def{floor}{\lfloor #1 \rfloor}
   p_n = 1 + \sum_{i=1}^{2^n} \floor{(\frac{n}{\sum_{j=1}^i \floor{\cos^2(\pi \frac{(j-1)! + 1}{j})}})^{\frac{1}{n}}}`

   ---

   Inline $`\mathcal{L}_\phi=\overline{(\partial_\mu\phi)}\partial^\mu\phi-\frac{\mu\phi-m_h^2[\bar{\phi}\phi-\frac{\nu^2}{2}]^2}{2\nu^2}`

   Lorem ipsum dolor sit amet, consectetur adipiscing elit.
   $`\mathcal{L}_\phi=\overline{(\partial_\mu\phi)}\partial^\mu\phi-\frac{\mu\phi-m_h^2[\bar{\phi}\phi-\frac{\nu^2}{2}]^2}{2\nu^2}`
   Sedeiusmod tempor incididunt.
   ```

 - A decent set of formulas in my booklet _SILE and the Hydra of Maths_... It's easy to miss something at unexpected places, but a lot more formulas than I thought are actually affected, and as far as I can tell, they all look better, except maybe one (but LaTeX does seem to give a very similar result here, though we don't use the same exact table constructs).[^1]

[^1]: Aside note, if I ever want to re-edit that booklet, the change on vertical delimiters is most of the time subtle, but very visible at some places... Sadly, page breaks occur at different places now, and sometimes my nice "filler" illustrations or quotes would have to be moved here and there. :rofl:

### DIFF
--- a/packages/math/base-elements.lua
+++ b/packages/math/base-elements.lua
@@ -510,7 +510,12 @@ function elements.stackbox:shape ()
       -- Handle stretchy operators
       for _, elt in ipairs(self.children) do
          if elt:is_a(elements.text) and elt.kind == "operator" and SU.boolean(elt.stretchy, false) then
-            elt:_vertStretchyReshape(self.depth, self.height)
+            local stretched = elt:_vertStretchyReshape(self.depth, self.height)
+            if stretched then
+               -- Stretched elements may have altered height and depth
+               self.height = maxLength(self.height, elt.height)
+               self.depth = maxLength(self.depth, elt.depth)
+            end
          end
       end
       -- Set self.width
@@ -1249,10 +1254,16 @@ function elements.text:_vertStretchyReshape (depth, height)
       -- We only do it if the scaling logic found constructions on the vertical block axis.
       -- It's a dirty hack until we properly implement assembly of glyphs in the case we couldn't
       -- find a big enough variant.
-      self.vertExpectedSz = height + depth
-      self.vertScalingRatio = (depth + height):tonumber() / (self.height:tonumber() + self.depth:tonumber())
-      self.height = height
-      self.depth = depth
+      -- At output, We will peform a symmetric scaling of the glyphs, centered on the height axis.
+      local constants = self:getMathMetrics().constants
+      local scaleDown = self:getScaleDown()
+      local axisHeight = constants.axisHeight * scaleDown
+
+      local midSz = maxLength(height - axisHeight, depth + axisHeight)
+      self.vertScalingRatio = 2 * midSz:tonumber() / (self.height + self.depth):tonumber()
+      self.vertScalingOffset = self.height - (midSz + axisHeight) / self.vertScalingRatio
+      self.height = midSz + axisHeight
+      self.depth = midSz - axisHeight
    end
    return hasStretched
 end
@@ -1285,6 +1296,9 @@ function elements.text:output (x, y, line)
       compensatedY = SILE.types.length(y.length + self.value.items[1].depth - self.value.items[1].fontDepth)
    else
       compensatedY = y
+   end
+   if self.vertScalingOffset then
+      compensatedY = compensatedY + self.vertScalingOffset
    end
    local xs = scaleWidth(x, line) -- account for stretchability/shrinkability in line justification
    SILE.outputter:setCursor(xs, compensatedY.length)
--- a/packages/math/base-elements.lua
+++ b/packages/math/base-elements.lua
@@ -1286,7 +1286,8 @@ function elements.text:output (x, y, line)
    else
       compensatedY = y
    end
-   SILE.outputter:setCursor(scaleWidth(x, line), compensatedY.length)
+   local xs = scaleWidth(x, line) -- account for stretchability/shrinkability in line justification
+   SILE.outputter:setCursor(xs, compensatedY.length)
    SILE.outputter:setFont(self.font)
    -- There should be no stretch or shrink on the width of a text
    -- element.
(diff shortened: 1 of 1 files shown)


Omikhleia:
And when one knows what caused an issue, here on #2300 ...

It's hard not wondering how one could have missed it, and what could have been a real-world simple equation showing it...

image

```
$$`\left\lfloor \frac{\left\lfloor \frac{x}{m} \right\rfloor}{n} \right\rfloor = \left\lfloor \frac{x}{mn} \right\rfloor`
$$`\left\lceil  \frac{\left\lceil  \frac{x}{m} \right\rceil }{n} \right\rceil = \left\lceil \frac{x}{mn} \right\rceil`
```

And we are now in line with Typst here too, perhaps even better :cat: -- in Typst's case, the right-side equations do not seem to stretch perfectly, heh...

image

EDIT: I picked the LaTeX formulas on [Wikipedia](https://en.wikipedia.org/wiki/Floor_and_ceiling_functions#Nested_divisions), but of course here SILE wouldn't need the left...right constructs, so the above example may be simplified:

```
$$`\lfloor \frac{\lfloor \frac{x}{m} \rfloor}{n} \rfloor = \lfloor \frac{x}{mn} \rfloor`
$$`\lceil \frac{\lceil \frac{x}{m} \rceil}{n} \rceil = \lceil \frac{x}{mn} \rceil`
```

alerque (approved):

PR #2306, opened 2025-08-22
fix(math): Properly handle empty final line in TeX-like math tables w…

### LINKED ISSUE #2303: Lost content in some math table without final backslashes
Djot MWE:

```
$$`\table[columnalign=left left]{
  a & 1 \\
  b & 2 \\
}`{numbered=true}

$$`\table[columnalign=left left]{
a & 1 \\
b & 2
}`{numbered=true}

$$`\table[columnalign=left left]{
a & 1 \\
  & 2 }`{numbered=true}

$$`\table[columnalign=left left]{
a & 1 \\
  & 2 \\}`{numbered=true}
```

Image

In formula (3) lacking a final `\\`, the second line of the table disappeared. Formula (2} also lacked it, but note that the line is present.

At a glance, the second part of the comparison below seems wrong:

https://github.com/sile-typesetter/sile/blob/7002613c5a4443764f3fd92ba4005178a01bdc93/packages/math/texlike.lua#L171-L173

The preceding comment shows that the intent here was clearly to support an optional `\\` at the end of the sequence.

Removing that second part of the comparison makes all 4 formulas show all their lines, but then (1) and (4) have an extra empty row. It's easy to check by adding brackets `[` ... `]` around the table; these brackets are stretched vertically and are then bigger in these two formulas.

BTW. On that occasion, also note that `dim2_arg_inner` and  `dim2_arg` are actually defined twice -- Luacheck linting did not raised an error as it is disabled around such grammars.

This should be an easy fix -- I haven't dumped the content of the mathlists in `t` but I'd guess the code is just not checking the right thing.

Omikhleia:
…ithout losing content

Closes #2303 

```
$$`\table[columnalign=left left]{
  a & 1 \\
  b & 2 \\
}`{numbered=true}

$$`\table[columnalign=left left]{
  a & 1 \\
  b & 2
}`{numbered=true}

$$`\table[columnalign=left left]{
  a & 1 \\
    & 2 }`{numbered=true}

$$`\table[columnalign=left left]{
  a & 1 \\
    & 2 \\}`{numbered=true}
```

image

### DIFF
--- a/packages/math/texlike.lua
+++ b/packages/math/texlike.lua
@@ -150,28 +150,14 @@ local mathGrammar = function (_ENV)
          (P"}" + E("`}` expected"))
          ) / function (...)
             local t = {...}
-            -- Remove the last mathlist if empty. This way,
-            -- `inner1 \\ inner2 \\` is the same as `inner1 \\ inner2`.
-            if not t[#t][1] or not t[#t][1][1] then table.remove(t) end
+            local last = t[#t]
+            -- Remove the last element if empty:
+            -- So that `inner1 \\ inner2 \\` is the same as `inner1 \\ inner2`.
+            if #last == 0 or (#last == 1 and #last[#last] == 0) then
+               -- Empty mathlist, or mathlist containing a single empty mathlist.
+               table.remove(t)
+            end
             return pl.utils.unpack(t)
-         end
-
-   local dim2_arg_inner = Ct(V"mathlist" * (P"&" * V"mathlist")^0) /
-      function (t)
-         t.id = "mathlist"
-         return t
-      end
-   local dim2_arg =
-      Cg(P"{" *
-         dim2_arg_inner *
-         (P"\\\\" * dim2_arg_inner)^1 *
-         (P"}" + E("`}` expected"))
-         ) / function (...)
-         local t = {...}
-         -- Remove the last mathlist if empty. This way,
-         -- `inner1 \\ inner2 \\` is the same as `inner1 \\ inner2`.
-         if not t[#t][1] or not t[#t][1][1] then table.remove(t) end
-         return pl.utils.unpack(t)
          end
 
    -- TeX uses the regular asterisk (* = U+002A) in superscripts or subscript:


packages/math/texlike.lua
```diff
@@ -150,30 +150,16 @@ local mathGrammar = function (_ENV)
...
+            -- Remove the last element if empty:
+            -- So that `inner1 \\ inner2 \\` is the same as `inner1 \\ inner2`.
+            if #last == 0 or (#last == 1 and #last[#last] == 0) then
+               -- Empty mathlist, or mathlist containing a single empty mathlist.
+               table.remove(t)
+            end
             return pl.utils.unpack(t)
          end
 
-   local dim2_arg_inner = Ct(V"mathlist" * (P"&" * V"mathlist")^0) /
```
alerque, line 159: Is the removal of this block related to the commit here or an unrelated cleanup of unused code?

alerque (changes requested):
Possibly split commit or document better in the commit message how the code removal is related.

packages/math/texlike.lua
Omikhleia: As far as I can remember, it seems to be unused code, that was somehow related to that very logic (it has the same kind of comment "Remove the last mathlist if empty" and likewise a similarly broken logic). I assumed it was an early attempt, but I can't remember (now), whether I (then) tried to check its origin. I would consider it as co-related clean-up / refactor, but feel free to split the commits if you prefer !

alerque: Fair enough.

image

alerque (approved):

packages/math/texlike.lua
Omikhleia: Lol :heart: -- I know for sure I am not as wise as that old wizard.

Actually; I realize I had mentioned it in the issue (a bit lost in all my text):

> BTW. On that occasion, also note that `dim2_arg_inner` and `dim2_arg` are actually defined twice -- Luacheck linting did not raised an error as it is disabled around such grammars.

The `dim2_xxx` stuff is exactly what is patched above. Well the second declaration, obviously. No wonder they looked alike. Removing the duplicate declaration is indeed co-related to that issue.

PR #2307, opened 2025-08-23
fix(math): Adjust math font size to match the current document font

### LINKED ISSUE #2145: Math font size does not follow the current normal text size

Omikhleia:
Math formula in block quotes, footnotes, headers, etc. now adapt to the current document size instead of using a fixed value. The adjustment is based on the eye (ex-height) for consistent optical alignment.

Closes #2145 

Better approach than #2149 :

**Advantages**

 - The user can still set math.font.size to a fixed value (globally or temporarily), and then this will be used.
 - Otherwise, adjustment is to the ex-height, matching the surrounding font (esp. interesting for inline formulas etc.)

**Scenario**

```
{.unnumbered .notoc}
## Regarding Euler's $`e^{\mathrm{i}\pi} + 1 = 0`

Some people say $`e^{\mathrm{i}\pi} + 1 = 0` is one of the most beautiful equation in mathematics.

> But all I want to check here is that $`e^{\mathrm{i}\pi} + 1 = 0` honors the current document font size.[^euler]

[^euler]: And $`e^{\mathrm{i}\pi} + 1 = 0` in a footnote is even smaller.
```

Run with `-u inputters.djot temp/euler-identity.dj -O papersize=a7 -Olandscape=true -c resilient.book`

I.e. adaptation between default text (Gentium) and math (Libertinus).

image

**Caveat**

 - In ex-height adjustment mode (default), performances are slightly degraded. We could do some better font caching, but that's a more general question, somewhat orthogonal to that specific case here. I might open a dedicated issue for discussion.
 - Existing tests will fail -- But ideally one would want to make a global pass with all of the proposed math-related PRs for this milestone, after #2297 is also fixed.

### DIFF
--- a/packages/math/base-elements.lua
+++ b/packages/math/base-elements.lua
@@ -209,22 +209,8 @@ function elements.mbox:_init ()
    self.value = {}
    self.mode = mathMode.display
    self.atom = atoms.types.ord
-   local font = {
-      family = SILE.settings:get("math.font.family"),
-      size = SILE.settings:get("math.font.size"),
-      style = SILE.settings:get("math.font.style"),
-      weight = SILE.settings:get("math.font.weight"),
-      -- https://learn.microsoft.com/en-us/typography/opentype/spec/math#opentype-layout-tags-used-with-the-math-table
-      --   "Script tag to be used for features in math layout.
-      --   The only language system supported with this tag is the default language system."
-      -- Thus, needed for the ssty feature in superscript/subscript to work properly.
-      script = "math",
-   }
-   local filename = SILE.settings:get("math.font.filename")
-   if filename and filename ~= "" then
-      font.filename = filename
-   end
-   self.font = SILE.font.loadDefaults(font)
+   -- The math font is assumed to be already set as current by the calling context.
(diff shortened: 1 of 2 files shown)


Omikhleia:
Push-forced: amended the typo in the commit log headline ("mattch" > "match"). (And fixed the PR title too, obviously)

alerque (approved):

PR #2309, opened 2025-08-24
refactor(fonts,shapers): Cache font adjustment for efficiency

Omikhleia:
WIP experiment -- Saved as PR before I switch to other topics.

Font adjustment is a bit expensive - each time it is used, we activate the target font at the same size as the current font, shape an "x" or "H" for both to compute ex-height or cap-height, compute the ratio. If it wasn't used often, that would be okay, but with it being used for code, and possibly math, it gets invoked a lot (in documents with, obviously, either lots of "code"-formatted strings or lots of formulas).
That's a lot of repetitive small operations: (building near to identical options, loading fonts (C involved in the process), decoding fonts (shaper but also possibly Lua OT parsing, etc.), calling the shaper (C involved again), etc.

On my math booklet, this caching gain is however but a few seconds for an approx. 60 second build with a "heavy" math font (Garamond Math); and less noticeable with some other fonts (Libertinus Math or STIX Two Math)... So while interesting, it's perhaps not that decisive towards performance - i.e. there are lots of other performance costly operations in the whole process...

SILE already does a lot of font caching by a very narrow key in many places. But it's not clear to me how "ad hoc" the caching is, and it would seem we still perform a lot of operations without much clarity whether it is actually needed for good reasons... Surely that's something that could be questioned and clarified, for proper generalization.

### DIFF
--- a/core/font.lua
+++ b/core/font.lua
@@ -12,42 +12,82 @@ local adjust_metric = P("ex-height") + P("cap-height")
 local adjustment = Ct(Cg(bits.number, "amount")^-1 * bits.ws * Cg(adjust_metric, "unit"))
 -- stylua: ignore end
 
-local function measureFontAdjustment (metric)
+--- Measure a font adjustment metric (ex-height or cap-height) for a given font options.
+-- @tparam string metric The metric to measure ("ex-height" or "cap-height")
+-- @tparam table fontoptions The font options to use
+local function measureFontAdjustment (metric, fontoptions)
    if metric == "ex-height" then
       -- Uses the height of lowercase letters.
       -- This is used to normalize lowercase letters across fonts.
       -- The height of the lowercase letter "x" is used as the reference.
       -- Another option would be to use the OS/2 font table sxHeight value when available.
-      return SILE.shaper:measureChar("x").height
+      return SILE.shaper:measureChar("x", fontoptions).height
    end
    if metric == "cap-height" then
       -- Uses the the height of uppercase letters.
       -- This is used to normalize uppercase letters across fonts.
       -- The height of the uppercase letter "H" is used as the reference.
       -- Another option would be to use the OS/2 font table sCapHeight value when available.
-      return SILE.shaper:measureChar("H").height
+      return SILE.shaper:measureChar("H", fontoptions).height
    end
    SU.error("Unknown font adjust metric " .. metric)
 end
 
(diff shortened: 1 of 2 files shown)


khaledhosny:
> shape an "x" or "H" for both to compute ex-height or cap-height

FWIW, x-height and cap-height are font properties set by font authors in the `OS/2` table. Applications are not supposed to measure glyphs to get them, and not all fonts have glyphs for “x” or “H”.

Omikhleia:
> FWIW, x-height and cap-height are font properties set by font authors in the `OS/2` table. Applications are not supposed to measure glyphs to get them, and not all fonts have glyphs for “x” or “H”.

It's probably wrong anyway to ask for ex-height adjustment in fonts not having an "x".

But yes, fundamentally, you are right, and an in-code comment mentions that fact:

https://github.com/sile-typesetter/sile/blob/7002613c5a4443764f3fd92ba4005178a01bdc93/core/font.lua#L20

But as far as I can tell, there are quite a lot of fonts where these OS/2 properties aren't properly set and seem off.[^1]

Moreover the "ex" unit in SILE elsewhere in the code base is also based on measuring a "x".... 

https://github.com/sile-typesetter/sile/blob/7002613c5a4443764f3fd92ba4005178a01bdc93/types/unit.lua#L201-L206

Truly, this could be a dedicated issue. Contributions to a saner approach would be welcome. --> EDIT: Issue opened #2317 

(The present draft PR _here_ is only about possibly adopting caching on existing code,[^2] so is not exactly appropriate for the discussion.)

[^1]: On another set of properties for instance:  https://github.com/alerque/libertinus/issues/511

[^2]: And as noted, I am not fully convinced by this implementation, the gain is not that great. But how SILE performs font caching here and there is questionable nonetheless.

Omikhleia:
By the way, in passing, SILE's OS/2 table parsing is quite partial at this date and does not extend to the whole table:[^edit]

https://github.com/sile-typesetter/sile/blob/7002613c5a4443764f3fd92ba4005178a01bdc93/core/opentype-parser.lua#L815

[^edit]: (EDIT) I actually added that partial OS/2 table support a while ago, but only as far as was needed for some feature at that time (for #1422)

So using sxHeight / sCapHeight would need to complete that parsing. That could be an interesting issue on its own too, and contributions would be welcome too, as a prerequisite. --> EDIT: Issue opened #2316

PR #2320, opened 2025-09-07
fix(types,fonts): Base "ex' unit on proper metrics and add "cap" unit

### LINKED ISSUE #2317: Reconsider how the ex unit is computed
**Blocked** by #2316 

From a discussion with @khaledhosny in https://github.com/sile-typesetter/sile/issues/2309#issuecomment-3222156139
> FWIW, x-height and cap-height are font properties set by font authors in the `OS/2` table. Applications are not supposed to measure glyphs to get them, and not all fonts have glyphs for “x” or “H”.

(...)
The "ex" unit in SILE is based on measuring a "x".... Maybe it should rather use the sxHeight property from the OS/2 font table.

https://github.com/sile-typesetter/sile/blob/7002613c5a4443764f3fd92ba4005178a01bdc93/types/unit.lua#L201-L206

Contributions to a saner approach would be welcome.

Omikhleia:
The "cap" unit is proposed in CSS4 Draft and makes sense for cap-height font adjustement.
Both "ex" and "cap" shall be based on font metrics when available, and only measure a character as a fallback heuristic.
Some fonts may not have the necessary Latin characters, in which case we shall resorti to a decent fallback (as per CSS recommendations).

Closes #2317 

**Note 1**: Obviously this only takes all its value with PR #2319 

**Note 2**: Some fonts have small differences between the sxHeight metric and the measured height of a lowercase “x”.
As a result, several regression tests (e.g., those using Libertinus Serif) will fail.
Minor variations in output are also to be expected.
But if I have interpreted the issue correctly, this represents better typography — and I am weighting for it.


alerque (changes requested):
Commit needs splitting, 1 fix + 1 feat.

I can do this, just commenting to myself since PR merging is blocked ATM while I'm working on test stuff locally.

alerque (approved):

PR #2340, opened 2025-11-05
Support more MathML attributes and other math fixes and enhancements

Omikhleia:
~~NOT READY FOR MERGE~~

~~Based on a temporary branch with cherry-picked commits from several other of my PRs for 0.15.14~~

Additions:
- refactor(math): Generalize attribute passing to MathML node elements
- feat(math): Support MathML mathcolor and mathbackground attributes = some progress on remaining items in #2138
- ~~fix(math): Properly apply 'cramping' styles in text mode and for radicands~~ = see #2341 EDIT: Moved to its own PR #2345

### DIFF
--- a/packages/math/typesetter.lua
+++ b/packages/math/typesetter.lua
@@ -111,30 +111,39 @@ function ConvertMathML (_, content)
    if content == nil or content.command == nil then
       return nil
    end
+   local attributes = content.options
+
    if content.command == "math" or content.command == "mathml" then -- toplevel
-      return b.stackbox("H", convertChildren(content))
-   elseif content.command == "mrow" then
-      local ret = b.stackbox("H", convertChildren(content))
+      return b.stackbox(attributes, "H", convertChildren(content))
+   end
+
+   if content.command == "mrow" then
+      local ret = b.stackbox(attributes, "H", convertChildren(content))
       -- Internal property to keep tracks or paired open/close in TeX-like syntax
       ret.is_paired = content.is_paired
       return ret
-   elseif content.command == "mphantom" then
-      local special = content.options.special
-      return b.phantom(convertChildren(content), special)
-   elseif content.command == "mi" then
-      local script = content.options.mathvariant and mathVariantToScriptType(content.options.mathvariant)
+   end
+
+   if content.command == "mphantom" then
+      return b.phantom(attributes, convertChildren(content))
+   end
+
+   if content.command == "mi" then
+      local script = attributes.mathvariant and mathVariantToScriptType(attributes.mathvariant)
       local text = content[1]
       if type(text) ~= "string" then
          SU.error("mi command contains content which is not text")
       end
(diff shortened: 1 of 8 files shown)


Omikhleia:
Rebased now that other math-related MRs got merged.

@alerque Thus, I assigned this to 0.15.14 rather than 0.15.x, but feel free to comment.

I do think it's a good refactor/addition though.

Omikhleia:
> I do think it's a good refactor/addition though.

Note to self: if all this math stuff is merged, I'll need to issue a new edition of my "SILE and the Hydra of Maths" (Showcase & Critical Assessment) booklet ^^ ;)

packages/math/typesetter.lua
```diff
@@ -111,30 +111,39 @@ function ConvertMathML (_, content)
...
+   local attributes = content.options
+
    if content.command == "math" or content.command == "mathml" then -- toplevel
-      return b.stackbox("H", convertChildren(content))
-   elseif content.command == "mrow" then
-      local ret = b.stackbox("H", convertChildren(content))
+      return b.stackbox(attributes, "H", convertChildren(content))
+   end
+
+   if content.command == "mrow" then
```
alerque, line 120: I don't think this coding style is an improvement. It looks  friendlier on the eyes sometimes, but it is also more fragile and can lead to mistakes. It's also slower. Lua doesn't have a switch statement, but it does have optimizations to skip unreachable control structures, and `if ... elseif ... end` is the right way to code these structures, not a sequence of `if ... end if ... end`

alerque (changes requested):

packages/math/typesetter.lua
Omikhleia: I tend to favor early returns over long sequences of if/elseif/end, but that's a coding style preference indeed. I find the former cleaner, personally. (EDIT and the long sequence of if/else, as we have it elsewhere, e.g. in the TeX-like adapter, is error-prone too, and can lead to mistakes as well, esp. as it is also interspersed with returns)

Omikhleia: N.B. both ways are in essence lame, this hints towards a table-oriented lookup, but I didn't want to go that far a refactor yet.

Omikhleia: > It's also slower. 

Are you sure this is true and significant ?

I'd expect performance to be virtually identical.

Both likely compile to the same underlying bytecode jumps.
Early returns might even be slightly better for LuaJIT (as a flat, predictable "hot path"). In general, CPU branch prediction favors early returns when filtering out rare states, so a proper ordering (most common tags first) _might_ even be better to that regard. But as noted, both paths probably scale poorly. Using a table lookup, while less readable (in some ways) would likely be the right move....

We can track this separately, if you wish. Currently performances are awful but for another reason: #2343  (here is a real performance bottleneck due to LPEG look-aheads / backtracking)

Omikhleia: (This being said, if you _really_ want if/return/elseif... sequences, I can comply and update the PR.)

PR #2346, opened 2025-11-15
feat(packages): Add scaling and anchoring to the background command

Omikhleia:
New options: `page` (for multi-page images such as a PDF), `scale` (for disabling scaling), `preserveaspect` (for aspect ratio preservation when scaling), `anchor` (for positioning of the image when needed).

Moreover a `color` and an image `src` can both be set, for cases where the image has transparent areas or isn't scaled to the full bleed area (due to aspect ratio preservation or absence of scaling).

---

The default behavior so far for images (corresponding to scale=true, preserveaspect=false) was nice in many cases (e.g. when using some old-paper-like texture on pages)...

But the extended proposal here is something I considered for some projects:
 - In one, I fancied having some sort of floral ornaments at the top-right corner of chapter pages (i.e. ideally anchor=ne, scale=true, preserveaspect=true to avoid distortion if changing the book paper size but using the same image file)
 - In another, I fancied having some markings (e.g. draft, confidential etc.) on pages (so anchor=center, scale=false).
 - When using a PDF as "background image", the ability to pick the page to use is also a nice option (e.g. in the just mentioned case for markings, using a PDF allows for scalable drawings, and one can have all backgrounds in a given file, etc.)

This is but a proposal, which I find interesting for my needs, but it also raises some good questions:
 - Whether allpages should be true by default...
 - Whether allpages and one-page backgrounds should be stacked rather that one disabling the other... On that matter see also in-code comment on the tricky question for the user for controlling where page breaks occur. That would lead to questioning the command(s) and their scope...
 - Whether the page (+ bleed) area is always the target, or some specific frame id could be used... But this needs to be _carefully_ considered, I played with it in another experiment but didn't include it here; it can quickly become a hornets' nest (due to switching masters and layouts etc.) and it is then hard to explain with good sense...

Feedbacks are welcome.

### DIFF
--- a/packages/background/init.lua
+++ b/packages/background/init.lua
@@ -5,18 +5,18 @@ package._name = "background"
 
 local background = {}
 
+local knownAnchorSet = pl.Set({
+   "center", "n", "ne", "e", "se", "s", "sw", "w", "nw"
+})
+
 local outputBackground = function ()
    local pagea = SILE.getFrame("page")
    local offset = SILE.documentState.bleed / 2
-   if type(background.bg) == "string" then
-      SILE.outputter:drawImage(
-         background.bg,
-         pagea:left() - offset,
-         pagea:top() - offset,
-         pagea:width() + 2 * offset,
-         pagea:height() + 2 * offset
-      )
-   elseif background.bg then
+
+   -- Background color first:
+   -- The image may not fully cover the area (depending on scaling
+   -- and aspect ratio preservation), and may anyway have transparent areas.
+   if background.bg then
       SILE.outputter:pushColor(background.bg)
       SILE.outputter:drawRule(
          pagea:left() - offset,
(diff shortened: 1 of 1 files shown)


packages/background/init.lua
```diff
@@ -63,20 +150,35 @@ package.documentation = [[
...
-The background extends to the page trim area (“page bleed”) if the latter is defined.
+The background color extends to the page trim area (“page bleed”) if the latter is defined.
 This is to ensure that it indeed “bleeds” off the sides of the page, so as to avoid thin white lignes on an otherwise full color page when the paper sheet is cut to dimension but some pages are trimmed slightly more than others.
-If setting only the current page background different from the default is desired, an extra parameter \autodoc:parameter{allpages=false} can be passed.
+
+When using an image as background, the following options are also available:
+\begin{itemize}
+\item{\autodoc:parameter{page=<number>} specifies which page of a multi-page image file to use as background (default is 1).}
+\item{\autodoc:parameter{scale=<boolean>} specifies whether to scale the image to fit the area (default is true).}
+\item{\autodoc:parameter{preserveaspect=<boolean>} specifies whether to preserve the image aspect ratio when scaling (default is false).
```
Omikhleia: Oops, missing `}` at the end of this item.

packages/background/init.lua
```diff
@@ -41,19 +109,38 @@ function package:registerCommands ()
...
+         background.scale = SU.boolean(options.scale, true)
+         background.preserveaspect = SU.boolean(options.preserveaspect, false)
       elseif color then
          background.bg = color
+         background.src = nil
       else
-         SU.error("background requires a color or an image src parameter")
+         SU.error("background requires at least a color or an image option")
       end
+      -- Changing the background immediately on the cuurrent page what one may
```
Omikhleia: typo > "cuurent" + missing "is"

> Changing the background immediately on the current page is what one may ....

packages/background/init.lua
Omikhleia: Fixed (amended)

packages/background/init.lua
Omikhleia: Fixed (amended)

PR #2360, opened 2026-02-06
fix(tableofcontents): Cast options.level to integer for TOC entry

leorosa:
When using the `tableofcontents` package, calling `\tocentry[level=1]{title}` stores the level number as a string instead of a number in the TOC file. Then, calling `\tableofcontents` crashes the document parsing.

This PR allows the package to work according to the indicated in documentation.


Omikhleia:
Good catch. (Same issue too in my own resilient.tableofcontents variant, but I never had the need to call `\tocentry` from SIL scripts)

Omikhleia (approved):

alerque (approved):

PR #2365, opened 2026-05-02
Add an image helper to replace libtexpdf's imagebbox and the vendored luasize library

### LINKED ISSUE #670: Extricate Lua imagesize module
Similar to #669, extricate the Lua `imagesize` module.

Omikhleia:
Closes #670 

Edit by Caleb: This PR now implements a Rust based helper that should accomplish the same goals.

----

This proposal adds Lua C bindings for a new minimal **imagehelper** library, intended to replace existing image size extraction mechanisms:
 - The current _imagebbox_ bindings based on **libtexpdf**, used via `outputter:getImageSize()`
 - The Lua implementation `lua-libraries/imagesize.lua`, used in other outputters, which: 
     - is a vendored copy of LGPL-licensed code
    - does not support the same set of formats

**What this provides**

 - Image bbox in pt (and resolution in dpi when available) for PDF, JPEG, JP2, PNG = as libtexpdf, exception made of BMP. (But is this bad format really still used in modern days? If so, it could be added, but heh...)
 - _As an extra_, number of pages for PDFs[^npages]

[^npages]: Quite handy for including a PDF in a document taking decisions (e.g. imposition) depending of the number of pages.

**Rationale**

 1. The current system is fragmented:
      - Lua-based `imagesize.lua` for some outputters
      - libtexpdf-based implementation for the corresponding outputter

  2. The libtexpdf path (derived from dvipdfmx):
       - reimplements image parsing instead of using standard libraries
       - is not designed for safe re-entrancy (relies on internal mutable state, with lots of "static" variables)
      - has required modifications (e.g. for `texpdf_doc_get_page` to work), which are difficult to maintain and update against newer dvipdfmx versions

Replacing the implementations with a unified module based on standard libraries improves correctness, maintainability and portability.

**Trade-offs**

This change introduces new dependencies for SILE: **libpng**, **libjpeg**, **libopenjp2** and **poppler-cpp**.

The first three are widely available in Linux distributions
However **poppler-cpp** may require additional installation in some environments. See comments below.

**Notes**

  - Only the C/C++ bindings are included in this change
  - Build system integration (Makefile / build scripts / packaging) is not updated. It still eludes me at this point, but see also further comments below.


Omikhleia:
**Comment on building the module**

For now, I built this module with the following Dockerfile

```
FROM ubuntu:24.04

ENV DEBIAN_FRONTEND=noninteractive

RUN apt-get update && apt-get install -y \
    build-essential \
    pkg-config \
    lua5.1 \
    liblua5.1-0-dev \
    libpoppler-cpp-dev \
    libpng-dev \
    libjpeg-dev \
    libopenjp2-7-dev \
    && rm -rf /var/lib/apt/lists/*

WORKDIR /src
COPY . /src

RUN g++ -O2 -fPIC -shared \
    -std=c++17 \
    -Wall -Wextra -Wpedantic -Werror \
    -Werror=return-type \
    -Wshadow \
    -Wconversion \
    -Wno-unused-parameter \
    imagehelper.cpp \
    -o imagehelper.so \
    $(pkg-config --cflags --libs poppler-cpp lua5.1 libpng libjpeg libopenjp2)

CMD ["/bin/bash"]
```

I build the library and extracted it from the container:

```
docker build -t pdf-lua .
docker run --rm pdf-lua cat /src/imagehelper.so > imagehelper.so
```

Finally, I copied it to my SILE environment (obviously I am using Ubuntu 24.04 here).

Omikhleia:
**Comments on smoke tests**

I had to add **libpoppler-cpp** to my Ubuntu environment. The rest of the dependencies was already satisfied (perhaps due to various extra tools I installed for my needs, not necessarily present on the base image).

```
$ add-apt-repository universe
$ apt install libpoppler-cpp0t64
$ sile
SILE v0.15.13 (LuaJIT 2.1.1703358377) [Rust]
> imagehelper = require("imagehelper")
> pdf = require("justenoughlibtexpdf")
> pdf.init("/tmp/hack", 500, 500, "producer") -- Just to enforce initialization (but creates a file)
> imagehelper.bbox("sherlock.png")
0	0	80.160160320321	80.160160320321	299.9994	299.9994
> pdf.imagebbox("sherlock.png", 1)
0	0	80.160160320321	80.160160320321	299.9994	299.9994
> imagehelper.bbox("paletted.jp2")
0	0	1024	1024	72	72
> pdf.imagebbox("paletted.jp2", 1)
0	0	1024	1024	72	72
> imagehelper.bbox("gutenberg.jpg")
0	0	189	240	72	72
> pdf.imagebbox("gutenberg.jpg", 1)
0	0	189	240	72	72
> imagehelper.bbox("resume-sample.pdf")
0	0	595.28	841.89	nil	nil
> pdf.imagebbox("resume-sample.pdf", 1)
0	0	595.28	841.89	nil	nil
> imagehelper.numpages("resume-sample.pdf")
2
```

Omikhleia:
N.B. The next step if this gets approved would be to work on an update of libtexpdf, without the burden of back-porting the re-entrancy hacks and other weird changes (partial renamings etc.)

I started experimenting in that direction, but the proposed independent image helper here is a first step to reduce our dependency on dvipfmx and the hacks it needed.

alerque (changes requested):
No, I'm quite against this exact approach. I'm utterly uninterested in adding more C code the project and this isn't even C, it is C++. I actually want to phase *out* the C stuff, not depend on it more. Additionally I've been down the road with libpoppler before and it's an ongoing pain point for packaging projects that depend on it because they regularly break things between releases.

That being said I totally agree with the need and what this is trying to accomplish, just not with the actual implementation.

My proposal would be to add the smoke tests demonstrating how you want this to work. I'll add the build stuff at least for local testing (not necessarily CI) just so I can see for sure how the Lua interface is supposed to work, then I'll write a Rust based provider that gets the job done without making it even harder than it already is to compile SILE as a static binary with embedded assets and/or use as a library. It should be pretty easy for me to work up the Rust based function and get it exported to Lua as long as we know what the function signature we want to have in Lua looks like.

alerque:
This should build on a host system that has the right libraries now. The Nix, Docker, and CI stuff is not fixed up to match but just using `just dev-conf` should be enough to reconfigure and build on the host system with the right libraries linked. I hope this helps at least for checking out the PoC and setting up unit tests to be the way we want them to work, then I can substitute the CPP for Rust.

alerque:
@Omikhleia I started some unit tests to compare the old and new image helper functions. Am I on the right track with the functionality that needs testing to have a drop in replacement that gives the same results?

Why did you drop the page number requirement for measuring a PDF bounding box?

Omikhleia:
> Why did you drop the page number requirement for measuring a PDF bounding box?

Did I do that ? Where ?

Omikhleia:
> > Why did you drop the page number requirement for measuring a PDF bounding box?
> 
> Did I do that ? Where ?

Oh perhaps here:

```
int page = (int)luaL_optinteger(L, 2, 1);
```

I.e. I didn't dropped it, but made it optional, defaulting to 1 ?

alerque:
> > > Why did you drop the page number requirement for measuring a PDF bounding box?
> > 
> > 
> > Did I do that ? Where ?
> 
> Oh perhaps here:
> 
> ```
> int page = (int)luaL_optinteger(L, 2, 1);
> ```
> 
> I.e. I didn't dropped it, but made it optional, defaulting to 1 ?

Okay, fair enough. In other words there wasn't really a mandate to have a different API.

The Rust implementation will (necessarily, as optional arguments are not a thing like they are in C++) make all arguments required on the Rust side, but we can probably wrap it in such a way that the Lua function call defaults to 1 if not provided. I was just wondering if there was a necessity for the API change over what libtexpdf was doing.

Omikhleia:
> The Rust implementation will (necessarily, as optional arguments are not a thing like they are in C++) make all arguments required on the Rust side, but we can probably wrap it in such a way that the Lua function call defaults to 1 if not provided

Wait, I am unsure I got the question then. I didn't use any C++ optional arguments.  The Lua wrapper accepts an optional page, defaulting to 1, but that's pure (standard) Lua C-API.

alerque:
I've re-implemented the logic for the image helper functions in Rust. It took a bit more code than I was expecting, largely because I used a lower level pair of crates that handle a bunch of image formats without very many high level functions. The main motivation for this was to not have to have a whole assortment of crates, some of them with potential system library dependencies — this implementation stays in pure Rust for all platforms.

I did make one major departure from the 3 other implementations: all of `pdf.imagebbox()`, the C++ implementation of `imagehelper.bbox()`, and the Lua `imagesize.imagesize()` functions returned bare floats for all values. For ease of migration and testing I made the Rust implementation of `rusile.imagebbox()` return the same 6 values, but the Lua binding for it returns the values that are points as `SILE.types.measurement(N, "pt")`. For comparing the test outputs to the original libtexpdf function I just cast this to a number. In the context of `packages.image` and its `\img{}` function (the only consumer of this outputter function distributed with SILE) no code change was required, the math functions being done on the results already just work. The difference is now it would be harder to accidentally use the point based size value in the wrong context. Does that small API change bother anything on your end @Omikhleia?

The intermediate C++ implementation is still in the history here if there is ever a reason to revive and compare it to anything, but I did drop it along with all the relevant build rules and tests. I also refactored all the outputters to use this implementation, and dropped the now obsolete vendored copy of the Lua shim.

alerque (approved):

PR #2373, opened 2026-07-13
Fix repeated hyphens logic for several languages, and update occitan and galician support

### LINKED ISSUE #2321: Quotes and punctuations in Occitan
_To be very honest and transparent,_ this comes from an interesting discussion on the Typst Discord 😁 

The Occitan language (which we do have actually in SILE, as language code "oc" with hyphenation patterns from TeX) is an old language from the south of France and areas in Spain, with revivals and involved communities.
 - In modern revivals, it _usually_ follows the French convention for punctuation marks (= primary quotes with _guillemets_, and the French "special" punctuation spaces rules applying: « ... », secondary quotes “…”, etc.)
 - But Aranese (a Pyrenean "Gascon" variety of the Occitan language, spoken in the Val d'Aran and in a few places in northwestern Catalonia, and by the way one of the "official" languages recognized by the Parliament of Catalonia, often uses guillemets too but without the punctuation spaces («...») for primary quotes. It's no fully authoritative here (some linguists advocate for the use of the curly braces for primary quotes in Aranese), but at least that's what the "Pedagogia - Portau deus ensenhaires d'occitan" website (promoting Occitan) says.

It's an interesting use case for SILE too, in the context of supporting minority languages... and an interesting case too for BCP47 support, ~~even not fully registered,~~ as one would need "oc" ~~to imply "oc-FR" (as Babel in LaTeX assumes "likely")~~ ; and ~~"oc-ES" (or something else)~~ "oc-aranese" for Aranese, ~~despite no "official" registration, but~~ to have a way for supporting the common usage (i.e. with or without punctuation spaces).

EDIT: I stand corrected, I hadn't looked at the IANA assignments in the language-subtag registry closely enough, Aranese has an official registration (`oc-aranes`) since 2018 ("Occitan variant spoken in the Val d'Aran").

### LINKED ISSUE #2372: Small issues with languages repeating explicit hyphens
**Observation**

1. In Polish, Czech, Spanish etc. the current language-specific node maker does:
   ``` 
   SILE.nodeMakers.pl.handlelineBreak = SILE.nodeMakers.unicode._handlelineBreakRepeatHyphen
   ```

   There's a double typo here, introduced in commit 45301559d: The method that was overridden in the previous commit was `handleLineBreak`  (with capital L). We are somewhat lucky that this override corresponded to edge-cases and is not needed in most usual cases (and hence also not triggered by the simple tests we have)...

1. I do have another "theoretical" issue with that commit, actually: The two underscored methods that were introduced aren't calling the parent (super) method (obviously, now that they resorted to prototype overrides / method-patching), but directly `SILE.nodeMakers.unicode.handleWordBreak` or `SILE.nodeMakers.unicode.handleLineBreak`. This is unfortunate for several reasons:
     - It breaks separation of concerns...
     - ... going via a global variable
     - ... and bypassing inheritance (e.g. so if one has derived another class for subtle changes on these methods, things start going bad)

**Expectations**

Besides the typo, there's some smell here.
 - I'd expect no method-patching here, and would rather propose to have but one generic unicode breaker, with appropriate feature flags.
 - Another (handled differently) language-specific / modified node breaker is the French one, but it would also benefiting from being generalized under feature flags. One may want a mix of those options (e.g. both French-like punctuation rules and repeated hyphens -- yes there's a case for that, though debated depending on the linguist being asked).
 - Moreover, I'd eventually be interested in trying to up-stream a generalization of the "experimental" handling of hanged punctuation introduced in re·sil·ient [v3.2](https://github.com/Omikhleia/resilient.sile/releases#release-v3.2.0) (Sept. 2025), and that would also benefit from a single implementation where all code paths are under control...
 - For similar reasons, #2279 would also be simplified, IMHO (so as to be able to mark both sides of a compound words, for instance, in a consistent way).

### LINKED ISSUE #2374: Galician (gl) should repeat lexical hyphens at line breaks
Galician (`gl`) follows the same type of rule as Portuguese and Spanish: When a lexical hyphen (one that is part of the word) falls at the end of a line, it is repeated at the beginning of the next line.

Source : https://ilg.usc.gal/corrector/docs/normas_galego_2003.pdf (Real Academia Galega (RAG), _Normas ortográficas e morfolóxicas do idioma galego,_ 20th ed., 2005, ISBN 978-84-87987-51-9, §3.2).

> Cando un guión que separa dúas palabras ou dous membros dunha palabra complexa coincide en final de liña, cómpre repetilo ao comezo da liña seguinte: _dicionario galego-/-inglés_.

That is (more or less):
> When a dash/hyphen that separates two words or two members of a complex word coincides at the end of a line, it must be repeated at the beginning of the following line: _ dicionario galego-/-inglés._

Image

Omikhleia:
Minimally addressing open issues:

Closes #2372 by refactoring the unicode segmenter used by languages cs, es, hr, pl, pt, sk (switching to a feature flag)

Closes #2321 by changing the language setup for "oc" and introducing support for "oc-aranes"; also adding a minimal i18n translation file for both languages.

Closes #2374 by enabling the repeated hyphen flag on language "gl"; also adding a minimal i18n translation file for this language.

### DIFF
--- a/languages/cs/init.lua
+++ b/languages/cs/init.lua
@@ -1,8 +1,7 @@
 SILE.nodeMakers.cs = pl.class(SILE.nodeMakers.unicode)
 
 -- According to Czech rules, when a break occurs at an explicit hyphen, the hyphen gets repeated on the next line...
-SILE.nodeMakers.cs.handleWordBreak = SILE.nodeMakers.unicode._handleWordBreakRepeatHyphen
-SILE.nodeMakers.cs.handlelineBreak = SILE.nodeMakers.unicode._handlelineBreakRepeatHyphen
+SILE.nodeMakers.cs.hasFeatureRepeatedHyphen = true
 
 local hyphens = require("languages.cs.hyphens-tex")
 SILE.hyphenator.languages["cs"] = hyphens
--- a/languages/es/init.lua
+++ b/languages/es/init.lua
@@ -1,10 +1,7 @@
 SILE.nodeMakers.es = pl.class(SILE.nodeMakers.unicode)
 
 -- According to Spanish rules, when a break occurs at an explicit hyphen, the hyphen gets repeated on the next line...
-SILE.nodeMakers.es.handleWordBreak = SILE.nodeMakers.unicode._handleWordBreakRepeatHyphen
-SILE.nodeMakers.es.handlelineBreak = SILE.nodeMakers.unicode._handlelineBreakRepeatHyphen
-
-SILE.hyphenator.languages["es"] = {}
+SILE.nodeMakers.es.hasFeatureRepeatedHyphen = true
 
 local hyphens = require("languages.es.hyphens-tex")
 SILE.hyphenator.languages["es"] = hyphens
--- a/languages/gl/init.lua
+++ b/languages/gl/init.lua
@@ -1,2 +1,15 @@
+SILE.nodeMakers.gl = pl.class(SILE.nodeMakers.unicode)
+
+-- Real Academia Galega (RAG), _Normas ortográficas e morfolóxicas do idioma
+-- galego,_ 20th ed., 2005, ISBN 978-84-87987-51-9, §3.2
+-- (https://ilg.usc.gal/corrector/docs/normas_galego_2003.pdf)
+-- "Cando un guión que separa dúas palabras ou dous membros dunha palabra
+-- complexa coincide en final de liña, cómpre repetilo ao comezo da liña
+-- seguinte: _dicionario galego-/-inglés._"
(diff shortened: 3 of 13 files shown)


Omikhleia:
This is a very conservative approach, just fixing the above-mentioned issues. As stated in #2372,  I'd eventually like to consider some other changes in the Unicode segmenter, but it can probably wait...

Omikhleia:
Moving to draft: I'll add #2374 here too.

Omikhleia:
> Moving to draft: I'll add #2374 here too.

Done.

alerque (approved):

PR #2375, opened 2026-07-18
feat(languages): Add basic support for Kazakh

Omikhleia:
Hyphenation patterns for Kazakh (kk) were added to the TeX hyphenation repository (https://github.com/hyphenation/tex-hyphen) in February 2026. 

They are here transpiled to Lua for use by SILE, and a minimal i18n translation file is also added.

See bullet point in #2104


alerque (approved):
Thanks. And I happen to be in Kazakhstan right now, so this seems like a good time to merge!

PR #2382, opened 2026-09-09
Add support for Lua 5.5

alerque:


alerque:
This branch is now fully function on Lua 5.5 including all unit and regression tests, but it is dependent on a patched version of vstruct. There is [an upstream PR waiting that adds Lua 5.5 support](https://github.com/ToxicFrog/vstruct/pull/42) but it isn't merged an released yet. Hopefully we can get some motion on that, if now I suppose the solution is we'll have to vendor a patched version or fork it and release it ourselves. I'd rather not have to do either of those things, but we can't just ignore 5.5 either.

I've backported the patch for vstruct to the Arch Linux packaging, so SILE would function an Lua 5.5 there, but I don't think we can release without some way for non Arch users to get rolling.

